PR #48, opened 2020-11-23
Add SPLUNK_MEMORY_TOTAL_MIB

flands:
Today, SPLUNK_BALLAST_SIZE_MIB needs to be computed and manually
configured by a user. Most users do not know or care about the ballast
and under normal circumstances they should not need to. Users will
understand the total amount of memory allocated to the Collector. As
such, introduce a SPLUNK_MEMORY_TOTAL_MIB env var that can be used
instead of or in addition to SPLUNK_BALLAST_SIZE_MIB. This has the added
bonus that now memory spike and limit env vars are no longer required
for non-linux systems. This change is backwards compatible.

A few other minor enhancements:

- Memory spike/limit env vars are now applied even if SPLUNK_CONFIG is
  passed
- Add logging for custom behavior so configuration is clear
- Add accessible endpoints to README
- Add tests

### DIFF
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -245,18 +245,18 @@ func checkMemorySettingsMiBFromEnvVar(envVar string, memTotalSizeMiB int) int {
 }
 
 func useMemorySettingsMiBFromEnvVar(memTotalSizeMiB int) {
-    // Check if memory limit is specified via environment variable
+	// Check if memory limit is specified via environment variable
 	memLimit := checkMemorySettingsMiBFromEnvVar(memLimitMiBEnvVarName, memTotalSizeMiB)
-    // Use if set, otherwise memory total size must be specified
+	// Use if set, otherwise memory total size must be specified
 	if memLimit == 0 {
 		if memTotalSizeMiB == 0 {
 			panic("PANIC: Both memory limit MiB and memory total size are set to zero. This should never happen.")
 		}
-        // If not set, compute based on memory total size specified
-        // and default memory limit percentage const
+		// If not set, compute based on memory total size specified
+		// and default memory limit percentage const
 		memLimitMiB := memTotalSizeMiB * defaultMemoryLimitPercentage / 100
-        // The memory limit should be set to defaultMemoryLimitPercentage of total memory
-        // while reserving a maximum of defaultMemoryLimitMaxMiB of memory.
+		// The memory limit should be set to defaultMemoryLimitPercentage of total memory
+		// while reserving a maximum of defaultMemoryLimitMaxMiB of memory.
 		if (memTotalSizeMiB - memLimitMiB) < defaultMemoryLimitMaxMiB {
 			memLimit = memLimitMiB
 		} else {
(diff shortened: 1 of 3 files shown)


cmd/otelcol/main.go
```diff
@@ -193,16 +235,42 @@ func checkMemorySettingsMiBFromEnvVar(envVar string) int {
...
+	if memLimit == 0 {
+		if memTotalSize == 0 {
+			panic("PANIC: Both memory limit MiB and memory total size are set to zero. This should never happen.")
+		}
+		memLimitMiB := memTotalSize * defaultMemoryLimitPercentage / 100
+		if (memTotalSize - memLimitMiB) < defaultMemoryLimitMaxMiB {
+			memLimit = memLimitMiB
+		} else {
+			memLimit = (memTotalSize - defaultMemoryLimitMaxMiB)
+		}
```
tigrannajaryan, line 258: It is not clear what this calculation does and what's relationship between defaultMemoryLimitPercentage and defaultMemoryLimitMaxMiB and why both exist.

cmd/otelcol/main.go
```diff
@@ -50,22 +51,50 @@ const (
...
+	defaultMemorySpikeMaxMiB        = 2048
 )
 
 func main() {
+	// TODO: Use same format as the collector
 	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
 
+	// Check if the total memory is specified via the env var.
+	memTotalEnvVarVal := os.Getenv(memTotalEnvVarName)
+	memTotalSize := 0
```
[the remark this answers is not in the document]
flands: Will update all instances

cmd/otelcol/main.go
```diff
@@ -98,18 +127,27 @@ func contains(arr []string, str string) bool {
 	return false
 }
 
-func useBallastSizeFromEnvVar() {
+func useMemorySizeFromEnvVar(memTotalSize int) {
```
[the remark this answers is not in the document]
flands: Will update all instances

cmd/otelcol/main.go
flands: percentage is percentage from total memory size and max is the maximum of amount of memory. I added a comment to hopefully clarify. In short, we want to set the limit to 90% of total memory reserving a maximum of 2048 of memory while for spike at 25% for a maximum of 2048 of memory.

cmd/otelcol/main.go
```diff
@@ -193,16 +235,54 @@ func checkMemorySettingsMiBFromEnvVar(envVar string) int {
...
+	// Use if set, otherwise memory total size must be specified
+	if memLimit == 0 {
+		if memTotalSizeMiB == 0 {
+			panic("PANIC: Both memory limit MiB and memory total size are set to zero. This should never happen.")
+		}
+		// If not set, compute based on memory total size specified
+		// and default memory limit percentage const
+		memLimitMiB := memTotalSizeMiB * defaultMemoryLimitPercentage / 100
+		// The memory limit should be set to defaultMemoryLimitPercentage of total memory
+		// while reserving a maximum of defaultMemoryLimitMaxMiB of memory.
```
tigrannajaryan, line 259: I am not sure I understand why we do this. So we specify let's say 8GB of total memory size, but this calculation will give only 2GB to memory limiter? Are we assuming the remaining 6GB will be used by something else?
Just to clarify: the memory limiter calculates the total size of the memory used by the process. I am confused about what we are trying to achieve here.

tigrannajaryan: OK, re-reading the code it appears this code tries to limit the unused memory. The unused memory is the 10% which we did not give to memory limiter despite the user telling us to do so.
Is this 10% reserved because we do not trust the memory to enforce it? And so we believe that the memory limiter may overshoot sometimes but never by more than 2GB, that's why if 10% is more than 2GB, we only reserve 2GB but no more.
Is this what we do?

tigrannajaryan (approved):

tigrannajaryan:
@flands something went wrong with Windows build: https://app.circleci.com/pipelines/github/signalfx/splunk-otel-collector/262/workflows/e59b6f5f-0c14-49dd-b819-e03f60aa7fc7/jobs/2141

PR #50, opened 2020-12-10
Disable starting fluentd with default config

jchengsfx:
The default fluentd config `/etc/td-agent/td-agent.conf` listens on port 8888:
```# HTTP input
# POST http://localhost:8888/<tag>?json=<json>
# POST http://localhost:8888/td.myapp.login?json={"user"%3A"me"}
# @see http://docs.fluentd.org/articles/in_http
<source>
  @type http
  @id input_http
  port 8888
</source>
```

This conflicts with the default port for the prometheus receiver, causing the collector to fail to start.

The installer script is updated to disable the fluentd service from starting and moving the default config.

### DIFF
--- a/internal/buildscripts/packaging/installer/install.sh
+++ b/internal/buildscripts/packaging/installer/install.sh
@@ -497,7 +497,13 @@ parse_args_and_install() {
   configure_realm "$realm" "$service_user" "$service_group"
   configure_ballast "$ballast" "$service_user" "$service_group"
 
-  systemctl start td-agent
+  # disable fluentd from starting with default config
+  systemctl stop td-agent
+  if [ -f /etc/td-agent/td-agent.conf ]; then
+    mv -f /etc/td-agent/td-agent.conf /etc/td-agent/td-agent.conf.default
+  fi
+  systemctl disable td-agent
+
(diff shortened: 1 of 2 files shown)


jrcamp:
Do we need the input_http plugin turned on at all?

jrcamp:
Maybe worth auditing the entire file to see if there's other stuff in it we don't need.

jchengsfx:
> Do we need the input_http plugin turned on at all?

Not that I'm aware of. Currently, the installer script and collector package doesn't interact with fluentd at all, other than installing the td-agent package. So the fluentd service just starts with the default config, which includes the input_http and other plugins enabled (see https://github.com/treasure-data/omnibus-td-agent/blob/master/templates/etc/td-agent/td-agent.conf).

Once we add our custom fluentd config to override the default config, we'll only enable the plugins/sources that are necessary for the collector.

Alternatively, instead of changing the port, I could just temporarily remove the default config and disable the fluentd service from auto-starting until we add our custom config?

jrcamp:
> Alternatively, instead of changing the port, I could just temporarily remove the default config and disable the fluentd service from auto-starting until we add our custom config?

Sounds good

PR #52, opened 2020-12-21
Update deb/rpm and installer script for fluentd

jchengsfx:
*Dependent on fixes for the `fluentfoward` receiver upstream*

- Add `fluentforward` receiver, `splunk_hec` exporter, and custom endpoint URL env vars to the default linux config
- Add default fluentd config to forward `@SPLUNK` labeled events to the collector
- Add sample fluentd source for journald (currently not enabled by default since it requires dependencies and permissions that we may not want to perform on the system)
- Update installer script to install/configure fluentd and add options for custom endpoint URLs
- Update deb/rpm packaging for new fluentd config files and set ownership/permissions

### DIFF
--- a/cmd/otelcol/config/collector/agent_config_linux.yaml
+++ b/cmd/otelcol/config/collector/agent_config_linux.yaml
@@ -5,10 +5,12 @@ extensions:
   health_check:
   http_forwarder:
     egress:
-      endpoint: "https://api.${SPLUNK_REALM}.signalfx.com"
+      endpoint: "${SPLUNK_API_URL}"
   zpages:
 
 receivers:
+  fluentforward:
+    endpoint: 0.0.0.0:8006
   hostmetrics:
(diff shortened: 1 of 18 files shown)


cmd/otelcol/config/collector/agent_config_linux.yaml
```diff
@@ -5,10 +5,12 @@ extensions:
   health_check:
   http_forwarder:
     egress:
-      endpoint: "https://api.${SPLUNK_REALM}.signalfx.com"
+      endpoint: "${SPLUNK_API_URL}"
   zpages:
 
 receivers:
+  fluentforward:
+    endpoint: 0.0.0.0:8006
```
jrcamp, line 13: localhost is probably sufficient since the primary usecase is collecting local logs from fluentd.

jrcamp (approved):

PR #54, opened 2021-01-05
Add MonitorCustomConfig support to Smart Agent Receiver

rmfitzpatrick:
These changes address the first remaining item detailed in https://github.com/signalfx/splunk-otel-collector/pull/53#issue-549208453.  In most cases they allow users to c/p SA monitor configs to corresponding `smartagent` receiver config blocks, which will be later used by the receiver to invoke the applicable monitor factories.  In cases where config tag* values are maps with case sensitive strings, this implementation does not correctly translate since the Collector config is based on viper, which destroys all record of tag key cases.  A translation approach for these cases will be implemented when necessary, though the majority of use cases are not planned to be supported by the Smart Agent Receiver (e.g. in-monitor metric transformations).

At a high level this implementation works similarly to the Smart Agent's dynamic config functionality, and is based on the reflection package to create and populate custom monitor config content from the provided yaml.  It uses a custom viper unmarshaller and the central SA monitor registration maps to meet this end.

Also including required go modules replace directives to import the agent and a required windows test job environment correction.


go.mod
```diff
@@ -33,10 +33,14 @@ require (
 	github.com/ory/go-acc v0.2.6
 	github.com/pavius/impi v0.0.3
 	github.com/securego/gosec/v2 v2.5.0
+	github.com/signalfx/defaults v1.2.2-0.20180531161417-70562fe60657
+	github.com/signalfx/signalfx-agent v0.0.0-00010101000000-000000000000
```
tigrannajaryan, line 37: Is `v0.0.0-00010101000000-000000000000` a correct version number of SA that we want to use? The latest version appears to be `v5.7.1`, but we are not referencing it here.

go.mod
```diff
@@ -50,3 +54,15 @@ replace (
...
 	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/redisreceiver v0.0.0-00010101000000-000000000000 => github.com/open-telemetry/opentelemetry-collector-contrib/receiver/redisreceiver v0.15.0
 )
+
+// each of these is required for the smartagentreceiver
+replace (
+	code.cloudfoundry.org/go-loggregator => github.com/signalfx/go-loggregator v1.0.1-0.20200205155641-5ba5ca92118d
+	github.com/dancannon/gorethink => gopkg.in/gorethink/gorethink.v4 v4.0.0
+	github.com/influxdata/telegraf => github.com/signalfx/telegraf v0.10.2-0.20201211214327-200738592ced
+	github.com/prometheus/prometheus => github.com/prometheus/prometheus v1.8.2-0.20201105135750-00f16d1ac3a4
+	github.com/signalfx/signalfx-agent => github.com/signalfx/signalfx-agent v1.0.1-0.20210104180832-f3147854c36c
```
tigrannajaryan, line 64: Can you please clarify version number used? Which SmartAgent version are we importing? The latest seems to be v5.7.1 but we use v1.0.1? I am probably misunderstanding something.

rmfitzpatrick: The Smart Agent didn't follow the expected practice of a major version suffix for go modules (https://blog.golang.org/v2-go-modules) so I believe now only pseudo-versions are supported.  Without them, the latest version available is the last one before adopting go modules.  These versions are correct for each modules within the signalfx-agent project per `go get`.

go.mod
rmfitzpatrick: It's the pseudo version that doesn't refer to an actual commit but is inserted by go when using local paths: https://golang.org/doc/go1.12#modules.  The subsequent replace directive will always apply for our builds so I don't believe it really matters what version is specified here.

go.mod
rmfitzpatrick: To clarify, the version is this commit: https://github.com/signalfx/signalfx-agent/commit/f3147854c36c45a7d73e52f4b94f57a83381df19.  It's required for the tests to pass.

go.mod
```diff
@@ -50,3 +54,15 @@ replace (
 	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/k8sclusterreceiver v0.0.0-00010101000000-000000000000 => github.com/open-telemetry/opentelemetry-collector-contrib/receiver/k8sclusterreceiver v0.15.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/redisreceiver v0.0.0-00010101000000-000000000000 => github.com/open-telemetry/opentelemetry-collector-contrib/receiver/redisreceiver v0.15.0
 )
+
+// each of these is required for the smartagentreceiver
```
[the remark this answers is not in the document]
rmfitzpatrick: I believe only for grpc (currently 1.29.1 vs 1.34.0 in the collector) since that requires ongoing changes in etcd: https://github.com/etcd-io/etcd/pull/12398 (and more likely https://github.com/etcd-io/etcd/pull/12597)

internal/receiver/smartagentreceiver/config.go
```diff
@@ -16,18 +16,114 @@ package smartagentreceiver
...
-		// validation placeholder for coverage
-		return fmt.Errorf("name and type are required")
+func (rCfg *Config) validate() error {
+	if rCfg.monitorConfig == nil {
+		return fmt.Errorf("you must supply a valid Smart Agent Monitor config")
 	}
+
+	monitorConfigCore := rCfg.monitorConfig.MonitorConfigCore()
+	if monitorConfigCore.IntervalSeconds == 0 {
+		monitorConfigCore.IntervalSeconds = 10
```
tigrannajaryan, line 44: Nit: we typically define config defaults as consts somewhere close to the top of the file for better visibility.

tigrannajaryan (approved):

rmfitzpatrick:
@jrcamp mind giving this a look w/ your SA expertise when possible?

jrcamp (approved):

PR #55, opened 2021-01-07
smartagentreceiver: add Dynamic Monitor creation

rmfitzpatrick:
These changes introduce the main receiver start and shutdown functionality (addressing the second remaining item detailed in https://github.com/signalfx/splunk-otel-collector/pull/53#issue-549208453).  They include dynamic SA monitor creation using the provided config and establish a means of accessing the SFx datapoints reported by the monitors.  These change don't include datapoint conversion, which will be introduced in the next proposed addition.

They also don't include the core collectd process creation, which will also be introduced in a later PR.

### DIFF
--- a/internal/receiver/smartagentreceiver/receiver.go
+++ b/internal/receiver/smartagentreceiver/receiver.go
@@ -16,8 +16,17 @@ package smartagentreceiver
 
 import (
 	"context"
+	"fmt"
+	"reflect"
+	"strings"
+	"sync"
 
+	"github.com/signalfx/signalfx-agent/pkg/core/config"
+	"github.com/signalfx/signalfx-agent/pkg/monitors"
+	"github.com/signalfx/signalfx-agent/pkg/monitors/types"
+	"github.com/signalfx/signalfx-agent/pkg/utils"
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/component/componenterror"
 	"go.opentelemetry.io/collector/consumer"
 	"go.uber.org/zap"
 )
@@ -25,7 +34,11 @@ import (
 type Receiver struct {
 	logger       *zap.Logger
 	config       *Config
+	monitor      interface{}
 	nextConsumer consumer.MetricsConsumer
+
+	startOnce sync.Once
+	stopOnce  sync.Once
 }
 
 var _ component.MetricsReceiver = (*Receiver)(nil)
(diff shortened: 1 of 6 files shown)


internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -39,9 +52,74 @@ func NewReceiver(logger *zap.Logger, config Config, nextConsumer consumer.Metric
...
 func (r *Receiver) Start(_ context.Context, host component.Host) error {
-	return nil
+	err := r.config.validate()
+	if err != nil {
+		return fmt.Errorf("config validation failed for %q: %w", r.config.Name(), err)
+	}
+
+	configCore := r.config.monitorConfig.MonitorConfigCore()
+	monitorType := configCore.Type
+	monitorName := strings.ReplaceAll(r.config.Name(), "/", "")
```
tigrannajaryan, line 62: Is / a prohibited character and that's why we replace it? Where is the monitor name visible? In the logs, in error messages? Perhaps replace by something that still looks like a delimiter but is a valid character?

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -39,9 +52,74 @@ func NewReceiver(logger *zap.Logger, config Config, nextConsumer consumer.Metric
...
+}
+
+func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
+	// retrieve registered MonitorFactory from agent's registration store
+	monitorFactory, ok := monitors.MonitorFactories[monitorType]
+	if !ok {
+		return nil, fmt.Errorf("unable to find MonitorFactory for %q", monitorType)
+	}
+
+	monitor := monitorFactory() // monitor is a pointer to a monitor struct
```
tigrannajaryan, line 100: So, I guess `interface{}` is all we have.

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -39,9 +52,74 @@ func NewReceiver(logger *zap.Logger, config Config, nextConsumer consumer.Metric
...
+
+	// monitorOutputValue is the monitor struct's "Output" field value obtained via reflection to be set to
+	// an Output instance.
```
tigrannajaryan, line 103: Is this the only way we can get the `Output` field? How is this done in SmartAgent codebase?
Looks like SmartAgent uses something called ActiveMonitor to wrap the monitor instance and then sets ActiveMonitor.Output field. Would this approach also work for us?

tigrannajaryan (approved):

internal/receiver/smartagentreceiver/receiver.go
rmfitzpatrick: The monitor id is used as a unique internal SFx datapoint dimension value in ~collectd and elsewhere, but it may be deleted before output in all cases.  I was thinking it could in some cases be a dimension value so instinctively stripped the slash as an unsupported character in SFx.

internal/receiver/smartagentreceiver/receiver.go
rmfitzpatrick: This effectively follows the practice of the ActiveMonitor, which informed its implementation.  We aren't incorporating ActiveMonitors directly because they're largely not exported and used as the mechanism to maintain monitor manager, which we're bypassing.

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -16,16 +16,29 @@ package smartagentreceiver
...
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/component/componenterror"
 	"go.opentelemetry.io/collector/consumer"
 	"go.uber.org/zap"
 )
 
 type Receiver struct {
 	logger       *zap.Logger
 	config       *Config
+	monitor      interface{}
```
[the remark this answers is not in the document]
rmfitzpatrick: Not by [`MonitorFactory`](https://github.com/signalfx/signalfx-agent/blob/ce1a2e0a25e75bb6623f29c7c23cd6271ca5defe/pkg/monitors/monitor.go#L17):
```
// MonitorFactory is a niladic function that creates an unconfigured instance
// of a monitor.
type MonitorFactory func() interface{}
```

Each monitor implements their own w/ a `Configure()` method that is invoked reflectively by `CallConfigure()`

internal/receiver/smartagentreceiver/receiver.go
rmfitzpatrick: Yeah seems so (https://github.com/signalfx/splunk-otel-collector/pull/55#discussion_r553615397)

internal/receiver/smartagentreceiver/receiver.go
rmfitzpatrick: https://github.com/signalfx/signalfx-agent/blob/4dd6371cc380f30012bc321519f02b86fd5b3e64/pkg/monitors/activemonitor.go#L121

PR #56, opened 2021-01-11
 Smart Agent Receiver: add datapoint conversion

rmfitzpatrick:
These changes address the 3rd remaining item detailed in https://github.com/signalfx/splunk-otel-collector/pull/53#issue-549208453 by adding internal SFx datapoint to pdata metric translation and forwarding to the next metric consumer.  They are largely based on the translation that occurs in [Collector Contrib's SFx receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/receiver/signalfxreceiver) but without using SFx's protobuf definitions.

In testing these changes, there is a minor form of end-to-end integration test that uses the native SA cpu monitor without vetting its metric content too heavily, which I deemed out of scope after earlier attempts in light of CI flake and cross-platform concerns.

### DIFF
--- a/internal/receiver/smartagentreceiver/convert.go
+++ b/internal/receiver/smartagentreceiver/convert.go
@@ -1,0 +1,185 @@
+// Copyright 2021, OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package smartagentreceiver
+
+import (
+	"fmt"
+
+	sfx "github.com/signalfx/golib/v3/datapoint"
+	"go.opentelemetry.io/collector/consumer/pdata"
+	"go.uber.org/zap"
+)
+
+var (
+	errUnsupportedMetricTypeTimestamp = fmt.Errorf("unsupported metric type timestamp")
+	errNoIntValue                     = fmt.Errorf("no valid value for expected IntValue")
+	errNoFloatValue                   = fmt.Errorf("no valid value for expected FloatValue")
+)
+
+type Converter struct {
+	logger *zap.Logger
+}
+
+// Based on https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.15.0/receiver/signalfxreceiver/signalfxv2_to_metricdata.go
+func (c *Converter) toMetrics(datapoints []*sfx.Datapoint) (pdata.Metrics, int) {
+	numDropped := 0
+	md := pdata.NewMetrics()
+	md.ResourceMetrics().Resize(1)
+	rm := md.ResourceMetrics().At(0)
+
+	rm.InstrumentationLibraryMetrics().Resize(1)
+	ilm := rm.InstrumentationLibraryMetrics().At(0)
+
+	metrics := ilm.Metrics()
+	metrics.Resize(len(datapoints))
+
+	i := 0
+	for _, datapoint := range datapoints {
+		if datapoint == nil {
+			continue
+		}
+
+		m := metrics.At(i)
+		err := setDataType(datapoint, m)
+		if err != nil {
+			numDropped++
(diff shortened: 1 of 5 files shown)


internal/receiver/smartagentreceiver/convert.go
```diff
@@ -0,0 +1,186 @@
...
+func setDataType(datapoint *sfx.Datapoint, m pdata.Metric) error {
+	sfxMetricType := datapoint.MetricType
+	if sfxMetricType == sfx.Timestamp {
+		return errUnsupportedMetricTypeTimestamp
+	}
+
+	var isFloat bool
+	switch datapoint.Value.(type) {
+	case sfx.IntValue:
+		isFloat = false
```
asuresh4, line 101: nit: this could be omitted since booleans are by default set to `false`.

internal/receiver/smartagentreceiver/convert.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, datapoint := range datapoints {
+		if datapoint == nil {
+			continue
+		}
+
+		m := metrics.At(i)
+		err := setDataType(datapoint, m)
+		if err != nil {
+			numDropped++
+			c.logger.Debug("SignalFx datapoint type conversion error",
```
asuresh4, line 57: Should this error and the one for the conversion be logged as error or a warning instead of debug?

internal/receiver/smartagentreceiver/convert.go
```diff
@@ -0,0 +1,186 @@
...
+func fillIntDatapoint(datapoint *sfx.Datapoint, dps pdata.IntDataPointSlice) error {
+	var intValue sfx.IntValue
+	var ok bool
+	if intValue, ok = datapoint.Value.(sfx.IntValue); !ok {
+		return errNoIntValue
+	}
+
+	dps.Resize(1)
+	dp := dps.At(0)
+	dp.SetTimestamp(pdata.TimestampUnixNano(uint64(datapoint.Timestamp.UnixNano())))
```
asuresh4, line 157: Should we check for `nil` timestamps on datapoint?

internal/receiver/smartagentreceiver/convert.go
rmfitzpatrick: Thanks, updated to warning.

internal/receiver/smartagentreceiver/convert.go
rmfitzpatrick: Still not used to go's lack of falling through in switches.  Removed.

internal/receiver/smartagentreceiver/convert.go
rmfitzpatrick: since `Timestamp` is a `time.Time` and not a pointer I don't believe nil can be assigned?

internal/receiver/smartagentreceiver/convert.go
```diff
@@ -0,0 +1,185 @@
...
+	for _, datapoint := range datapoints {
+		if datapoint == nil {
+			continue
+		}
+
+		m := metrics.At(i)
+		err := setDataType(datapoint, m)
+		if err != nil {
+			numDropped++
+			c.logger.Warn("SignalFx datapoint type conversion error",
```
tigrannajaryan, line 57: It is probably better to use Debug instead of Warn for conversion errors otherwise it can flood the logs: https://github.com/open-telemetry/opentelemetry-collector/blob/master/CONTRIBUTING.md#logging

internal/receiver/smartagentreceiver/output.go
```diff
@@ -60,6 +71,9 @@ func (output *Output) Copy() types.Output {
 
 func (output *Output) SendDatapoints(datapoints ...*datapoint.Datapoint) {
 	output.logger.Debug("SendDatapoints has been called.", zap.Any("datapoints", datapoints))
+	metrics, numDropped := output.converter.toMetrics(datapoints)
+	output.logger.Debug("SendDatapoints", zap.Any("metrics", metrics), zap.Int("numDropped", numDropped))
```
tigrannajaryan, line 75: We may want to remove the logger.Debug's in the final version so that Debug logs are not flooded for success cases. OK to keep for now if you are using them for debugging while you are actively developing.

tigrannajaryan (approved):

internal/receiver/smartagentreceiver/output.go
rmfitzpatrick: Good point, updated to use internal tracing and only debug log on dropped points.

internal/receiver/smartagentreceiver/convert.go
rmfitzpatrick: Makes sense, reverted change.

PR #57, opened 2021-01-12
Prometheus federation example

atoulme:
Add an example showing how to use the OpenTelemetry collector to collect metrics from a Prometheus server.

This example currently works well with the latest code of opentelemetry-collector-contrib (after the merge of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/1949), and will require a new release before they work here.


jrcamp:
If it lands in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/2007 in contrib I don't think we want it duplicated here as they risk getting out of sync. (There's perhaps an issue of discoverability of things that are in contrib repo but that applies to our docs as a whole.)

atoulme:
This example is vendor-specific as it introduces a Splunk instance, and that is quite useful to show. The example in the opentelemetry repository logs all metrics to stdout.

examples/prometheus-federation/prom-counter/main.go
```diff
@@ -0,0 +1,64 @@
...
+	meter.RecordBatch(ctx,
+		commonLabels,
+		valueRecorder.Measurement(counter))
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM, syscall.SIGQUIT)
+	ticker := time.NewTicker(1 * time.Second)
+	for {
+		select {
+		case <-ticker.C:
+			time.Sleep(1 * time.Second)
```
rmfitzpatrick, line 52: Same question as https://github.com/signalfx/splunk-otel-collector/pull/65#discussion_r563989525

rmfitzpatrick (approved):

examples/prometheus-federation/prom-counter/main.go
atoulme: Removed.

flands (approved):
LGTM

PR #62, opened 2021-01-20
Add logrus hook to redirect log entries by global logrus logger to zap

bjsignalfx:


internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -44,6 +46,9 @@ type Receiver struct {
 var _ component.MetricsReceiver = (*Receiver)(nil)
 
 func NewReceiver(logger *zap.Logger, config Config, nextConsumer consumer.MetricsConsumer) *Receiver {
+	logrus.SetOutput(ioutil.Discard)
+	logrus.AddHook(&ZapHooks{ZapLogger: logger})
```
jrcamp, line 50: If you have multiple instances of a smartagent receiver (not uncommon, especially when they're started dynamically with receiver_creator) if a receiver starts and stops it's still going to be registered with logrus. Also log messages will be duplicated because it's a global. So if you start 3 smartagent receivers I think the same log message would be logged 3 times. Unsure what the solution is.

bjsignalfx: Hey @jrcamp I did some refactoring. Please have a look and comment.

rmfitzpatrick:
Do you mind adding tests?

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -17,6 +17,8 @@ package smartagentreceiver
 import (
 	"context"
 	"fmt"
+	"github.com/signalfx/splunk-otel-collector/internal/zaputil"
+	"github.com/sirupsen/logrus"
```
rmfitzpatrick, line 21: `make lint` would reveal these are in undesired groups.  Also think you want to run `go fmt`.

internal/zaputil/logrus.go
```diff
@@ -0,0 +1,96 @@
...
+func RedirectLogrusLogs(from *logrus.Logger, to *zap.Logger) {
+	mu.Lock()
+	defer mu.Unlock()
+	r := redirect{from, to}
+	if redirects[r] {
+		return
+	}
+	from.ReportCaller = true
+	from.SetOutput(ioutil.Discard)
+	from.AddHook((*zapWrapper)(to))
```
rmfitzpatrick, line 50: `from` is always logrus's exported std logger right?  This would mean that all agent logs would be redirected to all smart agent receiver loggers.  Are we sure there is only one central zap logger that all components share?

internal/receiver/smartagentreceiver/receiver.go
jrcamp: I'm not sure how this helps the scenarios above. I don't really see a solution to making the two fit without modifying the smart agent code to have a per-monitor instance logger.

internal/zaputil/logrus.go
```diff
@@ -0,0 +1,96 @@
...
+	levelsMap = newLevelsMap()
+)
+
+// RedirectLogrusLogs mutes the output of the supplied logrus.Logger and adds a hook to it.
+// When fired the hook creates a zap entry using the supplied logrus entry.
+func RedirectLogrusLogs(from *logrus.Logger, to *zap.Logger) {
+	mu.Lock()
+	defer mu.Unlock()
+	r := redirect{from, to}
+	if redirects[r] {
```
rmfitzpatrick, line 45: nit: since the boolean isn't really used and only an existence check is performed I believe it's customary to do something like:
```go
_, ok := redirects[r] ; if ok {
  return
}
```

internal/zaputil/logrus.go
rmfitzpatrick: appears each receiver gets its own: https://github.com/open-telemetry/opentelemetry-collector/blob/1d38caf3b757b4d2f3db10ce9dc04187f170705f/service/builder/receivers_builder.go#L106

internal/receiver/smartagentreceiver/receiver.go
rmfitzpatrick: I think each monitor does use its own logger for the most part (std w/ identifying fields), but are scoped to their corresponding `Configure()` calls, which may be too late for purposes.  I think that a complete solution would be a good end goal, and imo should be investigated so we don't punt off the bat.  It would likely require an additional monitor method to return a wrapped logger ensuring we can access it before the monitor has started.

A short term stopgap could be to have a singleton smart agent receiver logger that is always used, overwriting the component name field of the first registered receiver logger.

bjsignalfx: Please explain "if a receiver starts and stops it's still going to be registered with logrus" a bit further. So newing a smartagent receiver hooks the supplied zap logger to the global logrus logger for redirection. zap logger instances are hooked only once throughout the life of the application. Different zap logger instances that write to the same destination will cause duplication.

internal/zaputil/logrus.go
bjsignalfx: Yes, there's no guarantee that all components share a single zap logger and we could run into duplicate log entries. We could new up a zap logger and redirect to it, I think.

internal/zaputil/logrus.go
bjsignalfx: Yes, you are correct. I was thinking since the map returns true only if the key exists and the value is true, let me be terse. Plus, think it's even better to use an empty struct instead of bool.

jrcamp: I think it's ok. `Using a map[string]bool` is pretty common when you're using it as a set so you can do `if redirects[r] { ... }`.

rmfitzpatrick:
Since this is quite local to the agent receiver it would likely make more sense as a part of its package instead of as a separate util with single consumer.

internal/receiver/smartagentreceiver/receiver.go
bjsignalfx: > I think each monitor does use its own logger for the most part (std w/ identifying fields), but are scoped to their corresponding `Configure()` calls, which may be too late for purposes. I think that a complete solution would be a good end goal, and imo should be investigated so we don't punt off the bat. It would likely require an additional monitor method to return a wrapped logger ensuring we can access it before the monitor has started.
> 
> A short term stopgap could be to have a singleton smart agent receiver logger that is always used, overwriting the component name field of the first registered receiver logger.

Wait, I only see monitors using the exported std logger!

rmfitzpatrick: Right, it's the same underlying logger, but the monitors use their own wrapper with fields.

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -43,6 +45,20 @@ type Receiver struct {
 
 var _ component.MetricsReceiver = (*Receiver)(nil)
 
+var logRedirectsVar logRedirects
+
+func init() {
+	cfg := zap.NewProductionConfig()
+	cfg.Level.SetLevel(zapcore.DebugLevel)
```
rmfitzpatrick, line 52: We want to base everything on the app logger and not our own right?

bjsignalfx: Essentially what tells me to which zap logger to redirect a logrus log entry in Fire() is the monitorType logrus entry field. So what if monitorType is unavailable? Which zap logger do I use? That's why I have that catchall logger.

internal/receiver/smartagentreceiver/log_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/sirupsen/logrus"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest"
+)
+
+func newLogRedirector(t *testing.T) *logRedirector {
+	return &logRedirector{
+		redirects:         make(map[srcLogger][]dstLogger),
+		mu:                sync.Mutex{},
+		dstCatchallLogger: zaptest.NewLogger(t),
```
rmfitzpatrick, line 17: zaptest provides observers to confirm that logged statements are emitted as expected.  Would you be able to add a more integration-level test to confirm that `Fire()` and redirection work with multiple loggers and logrus?

internal/receiver/smartagentreceiver/receiver.go
rmfitzpatrick: The smart agent shouldn't log anything until a receiver is created so I'm not sure this is a case we need to consider atm.  Having a default logger w/ debug level is probably not desired in general.

internal/receiver/smartagentreceiver/log.go
```diff
@@ -0,0 +1,167 @@
...
+
+	fields := make([]zapcore.Field, 0)
+
+	// Creating zap entry fields from logrus entry fields.
+	for k, v := range e.Data {
+		vStr := fmt.Sprintf("%v", v)
+		if k == "monitorType" {
+			monitorType = vStr
+		}
+		fields = append(fields, zapcore.Field{Key: k, Type: zapcore.StringType, String: vStr})
```
rmfitzpatrick, line 141: `zap.Any()` may be preferable to `%v`

internal/receiver/smartagentreceiver/log.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	// Creating loggerMap key from the logrus entry logger and field 'monitorType'
+	key := logrusKey{e.Logger, monitorType}
+	zapLogger := l.loggerMapValue0(key)
+	if zapLogger == nil {
+		fields = append(fields, zapcore.Field{Key: "monitorType", Type: zapcore.StringType, String: monitorType})
+		fields = append(fields, zapcore.Field{Key: "redirect_error", Type: zapcore.StringType, String: "Could not find zap logger in receiver for the monitorType. Using the catchall zap logger instead."})
```
rmfitzpatrick, line 149: I'm not sure it's necessary to flag this scenario as if it were actionable or an actual error scenario on the user's part.

internal/receiver/smartagentreceiver/log.go
```diff
@@ -0,0 +1,167 @@
...
+		zapLogger = l.catchallLogger
+	}
+
+	zapLevel := levelsMap[e.Level]
+	ce := zapLogger.Check(zapLevel, e.Message)
+	if ce == nil {
+		ce = zapLogger.Check(zap.ErrorLevel, fmt.Sprintf("Failed to redirect logrus logs at level %s. The matching zap log level %s is not enabled.", e.Level.String(), zapLevel.String()))
```
rmfitzpatrick, line 156: This logs error statements for every debug log statement if debug logging isn't enabled?

internal/receiver/smartagentreceiver/log_test.go
```diff
@@ -0,0 +1,238 @@
...
+	// Creating a typical monitor logrus entry/logger.
+	monitorLogger := logrus.WithFields(logrus.Fields{"monitorType": "monitor1"})
+
+	// Creating the monitor logrus key in the monitor receiver.
+	// The monitor type is known.
+	// The logger is assumed to be the standard logrus logger.
+	monitorLogrusKey := logrusKey{Logger: logrus.StandardLogger(), monitorType: "monitor1"}
+
+	// Checking that the monitor logger and the assumed logrus key logger are the same.
+	if monitorLogger.Logger != monitorLogrusKey.Logger {
```
rmfitzpatrick, line 39: All collector projects use [testify](https://github.com/stretchr/testify): `require.Same(t, monitorLogger.Logger, monitorLogrusKey.Logger)`

internal/receiver/smartagentreceiver/log.go
bjsignalfx: If debug logging is enabled for logrus and disabled for zap then yes.

rmfitzpatrick: Right, couldn't that flood logs with undesired errors?  Why not just not log the statements?

internal/receiver/smartagentreceiver/log.go
```diff
@@ -0,0 +1,166 @@
...
+	src.hookUnique(l)
+
+	l.loggerMap[src] = append(l.loggerMap[src], dst)
+}
+
+func (l *logrusToZap) unRedirect(src logrusKey, dst *zap.Logger) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	if l.loggerMap == nil {
```
rmfitzpatrick, line 97: If you provide a New function and assume its usage this won't be necessary, otherwise `redirect` is missing a nil check?

internal/receiver/smartagentreceiver/log.go
```diff
@@ -0,0 +1,166 @@
...
+	// Creating zap entry fields from logrus entry fields.
+	for k, v := range e.Data {
+		if k == "monitorType" {
+			monitorType = fmt.Sprintf("%v", v)
+		}
+		fields = append(fields, zap.Any(k, e.Data))
+	}
+
+	if monitorType == "" {
+		log.Panic("Expected non-zero log field monitorType")
```
rmfitzpatrick, line 148: This seems a little overzealous for something on our end imo.  Logging a warning with the contents maybe?

internal/receiver/smartagentreceiver/log.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	zapLogger := l.loggerMapValue0(logrusKey{e.Logger, monitorType})
+	if zapLogger == nil {
+		return nil
```
rmfitzpatrick, line 153: similar warning log request, otherwise we're missing an invalid configuration scenario

bjsignalfx: The thing is that we don't have a zap logger to use here!

rmfitzpatrick: I'd use either `log.Warn()` or create a catch all that is created from the first monitor's logger w/ a different name?

bjsignalfx: `log.Warn()` from which package? The standard log package does not have Warn().

rmfitzpatrick: oh, my mistake.  til

rmfitzpatrick (approved):
Thanks!

PR #68, opened 2021-01-27
Remove prometheus receiver

jrcamp:
This removes the prometheus receiver. Multiple instances of simpleprometheus should be used to monitor multiple endpoints. Only simpleprometheus is compatible with dynamically observed endpoints using observers/receiver_creator.


rmfitzpatrick:
Sure you've seen this, but config references need to be updated as well https://github.com/signalfx/splunk-otel-collector/tree/c23b586a9ba01fc36731be4b10df86b8d28eaf2b/cmd/otelcol/config/collector

PR #80, opened 2021-02-04
Update linux env file for SPLUNK_MEMORY_TOTAL_MIB

jchengsfx:


rmfitzpatrick (approved):

rmfitzpatrick:
Is this file documented?  I see the parent directory path is printed post-installation but not immediately seeing any mentions of its purpose.

jchengsfx:
> Is this file documented? I see the parent directory path is printed post-installation but not immediately seeing any mentions of its purpose.

It's mainly for users who install the deb/rpm packages manually instead of via the installer script, which will configure this file according to the command-line options.

I'll add instructions to the readme in a separate PR for installing the packages without the installer script, and how to use this file to set the env vars for the default config.

Issue #97, opened 2021-02-10
"otelcol --help" should not require env vars


jchengsfx:
```
$ otelcol --help
2021/02/10 10:29:43 main.go:197: Usage: SPLUNK_ACCESS_TOKEN=12345 SPLUNK_REALM=us0 SPLUNK_BALLAST_SIZE_MIB=684 SPLUNK_MEMORY_LIMIT_MIB=1024 SPLUNK_MEMORY_SPIKE_MIB=256 otelcol
2021/02/10 10:29:43 main.go:198: ERROR: Missing environment variable SPLUNK_MEMORY_LIMIT_MIB
```

flands:
Hmm this seems unexpected -- I will take a look

PR #98, opened 2021-02-10
tests: CollectorProcess helper

rmfitzpatrick:
These changes introduce the last basic helper for the integration testing module - a runner for the Collector.  It uses the subprocess manager from the JMX receiver, a specified or detected Collector binary path, and provided config to launch the collector process for the lifetime of a test or suite.

Also introduces the first `integration` tagged test file, but as with the previous changes I'm holding off on sourcing/running these in CI until a future PR.

PR #102, opened 2021-02-11
Ensure help CLI parameter can be passed

flands:
Addresses #97

Also add a helper function to keep main simple.

TODO: Add more tests


rmfitzpatrick:
> TODO: Add more tests

Tests could be added after https://github.com/signalfx/splunk-otel-collector/pull/98 lands

rmfitzpatrick (approved):

PR #107, opened 2021-02-17
Smart Agent receiver: Postgresql integration test

rmfitzpatrick:
These changes introduce the first integration test case for the tests module, and exercises the [native postgresql monitor](https://github.com/signalfx/signalfx-agent/blob/master/docs/monitors/postgresql.md).  Since output filtering is yet to be implemented it confirms that all metrics are received, though in the near future more restricted configs/tests could be added (@pmcollins).

requires: https://github.com/signalfx/splunk-otel-collector/pull/103 and https://github.com/signalfx/signalfx-agent/pull/1607

I will update the desired agent version once the PG13 support fix lands.

### DIFF
--- a/tests/receivers/smartagent/postgresql/testdata/client/requests.sh
+++ b/tests/receivers/smartagent/postgresql/testdata/client/requests.sh
@@ -1,4 +1,12 @@
 #!/bin/bash
+##################################################################################################
+# This file will exercise a target POSTGRES_SERVER's test_db whose schema is defined in the
+# corresponding server's initdb.d/db.sql.  It will create the necessary replication slots and make
+# a series of random insertions, updates, and deletions without end to help generate values for
+# produced metrics.
+# As implemented, test_db size will grow without bound if left is running, so this script isn't
+# currently suitable for extended demonstrations or soak test purposes.
+##################################################################################################
 
 echopsql() {
   set -x
@@ -25,7 +33,6 @@ random_float() {
 }
 
 modify_table_one() {
-x=1
 while true; do
   sleep .15
   s_one=$(random_str)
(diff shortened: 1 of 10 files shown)


jrcamp (approved):

tests/receivers/smartagent/postgresql/testdata/client/requests.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
```
tigrannajaryan, line 1: Does this file define what metrics we are going to observe as specified in all.yaml? If so can we explain what is going on here?

tests/receivers/smartagent/postgresql/testdata/server/init.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
tigrannajaryan, line 1: Please add comment to explain what this file does.

tests/receivers/smartagent/postgresql/testdata/server/initdb.d/db.sql
```diff
@@ -0,0 +1,25 @@
+drop schema if exists test_schema cascade;
```
tigrannajaryan, line 1: Please add comment to explain what this file does. It will be useful to explain what is the connection between these files and how data in one affects the data in others.

tests/receivers/smartagent/postgresql/testdata/client/requests.sh
rmfitzpatrick: Yes, it's responsible for populating most of the values for all generated metrics.  Updated to provide more info.

tigrannajaryan (approved):

PR #110, opened 2021-02-18
Add CodeCov config

flands:


.codecov.yml
```diff
@@ -0,0 +1,21 @@
...
+      default:
+        enabled: yes
+        target: 90%
+    patch:
+      default:
+        enabled: yes
+        target: 95%
+
+ignore:
+  - "tests/**/*"
```
rmfitzpatrick, line 21: Was this required?  testutils should have 90+% coverage and I don't believe there should be statements to cover in integration tests.

rmfitzpatrick (approved):

PR #111, opened 2021-02-18
Add smartagent extension

asuresh4:
This extension provides a centralized mechanism to configure things collectd that are common to all instances of the smartagent receiver. To begin with, this extension will enable users to override default config options such as "config_dir" and "bundle_dir" that are applicable for all collectd based monitors.

Also, note that it is not a requirement to configure this extension for collectd monitors to work. In such cases, the default collectd config will be applied.

### DIFF
--- a/internal/extension/smartagentextension/README.md
+++ b/internal/extension/smartagentextension/README.md
@@ -1,0 +1,24 @@
+# SignalFx Smart Agent Extension
+
+The `smartagent` extension provides a mechanism to specify config options that are not
+just specific to a single instance of the `smartagent` receiver but are applicable to
+all instances of the `smartagent` receiver.
+
+To begin with, this extension will provide a mechanism to specify config options to configure
+collectd. These options are mapped to the [collectd config options](https://docs.signalfx.com/en/latest/integrations/agent/config-schema.html#collectd)
+in the SignalFx Agent. Note that if this extension is not configured, the defaults in `smartagent`
+receiver will be used.
+
+In the below example configuration, `config_dir` and `bundle_dir` will be used for all instances
+of the `smartagent` receiver that wrap around a collectd based monitor.
+
+```yaml
+extensions:
+  smartagent:
+    collectd:
+      config_dir: /etc/collectd/collectd.conf
+      bundle_dir: /opt/bin/collectd/
+```
+
+The full list of settings exposed for this receiver are documented [here](./config.go)
+with detailed sample configurations [here](./testdata/config.yaml).
--- a/internal/extension/smartagentextension/README.md
+++ b/internal/extension/smartagentextension/README.md
@@ -15,7 +15,7 @@ of the `smartagent` receiver that wrap around a collectd based monitor.
 ```yaml
 extensions:
   smartagent:
-  bundleDir: /opt/bin/collectd/
+    bundleDir: /bundle/
     collectd:
       configDir: /etc/collectd/collectd.conf
 ```
--- a/internal/extension/smartagentextension/README.md
+++ b/internal/extension/smartagentextension/README.md
@@ -1,7 +1,7 @@
 # SignalFx Smart Agent Extension
 
 The `smartagent` extension provides a mechanism to specify config options that are not
-just specific to a single instance of the `smartagent` receiver but are applicable to
+just specific to a single instance of the [`smartagent` receiver](../../receiver/smartagentreceiver/README.md) but are applicable to
 all instances of the `smartagent` receiver.
 
 To begin with, this extension will provide a mechanism to specify config options to configure
--- a/internal/extension/smartagentextension/README.md
+++ b/internal/extension/smartagentextension/README.md
@@ -15,9 +15,9 @@ of the `smartagent` receiver that wrap around a collectd based monitor.
 ```yaml
 extensions:
   smartagent:
+  bundleDir: /opt/bin/collectd/
     collectd:
-      config_dir: /etc/collectd/collectd.conf
-      bundle_dir: /opt/bin/collectd/
+      configDir: /etc/collectd/collectd.conf
 ```
 
 The full list of settings exposed for this receiver are documented [here](./config.go)
--- a/internal/extension/smartagentextension/README.md
+++ b/internal/extension/smartagentextension/README.md
@@ -9,7 +9,7 @@ collectd. These options are mapped to the [collectd config options](https://docs
 in the SignalFx Agent. Note that if this extension is not configured, the defaults in `smartagent`
 receiver will be used.
 
-In the below example configuration, `config_dir` and `bundle_dir` will be used for all instances
+In the below example configuration, `configDir` and `bundleDir` will be used for all instances
 of the `smartagent` receiver that wrap around a collectd based monitor.
 
 ```yaml
@@ -17,7 +17,7 @@ extensions:
   smartagent:
     bundleDir: /bundle/
     collectd:
-      configDir: /etc/collectd/collectd.conf
+      configDir: /etc/collectd/
 ```
 
 The full list of settings exposed for this receiver are documented [here](./config.go)
--- a/internal/extension/smartagentextension/config.go
+++ b/internal/extension/smartagentextension/config.go
@@ -27,9 +27,8 @@ import (
 )
 
 // SmartAgentConfigProvider exposes config fields to other packages.
-// This is needed since fields such as fields such as bundleDir and
-// collectdConfig are mapped to camel case fields in the config and
-// hence are not exposed.
+// This is needed since fields such  as bundleDir and collectdConfig
+// are mapped to camel case fields in the config and hence are not exposed.
 type SmartAgentConfigProvider interface {
 	BundleDir() string
 	CollectdConfig() config.CollectdConfig
--- a/internal/extension/smartagentextension/config.go
+++ b/internal/extension/smartagentextension/config.go
@@ -53,13 +53,13 @@ type CollectdConfig struct {
 	// Collectd's log level -- info, notice, warning, or err
 	LogLevel string `mapstructure:"logLevel"`
 	// A default read interval for collectd plugins.  If zero or undefined,
-	// will default to the global agent interval.  Some collectd python
-	// monitors do not support overridding the interval at the monitor level,
-	// but this setting will apply to them.
+	// will default to 10 seconds. Some collectd python monitors do not
+	// support overridding the interval at the monitor level, but this
+	// setting will apply to them.
 	IntervalSeconds int `mapstructure:"intervalSeconds"`
 	// The local IP address of the server that the agent exposes to which
-	// collectd will send metrics.  This defaults to an arbitrary address in
-	// the localhost subnet, but can be overridden if needed.
+	// collectd will send metrics.  This defaults to 127.9.8.7, but can be
+	// overridden if needed.
 	WriteServerIPAddr string `mapstructure:"writeServerIPAddr"`
 	// The port of the agent's collectd metric sink server.  If set to zero
 	// (the default) it will allow the OS to assign it a free port.
--- a/internal/extension/smartagentextension/config.go
+++ b/internal/extension/smartagentextension/config.go
@@ -20,10 +20,11 @@ import (
 
 type Config struct {
 	configmodels.ExtensionSettings `mapstructure:",squash"`
+	BundleDir                      string         `mapstructure:"bundleDir"`
 	CollectdConfig                 CollectdConfig `mapstructure:"collectd"`
 }
 
-// Fields on this struct have a direct mapping to fields on
+// CollectdConfig fields have a direct mapping to fields on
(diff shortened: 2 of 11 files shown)


internal/extension/smartagentextension/README.md
```diff
@@ -0,0 +1,24 @@
+# SignalFx Smart Agent Extension
+
+The `smartagent` extension provides a mechanism to specify config options that are not
+just specific to a single instance of the `smartagent` receiver but are applicable to
```
rmfitzpatrick, line 4: Do you mind linking to the Smart Agent receiver readme?

internal/extension/smartagentextension/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+In the below example configuration, `config_dir` and `bundle_dir` will be used for all instances
+of the `smartagent` receiver that wrap around a collectd based monitor.
+
+```yaml
+extensions:
+  smartagent:
+    collectd:
+      config_dir: /etc/collectd/collectd.conf
+      bundle_dir: /opt/bin/collectd/
```
rmfitzpatrick, line 20: I think this should be top level in the extension since it's applicable to more than collectd and not able to be specified in its agent config: https://github.com/signalfx/signalfx-agent/blob/7bed2632de74e3e9359343d624dd6ce8d2452e07/pkg/core/config/config.go#L390

internal/extension/smartagentextension/config.go
```diff
@@ -0,0 +1,74 @@
...
+// collect.CollectdConfig in the SignalFx Agent.
+// https://github.com/signalfx/signalfx-agent/blob/v5.7.2/pkg/core/config/config.go#L342
+type CollectdConfig struct {
+	// How many read intervals before abandoning a metric. Doesn't affect much
+	// in normal usage.
+	// See [Timeout](https://collectd.org/documentation/manpages/collectd.conf.5.shtml#timeout_iterations).
+	Timeout int `mapstructure:"timeout"`
+	// Number of threads dedicated to executing read callbacks. See
+	// [ReadThreads](https://collectd.org/documentation/manpages/collectd.conf.5.shtml#readthreads_num)
+	ReadThreads int `mapstructure:"read_threads"`
```
rmfitzpatrick, line 36: The agent receiver supports the camel case struct tags from the agent to help w/ migrations.  Requiring the user to transform them here and not in the receiver seems a little counterintuitive to me.  Consistency is nice, but not entirely possible in this case, so inheriting the agent values is my preference.

internal/extension/smartagentextension/extension.go
```diff
@@ -0,0 +1,39 @@
...
+	"go.opentelemetry.io/collector/config/configmodels"
+)
+
+type smartAgentConfigExtension struct {
+}
+
+var _ component.ServiceExtension = (*smartAgentConfigExtension)(nil)
+
+func (sae *smartAgentConfigExtension) Start(_ context.Context, _ component.Host) error {
+	return nil
```
rmfitzpatrick, line 30: The extension functionality is come in a subsequent PR?

rmfitzpatrick: sry, just saw https://github.com/signalfx/splunk-otel-collector/pull/112

internal/extension/smartagentextension/README.md
asuresh4: That's a good point. I'll move this option to the receiver.

internal/extension/smartagentextension/config.go
asuresh4: Updated the config to use camel case.

jrcamp (commented):
Do we necessarily need this as a separate extension? I'm concerned about the config complexity. I was imagining for collectd the smartagent receiver would just have a global that would be lazily initialized. (ie when the first smartagent receiver starts it would start all the collectd stuff)

asuresh4:
> Do we necessarily need this as a separate extension? I'm concerned about the config complexity. I was imagining for collectd the smartagent receiver would just have a global that would be lazily initialized. (ie when the first smartagent receiver starts it would start all the collectd stuff)

@jrcamp I started out with that approach. However, I realized when there are multiple smartagent receiver instances that require collectd, users would have to provide the common collectd config/bundle directory value for each receiver. I took this approach to avoid that issue. Also, this extension is only required when one wants to override the defaults.

internal/extension/smartagentextension/README.md
rmfitzpatrick: Thanks, mind updating the example as well?

internal/extension/smartagentextension/config.go
```diff
@@ -0,0 +1,73 @@
...
+	// [WriteQueueLimitHigh](https://collectd.org/documentation/manpages/collectd.conf.5.shtml#writequeuelimithigh_highnum)
+	WriteQueueLimitHigh int `mapstructure:"writeQueueLimitHigh"`
+	// The lowest number of values in the collectd queue before which metrics
+	// begin being randomly dropped.  See
+	// [WriteQueueLimitLow](https://collectd.org/documentation/manpages/collectd.conf.5.shtml#writequeuelimitlow_lownum)
+	WriteQueueLimitLow int `mapstructure:"writeQueueLimitLow"`
+	// Collectd's log level -- info, notice, warning, or err
+	LogLevel string `mapstructure:"logLevel"`
+	// A default read interval for collectd plugins.  If zero or undefined,
+	// will default to the global agent interval.  Some collectd python
```
rmfitzpatrick, line 56: Know this is c/p, but there is no global agent level w/ the agent receiver.

internal/extension/smartagentextension/config.go
```diff
@@ -0,0 +1,73 @@
...
+	// monitors do not support overridding the interval at the monitor level,
+	// but this setting will apply to them.
+	IntervalSeconds int `mapstructure:"intervalSeconds"`
+	// The local IP address of the server that the agent exposes to which
+	// collectd will send metrics.  This defaults to an arbitrary address in
+	// the localhost subnet, but can be overridden if needed.
```
rmfitzpatrick, line 62: providing the actual ip would be more informative

tigrannajaryan:
@jrcamp is raising a good question. Let me rephrase:

Do we intend to use the extension instance in any way? If we are only using as a place where we hold the configuration how about instead we do this:
- Add collectd config settings in the smartagent receiver settings.
- The first smartagent receiver that starts initializes collectd
- All subsequent smartagent receivers check that their config is the same as for the first one and give an error if it does not.
- Documentation recommends to use yaml anchors to make it trivial to have the same config for all smartagents.

There are probably other variations of this possible.

If we _do_ intend to somehow use the extension instance then it may be worth keeping it.

Also, is there ever a need for 2 different collectd configs? I assume no, given that the code in the next PR uses only the first extension found.

Thoughts?

internal/extension/smartagentextension/factory.go
```diff
@@ -0,0 +1,66 @@
...
+		createExtension)
+}
+
+func createDefaultConfig() configmodels.Extension {
+	return &Config{
+		ExtensionSettings: configmodels.ExtensionSettings{
+			TypeVal: typeStr,
+			NameVal: string(typeStr),
+		},
+		BundleDir: os.Getenv(constants.BundleDirEnvVar),
```
rmfitzpatrick, line 44: Would it be preferable to adopt agent fallback behavior if env var not set? https://github.com/signalfx/signalfx-agent/blob/master/pkg/core/config/config.go#L181

tigrannajaryan:
I guess one other question I have is: how common is it to just use the default config? If it is vast majority of cases then we can afford if the non-default config use-case is a bit awkward to use.

rmfitzpatrick:
What I believe @jrcamp and @tigrannajaryan are proposing was my first inclination for doing this as well, but after discussing the matter with @asuresh4 and looking into the agent more, I agree this "optional" global config extension is preferable to the "use the receiver and have the first config value win" approach for at least three reasons:

1. The bundle dir will likely always need to be set for non-docker deployments whenever any non-native monitor is used.  Introducing this to the SA receiver config and having the "first win" could lead to heavily redundant values in all the SA receiver instances (potentially dozens if the fields are set in all of them) and tedious resolution and alerting logic.  To me it also signals poor modeling.  Separating the concern into a global configuration extension and throwing a warning that only a single instance should be used is far simpler to me.

2. In general, this extension provides a helpful transition point for the global agent config: https://github.com/signalfx/signalfx-agent/blob/master/docs/config-schema.md.  The vast majority of options are not going to be supported in the Splunk Collector as their functionality has migrated to SFx/sapm receivers, processors, and exporters.  The main exception to this are the procfs, etc, var, run, and sys path fields as well as collectd.  None of the monitors that use these system path fields actually allow them to be set in their own config and error out.  The env var set by the global config mechanism is required as implemented.  Altering all the applicable SA logic to allow the "first wins" receiver config approach seems less than optimal than a well functioning config extensions.

3. Migration and maintenance - It's quite likely we'll have a migration assistant in the near future that will need to translate global SA configs to equivalent Splunk Collector configs.  Having this operate by selecting an effectively random monitor to SA receiver config, or by heavy redundancy in applying to all monitor configs may be brittle and undesirable overall.

rmfitzpatrick (approved):
LGTM w/ example fields and comment nits updated.

internal/extension/smartagentextension/factory.go
asuresh4: Added the same behavior as in the agent.

asuresh4:
> LGTM w/ example fields and comment nits updated.

@rmfitzpatrick I made some updates to correctly handle camel case config values. The earlier update I made resulted in an invalid config since the keys were camel case. Can you take another look? (also updated #112 to make use of these latest updates)

internal/extension/smartagentextension/README.md
```diff
@@ -0,0 +1,24 @@
...
+in the SignalFx Agent. Note that if this extension is not configured, the defaults in `smartagent`
+receiver will be used.
+
+In the below example configuration, `config_dir` and `bundle_dir` will be used for all instances
+of the `smartagent` receiver that wrap around a collectd based monitor.
+
+```yaml
+extensions:
+  smartagent:
+  bundleDir: /opt/bin/collectd/
```
rmfitzpatrick, line 18: nit indentation

internal/extension/smartagentextension/config.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/signalfx/defaults"
+	"github.com/signalfx/signalfx-agent/pkg/core/config"
+	"github.com/spf13/viper"
+	"go.opentelemetry.io/collector/config/configmodels"
+	"gopkg.in/yaml.v2"
+)
+
+// SmartAgentConfigProvider exposes config fields to other packages.
+// This is needed since fields such as fields such as bundleDir and
```
rmfitzpatrick, line 30: "This is needed since fields such as ~~fields such as~~ bundleDir"

internal/extension/smartagentextension/config.go
```diff
@@ -0,0 +1,123 @@
...
+	err = defaults.Set(&collectdConfig)
+	if err != nil {
+		return fmt.Errorf("failed setting collectd config defaults: %w", err)
+	}
+
+	extensionCfg.collectdConfig = collectdConfig
+	return nil
+}
+
+// Copied from smartagent receiver.
```
rmfitzpatrick, line 101: Though there's an adage to break out helpers on the 3rd usage, I prefer on the 2nd to prevent redundancy when possible. May be time for an `internal/common`?

asuresh4: Agreed. Will open a separate PR for that though. To reduce the scope of changes in this one.

internal/extension/smartagentextension/README.md
```diff
@@ -0,0 +1,24 @@
...
+The `smartagent` extension provides a mechanism to specify config options that are not
+just specific to a single instance of the [`smartagent` receiver](../../receiver/smartagentreceiver/README.md) but are applicable to
+all instances of the `smartagent` receiver.
+
+To begin with, this extension will provide a mechanism to specify config options to configure
+collectd. These options are mapped to the [collectd config options](https://docs.signalfx.com/en/latest/integrations/agent/config-schema.html#collectd)
+in the SignalFx Agent. Note that if this extension is not configured, the defaults in `smartagent`
+receiver will be used.
+
+In the below example configuration, `config_dir` and `bundle_dir` will be used for all instances
```
rmfitzpatrick, line 12: camel case plz

internal/extension/smartagentextension/README.md
```diff
@@ -0,0 +1,24 @@
...
+of the `smartagent` receiver that wrap around a collectd based monitor.
+
+```yaml
+extensions:
+  smartagent:
+    bundleDir: /bundle/
+    collectd:
+      configDir: /etc/collectd/collectd.conf
```
rmfitzpatrick, line 20: `/etc/collectd` since just the directory right?

PR #112, opened 2021-02-18
smartagentreceiver: Add support for collectd based monitors

asuresh4:
Depends on #111 

Updates required to configure collectd singleton when the smartagent receiver is configured to run collectd based monitors.

The receiver picks up the default collectd config from the smartagent extension package. To override default values in the collectd config, one would configure the smartagent extension and the receiver will configure the collectd singleton based on those overrides.


rmfitzpatrick:
Do you mind updating the base branch to sa-extension to make this easier to review?

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +145,63 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+	// Note, that this receiver has to configure main collectd even for collectd/custom monitor,
+	// despite the fact that, that monitor stands up its own instance of collectd to prevent this
+	// panic "Main collectd instance should not be accessed before being configured".
+	if r.config.monitorConfig.MonitorConfigCore().IsCollectdBased() && !configuredCollectd {
+		r.setUpCollectdConfig(extensions)
+		if err := collectd.ConfigureMainCollectd(r.getCollectdConfig()); err != nil {
+			return nil, err
+		}
+		configuredCollectd = true
+	}
```
tigrannajaryan, line 159: Can we extract this as a func? I think it will help to understand why we need this, which is not quite clear. Does this initialize the a singleton collectd instance? (I read the comment, but it is still not completely clear to me what are we doing here).
From what I see the mutex+bool flag is used to execute this once, right? Maybe use once.Do instead to make it clearer? (https://golang.org/pkg/sync/#Once)

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +145,63 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+
 	return monitor, nil
 }
+
+func (r *Receiver) setUpCollectdConfig(extensions map[configmodels.Extension]component.ServiceExtension) {
+	f := smartagentextension.NewFactory()
+	defaultCfg := f.CreateDefaultConfig().(*smartagentextension.Config)
+	r.config.collectdConfig = defaultCfg.CollectdConfig
```
tigrannajaryan, line 167: Add comment that we fallback to the default if extension is not found. (That's what we do, right?)

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +145,63 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+		r.config.collectdConfig = cfg.CollectdConfig
+
+		// If there are multiple extensions configured, pick the first one. Ideally,
+		// there would only be one extension.
+		break
+	}
+}
+
+// Returns a configuration for collectd. Defaults provided by the receiver will overridden by
+// options specified on the smartagent extension.
```
tigrannajaryan, line 188: Which "defaults" are overridden? How does a receiver provide defaults? Can you please point to the code that does it?

internal/receiver/smartagentreceiver/receiver_linux_test.go
```diff
@@ -0,0 +1,174 @@
...
+		monitorConfig: &uptime.Config{
+			MonitorConfig: config.MonitorConfig{
+				Type: "collectd/uptime",
+			},
+		},
+	}
+	r := NewReceiver(zap.NewNop(), cfg, consumertest.NewMetricsNop())
+
+	// Test whether default config is correctly loaded from smartagent extension factory.
+	// Note, that the below invocation of Start method is bound to fail since collectd
```
tigrannajaryan, line 59: Can this succeed if the collectd happens to be installed on the machine? Would that be a problem or we don't care if it succeeds?

tigrannajaryan:
Do we have any automated tests that verify monitors that verify how everything works when collectd is present? From what I see that tests that we have only verify behavior when collectd is not present.

rmfitzpatrick:
> Do we have any automated tests that verify monitors that verify how everything works when collectd is present? From what I see that tests that we have only verify behavior when collectd is not present.

This is on the roadmap for the receiver integration tests since it will require the agent bundle to be available during test runs.  Ideally it would be similar to the proposed https://github.com/signalfx/splunk-otel-collector/pull/107 but utilize a collectd (and then sfxcollectd) integration and assert against all expected metrics.

tigrannajaryan:
LGTM, but I believe @rmfitzpatrick is in a better position to approve.

internal/receiver/smartagentreceiver/receiver.go
asuresh4: This comment is stale. The `getCollectdConfig` method just returns `*config.CollectdConfig` derived from the config that's setup in `setUpCollectdConfig`. Will update the comment.

internal/receiver/smartagentreceiver/receiver_linux_test.go
asuresh4: If collectd is already running on the machine when the receiver starts up, it will just create a new instance of collectd.

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +145,63 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
 		output.AddExtraDimension(k, v)
 	}
 
+	collectdLock.Lock()
+	defer collectdLock.Unlock()
+	// Note, that this receiver has to configure main collectd even for collectd/custom monitor,
+	// despite the fact that, that monitor stands up its own instance of collectd to prevent this
+	// panic "Main collectd instance should not be accessed before being configured".
+	if r.config.monitorConfig.MonitorConfigCore().IsCollectdBased() && !configuredCollectd {
+		r.setUpCollectdConfig(extensions)
```
rmfitzpatrick, line 154: Several monitors require specific env vars to be set by the final config to function.  Mind adding the core of https://github.com/signalfx/signalfx-agent/blob/master/pkg/core/config/config.go#L209 to this step as well?

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +145,63 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+	for c := range extensions {
+		if c.Type() != f.Type() {
+			continue
+		}
+
+		cfg, ok := c.(*smartagentextension.Config)
+		if !ok {
+			continue
+		}
+		r.config.collectdConfig = cfg.CollectdConfig
```
rmfitzpatrick, line 179: I think this should operate at individual fields instead of writing over all the defaults

asuresh4: For any field that does not have a user-specified value, the defaults would still correctly hold since the smartagent extension sets it, right?

internal/receiver/smartagentreceiver/receiver.go
asuresh4: Yes,  `collectd.ConfigureMainCollectd` should only be called once. This call sets up the collectd singleton. Updated to use `sync.Once`

internal/receiver/smartagentreceiver/receiver.go
rmfitzpatrick: Sry, you're right.  I was conflating validation and custom unmarshalling.

internal/receiver/smartagentreceiver/receiver.go
asuresh4: Updated to set `SIGNALFX_BUNDLE_DIR` and `JAVA_HOME` for now.

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +147,74 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
 
-	return monitor, nil
+	// Note, that this receiver has to configure main collectd even for collectd/custom monitor,
+	// despite the fact that, that monitor stands up its own instance of collectd to prevent this
+	// panic "Main collectd instance should not be accessed before being configured".
+	if r.config.monitorConfig.MonitorConfigCore().IsCollectdBased() {
+		configureCollectdOnce.Do(func() {
+			r.logger.Info("Configuring collectd")
+			r.setUpCollectdConfig(extensions)
+			r.setUpEnvironment()
```
rmfitzpatrick, line 157: I think `setUpEnvironment()` needs to happen even when collectd doesn't need to be started (which has a dependency on `setUpCollectdConfig()`) for sfxcollectd and gopsutil monitors.  Maybe via a new `configureEnvironmentOnce()`?

asuresh4: Yes, that's a good point. Will decouple the setup of `SmartConfigProvider` from the collectd use-case.

asuresh4: Updated.

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +148,79 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+	}
+
+	return monitor, err
+}
+
+func configureMainCollectd(collectdConfig *config.CollectdConfig) error {
+	return collectd.ConfigureMainCollectd(collectdConfig)
+}
+
+func (r *Receiver) setUpSmartConfigProvider(extensions map[configmodels.Extension]component.ServiceExtension) {
```
rmfitzpatrick, line 175: would prefer `setUpSmartAgentConfigProvider` or `setUpGlobalConfigProvider`

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +148,79 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+	// If smartagent extension is not configured, use the default config.
+	f := smartagentextension.NewFactory()
+	defaultCfg := f.CreateDefaultConfig().(smartagentextension.SmartAgentConfigProvider)
+	r.config.saConfigProvider = defaultCfg
```
rmfitzpatrick, line 179: Would it be better to make this a synchronized global instead of associated with the config for each receiver?  They aren't logically bound and config's are necessarily catch-alls.

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +148,79 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+		r.config.saConfigProvider = cfgProvider
+
+		// If there are multiple extensions configured, pick the first one. Ideally,
+		// there would only be one extension.
+		break
+	}
+}
+
+// getCollectdConfig returns a *config.CollectdConfig for r.config.collectdConfig.
+func (r *Receiver) getCollectdConfig() *config.CollectdConfig {
```
rmfitzpatrick, line 201: Does this need to be in the receiver?  Seems to be an extension responsibility (and potentially unnecessary?)

internal/receiver/smartagentreceiver/receiver.go
asuresh4: Fixed, that was not intentional.

internal/receiver/smartagentreceiver/receiver.go
asuresh4: Yeah, that's a fair point. Updated per suggestion.

internal/receiver/smartagentreceiver/receiver.go
asuresh4: You're right. This should be moved to the extension. I need to update the extension's default to set `BundleDir`, `HasGenericJMXMonitor` and `IntervalSeconds` appropriately. I'll open a subsequent PR for updating the extension, removing this method.

asuresh4: Addressed in #113

rmfitzpatrick: Wouldn't it be easier to update the defaults now and just call `configureMainCollectd(saConfigProvider.getCollectdConfig())` instead of an intermediary phase/commit?

rmfitzpatrick: sry, we raced each other...

internal/extension/smartagentextension/config.go
```diff
@@ -93,7 +93,19 @@ func customUnmarshaller(componentViperSection *viper.Viper, intoCfg interface{})
 		return fmt.Errorf("failed setting collectd config defaults: %w", err)
 	}
 
+	// The default on CollectdConfig is 0, use the default if this is the case.
+	if collectdConfig.IntervalSeconds == 0 {
+		collectdConfig.IntervalSeconds = defaultIntervalSeconds
+	}
+
+	// If the HasGenericJMXMonitor is unset and not user specified, set it to true.
+	if !collectdConfig.HasGenericJMXMonitor && collectdSettings["hasgenericjmxmonitor"] == nil {
```
rmfitzpatrick, line 102: If `hasgenericjmxmonitor` were in the provided config I believe the unmarshal strict would have errored since the yaml tag is set to omit.  I'm not 100% this field is necessary for us to worry about and think the collectd manager maintains it as genericjmx monitors call their `Configure()` methods.

rmfitzpatrick: at any rate, we can address it as a genericjmx integration test is added

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +149,57 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+		configureCollectdOnce.Do(func() {
+			r.logger.Info("Configuring collectd")
+			err = configureMainCollectd(saConfigProvider.CollectdConfig())
+		})
+	}
+
+	return monitor, err
+}
+
+func configureMainCollectd(collectdConfig config.CollectdConfig) error {
```
rmfitzpatrick, line 172: not sure this wrapper is necessary

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +149,57 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+
+		cfgProvider, ok := c.(smartagentextension.SmartAgentConfigProvider)
+		if !ok {
+			continue
+		}
+		saConfigProvider = cfgProvider
+
+		// If there are multiple extensions configured, pick the first one. Ideally,
+		// there would only be one extension.
+		break
```
rmfitzpatrick, line 197: Would it be better experience to iterate through all of them and log a warning when multiples are provided?

internal/receiver/smartagentreceiver/receiver.go
asuresh4: removed.

internal/receiver/smartagentreceiver/receiver.go
asuresh4: Updated.

internal/extension/smartagentextension/config.go
asuresh4: 👍

rmfitzpatrick: Mind removing this block for the PR?

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +149,62 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+	// Configure SmartAgentConfigProvider to gather any global config overrides and
+	// set required envs.
+	configureEnvironmentOnce.Do(func() {
+		r.setUpSmartAgentConfigProvider(extensions)
+		setUpEnvironment()
+	})
+
+	// Note, that this receiver has to configure main collectd even for collectd/custom monitor,
+	// despite the fact that, that monitor stands up its own instance of collectd to prevent this
+	// panic "Main collectd instance should not be accessed before being configured".
```
rmfitzpatrick, line 161: nit, not sure this comment is necessary or too informative since we don't really do any monitor type-based logic.

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -135,5 +149,62 @@ func (r *Receiver) createMonitor(monitorType string) (interface{}, error) {
...
+		cfgProvider, foundAtLeastOne = c.(smartagentextension.SmartAgentConfigProvider)
+		if !foundAtLeastOne {
+			continue
+		}
+		saConfigProvider = cfgProvider
+		r.logger.Info("Smart Agent Config provider configured", zap.String("extension_name", c.Name()))
+	}
+
+	if multipleSAExtensions {
+		r.logger.Warn("multiple smartagent extensions found, using the first one encountered")
```
rmfitzpatrick, line 202: We should probably keep track of and log the name of the extension we're using.

internal/receiver/smartagentreceiver/receiver_test.go
```diff
@@ -242,3 +264,102 @@ func TestConfirmStartingReceiverWithInvalidMonitorInstancesDoesntPanic(t *testin
...
+
+func TestSmartAgentConfigProviderOverrides(t *testing.T) {
+	t.Cleanup(cleanUp)
+	cfg := newConfig("valid", "cpu", 1)
+	observedLogger, logs := observer.New(zapcore.WarnLevel)
+	logger := zap.New(observedLogger)
+	r := NewReceiver(logger, cfg, consumertest.NewMetricsNop())
+	host := &mockHost{
+		smartagentextensionConfig:  getSmartAgentExtensionConfig(t),
+		smartagentextensionConfig2: getSmartAgentExtensionConfig(t),
```
rmfitzpatrick, line 276: Would it make more sense to have two extension definitions in the config file with different values and verify we're using the one we expect?

internal/receiver/smartagentreceiver/receiver.go
asuresh4: There's an info log above in the for loop. Would that not sufficient?

rmfitzpatrick: Right, I'm referring to the "using the first one encountered" statement, which is a bit of an implementation detail

internal/receiver/smartagentreceiver/testdata/extension_config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  examplereceiver:
+
+processors:
+  exampleprocessor:
+
+exporters:
+  exampleexporter:
+
+service:
+  extensions: [smartagent/partial_settings]
```
rmfitzpatrick, line 29: missing the extra one?

rmfitzpatrick: oh, guessing this mechanism doesn't propagate to the test host...  disregard

internal/receiver/smartagentreceiver/receiver_test.go
```diff
@@ -242,3 +264,100 @@ func TestConfirmStartingReceiverWithInvalidMonitorInstancesDoesntPanic(t *testin
...
+func (m *mockHost) GetFactory(kind component.Kind, componentType configmodels.Type) component.Factory {
+	return nil
+}
+
+func (m *mockHost) GetExtensions() map[configmodels.Extension]component.ServiceExtension {
+	randomExtensionConfig := &healthcheckextension.Config{}
+	return map[configmodels.Extension]component.ServiceExtension{
+		m.smartagentextensionConfig:      getExtension(smartagentextension.NewFactory(), m.smartagentextensionConfig),
+		randomExtensionConfig:            getExtension(healthcheckextension.NewFactory(), randomExtensionConfig),
+		m.smartagentextensionConfigExtra: nil,
```
rmfitzpatrick, line 349: Curious why this is nil and not using a created extension from `smartagentextensionConfigExtra`?

internal/receiver/smartagentreceiver/receiver.go
asuresh4: Updated.

internal/receiver/smartagentreceiver/testdata/extension_config.yaml
asuresh4: yeah, was about say but updated regardless.

internal/receiver/smartagentreceiver/receiver_test.go
asuresh4: The receiver uses only the config of the extension, so we don't really about the actual extension itself.

internal/receiver/smartagentreceiver/receiver_test.go
```diff
@@ -67,6 +83,7 @@ func newConfig(nameVal, monitorType string, intervalSeconds int) Config {
 }
 
 func TestSmartAgentReceiver(t *testing.T) {
+	t.Cleanup(cleanUp)
```
rmfitzpatrick, line 86: Not requesting, but noting testify provides fixtures in its suite helper: https://pkg.go.dev/github.com/stretchr/testify/suite#BeforeTest

rmfitzpatrick (approved):
lgtm with https://github.com/signalfx/splunk-otel-collector/pull/112/files#r580296650 removed

internal/extension/smartagentextension/config.go
asuresh4: Missed this comment initially. Updated.

PR #114, opened 2021-02-19
Update agent linux config for deb/rpm

jchengsfx:
- Move the separate deb/rpm config to `cmd/otelcol/config/collector/agent_config_linux.yaml` so that there is only a single file
- Update the config at deb/rpm build time so that the default values can be explicitly overridden by separate environment variables


internal/buildscripts/packaging/fpm/common.sh
```diff
@@ -43,6 +43,17 @@ create_user_group() {
         sudo useradd --system --user-group --no-create-home --shell /sbin/nologin $SERVICE_USER
 }
 
+update_config() {
+    local config_path="$1"
+
+    # update the following in the deb/rpm config to use these env vars instead
+    # so that they can be overridden by the installer script options
```
jrcamp, line 50: Why do these need to be updated here instead of changing configs?

jchengsfx: Since the endpoints in the config all default to URLs determined by the required `SPLUNK_REALM` env var, e.g. `https://ingest.${SPLUNK_REALM}.signalfx.com`, I wanted to leave the defaults as-is in the agent config so that users don't have to set new env vars, like `SPLUNK_TRACE_URL` or `SPLUNK_API_URL`.

But for the installer script, I wanted to easily override these defaults via the new env vars based on the installation parameters if the user needs to explicitly set the URLs to something other than `https://ingest.${SPLUNK_REALM}.signalfx.com` (like when testing in lab which have different URL formats than prod realms).

Alternatively, I could just change the agent config directly and add comments for the new required env vars?

Or is it possible to set defaults for env vars in the config like with bash, e.g.
`${SPLUNK_API_URL:-https://api.$SPLUNK_REALM.signalfx.com}`?

PR #117, opened 2021-02-22
Simplify and automate memory limiter

flands:
While the previous implementation was functionally complete, it was
unnecessarily complex. Instead of requiring that a memory environment
variable be set and allowing customization via every possible memory
limiting setting, instead offer a sane default and allow the most common
configuration changes via environment variable.

- If `SPLUNK_MEMORY_TOTAL_MIB` is not set, default to `512`. This should
  be sufficient the majority of use-cases from getting started to agent
  deployment as a basic gateway setup.
- Only allow the ballast and limit to be configured via environment
  variables. It is still possible to use a custom config given env vars
  are expanded.
- Increase the minimum thresholds. The limit must be at least 100MiB and
  the ballast must be at least 33MiB.

Technically, this is a breaking change as environment variable passed
may no longer be respected.

In addition, cleanup default config. Most notably, no longer enable the
`opencensus` receiver by default -- another breaking change.


cmd/otelcol/config/collector/otlp_config_linux.yaml
```diff
@@ -34,32 +34,29 @@ processors:
...
     check_interval: 2s
-    limit_percentage: ${SPLUNK_MEMORY_LIMIT_PERCENTAGE}
-    spike_limit_percentage: ${SPLUNK_MEMORY_SPIKE_PERCENTAGE}
+    limit_mib: ${SPLUNK_MEMORY_LIMIT_MIB}
 
 exporters:
   # Traces
-  otlp:
+  otlphttp:
     endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace/otlp"
```
rmfitzpatrick, line 52: should this be `traces_endpoint` to avoid it becoming `v2/trace/otlp/v1/traces`?  https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlphttpexporter#otlphttp-exporter

cmd/otelcol/config/collector/splunk_config_linux.yaml
```diff
@@ -34,20 +34,17 @@ processors:
...
   # Configuration is based on the amount of memory allocated to the collector.
-  # The configuration below assumes 2GB of memory. In general, the ballast
-  # should be set to 1/3 of the collector's memory, the limit should be 90% of
-  # the collector's memory up to 2GB, and the spike should be 25% of the
-  # collector's memory up to 2GB. The simplest way to specify the ballast size is
-  # set the value of SPLUNK_BALLAST_SIZE_MIB env variable. This will overrides
+  # In general, the ballast should be set to 1/3 of the collector's memory, the limit
+  # should be 90% of the collector's memory up to 2GB. The simplest way to specify the
+  # ballast size is set the value of SPLUNK_BALLAST_SIZE_MIB env variable. This will override
   # the value of --mem-ballast-size-mib command line flag. If SPLUNK_BALLAST_SIZE_MIB
```
rmfitzpatrick, line 40: Not introduced here or blocking, but don't cli args usually take priority over env vars since they are the most directly applied option for a given process and not potentially inherited?

cmd/otelcol/config/collector/otlp_config_linux.yaml
flands: nice catch -- yes!

cmd/otelcol/config/collector/splunk_config_linux.yaml
flands: Hmm this comment is wrong -- CLI does take priority. See cmd/otelcol/main.go line 111

cmd/otelcol/config/collector/otlp_config_linux.yaml
```diff
@@ -34,32 +34,29 @@ processors:
...
-  # The configuration below assumes 2GB of memory. In general, the ballast
-  # should be set to 1/3 of the collector's memory, the limit should be 90% of
-  # the collector's memory up to 2GB, and the spike should be 25% of the
-  # collector's memory up to 2GB. The simplest way to specify the ballast size is
-  # set the value of SPLUNK_BALLAST_SIZE_MIB env variable. This will overrides
+  # In general, the ballast should be set to 1/3 of the collector's memory, the limit
+  # should be 90% of the collector's memory up to 2GB. The simplest way to specify the
+  # ballast size is set the value of SPLUNK_BALLAST_SIZE_MIB env variable. This will override
   # the value of --mem-ballast-size-mib command line flag. If SPLUNK_BALLAST_SIZE_MIB
   # is not defined then --mem-ballast-size-mib command line flag must be manually specified.
```
flands, line 41:
```suggestion
  # ballast size is set the value of SPLUNK_BALLAST_SIZE_MIB env variable. Alternatively, the
  # --mem-ballast-size-mib command line flag can be passed and take priority.
```

cmd/otelcol/config/collector/splunk_config_linux.yaml
```diff
@@ -34,20 +34,17 @@ processors:
...
   # is not defined then --mem-ballast-size-mib command line flag must be manually specified.
```
flands, line 41:
```suggestion
  # ballast size is set the value of SPLUNK_BALLAST_SIZE_MIB env variable. Alternatively, the
  # --mem-ballast-size-mib command line flag can be passed and take priority.
```

cmd/otelcol/config/collector/otlp_config_linux.yaml
flands, line 52:
```suggestion
    traces_endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace/otlp"
```

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -403,30 +403,30 @@ If access_token is not provided, it will prompted for on stdin.
```
rmfitzpatrick, line 403: Not introduced here, but mind correcting `it will prompted for on stdin`?

rmfitzpatrick (approved):

mstumpfx (approved):

jrcamp (approved):

PR #120, opened 2021-02-22
Convert Smart Agent dimensions to MetadataUpdates

rmfitzpatrick:
These changes introduce an unused Smart Agent dimension to contrib metadata update converter that will be used in upcoming changes to propagate dimension updates to the configured SFx exporter.

Staging until next Collector Contrib release, and dep versions will be updated to latest from* https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/5de076e9773bdb7617b544a57fa0a4b848cec92c.

PR #121, opened 2021-02-22
Smart Agent: Send dimension updates

rmfitzpatrick:
These changes adopt the dimension to MetadataUpdate converter introduced in https://github.com/signalfx/splunk-otel-collector/pull/120 and send the resulting updates to an applicable metadata exporter if the provided exporter component implements its interface.  This presumes the next consumer is an SFx exporter and will not work w/ intermediary processors.  More robust exporter selection like that in the k8s cluster receiver will be proposed in a subsequent PR.

Also updates a required test case and another to allow `-count` flag (previously tested extension has single usage limit that leads to panic).

### DIFF
--- a/internal/receiver/smartagentreceiver/output.go
+++ b/internal/receiver/smartagentreceiver/output.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"time"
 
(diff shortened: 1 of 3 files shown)


tigrannajaryan (approved):
LGTM as temporary state until the next PR makes this work even when there are processors.

asuresh4 (approved):

PR #124, opened 2021-02-23
WIP: smartagent: Allow specifying metadata exporters for dimension updates

rmfitzpatrick:
These changes build upon https://github.com/signalfx/splunk-otel-collector/pull/120 but add a `metadataClients` config field to denote the SFx exporter to use to send dimension updates.  If a client array is isn't provided the current behavior of defaulting to the nextConsumer if possible remains.

Camelcasing is inherited by the Agent config conventions to avoid mixing it w/ snake case.

I am not happy with the configuration experience here and hope something comes out of https://github.com/open-telemetry/opentelemetry-collector/issues/2536 or the existing sdk has features that can improve it I wasn't able to find.  It's unfortunate that we'd need to require the user to specify the exporter they're using without otherwise having to make the assumption that all SFx exporters are desired and in our pipeline(s).

I am opening for a review now and will add a more robust test to get coverage up.

### DIFF
--- a/internal/receiver/smartagentreceiver/README.md
+++ b/internal/receiver/smartagentreceiver/README.md
@@ -17,10 +17,17 @@ and associated [Observer extensions](https://github.com/open-telemetry/opentelem
 should be used.
 1. All metric content replacement and transformation rules should utilize existing
 [Collector processors](https://github.com/open-telemetry/opentelemetry-collector/blob/master/processor/README.md).
+1. Monitors with [dimension property and tag update
+functionality](https://dev.splunk.com/observability/docs/datamodel#Creating-or-updating-custom-properties-and-tags)
+require an associated `metadataClients` field that references the name of the SignalFx exporter you are using in your
+pipeline.
 
 Example:
 
 ```yaml
+exporters:
+  signalfx:
+
 receivers:
   smartagent/haproxy:
     type: haproxy
@@ -30,6 +37,7 @@ receivers:
     type: postgresql
     host: mypostgresinstance
     port: 5432
+    metadataClients: [signalfx]
 ```
 
 The full list of settings exposed for this receiver are documented for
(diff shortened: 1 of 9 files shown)


internal/receiver/smartagentreceiver/README.md
```diff
@@ -30,6 +37,7 @@ receivers:
     type: postgresql
     host: mypostgresinstance
     port: 5432
+    metadataClients: [signalfx]
```
tigrannajaryan, line 40: It's unfortunate that we need to use a different casing than what is used by Otelcol. This should be temporary. Once we settle on refactoring MetadataUpdates into Entity events this should be removed completely. If that happens then we will not need GetPipelines() either.

internal/receiver/smartagentreceiver/config.go
```diff
@@ -77,6 +80,23 @@ func mergeConfigs(componentViperSection *viper.Viper, intoCfg interface{}) error
...
+				} else {
+					return errMetadataClientValue
+				}
+			}
+			receiverCfg.MetadataClients = &mClients
+		} else {
+			return errMetadataClientValue
+		}
+		delete(allSettings, "metadataclients")
+	}
```
tigrannajaryan, line 98: What is this code for?

tigrannajaryan (approved):

internal/receiver/smartagentreceiver/README.md
rmfitzpatrick: That was my one of my thoughts (this is the y of the x/y problem).  It would be even be preferable to send meaningful datapoints w/ special labels and no metric value that can be converted to dimension updates if entities don't land.

internal/receiver/smartagentreceiver/config.go
rmfitzpatrick: Custom unmarshalling of `metadataClients` to a `*[]string`.  In viper unmarshallers `allSettings` is a `map[string]interface{}`.

rmfitzpatrick: Will add comment

PR #125, opened 2021-02-24
receiver/smartagent: Use default logger when monitorType field does not exist

asuresh4:
Currently the receiver logs a warning whenever `monitorType` is not present as a field on the underlying logrus logs. Use the default logger instead.

### DIFF
--- a/internal/receiver/smartagentreceiver/log_test.go
+++ b/internal/receiver/smartagentreceiver/log_test.go
@@ -77,9 +77,7 @@ func TestRedirectMonitorLogs(t *testing.T) {
 		defaultZapLogger, defaultLogs := newObservedLogs(zapLevel)
 
 		// Simulating logrus to zap redirections in receiver.
-		logToZap := newLogrusToZap(func() *zap.Logger {
-			return defaultZapLogger
-		})
+		logToZap := newLogrusToZap(func() *zap.Logger { return defaultZapLogger })
 		logToZap.redirect(monitor1LogrusKey, monitor1ZapLogger)
 		logToZap.redirect(monitor2LogrusKey, monitor2ZapLogger)
 		logToZap.redirect(monitor3LogrusKey, monitor3ZapLogger)
@@ -326,7 +324,7 @@ func TestLogrusToZapLevel(t *testing.T) {
 			logger, err := test.zapConfig.Build()
 			require.NoError(t, err)
 
-			logToZap := newLogrusToZap(func() *zap.Logger { return logger })
+			logToZap := newLogrusToZap(loggerProvider(logger.Core()))
 			defaultLoggerCore := logToZap.defaultLogger.Core()
 
 			for _, level := range test.levelsEnabled {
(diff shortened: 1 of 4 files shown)


internal/receiver/smartagentreceiver/output.go
```diff
@@ -89,7 +89,7 @@ func getMetadataExporters(config Config, host component.Host, nextConsumer consu
 	}
 
 	if len(exporters) == 0 {
-		logger.Warn("no dimension updates are possible as no valid metadataClients have been provided and next pipeline component isn't a MetadataExporter")
```
asuresh4, line 92: @rmfitzpatrick seems like this results in a warning for any receiver that does not use metadata updates. I changed the logging level to `Info`. But do we really need to log this out?

rmfitzpatrick: Good call, debug makes sense to me

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -60,7 +60,7 @@ var (
 )
 
 func init() {
-	rusToZap = newLogrusToZap()
+	rusToZap = newLogrusToZap(nil)
```
rmfitzpatrick, line 63: I'm still not sure this init block is necessary and that we shouldn't base our default logger on the first receiver instances.  We aren't complying w/ app log level otherwise.

I can take a crack at it in a later PR if you don't see a quick way/want to go for it.

asuresh4: I wasn't aware the current approach is something we're looking to update. So, we would ideally use the first receiver logger's settings for any subsequent receivers?

rmfitzpatrick: There was some back and forth in the initial PR, but the experience needs improvement imo.  thanks for taking care of this

asuresh4: I played around a bit with this and I wasn't able to find a way to use the first receiver's settings to create the default logger as the receiver name field on the logger cannot be updated (relevant info: https://github.com/uber-go/zap/issues/765#issuecomment-558425090). However, like you suggested offline, I've set the level of the default logger based on the settings from the first receiver instance.

internal/receiver/smartagentreceiver/log_test.go
```diff
@@ -187,6 +202,144 @@ func TestRedirectShouldUniquelyAddHooksToLogrusKeyLogger(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			test.zapConfig.Encoding = "json"
+			logger, err := test.zapConfig.Build()
+			require.NoError(t, err)
+
+			logToZap := newLogrusToZap(func() *zap.Logger { return logger })
```
rmfitzpatrick, line 329: Why aren't we using the real `loggerProvider` with `newDefaultLoggerCfg` call?  I think otherwise this test confirms that `logToZap.defaultLogger` is the same logger instance, which we don't want.

rmfitzpatrick: I think this would also help the patch coverage

asuresh4: Ah good catch, fixed.

rmfitzpatrick (approved):
thanks!

PR #126, opened 2021-02-25
Update logging config

flands:
SFx events and fluent to hec need to be separate pipelines


jrcamp (approved):
I didn't know this file existed, will add to list of config files to deduplicate.

PR #129, opened 2021-02-25
Add default config for popular sources

flands:
The goal is to improve the initial experience for users by providing
sane defaults. The user may need to update the paths depending on their
environment and configuration. Attempts to properly configure
permissions.

Limitations:

- Paths may not be accurate for all environments
- Parsing and multiline support not validated
- File rotation not configured so wildcard source paths may not work as
  expected


jchengsfx:
My only concern is that the `td-agent` user/group will require permissions to access these sources if they exist.  Is this something we should handle (i.e. in the installer script), or something that the user should handle themselves?

flands:
@jchengsfx I think we have three options:
- attempt to use https://docs.fluentd.org/deployment/linux-capability
- attempt to add to relevant groups
- leave it to the user

flands:
@jchengsfx I believe your concern is addressed -- PTAL

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -386,12 +389,28 @@ install() {
...
+
+  if [ -f "/opt/td-agent/bin/fluent-cap-ctl" ]; then
+    td-agent-gem install capng_c
+    /opt/td-agent/bin/fluent-cap-ctl --add dac_read_search -f /opt/td-agent/bin/ruby
+    /opt/td-agent/bin/fluent-cap-ctl --add dac_override -f /opt/td-agent/bin/ruby
+  else
+    getent group adm >/dev/null 2>&1
+    if [ $? == 0 ]; then
+      usermod -a -G adm td-agent
+    fi
```
jchengsfx, line 412: Since we have the `set -e` option, this will cause the script to fail and exit immediately if the `getent` command returns non-zero.  This block should be changed to:
```
if getent group adm >/dev/null 2>&1; then
  usermod -a -G adm td-agent
fi
```

jchengsfx (changes requested):

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -373,6 +373,9 @@ install() {
         install_apt_package "td-agent" "$td_agent_version"
         systemctl stop td-agent
       fi
+      if [ -f "/opt/td-agent/bin/fluent-cap-ctl" ]; then
+        apt-get -y install build-essential pkg-config libcap-ng-deb
```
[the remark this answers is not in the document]
flands: nice catch. dev pulls in main package: https://packages.debian.org/sid/libcap-ng-dev

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -386,12 +389,28 @@ install() {
         install_yum_package "td-agent" "$td_agent_version"
         systemctl stop td-agent
       fi
+      if [ -f "/opt/td-agent/bin/fluent-cap-ctl" ]; then
+        for package in build-essential pkg-config libcap-ng-deb; do
```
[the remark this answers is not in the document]
flands: ah yes, knew this and did not up PR

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -386,12 +389,26 @@ install() {
         install_yum_package "td-agent" "$td_agent_version"
         systemctl stop td-agent
       fi
+      if [ -f "/opt/td-agent/bin/fluent-cap-ctl" -a command -v yum >/dev/null 2>&1; ]; then
+        yum group install "Developement Tools"
```
jchengsfx, line 393: Need to also support `dnf` for systems without `yum`.

jchengsfx (changes requested):

internal/buildscripts/packaging/installer/install.sh
jchengsfx: We should be able to run `dnf group install "Development Tools"` if `yum` is not installed, similar to what the `install_yum_package` function does. I have it in https://github.com/signalfx/splunk-otel-collector/pull/139.

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -408,6 +408,9 @@ install() {
         apt-get -y install build-essential libcap-ng0 libcap-ng-dev pkg-config
         systemctl stop td-agent
       fi
+      if [ -f "/opt/td-agent/bin/fluent-cap-ctl" ]; then
+        apt-get -y install build-essential pkg-config libcap-ng-dev
+      fi
```
flands, line 413:
```suggestion
```

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -429,12 +432,26 @@ install() {
         done
         systemctl stop td-agent
       fi
+      if [ -f "/opt/td-agent/bin/fluent-cap-ctl" -a command -v yum >/dev/null 2>&1; ]; then
+        yum group install "Developement Tools"
+        install_yum_package "libcap-ng-devel" ""
+      fi
```
flands, line 438:
```suggestion
```

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -429,12 +432,26 @@ install() {
...
+
+  if [ -f "/opt/td-agent/bin/fluent-cap-ctl" ]; then
+    td-agent-gem install capng_c
+    /opt/td-agent/bin/fluent-cap-ctl --add dac_read_search -f /opt/td-agent/bin/ruby
+    /opt/td-agent/bin/fluent-cap-ctl --add dac_override -f /opt/td-agent/bin/ruby
+  else
+    if getent group adm >/dev/null 2>&1; then
+      usermod -a -G adm td-agent
+    fi
+  fi
```
flands, line 454:
```suggestion
```

jchengsfx (approved):

PR #130, opened 2021-02-26
translate Smart Agent events into log records

rmfitzpatrick:
These changes introduce a conversion helper that will be later adopted by the Smart Agent receiver's Output `SendEvent` method to provide the configured SFx exporter monitor-created events as log records.  They are based on the logic in the SFx receiver, but are unable to source it directly since it isn't exported and operates on the protobuf-generated types instead of the type defined in SFx golib as used by the agent.

### DIFF
--- a/internal/receiver/smartagentreceiver/event.go
+++ b/internal/receiver/smartagentreceiver/event.go
@@ -1,0 +1,119 @@
+// Copyright 2021 Splunk, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package smartagentreceiver
+
+import (
+	"fmt"
+
+	"github.com/signalfx/golib/v3/event"
+	"go.opentelemetry.io/collector/consumer/pdata"
+	"go.uber.org/zap"
+)
+
+const (
+	SFxEventCategoryKey   = "com.splunk.signalfx.event_category"
+	SFxEventPropertiesKey = "com.splunk.signalfx.event_properties"
+)
+
+// eventToLog converts a SFx event to a pdata.Logs entry suitable for consumption by LogConsumer.
+// based on https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5de076e9773bdb7617b544a57fa0a4b848cec92c/receiver/signalfxreceiver/signalfxv2_event_to_logdata.go#L27
+func eventToLog(event *event.Event, logger *zap.Logger) pdata.Logs {
+	logs, lr := newLogs()
+
+	lr.SetName(event.EventType)
+
+	var unixNano int64
+	if !event.Timestamp.IsZero() {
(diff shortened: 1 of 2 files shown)


internal/receiver/smartagentreceiver/event.go
```diff
@@ -0,0 +1,118 @@
...
+	"go.opentelemetry.io/collector/consumer/pdata"
+)
+
+const (
+	SFxEventCategoryKey   = "com.splunk.signalfx.event_category"
+	SFxEventPropertiesKey = "com.splunk.signalfx.event_properties"
+)
+
+// eventToLog converts a SFx event to a pdata.Logs entry suitable for consumption by LogConsumer.
+// based on https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5de076e9773bdb7617b544a57fa0a4b848cec92c/receiver/signalfxreceiver/signalfxv2_event_to_logdata.go#L27
```
asuresh4, line 30: Do you know if it would be possible to just reuse the method from the signalfxreceiver package? I know it would need to be exposed but is that something we should consider given that the conversion logic would essentially be the same?

asuresh4: nvm, read the PR description :)

rmfitzpatrick: I considered updating the agent to use the protobuf type but this code would then be in that project and I'd have to export the receiver helper.  Seemed simpler to do it here.

internal/receiver/smartagentreceiver/event.go
```diff
@@ -0,0 +1,118 @@
...
+		attrs.InsertString(k, v)
+	}
+
+	if len(event.Properties) > 0 {
+		propMapVal := pdata.NewAttributeValueMap()
+		propMap := propMapVal.MapVal()
+		propMap.InitEmptyWithCapacity(len(event.Properties))
+
+		for property, value := range event.Properties {
+			if value == nil {
```
asuresh4, line 67: When the value is `nil` should we still account for the key? If I understand correctly it would be equivalent to this [line](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5de076e9773bdb7617b544a57fa0a4b848cec92c/receiver/signalfxreceiver/signalfxv2_event_to_logdata.go#L75), no?

rmfitzpatrick: It's never serialized with the final event, but I guess we could keep it for the error path? https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/signalfxexporter/translation/logdata_to_signalfxv2.go#L115.

Seems preferable to continue to drop it but log here instead of deferring to the exporter.

internal/receiver/smartagentreceiver/event.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+		}
+
+		attrs.Insert(SFxEventPropertiesKey, propMapVal)
+	}
+
+	return logs
+}
+
+func newLogs() (*pdata.Logs, *pdata.LogRecord) {
```
tigrannajaryan, line 105: Nit: `pdata.LogRecord` is designed to be efficient to [pass by value](https://github.com/open-telemetry/opentelemetry-collector/blob/e6319ac4c6fce274b8ba2d0e3b8e5eb0fd0d6142/consumer/pdata/generated_log.go#L446), it is a reference type itself, so no need to return the pointer.

tigrannajaryan (approved):

internal/receiver/smartagentreceiver/event.go
asuresh4: Oh ok, yeah, I agree we should log here and continue in that case.

asuresh4 (approved):

PR #131, opened 2021-02-26
Make ensure_not_installed more robust

flands:
The Linux installer script has a check to see whether otelcol is
installed. td-agent should also be checked and now is. Also make echo
cleaner.


jchengsfx (approved):
LGTM.  I was actually in process of making the same changes.  I also need to fix the installer test for the new memory requirements, but shouldn't be related to these changes.

PR #134, opened 2021-02-26
Smart Agent: send SignalFx events to event clients

rmfitzpatrick:
These changes adopt SignalFx Event support using the same `MetadataClients` mechanism provided by https://github.com/signalfx/splunk-otel-collector/pull/124 but with the conversion provided by https://github.com/signalfx/splunk-otel-collector/pull/130.  The consumer selection logic is effectively the same as that of the metadata exporter path, but checking for `LogsConsumer` implementors and still defaulting to the next consumer if applicable.  **edit: Now these changes split dimension clients from event clients, but have the same functionality as described**

Not included, but associated with these changes will be a future PR proposing to default to any lone SignalFx exporter if no MetadataClients are specified to avoid the edge cases with processor usage and not requiring users specify metadata clients for each Smart Agent receiver instance.

edit: forgot to mention if not clear, this is designed to work w/ the SFx exporter's log to SFx event api feature without requiring standing up superfluous logging pipeline (after conversation w/ @jrcamp and @flands): https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/signalfxexporter/exporter.go#L193

additional edit: These changes split the proposal for a unified metadata client into dimension and event clients, but the functionality is unchanged.

### DIFF
--- a/internal/receiver/smartagentreceiver/README.md
+++ b/internal/receiver/smartagentreceiver/README.md
@@ -19,7 +19,11 @@ should be used.
 [Collector processors](https://github.com/open-telemetry/opentelemetry-collector/blob/master/processor/README.md).
 1. Monitors with [dimension property and tag update
 functionality](https://dev.splunk.com/observability/docs/datamodel#Creating-or-updating-custom-properties-and-tags)
-require an associated `metadataClients` field that references the name of the SignalFx exporter you are using in your
+require an associated `dimensionClients` field that references the name of the SignalFx exporter you are using in your
+pipeline.
+1. Monitors with [event-sending
+functionality](https://dev.splunk.com/observability/docs/datamodel/ingest#Send-custom-events)
+require an associated `eventClients` field that references the name of the SignalFx exporter you are using in your
 pipeline.
 
 Example:
@@ -37,7 +41,8 @@ receivers:
     type: postgresql
     host: mypostgresinstance
     port: 5432
-    metadataClients: [signalfx]
+    dimensionClients: [signalfx]
+    eventClients: [signalfx]
 ```
 
 The full list of settings exposed for this receiver are documented for
--- a/internal/receiver/smartagentreceiver/config.go
+++ b/internal/receiver/smartagentreceiver/config.go
@@ -32,15 +32,17 @@ import (
 
 const defaultIntervalSeconds = 10
 
-var errMetadataClientValue = fmt.Errorf("metadataClients must be an array of compatible exporter names")
+var errDimensionClientValue = fmt.Errorf("dimensionClients must be an array of compatible exporter names")
+var errEventClientValue = fmt.Errorf("eventClients must be an array of compatible exporter names")
 
 type Config struct {
 	configmodels.ReceiverSettings `mapstructure:",squash"`
 	// Generally an observer/receivercreator-set value via Endpoint.Target.
 	// Will expand to MonitorCustomConfig Host and Port values if unset.
-	Endpoint        string    `mapstructure:"endpoint"`
-	MetadataClients *[]string `mapstructure:"metadataclients"`
-	monitorConfig   config.MonitorCustomConfig
+	Endpoint         string    `mapstructure:"endpoint"`
+	DimensionClients *[]string `mapstructure:"dimensionclients"`
+	EventClients     *[]string `mapstructure:"eventclients"`
+	monitorConfig    config.MonitorCustomConfig
 }
 
 func (rCfg *Config) validate() error {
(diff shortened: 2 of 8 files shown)


tigrannajaryan:
> These changes adopt SignalFx Event support using the same `MetadataClients` mechanism provided by #124 but with the conversion provided by #130. The consumer selection logic is effectively the same as that of the metadata exporter path, but checking for `LogsConsumer` implementors and still defaulting to the next consumer if applicable.

I don't understand why we need this. Logs are a supported pipeline type and there is no need to have a parallel mechanism to connect receivers with exporters for logs. I may be missing something.

tigrannajaryan:
> edit: forgot to mention if not clear, this is designed to work w/ the SFx exporter's log to SFx event api feature without requiring standing up superfluous logging pipeline (after conversation w/ @jrcamp and @flands): https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/signalfxexporter/exporter.go#L193

Why is it "superfluous"?

rmfitzpatrick:
@tigrannajaryan there are a couple of factors that led to this approach:

1) Requiring the user to specify a logs pipeline and requiring all SA receivers to be listed in it is poor UX imo w/o agent equivalent.  The aim of this effort is to assist existing customers and not require them to rewrite any more of their config than is necessary, not including observers, receiver creator, new receivers/exporters etc.
2) The SFx exporter's only supported logs functionality is for SFx events, so it's really just an interface to an existing event client.  No additional pipeline components have a functional equivalent in the agent so are unnecessary to support here.
3) No other log exporter will ever need to be supported for SA receiver-generated SFx events.
4) Requiring the specification of metadata clients is more than likely a needless configuration requirement if we can make better decisions internally (~https://github.com/open-telemetry/opentelemetry-collector/issues/2536).  There is most likely only one "correct" configuration and the configuration burden is pushed to the user (which I will add in the proposed followup PR).  This being the case, I consider any configuration requirement to denote what we probably already know if there's a single SFx exporter as "superfluous".  If users are sending metrics in another format and don't want metadata clients, they can explicitly set the empty array.

tigrannajaryan:
> @tigrannajaryan there are a couple of factors that led to this approach:
> 
> 1. Requiring the user to specify a logs pipeline and requiring all SA receivers to be listed in it is poor UX imo w/o agent equivalent.  The aim of this effort is to assist existing customers and not require them to rewrite any more of their config than is necessary, not including observers, receiver creator, new receivers/exporters etc.
> 2. The SFx exporter's only supported logs functionality is for SFx events, so it's really just an interface to an existing event client.  No additional pipeline components have a functional equivalent in the agent so are unnecessary to support here.
> 3. No other log exporter will ever need to be supported for SA receiver-generated SFx events.
> 4. Requiring the specification of metadata clients is more than likely a needless configuration requirement if we can make better decisions internally (~[open-telemetry/opentelemetry-collector#2536](https://github.com/open-telemetry/opentelemetry-collector/issues/2536)).  There is most likely only one "correct" configuration and the configuration burden is pushed to the user (which I will add in the proposed followup PR).  This being the case, I consider any configuration requirement to denote what we probably already know if there's a single SFx exporter as "superfluous".  If users are sending metrics in another format and don't want metadata clients, they can explicitly set the empty array.

I believe this is a step in the wrong direction. I can be convinced but I would like to discuss this.

rmfitzpatrick:
For posterity, after a huddle I believe we've agreed to keep this as is since logging pipelines are supported and auto-configuration is desired where possible.

internal/receiver/smartagentreceiver/output.go
```diff
@@ -95,6 +88,69 @@ func getMetadataExporters(config Config, host component.Host, nextConsumer consu
...
+	for _, client := range metadataClients {
+		if logsExporter, ok := (*client).(consumer.LogsConsumer); ok {
+			consumers = append(consumers, &logsExporter)
+		} else {
+			logger.Info("cannot send events to metadataClient", zap.Any("client", *client))
+		}
+	}
+
+	if len(consumers) == 0 {
+		logger.Debug("no SFx events are possible as no valid metadataClients have been provided and next pipeline component isn't a LogsConsumer")
```
asuresh4, line 109: I may have missed this when it was discussed but why don't we have separate fields for metadata clients and log consumers in the config?

rmfitzpatrick: They are almost certainly the same exporter, and when OTel has entities or another change to trigger removing the experimental metric metadata update consumer and open their functionality to other components beyond SFx exporter, we'll likely gain another configuration method via pipelines or component sets.

rmfitzpatrick: Updated per our discussion

internal/receiver/smartagentreceiver/README.md
```diff
@@ -19,7 +19,11 @@ should be used.
 [Collector processors](https://github.com/open-telemetry/opentelemetry-collector/blob/master/processor/README.md).
 1. Monitors with [dimension property and tag update
 functionality](https://dev.splunk.com/observability/docs/datamodel#Creating-or-updating-custom-properties-and-tags)
-require an associated `metadataClients` field that references the name of the SignalFx exporter you are using in your
+require an associated `dimensionClients` field that references the name of the SignalFx exporter you are using in your
+pipeline.
+1. Monitors with [event-sending
+functionality](https://dev.splunk.com/observability/docs/datamodel/ingest#Send-custom-events)
+require an associated `eventClients` field that references the name of the SignalFx exporter you are using in your
```
asuresh4, line 26: Maybe for a later PR, but would it be a better experience to list out monitors that support properties and events explicitly in this README?

internal/receiver/smartagentreceiver/config.go
```diff
@@ -145,6 +140,29 @@ func mergeConfigs(componentViperSection *viper.Viper, intoCfg interface{}) error
 	return nil
 }
 
+type viperAllSettings map[string]interface{}
+
+func (allSettings viperAllSettings) getPointerToStringSliceFromYaml(key string, errToReturn error) (*[]string, error) {
```
asuresh4, line 145: Why not just return `[]string` instead of `*[]string`?

internal/receiver/smartagentreceiver/output.go
```diff
@@ -34,67 +35,117 @@ import (
...
-					"provided metadataClients item not a MetadataExporter.  cannot send dimension updates",
-					zap.String("client", client),
-				)
-			}
+	metadataExporters := getClientsFromMetricsExporters(config.DimensionClients, host, nextConsumer, "dimensionClients", logger)
+	for _, client := range metadataExporters {
+		if metadataExporter, ok := (*client).(metadata.MetadataExporter); ok {
+			exporters = append(exporters, &metadataExporter)
+		} else {
+			logger.Info("cannot send dimension updates to dimension client", zap.Any("client", *client))
```
asuresh4, line 81: Assuming that metadata clients are only configured for receiver instances that wrap monitors emitting metadata, should this be logged as a warning? Same with the log for event clients.

internal/receiver/smartagentreceiver/config.go
rmfitzpatrick: We need to know if it was unset for default behavior

internal/receiver/smartagentreceiver/output.go
rmfitzpatrick: This will occur for all SA receivers since they all have an Output and their nextConsumer default check will be attempted if no clients are configured.

internal/receiver/smartagentreceiver/config.go
asuresh4: We can check for `nil` value on `[]string` too, right?

rmfitzpatrick: Oh definitely.  My reasoning was muddled about nil slices not being possible in viper unmarshalling and it led to a bug.  thanks!

asuresh4 (approved):

PR #137, opened 2021-03-01
Fix installer test

jchengsfx:


flands (approved):
Nice! Didn't realize this existed.

PR #139, opened 2021-03-01
Enable journald log collection by default

jchengsfx:


internal/buildscripts/packaging/fpm/etc/otel/collector/fluentd/conf.d/journald.conf
```diff
@@ -25,30 +25,40 @@
...
 #
 
 <source>
+  @id journald
   @type systemd
   @label @SPLUNK
   tag "journald"
-  path "/run/log/journal"
+  path "/var/log/journal"
+  read_from_head true
```
flands, line 54: Is this required?

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -379,6 +405,7 @@ install() {
         install_td_agent_apt_repo "$td_agent_version"
         apt-get -y update
         install_apt_package "td-agent" "$td_agent_version"
+        apt-get -y install build-essential libcap-ng0 libcap-ng-dev pkg-config
```
flands, line 408: is `libcap-ng0` required? is the `0` expected?

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -392,6 +419,14 @@ install() {
       if [ -n "$td_agent_version" ]; then
         install_td_agent_yum_repo "$td_agent_version"
         install_yum_package "td-agent" "$td_agent_version"
+        if command -v yum >/dev/null 2>&1; then
+          yum group install -y 'Development Tools'
+        else
+          dnf group install -y 'Development Tools'
```
flands, line 425: Does this work?

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -392,6 +419,14 @@ install() {
...
+        fi
+        for pkg in libcap-ng libcap-ng-devel pkgconfig; do
```
flands, line 427: is `libcap-ng` required?

internal/buildscripts/packaging/fpm/etc/otel/collector/fluentd/conf.d/journald.conf
jchengsfx: Not sure. I just copied that option from https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/configmap-fluentd.yaml#L148.

internal/buildscripts/packaging/installer/install.sh
jchengsfx: According to https://docs.fluentd.org/deployment/linux-capability, both the `libcap-ng` and the dev packages are required.  Installing the dev package usually installs the main package, but I include it just in case.  The `0` in the debian package name is correct.

internal/buildscripts/packaging/installer/install.sh
jchengsfx: I tested this command in a oraclelinux:8 docker image which doesn't have `yum`, and it installed correctly.

internal/buildscripts/packaging/installer/install.sh
jchengsfx: See https://github.com/signalfx/splunk-otel-collector/pull/139#discussion_r585830676

internal/buildscripts/packaging/fpm/etc/otel/collector/fluentd/conf.d/journald.conf
flands: My understanding is that this is for file rotation and startup -- results in the entire thing being read and sent on every startup. I am under the impression we do not want this and that it could lead to performance issues.

internal/buildscripts/packaging/installer/install.sh
flands: Yes, both are required, but dev has a dependency on the main package. Seems fine to be explicit regardless.

flands (approved):
LGTM

PR #142, opened 2021-03-03
Fluentd updates

jchengsfx:


internal/buildscripts/packaging/fpm/etc/otel/collector/fluentd/conf.d/tomcat.conf
```diff
@@ -17,7 +17,6 @@
     format_firstline /^(\w+\s\d+,\s\d+)|(\d+-\d+-\d+\s)/
     format1 /(?<message>.*)/
   </parse>
-  format multiline
```
flands, line 20: Is this not needed?

flands (approved):
Looks good -- one question

internal/buildscripts/packaging/fpm/etc/otel/collector/fluentd/conf.d/tomcat.conf
jchengsfx: According to https://docs.fluentd.org/input/tail#format, it is deprecated.  I assume that the parser section you already have in there should handle multilines.  Also the fluentd log says that it is ignored if left in.

flands: oh this is format which is deprecated -- good call.

PR #144, opened 2021-03-05
Add and adopt unified integration test helper

rmfitzpatrick:
These changes introduce and source a new `testutils.AssertAllMetricsReceived()` helper to remove unnecessary boilerplate in integration tests, which is largely relocated from the existing postgresql monitor test.  The major change is that they enable a random OTLP receiver address to be used for each case and disable internal telemetry to avoid port conflicts.

Also adds a new `integration-test` make target to be utilized in future Circle CI change.

### DIFF
--- a/tests/receivers/smartagent/postgresql/postgresql_test.go
+++ b/tests/receivers/smartagent/postgresql/postgresql_test.go
@@ -15,117 +15,37 @@
 package tests
 
 import (
-	"context"
-	"fmt"
 	"path"
 	"testing"
-	"time"
-
-	"github.com/stretchr/testify/require"
-	"go.uber.org/zap"
-	"go.uber.org/zap/zapcore"
-	"go.uber.org/zap/zaptest/observer"
 
 	"github.com/signalfx/splunk-otel-collector/tests/testutils"
 )
 
 func TestPostgresReceiverProvidesAllMetrics(t *testing.T) {
-	tc := newTestcase(t)
-	defer tc.printLogsOnFailure()
-	defer func() { require.NoError(t, tc.otlp.Shutdown()) }()
+	containers := []testutils.Container{
+		testutils.NewContainer().WithContext(
+			path.Join(".", "testdata", "server"),
+		).WithEnv(map[string]string{
+			"POSTGRES_DB":       "test_db",
+			"POSTGRES_USER":     "postgres",
+			"POSTGRES_PASSWORD": "postgres",
+		}).WithExposedPorts(
+			"5432:5432",
+		).WithName("postgres-server").WithNetworks(
+			"postgres",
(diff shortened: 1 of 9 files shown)


tests/receivers/smartagent/postgresql/postgresql_test.go
```diff
@@ -15,117 +15,37 @@
...
+		).WillWaitForPorts("5432").WillWaitForLogs(
+			"database system is ready to accept connections",
+		),
+		testutils.NewContainer().WithContext(
+			path.Join(".", "testdata", "client"),
+		).WithEnv(map[string]string{
+			"POSTGRES_SERVER": "postgres-server",
+		}).WithName("postgres-client").WithNetworks(
+			"postgres",
+		).WillWaitForLogs("Beginning psql requests"),
```
tigrannajaryan, line 45: This is difficult to read. I can't tell where slice items start and where they end. Or is it a one item? Hard to tell, looks like a wall of text.
Maybe try formatting differently or extract bits into variables.

tests/receivers/smartagent/postgresql/testdata/all_metrics_config.yaml
```diff
@@ -12,17 +12,12 @@ receivers:
 
 exporters:
   otlp:
-    endpoint: localhost:23456
+    endpoint: "${OTLP_ENDPOINT}"
```
tigrannajaryan, line 15: 👍

tests/testutils/testcase.go
```diff
@@ -0,0 +1,123 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/testutil"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest/observer"
+)
+
+type Testcase struct {
```
tigrannajaryan, line 31: Please document exported struct. Also, does it need to be exported?

rmfitzpatrick: Will do.  I'd say so since some proposed cases use the builder and its receiver methods (e.g. https://github.com/signalfx/splunk-otel-collector/pull/145/files#diff-10deb04b733fce760ce0b30c42f5862339f0e2d7aced0e2f8a934e637eca3adb) and I believe golint discourages this.

tests/receivers/smartagent/postgresql/postgresql_test.go
rmfitzpatrick: This better?  Unfortunately (imo) `gofmt` collapses things to one endless line without giving arguments their own individual line.

tigrannajaryan (approved):

PR #149, opened 2021-03-08
Smart Agent: Default to lone SFx Exporter for dimension/event client if applicable

rmfitzpatrick:
These changes add an additional path for default dimension and event client selection in the Smart Agent receiver.  The final proposed behavior is:

1.  If dimension/event clients config field is not null, try to use all specified exporters.  If it is not null, but empty don't use any exporters. (already the case)
2. If dimension/event clients is null, try to use the next consumer for the associated pipeline. (already the case)
3. If dimension/event clients is null but the next consumer is not compatible, use the lone SFx exporter on the host. (this PR)
4. If there is more than one configured SFx exporter, don't use any exporters. (this PR)

edit: To clarify, the aim of these changes is to not require specifying a value for this field for most usages, but to not interfere with or prevent any desired configuration.

### DIFF
--- a/internal/receiver/smartagentreceiver/README.md
+++ b/internal/receiver/smartagentreceiver/README.md
@@ -19,12 +19,18 @@ should be used.
 [Collector processors](https://github.com/open-telemetry/opentelemetry-collector/blob/master/processor/README.md).
 1. Monitors with [dimension property and tag update
 functionality](https://dev.splunk.com/observability/docs/datamodel#Creating-or-updating-custom-properties-and-tags)
-require an associated `dimensionClients` field that references the name of the SignalFx exporter you are using in your
-pipeline.
+allow an associated `dimensionClients` field that references the name of the SignalFx exporter you are using in your
+pipeline.  If you do not specify any exporters via this field, the receiver will attempt to use the associated
+pipeline.  If the next element of the pipeline isn't compatible with dimension update behavior, if a lone SignalFx
+exporter was configured for your deployment, it will be selected.  If no dimension update behavior is desired,
+you can specify the empty array `[]` to disable.
 1. Monitors with [event-sending
 functionality](https://dev.splunk.com/observability/docs/datamodel/ingest#Send-custom-events)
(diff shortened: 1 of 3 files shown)


tigrannajaryan:
>If there is more than one configured SFx exporter, don't use any exporters.

This may be a surprising behavior: you add an Sfx exporter to the config and suddenly a previous one that worked fine stops working. I would argue that a less surprising approach is to just use all Sfx exporters and if that's undesirable you can always override using empty setting.

rmfitzpatrick:
> > If there is more than one configured SFx exporter, don't use any exporters.
> 
> This may be a surprising behavior: you add an Sfx exporter to the config and suddenly a previous one that worked fine stops working. I would argue that a less surprising approach is to just use all Sfx exporters and if that's undesirable you can always override using empty setting.

I considered this, but it seemed unlikely that two different SFx exporters would be in the same pipeline, so it's likely that one would receive dimension updates and events for a receiver it's not receiving metrics for if all SFx exporters are selected.

tigrannajaryan:
> > > If there is more than one configured SFx exporter, don't use any exporters.
> > 
> > 
> > This may be a surprising behavior: you add an Sfx exporter to the config and suddenly a previous one that worked fine stops working. I would argue that a less surprising approach is to just use all Sfx exporters and if that's undesirable you can always override using empty setting.
> 
> I considered this, but it seemed unlikely that two different SFx exporters would be in the same pipeline, so it's likely that one would receive dimension updates and events for a receiver it's not receiving metrics for if all SFx exporters are selected.

That's a good point. OK, I don't have a strong opinion. It looks like there is no ideal approach here.

rmfitzpatrick:
I've realized there's an issue with events -> log records and the factory-created SFx exporter is not compatible.  These changes ensure that both dimension updates and events retrieve the desired exporter instance by `configmodels.DataType`.

tigrannajaryan (approved):

rmfitzpatrick:
Closing this and will require event-producers to be in log pipelines in later PR.

PR #151, opened 2021-03-09
Port monitor-level filtering functionality from signalfx-agent

dmitryax:
The following files ported mostly as is from signalfx-agent:
- internal/receiver/smartagentreceiver/filtering.go 
- internal/receiver/smartagentreceiver/filtering.go

### DIFF
--- a/internal/receiver/smartagentreceiver/config_test.go
+++ b/internal/receiver/smartagentreceiver/config_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/signalfx/signalfx-agent/pkg/monitors/collectd/hadoop"
 	"github.com/signalfx/signalfx-agent/pkg/monitors/collectd/python"
 	"github.com/signalfx/signalfx-agent/pkg/monitors/collectd/redis"
+	"github.com/signalfx/signalfx-agent/pkg/monitors/filesystems"
 	"github.com/signalfx/signalfx-agent/pkg/monitors/haproxy"
 	"github.com/signalfx/signalfx-agent/pkg/monitors/prometheusexporter"
 	"github.com/signalfx/signalfx-agent/pkg/monitors/telegraf/monitors/ntpq"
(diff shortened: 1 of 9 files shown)


internal/receiver/smartagentreceiver/filtering.go
```diff
@@ -0,0 +1,256 @@
...
+
+// HasEnabledMetricInGroup returns true if there are any metrics enabled that
+// fall into the given group.
+func (mf *monitorFiltering) HasEnabledMetricInGroup(group string) bool {
+	if mf.metadata == nil {
+		return false
+	}
+
+	for _, m := range mf.EnabledMetrics() {
+		// TODO: If metric names in metadata.yaml ever support wildcards this
```
rmfitzpatrick, line 82: Is this TODO still valid for this project?

rmfitzpatrick:
Are you able to add config file tests using `datapointsToExclude`, `extraMetrics`, `extraGroups` and other applicable fields?  Ideally the diff coverage would also be more complete.

internal/receiver/smartagentreceiver/filtering_test.go
```diff
@@ -0,0 +1,197 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/signalfx/golib/v3/datapoint"
+	"github.com/signalfx/signalfx-agent/pkg/monitors"
+	"github.com/signalfx/signalfx-agent/pkg/utils"
+)
+
+func testMetadata(sendUnknown bool) *monitors.Metadata {
```
rmfitzpatrick, line 26: Mind attributing this test file content as well?

internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -131,7 +131,20 @@ func (r *Receiver) createMonitor(monitorType string, host component.Host) (monit
...
 	monitor = monitorFactory() // monitor is a pointer to a monitor struct
 
-	output := NewOutput(*r.config, r.nextConsumer, host, r.logger)
+	// Make metadata nil if we aren't using built in filtering and then none of
+	// the new filtering logic will apply.
+	metadata, ok := monitors.MonitorMetadatas[monitorType]
+	if !ok || metadata == nil {
+		// This indicates a programming error in not specifying metadata, not
+		// bad user input
+		panic(fmt.Sprintf("could not find monitor metadata of type %s", monitorType))
```
rmfitzpatrick, line 140: The current pattern is to return errors and not panic.  Looking through collector and collector-contrib panicking doesn't seem too common either.

internal/receiver/smartagentreceiver/filtering.go
```diff
@@ -0,0 +1,256 @@
...
+func (mf *monitorFiltering) AddDatapointExclusionFilter(filter dpfilters.DatapointFilter) {
+	mf.filterSet.ExcludeFilters = append(mf.filterSet.ExcludeFilters, filter)
+}
+
+func (mf *monitorFiltering) EnabledMetrics() []string {
+	if mf.metadata == nil {
+		return nil
+	}
+
+	// This can be optimized to cache results on the struct if it is worth it.
```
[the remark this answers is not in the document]
dmitryax: Looks like it is not, removing the comment

internal/receiver/smartagentreceiver/filtering.go
dmitryax: removing

internal/receiver/smartagentreceiver/filtering_test.go
dmitryax: done

internal/receiver/smartagentreceiver/receiver.go
dmitryax: converted to a returnen error

dmitryax:
@rmfitzpatrick , I added the config tests and bumped the coverage. PTAL

internal/receiver/smartagentreceiver/config_test.go
```diff
@@ -416,3 +417,77 @@ func TestLoadInvalidConfigWithNonStringArrayDimensionClients(t *testing.T) {
...
+						MetricNames: []string{"./[0-"},
+					},
+				},
+			},
+		},
+	}, fsCfg)
+
+	err = fsCfg.validate()
+	require.Error(t, err)
+	require.EqualError(t, err, "unexpected end of input")
```
rmfitzpatrick, line 492: Not necessary to address here, but we should probably provide an error wrapper for validation errors so they provide a bit more context.  This is pretty vague.

internal/receiver/smartagentreceiver/filtering.go
```diff
@@ -0,0 +1,255 @@
...
+				return nil
+			}
+		}
+
+		logrus.Warnf("extraMetrics: metric pattern '%s' did not match any available metrics for monitor %s",
+			metricName, metadata.MonitorType)
+	}
+
+	if !metadata.HasMetric(metricName) {
+		logrus.Warnf("extraMetrics: metric '%s' does not exist for monitor %s", metricName, metadata.MonitorType)
```
rmfitzpatrick, line 171: We should be using the receiver's zap logger and not the global logrus logger.

rmfitzpatrick:
@dmitryax thanks for this and your updates.  It lgtm w/ all logrus usage converted to zap (we have a shim and I think it would end up using the default instance which wouldn't be ideal).

internal/receiver/smartagentreceiver/config_test.go
dmitryax: Makes sense, submitted an issue https://github.com/signalfx/splunk-otel-collector/issues/157. Will cover it in a following PR

internal/receiver/smartagentreceiver/filtering.go
dmitryax: addressed

rmfitzpatrick (approved):

PR #161, opened 2021-03-11
Smart Agent log receiver

rmfitzpatrick:
(In lieu of https://github.com/signalfx/splunk-otel-collector/pull/149)

These changes allow the Smart Agent receiver to act as both a metrics and logs receiver, as well as removing the notion of an event client in config and Output functionality.  Instead they use standard log pipelines after I discovered that the main user of events in the agent, the processlist monitor, requires a host dimension that must be added via processor in a log pipeline unless hardcoded in agent config (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/2631 is necessary in SFx exporter for this to take effect).

Also includes the basic SFx exporter dimension client defaulting logic, which is simpler without event clients.

### DIFF
--- a/internal/receiver/smartagentreceiver/README.md
+++ b/internal/receiver/smartagentreceiver/README.md
@@ -19,30 +19,54 @@ should be used.
 [Collector processors](https://github.com/open-telemetry/opentelemetry-collector/blob/master/processor/README.md).
 1. Monitors with [dimension property and tag update
 functionality](https://dev.splunk.com/observability/docs/datamodel#Creating-or-updating-custom-properties-and-tags)
-require an associated `dimensionClients` field that references the name of the SignalFx exporter you are using in your
-pipeline.
+allow an associated `dimensionClients` field that references the name of the SignalFx exporter you are using in your
+pipeline.  If you do not specify any exporters via this field, the receiver will attempt to use the associated
+pipeline.  If the next element of the pipeline isn't compatible with dimension update behavior, if a lone SignalFx
+exporter was configured for your deployment, it will be selected.  If no dimension update behavior is desired,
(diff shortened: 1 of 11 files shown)


internal/receiver/smartagentreceiver/README.md
```diff
@@ -19,30 +19,58 @@ should be used.
...
+to use a Resource Detection to ensure that host identity and other useful information is made available as event
+dimensions.
 
 Example:
 
 ```yaml
-exporters:
-  signalfx:
+extensions:
+  smartagent:
```
tigrannajaryan, line 39: Is the extension required even if we don't specify any settings for it?

internal/receiver/smartagentreceiver/README.md
```diff
@@ -19,30 +19,58 @@ should be used.
...
+
+service:
+  extensions: [smartagent]
+  pipelines:
+    metrics:
+      receivers: [smartagent/postgresql, smartagent/kafka]
+      processors: [resourcedetection]
+      exporters: [signalfx]
+    logs:
+      receivers: [smartagent/processlist]
```
tigrannajaryan, line 71: So, just to clarify. The logs pipeline is required for events to work, correct?
Do we want to document which monitors require it?

rmfitzpatrick: Yes, it's required and I will provide a list of dimension update and event using monitors w/ my next full documentation PR.

internal/receiver/smartagentreceiver/README.md
rmfitzpatrick: Not necessary unless it changes settings.  Will remove for now.

tigrannajaryan (approved):

PR #163, opened 2021-03-11
Add host to windows eventlog

jchengsfx:


jrcamp:
Seems like host.name should be added by resourcedetection processor to all logs going through collector. Is it not?

jchengsfx:
> Seems like host.name should be added by resourcedetection processor to all logs going through collector. Is it not?

Looks like we only need to add `host` instead of `host.name`. Without it, both `host` and `host.name` are "Unknown".

jrcamp (approved):

PR #169, opened 2021-03-17
rename collector config and disable exporter rules

mstumpfx:
translation_rules and exclude_metrics should not be used in a gateway
style collector, and should be applied only at the agent level.


flands (commented):
Change is fine, but rename of config requires a bunch of other changes including the Dockerfile :) how about we use gateway_config.yaml and change everywhere?

mstumpfx:
I messed this up locally, easier for me to just wipe and reset.

PR #170, opened 2021-03-17
rename splunk_config_linux.yaml to gateway_config.yaml

mstumpfx:
I used ```rg "splunk_config_linux.yaml" -l | xargs sed -i '' "s/splunk_config_linux.yaml/gateway_config.yaml/g"``` I'm not sure if this needs to be changed elsewhere.


jrcamp:
I'm confused about https://github.com/signalfx/splunk-otel-collector/blob/758688e2eb0bf3c49a2bc0058387095fc78d12fb/cmd/otelcol/main.go#L42 I don't see it referencing agent_config.yaml anywhere in this section 🤔 Either I don't understand how this is supposed to work or maybe it was never updated?

I feel like defaultDockerSAPMConfig, etc. should be renamed to defaultGatewayConfig to make it more clear.

mstumpfx:
> I'm confused about
> 
> https://github.com/signalfx/splunk-otel-collector/blob/758688e2eb0bf3c49a2bc0058387095fc78d12fb/cmd/otelcol/main.go#L42
> I don't see it referencing agent_config.yaml anywhere in this section 🤔 Either I don't understand how this is supposed to work or maybe it was never updated?
> 
> I feel like defaultDockerSAPMConfig, etc. should be renamed to defaultGatewayConfig to make it more clear.

hm yeah, some of the wording around these probably needs to be changed. I would think for the install sections we would want a section for using collector as a gateway and a section for using collector as an agent.

flands (approved):

internal/buildscripts/packaging/fpm/common.sh
```diff
@@ -16,7 +16,7 @@ SERVICE_GROUP="splunk-otel-collector"
 
 OTELCOL_INSTALL_PATH="/usr/bin/otelcol"
 CONFIG_REPO_PATH="$REPO_DIR/cmd/otelcol/config/collector/agent_config.yaml"
-CONFIG_INSTALL_PATH="/etc/otel/collector/splunk_config_linux.yaml"
+CONFIG_INSTALL_PATH="/etc/otel/collector/gateway_config.yaml"
```
jchengsfx, line 19: For the default config file in the deb/rpm packages, maybe we just package the file as `config.yaml` instead of `gateway_config.yaml`. We'll then need to add the `SPLUNK_CONFIG=/etc/otel/collector/config.yaml` env var to the collector service environment file to override whatever is the default in main.go.  I can make the necessary changes after this PR is merged.

PR #174, opened 2021-03-18
Update collector core and contrib to latest and fix example components usage

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>


internal/receiver/smartagentreceiver/output_test.go
```diff
@@ -264,7 +264,18 @@ type mockMetadataClient struct {
...
+func (me *mockMetadataClient) Start(_ context.Context, _ component.Host) error {
+	panic("implement me")
+}
+
+func (me *mockMetadataClient) Shutdown(_ context.Context) error {
+	panic("implement me")
+}
+
+func (me *mockMetadataClient) ConsumeMetrics(_ context.Context, _ pdata.Metrics) error {
+	panic("implement me")
```
bogdandrutu, line 278: If anyone can file an issue in the core about this use-case to be provided/facilitate in core would be appreciated.

dmitryax (approved):

internal/receiver/smartagentreceiver/output_test.go
```diff
@@ -158,7 +158,7 @@ func TestSendDimensionUpdateFromConfigMetadataExporters(t *testing.T) {
 			DimensionClients: []string{"mockmetadataexporter", "exampleexporter", "metricsreceiver", "notareceiver", "notreal"},
 		},
 		fakeMonitorFiltering(),
-		&componenttest.ExampleExporterConsumer{},
+		consumertest.NewMetricsNop(),
```
bogdandrutu, line 161: Here and the next 3 similar usages are just wrong usages of the exporter example (one of the reason to be removed)

internal/receiver/smartagentreceiver/output_test.go
bogdandrutu: Also an investigation if this is required at all to implement that interface will be good

internal/receiver/smartagentreceiver/output_test.go
rmfitzpatrick: What is wrong with its usage?  This test exercises usage of incompatible components and confirms they don't conflict with valid ones.

PR #177, opened 2021-03-18
Include smart agent bundle in amd64 deb/rpm

jchengsfx:
- ~~Download and install the smart agent bundle to `/usr/lib/splunk-otel-collector` (latest SA release by default)~~ Download and install the smart agent v5.9.1 bundle to `/usr/lib/splunk-otel-collector/agent-bundle`
- Add the `smartagent` extension to the default config with the `SPLUNK_BUNDLE_DIR` and `SPLUNK_COLLECTD_DIR` env vars
- Rename the systemd environment file from `splunk_env` to `splunk-otel-collector.conf`


internal/buildscripts/packaging/fpm/common.sh
```diff
@@ -43,6 +48,36 @@ create_user_group() {
...
+        exit 1
+    fi
+
+    echo "Downloading $dl_url ..."
+    curl -sL "$dl_url" -o "$buildroot/signalfx-agent.tar.gz"
+
+    mkdir -p "$buildroot/usr/lib"
+    tar -xzf "$buildroot/signalfx-agent.tar.gz" -C "$buildroot/usr/lib"
+    mv "$buildroot/usr/lib/signalfx-agent" "$buildroot/$SMART_AGENT_INSTALL_PATH"
+    rm -f "$buildroot/signalfx-agent.tar.gz"
```
rmfitzpatrick, line 78: If we remove the agent binary we should free up a fair amount of space.

jchengsfx: Wasn't sure if we needed it or not. I'll remove it.

rmfitzpatrick: 🤞

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -4,6 +4,8 @@
 # environment variables are required to be manually defined or configured below:
 # - SPLUNK_ACCESS_TOKEN: The Splunk access token to authenticate requests
 # - SPLUNK_API_URL: The Splunk API URL, e.g. https://api.us0.signalfx.com
+# - SPLUNK_BUNDLE_DIR: The path to the Smart Agent bundle, e.g. /usr/lib/splunk-otel-collector
```
rmfitzpatrick, line 7: Does `/usr/lib/splunk-otel-collector/agent-bundle` make more sense?  We will likely have more artifacts over time and having everything all together in one directory seems undesirable.

internal/buildscripts/packaging/fpm/Dockerfile
```diff
@@ -12,5 +12,6 @@ ENV PACKAGE="deb"
 ENV VERSION=""
 ENV ARCH="amd64"
 ENV OUTPUT_DIR="/repo/dist/"
+ENV SMART_AGENT_RELEASE="latest"
```
rmfitzpatrick, line 15: If we wanted to keep in step with our agent dep can we use the output of `go list -m github.com/signalfx/signalfx-agent` instead of pulling from latest?

rmfitzpatrick: per offline convo: pinning to v5.9.0 for now and using `20210308172704-d05049330d0d` in go.mod is good workaround (though looks like there's a 5.9.1 release on the way)

rmfitzpatrick (approved):
Thanks!

PR #178, opened 2021-03-18
Smart Agent: translate SFx traces to pdata

rmfitzpatrick:
These change provide some minor refactoring of the conversion functions by breaking out to a new `converter` subpackage and add SFx->pdata trace conversion.

SFx trace format is an implementation of Zipkin v2, so I coerce to its json representation before using the existing Collector's zipkin v2 to pdata translator.  If performance issues become a concern we can optimize as needed.

This PR doesn't adopt any of this translation, which will be done shortly after.


rmfitzpatrick:
Will improve diff coverage, but reviews still appreciated.

mstumpfx (approved):
less familiar with tracing still but lgtm

rmfitzpatrick:
Bypassing diff coverage since the remaining uncovered is metric fallout from rename.

PR #187, opened 2021-03-19
Support gateway mode for linux installer

jchengsfx:
- Updated installer script for the `--mode <agent|gateway>` option (defaults to agent)
- Sets the `SPLUNK_CONFIG` env var for the collector service to `agent_config.yaml` if running the installer script with `--mode agent`, or `gateway_config.yaml` if `--mode gateway`
- Updated the installer test for the `--mode` and `--without-fluentd` options

### DIFF
--- a/internal/buildscripts/packaging/fpm/common.sh
+++ b/internal/buildscripts/packaging/fpm/common.sh
@@ -29,8 +29,10 @@ SERVICE_USER="splunk-otel-collector"
 SERVICE_GROUP="splunk-otel-collector"
 
 OTELCOL_INSTALL_PATH="/usr/bin/otelcol"
-CONFIG_REPO_PATH="$REPO_DIR/cmd/otelcol/config/collector/agent_config.yaml"
-CONFIG_INSTALL_PATH="/etc/otel/collector/agent_config.yaml"
+AGENT_CONFIG_REPO_PATH="$REPO_DIR/cmd/otelcol/config/collector/agent_config.yaml"
+AGENT_CONFIG_INSTALL_PATH="/etc/otel/collector/agent_config.yaml"
+GATEWAY_CONFIG_REPO_PATH="$REPO_DIR/cmd/otelcol/config/collector/gateway_config.yaml"
+GATEWAY_CONFIG_INSTALL_PATH="/etc/otel/collector/gateway_config.yaml"
 SERVICE_REPO_PATH="$FPM_DIR/$SERVICE_NAME.service"
 SERVICE_INSTALL_PATH="/lib/systemd/system/$SERVICE_NAME.service"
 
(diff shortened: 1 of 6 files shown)


jrcamp:
Can we make agent vs gateway explicit instead of inferring based on fluentd option? ie, have an option `--mode (agent|gateway)` that defaults to agent. This would also allow people to install the agent without fluentd (which some may want if they're not using log observer).

jrcamp:
Actually.. we would probably need to do something similar to how helm chart works as you may want both agent and gateway at the same time. So it'd probably need to be something like `--{enable,disable} {agent,gateway,fluentd}` where agent and fluentd is on by default.

jchengsfx:
> Actually.. we would probably need to do something similar to how helm chart works as you may want both agent and gateway at the same time. So it'd probably need to be something like `--{enable,disable} {agent,gateway,fluentd}` where agent and fluentd is on by default. @flands thoughts?

@jrcamp Which config file should the collector service be configured for if we allow enabling "both agent and gateway at the same time"?  Also, other than setting the `SPLUNK_CONFIG` env var for either the agent or gateway config, is there anything else that the installer script should do when enabling agent vs gateway?

flands (approved):
LGTM -- should support one mode or the other and should default to agent. Besides configuration, everything else should be the same.

jrcamp:
> LGTM -- should support one mode or the other and should default to agent. Besides configuration, everything else should be the same.

Though what do you do if you want to monitor the host your gateway is running on if you can't install the agent? 😛 I don't think that's covered today is it? It means there's 3 states (which is what the helm chart covers). agent, gateway, agent + gateway.

jchengsfx:
> > LGTM -- should support one mode or the other and should default to agent. Besides configuration, everything else should be the same.
> 
> Though what do you do if you want to monitor the host your gateway is running on if you can't install the agent? 😛 I don't think that's covered today is it? It means there's 3 states (which is what the helm chart covers). agent, gateway, agent + gateway.

@jrcamp Sorry, I'm still not understanding how we run the collector in both modes at the same time on the same host.  Are you suggesting that we run 2 separate collector services (one for agent mode and one for gateway), or is there a way to run a single collector service with both agent and gateway configs?

mstumpfx:
@jchengsfx I think he means a tertiary config, where the behavior is like the gateway, but also has enabled like the host metrics receiver so that the host the gateway is running on is also monitored. I could also see us just enabling this by default in the gateway config.

jrcamp:
> @jchengsfx I think he means a tertiary config, where the behavior is like the gateway, but also has receivers enabled like the host metrics receiver, so that the host the gateway is running on is also monitored. I could also see us just enabling this behavior by default in the gateway config. If not, maybe we should at least doc it.

Correct. I believe there are APM-only customers (though those might actually come with host metrics for free, I can't remember).

jchengsfx:
> > @jchengsfx I think he means a tertiary config, where the behavior is like the gateway, but also has receivers enabled like the host metrics receiver, so that the host the gateway is running on is also monitored. I could also see us just enabling this behavior by default in the gateway config. If not, maybe we should at least doc it.
> 
> Correct. I believe there are APM-only customers (though those might actually come with host metrics for free, I can't remember).

Ok, thanks.  We can add a third mode in a separate PR, maybe something like `--mode <agent|gateway|both>`, once we have the tertiary config available to include in the packages.

Issue #153, opened 2021-03-09
Windows service event logging issue


jchengsfx:
Messages from the collector when running as a Windows service are logged in the Windows Event Log in the following format:
```
The description for Event ID 1 from source splunk-otel-collector cannot be found. Either the component that raises this event is not installed on your local computer or the installation is corrupted. You can install or repair the component on the local computer.

If the event originated on another computer, the display information had to be saved with the event.

The following information was included with the event: 

Everything is ready. Begin running and processing data.

Stack Trace:
```

Tested with the v0.21.0 MSI on both Server 2016 and 2019.

The core and contrib MSI's also have the same issue.

![Screen Shot 2021-03-09 at 4 40 17 PM](https://user-images.githubusercontent.com/36963516/110541790-384fdc00-80f6-11eb-82f2-a8b4b38f0a7b.png)

mtwo:
I'll try to work my way through my network, but I don't think that I know anyone at Microsoft who could answer this.

I assume that you've tried restarting the event viewer after the Collector registers itself as a log source? Looks like that's a common cause of this issue, as the event viewer won't see any sources that registered after launch. I also found this: https://stackoverflow.com/questions/3412463/description-for-event-id-from-source-cannot-be-found/32653571

jchengsfx:
Thanks @mtwo.  I can try to implement a workaround by updating the MSI to create the registry entry in `HKLM:SYSTEM\CurrentControlSet\Services\EventLog\Application` for the service.  However, I think the better long-term solution may be for the collector app itself to register to the eventlog, maybe incorporate something like https://github.com/golang/sys/blob/master/windows/svc/eventlog/install.go or try to emulate the smart agent https://github.com/signalfx/signalfx-agent/blob/master/cmd/agent/agentmain/windows.go?  I'm not a golang developer, so not sure what the best course of action would be.

atoulme:
Closing this issue for inactivity. Please reopen if further work is required.

PR #188, opened 2021-03-19
Create registry key for collector service events

jchengsfx:
- Workaround for #153 
- The registry key is created on MSI installation and removed on uninstall
- Adding the registry key changes the events to:
Screen Shot 2021-03-19 at 5 10 38 PM

### DIFF
--- a/internal/buildscripts/packaging/msi/splunk-otel-collector.wxs
+++ b/internal/buildscripts/packaging/msi/splunk-otel-collector.wxs
@@ -19,7 +19,7 @@
                      <File Id="ExecutableFile" Name="otelcol.exe" KeyPath="yes" Source="$(var.Otelcol)"/>
                      <!-- Install the collector service with the config file in ProgramData -->
                      <ServiceInstall
-                        Id="Sevice"
+                        Id="Service"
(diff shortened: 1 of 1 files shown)


internal/buildscripts/packaging/msi/splunk-otel-collector.wxs
```diff
@@ -36,6 +36,9 @@
                         Stop="both"
                         Remove="uninstall"
                         Wait="yes" />
+                     <RegistryKey Root="HKLM" Key="SYSTEM\CurrentControlSet\Services\EventLog\Application\splunk-otel-collector">
+                        <RegistryValue Type="expandable" Name="EventMessageFile" Value="C:\Windows\Microsoft.NET\Framework64\v4.0.30319\EventLogMessages.dll"/>
```
rmfitzpatrick, line 40: How reliable is this path across supported OS versions?

jchengsfx: Good point. I'll change it to `%SystemRoot%\System32\EventCreate.exe` which is the same as the smart agent.

rmfitzpatrick (approved):

PR #167, opened 2021-03-16
Provide introductory Smart Agent migration example

rmfitzpatrick:
These changes provide a basic help section for SA Receiver adoption and migration.  They also provide more details on the required usage of dimension clients and when to configure the receiver as part of a logs pipeline.

I have not tested the recommended config as is, but will be doing so when possible.  Deferring to @flands for when to merge.

PR #192, opened 2021-03-22
Add Trace support to SA Receiver

rmfitzpatrick:
These changes adopt the converter introduced in https://github.com/signalfx/splunk-otel-collector/pull/178 and add trace receiver support to the Smart Agent Receiver.  They also port the extra and default span tags functionality from the SA codebase.

This is largely done to support the signalfx-forwarder monitor, whose documentation I will add in a subsequent PR once https://github.com/signalfx/splunk-otel-collector/pull/167 has landed.

### DIFF
--- a/internal/receiver/smartagentreceiver/output.go
+++ b/internal/receiver/smartagentreceiver/output.go
@@ -43,7 +43,10 @@ const internalTransport = "internal"
 type Output struct {
 	nextMetricsConsumer  consumer.MetricsConsumer
 	nextLogsConsumer     consumer.LogsConsumer
+	nextTracesConsumer   consumer.TracesConsumer
 	extraDimensions      map[string]string
+	extraSpanTags        map[string]string
+	defaultSpanTags      map[string]string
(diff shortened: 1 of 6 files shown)


internal/receiver/smartagentreceiver/output.go
```diff
@@ -223,7 +232,28 @@ func (output *Output) SendEvent(event *event.Event) {
...
 
 func (output *Output) SendSpans(spans ...*trace.Span) {
-	output.logger.Debug("SendSpans has been called.", zap.Any("Span", spans))
+	if output.nextTracesConsumer == nil {
+		return
+	}
+
+	for _, span := range spans {
+		for name, value := range output.defaultSpanTags {
+			if span.Tags == nil {
```
pmcollins, line 241: Would it work if this block were in the outer loop?

rmfitzpatrick: definitely, thanks for catching.

pmcollins (approved):
LGTM

PR #218, opened 2021-03-29
Add SignalFx forwarder to default configs

rmfitzpatrick:
This will allow existing tracing library users to not need to update their instrumented deployments' configurations when using the Collector.

Also pork-barrelling layer caching on fpm image builds.

### DIFF
--- a/cmd/otelcol/config/collector/agent_config.yaml
+++ b/cmd/otelcol/config/collector/agent_config.yaml
@@ -66,6 +66,9 @@ receivers:
             regex: '.*grpc_io.*'
             action: drop
   sapm:
+  smartagent/signalfx-forwarder:
+    type: signalfx-forwarder
+    listenAddress: 0.0.0.0:9080
   signalfx:
   zipkin:
 
@@ -120,7 +123,7 @@ service:
   extensions: [health_check, http_forwarder, zpages]
   pipelines:
     traces:
-      receivers: [jaeger, otlp, sapm, zipkin]
+      receivers: [jaeger, otlp, sapm, smartagent/signalfx-forwarder, zipkin]
       processors:
       - memory_limiter
       - batch
@@ -128,7 +131,7 @@ service:
       #- resource/add_environment
       exporters: [sapm, signalfx]
     metrics:
-      receivers: [otlp, signalfx, prometheus, hostmetrics]
+      receivers: [hostmetrics, otlp, prometheus, signalfx, smartagent/signalfx-forwarder]
       processors: [memory_limiter, batch, resourcedetection]
       exporters: [signalfx]
     logs/signalfx:
--- a/cmd/otelcol/config/collector/full_config_linux.yaml
+++ b/cmd/otelcol/config/collector/full_config_linux.yaml
@@ -76,6 +76,12 @@ receivers:
   sapm:
     #access_token_passthrough: true
 
+  # Enables the Smart Agent's SignalFx Forwarder with default settings
+  # - listenAddress: 127.0.0.1:9080
+  # - serverTimeout: 5 (seconds)
+  smartagent/signalfx-forwarder:
+    type: signalfx-forwarder
+
   # Enables the zipkin receiver with default settings
   #  - grpc (default endpoint = 0.0.0.0:9411)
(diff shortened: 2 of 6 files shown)


cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -66,6 +66,9 @@ receivers:
             regex: '.*grpc_io.*'
             action: drop
   sapm:
+  smartagent/signalfx-forwarder:
+    type: signalfx-forwarder
+    listenAddress: 127.0.0.1:9080
```
flands, line 71:
```suggestion
    listenAddress: 0.0.0.0:9080
```

flands: Clearly debatable, but I believe all the other receivers above default to 0.0.0.0 -- if so we should do the same thing here

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -120,7 +123,7 @@ service:
   extensions: [health_check, http_forwarder, zpages]
   pipelines:
     traces:
-      receivers: [jaeger, otlp, sapm, zipkin]
+      receivers: [jaeger, otlp, sapm, smartagent/signalfx-forwarder, zipkin]
```
flands, line 126: I believe we should add this to metrics receivers below as well -- otherwise metric client libraries will not work by default

cmd/otelcol/config/collector/full_config_linux.yaml
```diff
@@ -533,7 +539,7 @@ service:
   # NOTE: Receivers expose ports so ensure no other application uses the same ports
   pipelines:
     traces:
-      receivers: [jaeger, otlp, sapm, zipkin]
+      receivers: [jaeger, otlp, sapm, smartagent/signalfx-forwarder, zipkin]
```
flands, line 542: Add to metrics?

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -27,6 +27,9 @@ receivers:
             regex: '.*grpc_io.*'
             action: drop
   sapm:
+  smartagent/signalfx-forwarder:
+    type: signalfx-forwarder
+    listenAddress: 0.0.0.0:9080
```
flands, line 32: Do we need this for gateway mode? We did not have this in signalfx gateway to my knowledge -- I think we should only enable this for agent mode.

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -83,7 +86,7 @@ service:
   extensions: [health_check, http_forwarder, zpages]
   pipelines:
     traces:
-      receivers: [jaeger, otlp, sapm, zipkin]
+      receivers: [jaeger, otlp, sapm, smartagent/signalfx-forwarder, zipkin]
```
flands, line 89: Remove?

cmd/otelcol/config/collector/otlp_config_linux.yaml
```diff
@@ -80,7 +83,7 @@ service:
   extensions: [health_check, http_forwarder, zpages]
   pipelines:
     traces:
-      receivers: [jaeger, otlp, sapm, zipkin]
+      receivers: [jaeger, otlp, sapm, smartagent/signalfx-forwarder, zipkin]
```
flands, line 86: Add to metrics?

cmd/otelcol/config/collector/agent_config.yaml
rmfitzpatrick: Thanks for spotting this.  Updating docs as well.

cmd/otelcol/config/collector/gateway_config.yaml
rmfitzpatrick: The smart gateway listened at this so I assumed we'd want the collector to as well.  If everything is reporting otlp then it's not necessary though.

flands: I think we should only do this for agent -- gateway mode config will be different and in gateway mode we should not enable signalfx-forwarder

flands (approved):
LGTM

PR #222, opened 2021-03-30
Include SA bundle in msi

rmfitzpatrick:
requires https://github.com/signalfx/splunk-otel-collector/pull/210

These changes add a slightly modified Smart Agent bundle to msi installations.  There is a path length limitation w/ WiX and Wine that prevents copying it completely that I wasn't able to find a workaround to while preserving signed builds in linux.


flands (approved):
LGTM

PR #225, opened 2021-03-31
Update splunk-forwarder-integration.md

rblake-splunk:
Removed _raw field from parsing that was present during testing - This will not work with log observer as it isn't present. - @flands


flands (approved):
Thanks!

PR #228, opened 2021-04-01
use SA Config directly in Smart Agent extension

rmfitzpatrick:
These changes update the Smart Agent Extension's config and custom unmarshaller to be based on the global agent config.  By doing so, we gain the defaults for all observed config values instead of just collectd and can provide better migration assistance and messaging in the future based on provided global configs.

Also includes setting the required path env vars that some monitors rely on: https://github.com/signalfx/signalfx-agent/blob/7bed2632de74e3e9359343d624dd6ce8d2452e07/pkg/core/config/config.go#L209


rmfitzpatrick:
I'm probably not going to be able to meet the diff coverage requirement, since three missing lines are inaccessible error paths and the last requires a collectd integration test.

internal/extension/smartagentextension/config.go
```diff
@@ -26,50 +26,62 @@ import (
...
+	}
+	config.Collectd.BundleDir = config.BundleDir
+
+	extensionCfg.Config = *config
+	return nil
+}
+
+func smartAgentConfigFromSettingsMap(settings map[string]interface{}) (*saconfig.Config, error) {
+	var config saconfig.Config
+	yamlTags := yamlTagsFromStruct(reflect.TypeOf(config))
```
asuresh4, line 67: Do you know if the SA top level config requires the recursive version of this method available in the receiver code? For the collectd config it wasn't required. I believe we discussed moving this method to a utils package. If you require the recursive version here, I can open a PR moving it since I said I'll do it but haven't :)

rmfitzpatrick: It would probably be preferable to what I have here.  Will do so in this PR if that's ok w/ you.

asuresh4: That's fine as well.

rmfitzpatrick: @asuresh4 after thinking about this I think the recursive version is unnecessary since the case correction only occurs at the root level, or here manually for the additional collectd updates.  In the future we will probable want a more robust helper to coerce arbitrary maps of maps to desired cases but that's not needed here imo.

asuresh4 (approved):

PR #229, opened 2021-04-01
Fix MSI release build for signed exe

jchengsfx:
- Move Dockerfile entrypoint command to `docker-entrypoint.sh`
- Copy the signed exe before the MSI release build to the staging directory and pass the `--otelcol` option to the build script


rmfitzpatrick (approved):
😅 thanks!

PR #235, opened 2021-04-02
SA Extension: Set platform specific bundleDir default

rmfitzpatrick:
These changes use saner defaults for the agent bundle path, which should match standard installation paths.  For windows, since the bundle is adjacent to the executable in standard installation we check for it and source it if found.

edit: I've also added uploading a windows run coverage report, but don't think I'll be hitting the requirement for this PR since one diff path requires running from an installed msi's perspective.

The significant drop in project coverage w/ these changes results from the windows build files now revealing they are without coverage.  The actual missing coverage from the diff is from a runtime only path, and one where a critical env var has been cleared (both relative edge cases).

### DIFF
--- a/internal/extension/smartagentextension/factory.go
+++ b/internal/extension/smartagentextension/factory.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/signalfx/signalfx-agent/pkg/core/common/constants"
 	saconfig "github.com/signalfx/signalfx-agent/pkg/core/config"
@@ -41,18 +42,26 @@ func NewFactory() component.ExtensionFactory {
 }
 
 var bundleDir = func() string {
-	out := os.Getenv(constants.BundleDirEnvVar)
-	if out == "" {
-		exePath, err := os.Executable()
-		if err != nil {
-			panic("Cannot determine agent executable path, cannot continue")
-		}
-		out, err = filepath.Abs(filepath.Join(filepath.Dir(exePath), ".."))
-		if err != nil {
-			panic("Cannot determine absolute path of executable parent dir " + exePath)
+	dir := os.Getenv(constants.BundleDirEnvVar)
+	if dir == "" {
+		if runtime.GOOS == "windows" {
+			pfDir := os.Getenv("programfiles")
+			if pfDir == "" {
+				pfDir = "C:\\Program Files"
+			}
+			dir = filepath.Join(pfDir, "Splunk", "OpenTelemetry Collector", "agent-bundle")
+			if exePath, err := os.Executable(); err == nil {
+				if colocatedBundle, err := filepath.Abs(filepath.Join(filepath.Dir(exePath), "agent-bundle")); err == nil {
+					if info, err := os.Stat(colocatedBundle); err == nil && info.IsDir() {
+						dir = colocatedBundle
+					}
+				}
+			}
+		} else {
+			dir = "/usr/lib/splunk-otel-collector/agent-bundle"
 		}
 	}
-	return out
+	return dir
 }()
 
 func createDefaultConfig() config.Extension {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -377,9 +377,14 @@ jobs:
             choco upgrade golang --version=1.15
             refreshenv
             go env -w CGO_ENABLED=0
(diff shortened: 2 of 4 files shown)


internal/extension/smartagentextension/factory.go
```diff
@@ -41,18 +42,24 @@ func NewFactory() component.ExtensionFactory {
...
-		if err != nil {
-			panic("Cannot determine agent executable path, cannot continue")
-		}
-		out, err = filepath.Abs(filepath.Join(filepath.Dir(exePath), ".."))
-		if err != nil {
-			panic("Cannot determine absolute path of executable parent dir " + exePath)
+	dir := os.Getenv(constants.BundleDirEnvVar)
+	if dir == "" {
+		if runtime.GOOS == "windows" {
+			dir = filepath.Join("C:\\", "Program Files", "Splunk", "OpenTelemetry Collector", "agent-bundle")
```
pmcollins, line 48:
```suggestion
			dir = filepath.Join("C:", "Program Files", "Splunk", "OpenTelemetry Collector", "agent-bundle")
```
I am not familiar with Windows -- will this work?

rmfitzpatrick: I think `C:relative_path` and `C:\\absolute_path` in windows, but I'm having trouble getting the desired absolute value w/ Join.

pmcollins: `filepath.Join("C:", "Program Files")` seems to work for me on a Windows VM -- it returns an absolute path and I can take a directory listing from it 🤷

rmfitzpatrick: Weird, the windows-test unit test run failed with that and `"C:\\", "Program Files"` but `"C:\\Program Files", ...` works

jrcamp: I think you'll want to `os.Getenv("programfiles")` since the default location can vary.

internal/extension/smartagentextension/factory.go
```diff
@@ -41,18 +42,24 @@ func NewFactory() component.ExtensionFactory {
...
-		if err != nil {
-			panic("Cannot determine agent executable path, cannot continue")
-		}
-		out, err = filepath.Abs(filepath.Join(filepath.Dir(exePath), ".."))
-		if err != nil {
-			panic("Cannot determine absolute path of executable parent dir " + exePath)
+	dir := os.Getenv(constants.BundleDirEnvVar)
+	if dir == "" {
+		if runtime.GOOS == "windows" {
+			dir = filepath.Join("C:\\Program Files", "Splunk", "OpenTelemetry Collector", "agent-bundle")
```
jrcamp, line 48: I think you'll want to os.Getenv("programfiles") since the default location can vary.

jrcamp (changes requested):

internal/extension/smartagentextension/factory.go
```diff
@@ -41,18 +42,26 @@ func NewFactory() component.ExtensionFactory {
...
+			dir = filepath.Join(pfDir, "Splunk", "OpenTelemetry Collector", "agent-bundle")
+			if exePath, err := os.Executable(); err == nil {
+				if colocatedBundle, err := filepath.Abs(filepath.Join(filepath.Dir(exePath), "agent-bundle")); err == nil {
+					if info, err := os.Stat(colocatedBundle); err == nil && info.IsDir() {
+						dir = colocatedBundle
+					}
+				}
+			}
+		} else {
+			dir = filepath.Join("/", "usr", "lib", "splunk-otel-collector", "agent-bundle")
```
pmcollins, line 61: Not sure how much value we're getting from `filepath.Join` since we're not on Windows in this block, but if we keep it, looks like we could also use `os.PathSeparator` instead of the explicit "/".

rmfitzpatrick: Updated to just use the unjoined/cleaned string.  I think I've mistaken this practice from the collector as a go convention but there's no real reason for it in this case.

pmcollins (approved):

internal/extension/smartagentextension/factory.go
rmfitzpatrick: Good idea, updated.

rmfitzpatrick:
@jrcamp mind taking another look?

jrcamp (approved):

PR #241, opened 2021-04-06
Include smart agent bundle in docker image

bjsignalfx:
- Install smart agent bundle in image.
- Set environment variables:
   - SPLUNK_BUNDLE_DIR="/usr/lib/splunk-otel-collector/agent-bundle"
   - SPLUNK_COLLECTD_DIR="/usr/lib/splunk-otel-collector/agent-bundle/run/collectd

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,7 @@ delete-tag:
 docker-otelcol:
 	GOOS=linux $(MAKE) otelcol
 	cp ./bin/otelcol_linux_amd64 ./cmd/otelcol/otelcol
-	docker build -t otelcol ./cmd/otelcol/
+	docker build -t otelcol --build-arg SMART_AGENT_RELEASE=$(SMART_AGENT_RELEASE) ./cmd/otelcol/
 	rm ./cmd/otelcol/otelcol
 
 .PHONY: binaries-all-sys
--- a/cmd/otelcol/Dockerfile
+++ b/cmd/otelcol/Dockerfile
@@ -1,16 +1,37 @@
-FROM alpine:3.12 as certs
+ARG SMART_AGENT_RELEASE=v5.9.1
+ARG ALPINE_VERSION=3.12
+
+FROM alpine:${ALPINE_VERSION} as certs
 RUN apk --update add ca-certificates
 
-FROM alpine:3.12 AS otelcol
+FROM alpine:${ALPINE_VERSION} AS otelcol
 COPY otelcol /
 # Note that this shouldn't be necessary, but in some cases the file seems to be
 # copied with the execute bit lost (see #1317)
 RUN chmod 755 /otelcol
 
+FROM alpine:${ALPINE_VERSION} AS smartagent
+ARG SMART_AGENT_RELEASE
+RUN mkdir -p "$(dirname "/usr/lib/splunk-otel-collector/agent-bundle")" && \
+    SMART_AGENT=signalfx-agent-${SMART_AGENT_RELEASE#v}.tar.gz && \
+    URL=https://github.com/signalfx/signalfx-agent/releases/download/${SMART_AGENT_RELEASE}/$SMART_AGENT && \
+    wget "$URL" && tar -xzf $SMART_AGENT && mv signalfx-agent /usr/lib/splunk-otel-collector/agent-bundle && \
+    # Absolute path of interpreter in smart agent dir is set in dependent binaries
+    # requiring the interpreter location not to change.
+    /usr/lib/splunk-otel-collector/agent-bundle/bin/patch-interpreter /usr/lib/splunk-otel-collector/agent-bundle && \
+    rm -f /usr/lib/splunk-otel-collector/agent-bundle/bin/signalfx-agent \
+          /usr/lib/splunk-otel-collector/agent-bundle/bin/agent-status \
+          $SMART_AGENT
+
 FROM scratch
+ENV SPLUNK_BUNDLE_DIR=/usr/lib/splunk-otel-collector/agent-bundle
+ENV SPLUNK_COLLECTD_DIR=$SIGNALFX_BUNDLE_DIR/run/collectd
 COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=otelcol /otelcol /
-COPY config/collector/gateway_config.yaml /etc/otel/collector/gateway_config.yaml
+# Maintaining interpreter location as required.
+COPY --from=smartagent /usr/lib/splunk-otel-collector/agent-bundle /usr/lib/splunk-otel-collector/agent-bundle
+COPY config/collector/gateway_config.yaml    /etc/otel/collector/gateway_config.yaml
 COPY config/collector/otlp_config_linux.yaml /etc/otel/collector/otlp_config_linux.yaml
+COPY config/collector/agent_config.yaml      /etc/otel/collector/agent_config.yaml
 ENTRYPOINT ["/otelcol"]
 EXPOSE 13133 14250 14268 55680 4317 6060 7276 8888 9411 9443


cmd/otelcol/Dockerfile
```diff
@@ -7,10 +9,16 @@ COPY otelcol /
 # copied with the execute bit lost (see #1317)
 RUN chmod 755 /otelcol
 
-FROM scratch
-COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+FROM quay.io/signalfx/signalfx-agent:${SMART_AGENT_VERSION_NUMBER}
```
rmfitzpatrick, line 12: I would advise against adding this dependency.  Using the standalone agent bundle and running the patch-interpreter script should be fine (what the deb and rpm jobs do): https://github.com/signalfx/signalfx-agent#linux-standalone-targz

rmfitzpatrick: https://github.com/signalfx/splunk-otel-collector/blob/a4dd11d6d68bfe3d408c49dd33b8cf260f0aa684/internal/buildscripts/packaging/fpm/common.sh#L68

bjsignalfx: Ok, I'm now using alpine:3.12 as parent image.

Makefile
```diff
@@ -42,7 +42,8 @@ endif
 BUILD_X3=-X $(BUILD_INFO_IMPORT_PATH).BuildType=$(BUILD_TYPE)
 BUILD_INFO=-ldflags "${BUILD_X1} ${BUILD_X2} ${BUILD_X3}"
 
-SMART_AGENT_RELEASE=v5.9.1
+SMART_AGENT_VERSION_NUMBER=5.9.1
+SMART_AGENT_RELEASE=v$(SMART_AGENT_VERSION_NUMBER)
```
rmfitzpatrick, line 46: seems unnecessary if only used once

cmd/otelcol/Dockerfile
```diff
@@ -1,16 +1,30 @@
...
 RUN apk --update add ca-certificates
 
-FROM alpine:3.12 AS otelcol
 COPY otelcol /
-# Note that this shouldn't be necessary, but in some cases the file seems to be
-# copied with the execute bit lost (see #1317)
+# Note that this shouldn't be necessary, but in some cases the file seems to be copied with the execute bit lost (see #1317)
 RUN chmod 755 /otelcol
 
-FROM scratch
```
rmfitzpatrick, line 10: I think having a scratch base is desired overall.  Is this still possible?

Makefile
bjsignalfx: I removed SMART_AGENT_VERSION_NUMBER and used the "subst" to extract the version number.

cmd/otelcol/Dockerfile
```diff
@@ -1,16 +1,34 @@
...
 COPY otelcol /
 # Note that this shouldn't be necessary, but in some cases the file seems to be
 # copied with the execute bit lost (see #1317)
 RUN chmod 755 /otelcol
 
+FROM alpine:${ALPINE_VERSION} AS smartagent
+ARG SMART_AGENT_RELEASE_URL
+RUN  wget -O /signalfx-agent.tar.gz "$SMART_AGENT_RELEASE_URL" && \
+     tar -xzf /signalfx-agent.tar.gz && \
+     /signalfx-agent/bin/patch-interpreter /signalfx-agent && \
```
rmfitzpatrick, line 16: does this need to be run in the scratch stage?

Makefile
```diff
@@ -191,7 +191,7 @@ delete-tag:
 docker-otelcol:
 	GOOS=linux $(MAKE) otelcol
 	cp ./bin/otelcol_linux_amd64 ./cmd/otelcol/otelcol
-	docker build -t otelcol ./cmd/otelcol/
+	docker build -t otelcol ./cmd/otelcol/ --build-arg SMART_AGENT_RELEASE_URL=https://github.com/signalfx/signalfx-agent/releases/download/$(SMART_AGENT_RELEASE)/signalfx-agent-$(subst v,,$(SMART_AGENT_RELEASE)).tar.gz
```
rmfitzpatrick, line 194: nit, mind specifying the context as last arg?

rmfitzpatrick:
Is this "ready for review"?

cmd/otelcol/Dockerfile
```diff
@@ -1,16 +1,34 @@
...
+
+ENV SPLUNK_BUNDLE_DIR="/usr/lib/splunk-otel-collector/agent-bundle"
+COPY --from=smartagent /signalfx-agent $SPLUNK_BUNDLE_DIR
+ENV SPLUNK_COLLECTD_DIR=$SPLUNK_BUNDLE_DIR/run/collectd
+
+COPY config/collector/gateway_config.yaml    /etc/otel/collector/gateway_config.yaml
 COPY config/collector/otlp_config_linux.yaml /etc/otel/collector/otlp_config_linux.yaml
+COPY config/collector/agent_config.yaml      /etc/otel/collector/agent_config.yaml
+
+ENV SPLUNK_CONFIG=/etc/otel/collector/agent_config.yaml
```
rmfitzpatrick, line 31: @jchengsfx @jrcamp can you confirm this is desired?

cmd/otelcol/Dockerfile
rmfitzpatrick: I missed the point entirely

cmd/otelcol/Dockerfile
rmfitzpatrick: If so I think we need to set a number of default env vars here.  May be best to defer to a subsequent PR.

jchengsfx: Yes, the agent config requires more env vars to be set.  Not sure if we want to change the default config from gateway to agent, but maybe @jrcamp will know best.

Makefile
jchengsfx, line 194: Instead of passing the url as a build arg, would it be simpler to pass `SMART_AGENT_RELEASE` as a build arg, and set the url in the dockerfile, i.e. `--build-arg SMART_AGENT_RELEASE=$(SMART_AGENT_RELEASE)`?

Also, we'll need to set a default for `SMART_AGENT_RELEASE` either in the Makefile, and/or update the dockerfile to handle `SMART_AGENT_RELEASE` correctly if it is empty, so that we can run `make docker-otelcol` without any args.

jchengsfx (changes requested):

cmd/otelcol/Dockerfile
bjsignalfx: I'll to remove this env var.

Makefile
bjsignalfx: No, I don't mind.

bjsignalfx: > Also, we'll need to set a default for `SMART_AGENT_RELEASE` either in the Makefile, and/or update the dockerfile to handle `SMART_AGENT_RELEASE` correctly if it is empty, so that we can run `make docker-otelcol` without any args.

So, what should the default SMART_AGENT_RELEASE be? v5.9.1?

jchengsfx: For now, I think v5.9.1 is fine since that is what the deb, rpm, and msi packages include.  I'll defer to @rmfitzpatrick if we want to upgrade to the latest v5.9.2 release.

cmd/otelcol/Dockerfile
```diff
@@ -1,16 +1,36 @@
...
+FROM alpine:${ALPINE_VERSION} AS otelcol
 COPY otelcol /
 # Note that this shouldn't be necessary, but in some cases the file seems to be
 # copied with the execute bit lost (see #1317)
 RUN chmod 755 /otelcol
 
+FROM alpine:${ALPINE_VERSION} AS smartagent
+ARG SMART_AGENT_RELEASE
+RUN mkdir -p "$(dirname "/usr/lib/splunk-otel-collector/agent-bundle")" && \
+    SMART_AGENT=signalfx-agent-${SMART_AGENT_RELEASE//"v"/""}.tar.gz && \
```
jchengsfx, line 16:
```suggestion
    SMART_AGENT=signalfx-agent-${SMART_AGENT_RELEASE#v}.tar.gz && \
```
Just a nit and probably wouldn't ever be an issue, but maybe best to just strip the leading `v` instead of replacing all occurrences.

cmd/otelcol/Dockerfile
```diff
@@ -1,16 +1,37 @@
...
+    wget "$URL" && tar -xzf $SMART_AGENT && mv signalfx-agent /usr/lib/splunk-otel-collector/agent-bundle && \
+    # Absolute path of interpreter in smart agent dir is set in dependent binaries
+    # requiring the interpreter location not to change.
+    /usr/lib/splunk-otel-collector/agent-bundle/bin/patch-interpreter /usr/lib/splunk-otel-collector/agent-bundle && \
+    rm -f /usr/lib/splunk-otel-collector/agent-bundle/bin/signalfx-agent \
+          /usr/lib/splunk-otel-collector/agent-bundle/bin/agent-status \
+          $SMART_AGENT
+
 FROM scratch
+ENV SIGNALFX_BUNDLE_DIR=/usr/lib/splunk-otel-collector/agent-bundle
```
jchengsfx, line 27:
```suggestion
ENV SPLUNK_BUNDLE_DIR=/usr/lib/splunk-otel-collector/agent-bundle
```

jchengsfx (dismissed):

internal/extension/smartagentextension/factory.go
```diff
@@ -64,6 +66,12 @@ func createDefaultConfig() config.Extension {
 	}
 	cfg.BundleDir = bundleDir
 	cfg.Collectd.BundleDir = bundleDir
+
+	dir := strings.TrimSpace(os.Getenv(collectdConfigDirEnvVarName))
+	if dir != "" && strings.TrimSpace(cfg.Collectd.ConfigDir) == "" {
```
rmfitzpatrick, line 71: A few things going on here:

`cfg.Collectd.ConfigDir` is the agent's default value so this empty string value should never happen: https://github.com/signalfx/signalfx-agent/blob/master/pkg/core/config/config.go#L387

The custom unmarshaller makes the default config not actually the default and needs to be pulled from directly like is happening w/ the bundleDir here: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/extension/smartagentextension/config.go#L56

I'm not 100% we want to use that env var value here since it's just used by the connector installation path (I see that the readme is vague though).  What if we were to use `Filepath.Join(cfg.BundleDir, "run", "collectd")` in the custom unmarshaller and leave this code unchanged?

rmfitzpatrick: You could also take care of this in a separate PR since it's not required for bundling in docker afaict and just rely on your added `ENV SPLUNK_COLLECTD_DIR=$SIGNALFX_BUNDLE_DIR/run/collectd` line, unless that's not taking effect.

bjsignalfx: I deleted this code and moved the logic to configure collectd config dir to smartagentextension/config.go.

internal/extension/smartagentextension/config.go
```diff
@@ -58,6 +59,10 @@ func customUnmarshaller(componentViperSection *viper.Viper, intoCfg interface{})
 	}
 	config.Collectd.BundleDir = config.BundleDir
 
+	if config.Collectd.ConfigDir == "/var/run/signalfx-agent/collectd" {
```
rmfitzpatrick, line 62: This prevents using the default value which may be specified by the user config if they are using an existing agent installation (edge case).  You can check for it being set in the `allSettings` map before it's coerced to a saconfig.Config and simply use this "was specified" check instead of tracking the default value.

rmfitzpatrick: You can also open these changes in another PR unless they are necessary (not sure they are).

bjsignalfx: Will address these changes in another PR.

rmfitzpatrick (approved):

PR #251, opened 2021-04-07
Fix support bundle script and tests

jchengsfx:
- The support bundle script currently exits with non-zero code because `TMPDIR` does not exist before `tee -a "$TMPDIR"/stdout.log`
- Need to temporarily skip bundle support script tests until these changes are included in the next release, otherwise CI tests will continue to fail
- Install wget in the test images as a requirement for the support bundle script


flands (approved):
LGTM

PR #261, opened 2021-04-12
docker-otelcol circleci job and Makefile updates

jchengsfx:
- Add `docker-otelcol` job to build and run the image before publishing
- Add `msi` make target
- Add `SKIP_COMPILE` Makefile var so that we can use the make targets to build the deb, rpm, msi, and docker image in circleci without recompiling the otelcol binaries

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -349,7 +349,7 @@ jobs:
       - run:
           name: Run docker image
           command: |
-            docker run -d -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 --name otelcol otelcol:latest
+            docker run -d -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=fake-realm --name otelcol otelcol:latest
             sleep 10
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -328,6 +337,33 @@ jobs:
...
+    steps:
+      - attach_to_workspace
+      - setup_remote_docker
+      - run:
+          name: Build docker image
+          command: make docker-otelcol SKIP_COMPILE=true
+      - run:
+          name: Run docker image
+          command: |
+            docker run -d -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 --name otelcol otelcol:latest
```
rmfitzpatrick, line 352: Don't need to address now, but it's preferable to not spam us0 and instead use a hosts entry and mock service.

jchengsfx: Makes sense. I'll just change it to some fake realm for now since we're not validating anything here other than the docker image can start and run.

rmfitzpatrick (approved):

PR #277, opened 2021-04-15
Add configsource from core

pjanotti:
This brings config source code from core that is not public yet. The goal of having it here is that the config sources can be, at least initially, added to the repo/ This code was already mostly reviewed on core the parts that are new here are the parser, builder, and helpers. The first two simply follow the pattern of config loader and builder on the core repo but specific for config sources. The helpers have just a struct and a very simple function - usage for those will become clear as code for actual config sources are brought to the repo.

cc @owais

### DIFF
--- a/internal/configsource/builder_test.go
+++ b/internal/configsource/builder_test.go
@@ -1,0 +1,152 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configsource
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config"
+)
+
+func TestConfigSourceBuild(t *testing.T) {
+	ctx := context.Background()
+	params := CreateParams{
+		Logger:               zap.NewNop(),
+		ApplicationStartInfo: component.ApplicationStartInfo{},
+	}
+
+	testFactories := Factories{
+		"tstcfgsrc": &mockCfgSrcFactory{},
+	}
+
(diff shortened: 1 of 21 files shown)


internal/configsource/builder_test.go
```diff
@@ -38,11 +37,11 @@ func TestConfigSourceBuild(t *testing.T) {
 	}
 
 	tests := []struct {
-		name               string
 		configSettings     map[string]ConfigSettings
 		factories          Factories
 		expectedCfgSources map[string]ConfigSource
 		wantErr            error
+		name               string
```
pjanotti, line 44: I got errors like `fieldalignment: struct with 56 pointer bytes could be 48 (govet)` for structs like this one that's why the strange ordering.

rmfitzpatrick: It would be nice if we had a formatter make target that runs the necessary fieldalignment commands.

tigrannajaryan:
Is this a duplicate of the code that already exists in the core? I am not sure I understand.

rmfitzpatrick:
@pjanotti I'm not sure it's totally necessary for us to review what is already in core in addition to the new bits.  Maybe if this approach is agreed on by @tigrannajaryan you can open an initial PR that's c/p and then it's easier to know what hasn't been reviewed?

I do wonder what other internal core items are necessary for this to be functional and it'd be good to confirm that we don't need to update much of how the application is loaded in general (ideally this wouldn't be internal in core).  Hopefully tracking from core directly will make this easier.

internal/configsource/helpers.go
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configsource
+
+// WatcherNotSupported is the a watcher function that always returns ErrWatcherNotSupported.
+func WatcherNotSupported() error {
+	return ErrWatcherNotSupported
```
owais, line 19: Is this supported to be used by sources? Why not just return ErrWatcherNotSupported? Doesn't look like this will help much.

owais (approved):

tigrannajaryan:
We can now use results of https://github.com/open-telemetry/opentelemetry-collector/pull/2948 instead of copying the code.

internal/configsource/helpers.go
pjanotti: Yeah, that is very trivial to implement but I found myself repeating that code many times. I was repeating that in a few places - at least vault and an environment config source will use it. Let's re-evaluate as we bring these to the repo.

pjanotti:
@rmfitzpatrick @tigrannajaryan let me close this one and bring only the parts not exposed on core - but will actually start with the actual vault config source.

PR #281, opened 2021-04-15
Add filelog receiver

flands:


tigrannajaryan:
I think we need a line for filelog in go.mod too.

PR #295, opened 2021-04-16
Add configparser package

pjanotti:
This package will be used to implement configuration loading that will leverage config sources. This PR only includes code that was not reviewed on the core repository. This is basically the boiler-plate code needed to load config sources from the configuration file: read config sources from configuration and build them. For these, we need factory and default config settings (for the other components these are defined on core).


pjanotti:
This one supersedes https://github.com/signalfx/splunk-otel-collector/pull/277

internal/configprovider/parser_test.go
```diff
@@ -0,0 +1,141 @@
+// Copyright The OpenTelemetry Authors
```
dmitryax, line 1: Should we have on top as well?
```
// Copyright Splunk Inc.
```

dmitryax (approved):

internal/configprovider/parser_test.go
pjanotti: Ah good point, I was doing this initially on the OTel repo. Let me update that.

PR #301, opened 2021-04-17
Explicitly define receivers endpoints

dmitryax:
For better user experience, we want to have default endpoint values always to be explicitly set in the configs, so user don't have to check receiver docs in order to find out which port data should be sent to.

### DIFF
--- a/examples/kubernetes-yaml/splunk-otel-collector-agent.yaml
+++ b/examples/kubernetes-yaml/splunk-otel-collector-agent.yaml
@@ -10,7 +10,10 @@ data:
   otel-collector-config: |
     extensions:
       health_check:
+        endpoint: 0.0.0.0:13133
       http_forwarder:
+        ingress:
+          endpoint: 0.0.0.0:6060
         egress:
           # TODO: Ensure this is set properly
           endpoint: "https://api.${SPLUNK_REALM}.signalfx.com"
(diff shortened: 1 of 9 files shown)


examples/kubernetes-yaml/splunk-otel-collector-agent.yaml
```diff
@@ -104,6 +113,7 @@ data:
         endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace"
       # Metrics
       signalfx:
+        endpoint: 0.0.0.0:9943
```
rmfitzpatrick, line 116: This is an exporter

examples/kubernetes-yaml/splunk-otel-collector-gateway.yaml
```diff
@@ -104,6 +113,7 @@ data:
         endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace"
       # Metrics
       signalfx:
+        endpoint: 0.0.0.0:9943
```
rmfitzpatrick, line 116: also an exporter

flands:
@dmitryax Please add http_forwarder ingress endpoint too

dmitryax:
@flands http_forwarder ingress is added

flands (approved):
LGTM @rmfitzpatrick ?

rmfitzpatrick (approved):

PR #302, opened 2021-04-17
Fix config examples

dmitryax:
- open gateway service ports for defined receivers
- add required fields in example exporters

### DIFF
--- a/examples/kubernetes-yaml/splunk-otel-collector-gateway.yaml
+++ b/examples/kubernetes-yaml/splunk-otel-collector-gateway.yaml
@@ -137,10 +137,18 @@ spec:
     port: 13133
   - name: jaeger-grpc # Default endpoint for Jaeger gRPC receiver.
     port: 14250
-  - name: jaeger-thrift # Default endpoint for Jaeger Thrift receiver.
+  - name: thrift-binary # Default endpoint for Jaeger Thrift receiver.
+    port: 6832
+    protocol: UDP
+  - name: thrift-compact
+    port: 6831
+    protocol: UDP
+  - name: thrift-http # Default endpoint for Jaeger Thrift receiver.
     port: 14268
   - name: otlp # Default endpoint for OpenTelemetry gRPC receiver.
     port: 4317
+  - name: otlp-http # Default endpoint for OpenTelemetry gRPC receiver.
+    port: 55681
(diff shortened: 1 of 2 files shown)


examples/kubernetes-yaml/splunk-otel-collector-gateway.yaml
```diff
@@ -137,10 +137,16 @@ spec:
     port: 13133
   - name: jaeger-grpc # Default endpoint for Jaeger gRPC receiver.
     port: 14250
-  - name: jaeger-thrift # Default endpoint for Jaeger Thrift receiver.
+  - name: thrift-binary # Default endpoint for Jaeger Thrift receiver.
+    port: 6832
```
flands, line 141: Are these needed for gateway mode? They are udp

dmitryax: Yes, otherwise ports are not exposed on the k8s service. If UDP I need to update protocol then

dmitryax: updated

rmfitzpatrick: Does removing the jaeger prefix shift some guessing work to the user?  `jaeger-thrift-binary`, etc.

dmitryax: K8s has a limit for port names - it must be less then 16 characters

flands: This is gateway mode -- why would udp ports be required?

dmitryax: It’s not required but nice to have if user wants to send traces directly to gateway I guess. The PR is just to update service deployment to what’s already configured on the gateway. We can either change the gateway config - remove jaeger receivers or expose ports for them in the service deployment

rmfitzpatrick (approved):

examples/kubernetes-yaml/splunk-otel-collector-gateway.yaml
dmitryax: just to clarify: the getaway already has the binary thrift receiver configured, but the port is not enabled on the service deployment - this PR fixes it

PR #303, opened 2021-04-17
Add the config source manager

pjanotti:
A follow-up to #295 that superseded #277.

This brings the config source manager from the core repository that is not publicly exposed. The manager will be used to implement a provider that uses config sources to inject data in configuration. This code was originally reviewed on the collector core repository.

/cc @owais

### DIFF
--- a/internal/configprovider/manager.go
+++ b/internal/configprovider/manager.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package configsource
+package configprovider
 
 import (
 	"bytes"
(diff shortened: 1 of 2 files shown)


tigrannajaryan:
I do not quite understand why are we not implementing the configsource support in Collector core. Why do we need the config manager here? I thought it can be internal in Collector core and still do its job. I am confused.

pjanotti:
> I do not quite understand why are we not implementing the configsource support in Collector core. Why do we need the config manager here? I thought it can be internal in Collector core and still do its job. I am confused.

It is a good question: internal is not visible on other repositories - so until that is moved outside internal we can't use it. The obvious risk is the code diverging, but, that is the point here per my understanding: we want to experiment before making it public on the core repository.

PR #304, opened 2021-04-18
Add the Vault config source

pjanotti:
Add a config source for Vault. This is based on the SA implementation for a config source.


internal/configsource/vaultconfigsource/README.md
```diff
@@ -0,0 +1,88 @@
...
+    # path is the Vault path to the secret location.
+    path: secret/data/kv
+    # poll_interval is used only for non-dynamic V2 K/V secret stores. It is
+    # the interval in which the config source will check for changes on the
+    # data on the given Vault path. Defaults to 1 minute if not specified.
+    poll_interval: 90s
+    # auth is a section used to indicate the authentication method to be used.
+    # Exactly one method must be specified, it must be one of the following:
+    # "token", "iam", or "gcp".
+    auth:
```
tigrannajaryan, line 31: There is an initiative at Otel Collector to define extensible auth mechanism for outgoing connections. It would be good to see if we can reuse it here. It is not yet ready at Otel so hard to tell but once it is I think we should evaluate it. This means that the config here may need to break in the near future, but I think that's OK.

internal/configsource/vaultconfigsource/README.md
```diff
@@ -0,0 +1,88 @@
+# Vault Config Source
```
tigrannajaryan, line 1: Can we label this component as "Alpha"?

internal/configsource/vaultconfigsource/README.md
```diff
@@ -0,0 +1,88 @@
...
+      # gcp is used on GCP deployments to generate the required Vault token.
+      # For details about each of the settings below, see
+      # https://github.com/hashicorp/vault-plugin-auth-gcp/blob/e1f6784b379d277038ca0661606aa8d23791e392/plugin/cli.go#L138
+      gcp:
+        role: role
+        mount: gcp
+        credentials: json_string # This setting is not recommended.
+        jwp_ext: 10m
+        service_account: some_account
+        project: project_id
```
tigrannajaryan, line 54: I assume these are copied from SmartAgent config, right?

internal/configsource/vaultconfigsource/session.go
```diff
@@ -0,0 +1,318 @@
...
+	// value read from the vault secret.
+	watchForUpdateFn := watcherNotSupported
+
+	if v.secret == nil {
+		if err := v.readSecret(); err != nil {
+			return nil, err
+		}
+
+		// The keys come all from the same secret so creating a watcher only for the
+		// first it is fine.
```
tigrannajaryan, line 69:
```suggestion
		// first is fine.
```

internal/configsource/vaultconfigsource/session.go
```diff
@@ -0,0 +1,318 @@
...
+	return &vaultSession{
+		logger:       logger,
+		client:       client,
+		path:         path,
+		pollInterval: pollInterval,
+		doneCh:       make(chan struct{}),
+	}, nil
+}
+
+func (v *vaultSession) readSecret() error {
```
tigrannajaryan, line 110: Is this function expected to leave a `nil` in `v.secret` in case of error?

internal/configsource/vaultconfigsource/session.go
```diff
@@ -0,0 +1,318 @@
...
+		Version:   versionInt,
+	}
+}
+
+// Allows key to be dot-delimited to traverse nested maps.
+func traverseToKey(data map[string]interface{}, key string) interface{} {
+	parts := strings.Split(key, ".")
+
+	for i := 0; ; i++ {
+		partVal := data[parts[i]]
```
tigrannajaryan, line 303: `parts` may be empty, right?

internal/configsource/vaultconfigsource/README.md
pjanotti: Right, this is following the same docs/guidelines.

internal/configsource/vaultconfigsource/session.go
pjanotti: Yes, the expectation is that the caller will stop trying to retrieve data, since this an error that triggers configuration failure, but if called again it will try once more to retrieve the secret.

internal/configsource/vaultconfigsource/session.go
pjanotti: Because the `sep` parameter is never empty it parts always have at least one element. I will add a comment to make it clear.

internal/configsource/vaultconfigsource/README.md
pjanotti: Done

internal/configsource/vaultconfigsource/session.go
pjanotti: Added a comment to make clear that this is the intended behavior.

internal/configsource/vaultconfigsource/session.go
pjanotti: Added comment.

tigrannajaryan (approved):

PR #310, opened 2021-04-20
Linux installer updates

jchengsfx:
- Update doc for fluentd dependencies and uninstall info
- Update uninstall to stop services before removing packages and for `apt-get purge` to remove unmodified managed config files
- Re-enable support-bundle tests


flands (approved):
LGTM

PR #313, opened 2021-04-20
Set PATH to include smart agent bundled utilities

jrcamp:
We already ship bash, tar, curl etc. as part of smart agent bundle. Make this
easier to use by adding it to the default PATH. This allows you to exec into a
running container. (You could do this already by supplying the whole path if
you knew where they were.)

Added dependabot for dockerfiles to ensure things stay updated.

Cleanup Dockerfile.


cmd/otelcol/Dockerfile
```diff
@@ -1,18 +1,15 @@
-ARG SMART_AGENT_RELEASE=v5.9.1
```
dmitryax, line 1: Don't we need the default value for `SMART_AGENT_RELEASE`?

jrcamp: It's set in the Makefile so removed it from here so it's not defined in multiple places.

rmfitzpatrick (approved):

PR #315, opened 2021-04-21
include kafkametricsreceiver to collect and ingest kafka consumer lag

harnitsignalfx:
This receiver was added by Splunk to be able to collect and ingest consumer lag from kafka. Please let me know if you require changes to go.mod or should I only include changes to the components file.


jrcamp:
You'll want to wait for #314 to be merged then rebase on that. https://github.com/signalfx/splunk-otel-collector/blob/main/docs/components.md will need updating as well.

harnitsignalfx:
ack, thanks Jay! Will do. we can close this one.

PR #325, opened 2021-04-23
Expose initial and effective config for debugging purposes

pjanotti:
Exposes config via an HTTP server at localhost if explicitly requested. Expose the config by setting the environment variable "SPLUNK_DEBUG_CONFIG_SERVER" to "true" (it won't work for any other value). By default it exposes the HTTP server at localhost:55555 with the following paths:

- /debug/configz/initial
- /debug/configz/effective

The first shows the initial configuration used by the collector the second the configuration actually in use by the collector. If the configuration is reloaded due to some update triggered by a config source the endpoint is briefly unavailable until the updated configuration is reloaded.

The exposed port is controlled by the env var "SPLUNK_CONFIG_SERVER_PORT" - if it is explicitly set to the empty string, e.g. SPLUNK_DEBUG_CONFIG_SERVER_PORT="" it disables the HTTP server.

On the effective configuration, it performs a simple attempt to redact all string values for which the key contains strings like "token", "secret", etc. This covers all components currently in use by the collector a more generic solution is desirable.

### DIFF
--- a/internal/configprovider/config_server.go
+++ b/internal/configprovider/config_server.go
@@ -152,6 +152,7 @@ func shouldRedactKey(k string) bool {
 		"login",
 		"password",
 		"pwd",
+		"token",
 		"user",
 	}
 
--- a/internal/configprovider/config_server.go
+++ b/internal/configprovider/config_server.go
@@ -1,0 +1,145 @@
+// Copyright Splunk, Inc.
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configprovider
+
+import (
+	"net"
+	"net/http"
+	"os"
+	"strings"
+
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	defaultEndpoint = "localhost:5555"
+)
+
+type configServer struct {
+	logger    *zap.Logger
+	initial   map[string]interface{}
+	effective map[string]interface{}
+	server    *http.Server
+	doneCh    chan struct{}
+}
+
+func newConfigServer(logger *zap.Logger, initial, effective map[string]interface{}) *configServer {
+	return &configServer{
+		logger:    logger,
+		initial:   initial,
+		effective: effective,
+	}
+}
+
+func (cs *configServer) start() error {
+	endpoint := defaultEndpoint
+	if portOverride, ok := os.LookupEnv("SPLUNK_CONFIG_SERVER_PORT"); ok {
+		if portOverride == "" {
+			// If explicitly set to empty do not start the server.
+			return nil
+		}
+
+		endpoint = "localhost:" + portOverride
+	}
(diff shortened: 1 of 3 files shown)


internal/configprovider/config_server.go
```diff
@@ -0,0 +1,165 @@
...
+		redactedConfig[k] = v
+	}
+
+	return redactedConfig
+}
+
+// shouldRedactKey applies a simple check to see if the contents of the given key
+// should be redacted or not.
+func shouldRedactKey(k string) bool {
+	fragments := []string{
```
pjanotti, line 145: This is a simple solution for the time being. A better solution in my view will be to have a metadata annotation indicating which configuration fields must not be displayed on the configuration.

internal/configprovider/config_server.go
```diff
@@ -0,0 +1,165 @@
...
+		logger:    logger,
+		initial:   initial,
+		effective: effective,
+	}
+}
+
+func (cs *configServer) start() error {
+	endpoint := defaultConfigServerEndpoint
+	if portOverride, ok := os.LookupEnv(defaultConfigServerPortEnvVar); ok {
+		if portOverride == "" {
```
pjanotti, line 53: I am making the default to expose the config on the default port let me know if this should be disabled by default.

internal/configprovider/config_server.go
```diff
@@ -0,0 +1,165 @@
...
+		"access",
+		"api_key",
+		"apikey",
+		"auth",
+		"credential",
+		"creds",
+		"login",
+		"password",
+		"pwd",
+		"user",
```
tigrannajaryan, line 155: Maybe also "token"?

tigrannajaryan (approved):
Do we want to upstream this too?

internal/configprovider/config_server.go
pjanotti: 👁️ deleted by accident when reformating 🤦

pjanotti:
> Do we want to upstream this too?

My idea at this moment is to add a mechanism for the insertion of custom zPages and use it to add these pages but I'm not sure yet about the best path to add this support on the zPages extension.

jrcamp:
> My idea at this moment is to add a mechanism for the insertion of custom zPages and use it to add these pages but I'm not sure yet about the best path to add this support on the zPages extension.

@pjanotti I'm interested in this for a different use case but have some ideas if you want to discuss sometime.

internal/configprovider/config_server.go
```diff
@@ -0,0 +1,166 @@
...
+	"strings"
+
+	"github.com/spf13/cast"
+	"go.uber.org/zap"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	defaultConfigServerEndpoint   = "localhost:55555"
+	defaultConfigServerPortEnvVar = "SPLUNK_CONFIG_SERVER_PORT"
```
rmfitzpatrick, line 31:
```suggestion
	configServerPortEnvVar = "SPLUNK_CONFIG_SERVER_PORT"
```
the env var isn't configurable, so isn't a default?

internal/configprovider/config_server.go
rmfitzpatrick: Even w/ redacted fields, potentially exposing sensitive info by default doesn't seem ideal imo.  What about adding a flag like `SPLUNK_DEBUG_CONFIG_SERVER` that defaults to false if this can't easily be integrated w/ the config providers upstream?

pjanotti: This will require a re-start but it seems fair to do it at least for initial implementation.

internal/configprovider/config_server.go
pjanotti: @pmcollins this may be something to be done together with annotating the configuration for documentation, i.e., we could annotate the fields with info to indicate if they are secrets and not display those annotated as secrets.

PR #339, opened 2021-04-27
Add links to component READMEs

asuresh4:
Add links to components in core/contrib.

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -8,21 +8,54 @@ These components are considered stable. While in beta, breaking changes may be
 introduced in a new release. In addition, any of these components may be
 removed prior to the 1.0 release.
 
-| Receivers        | Processors           | Exporters  | Extensions    |
-| :--------------: | :--------:           | :-------:  | :--------:    |
-| fluentforward    | attributes           | file       | healthcheck   |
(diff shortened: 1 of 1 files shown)


pjanotti (approved):
LGTM

jrcamp:
The links would still work in the table wouldn't they?

asuresh4:
> The links would still work in the table wouldn't they?

Yes, they would. In fact, I did that originally but the diff was hard to read. That's how I ended up with a list. Do we prefer having a table instead of a list?

jrcamp:
Yeah I can see how diff would be hard to read :(. I do think the end result was easier to read and more compact though. I think we should autogenerate this file anyway (or at least this part). If don't want to do that now I'd err on the side of end result being better and can autogenerate later.

jrcamp:
Actually I guess we'd need a way to mark components alpha/beta. Would need to be done upstream in a consistent way. (Maybe factory can return metadata that contains stability status). Not worth blocking this though obviously.

asuresh4:
Ok, in that case I'll restore the table formatting for now. Will update PR shortly.

asuresh4:
@jrcamp ptal

jrcamp (approved):

Issue #322, opened 2021-04-22
how to fetch kubernetes volumes metrics


xp-1000:
Hello,

I am trying to replace signalfx smart agent by otel collector since its deprecation.

I started with a simple configuration for host, kubernetes and nginx metrics.

```
exporters:
  sapm:
    access_token: ${SPLUNK_ACCESS_TOKEN}
    endpoint: https://ingest.eu0.signalfx.com/v2/trace
  signalfx:
    access_token: ${SPLUNK_ACCESS_TOKEN}
    api_url: https://api.eu0.signalfx.com
    correlation: null
    ingest_url: https://ingest.eu0.signalfx.com
    sync_host_metadata: true
extensions:
  health_check: null
  k8s_observer:
    auth_type: serviceAccount
    node: ${K8S_NODE_NAME}
  zpages: null
processors:
  batch: null
  k8s_tagger:
    extract:
      metadata:
      - namespace
      - node
      - podName
      - podUID
    filter:
      node_from_env_var: K8S_NODE_NAME
  memory_limiter:
    ballast_size_mib: ${SPLUNK_BALLAST_SIZE_MIB}
    check_interval: 5s
    limit_mib: ${SPLUNK_MEMORY_LIMIT_MIB}
  resource:
    attributes:
    - action: insert
      key: host.name
      value: ${K8S_NODE_NAME}
    - action: insert
      key: k8s.node.name
      value: ${K8S_NODE_NAME}
    - action: insert
      key: k8s.cluster.name
      value: kubeoteltest
  resource/add_agent_k8s:
    attributes:
    - action: insert
      key: k8s.pod.name
      value: ${K8S_POD_NAME}
    - action: insert
      key: k8s.pod.uid
      value: ${K8S_POD_UID}
    - action: insert
      key: k8s.namespace.name
      value: ${K8S_NAMESPACE}
  resourcedetection:
    detectors:
    - system
    - env
    override: false
    timeout: 10s
receivers:
  hostmetrics:
    collection_interval: 10s
    scrapers:
      cpu: null
      disk: null
      filesystem: null
      load: null
      memory: null
      network: null
      paging: null
      processes: null
  jaeger:
    protocols:
      grpc:
        endpoint: 0.0.0.0:14250
      thrift_http:
        endpoint: 0.0.0.0:14268
  kubeletstats:
    auth_type: serviceAccount
    collection_interval: 10s
    endpoint: ${K8S_NODE_IP}:10250
[...]

PR #345, opened 2021-04-28
Fix config parsing for structs and pointers to structs

asuresh4:
Addresses the following -

- Currently yamlTags from pointers to structs aren't accounted for
- Capitalized yaml tags aren't applied recursively on structs

### DIFF
--- a/internal/receiver/smartagentreceiver/config.go
+++ b/internal/receiver/smartagentreceiver/config.go
@@ -27,6 +27,8 @@ import (
 	"github.com/signalfx/signalfx-agent/pkg/monitors"
 	"go.opentelemetry.io/collector/config"
 	"gopkg.in/yaml.v2"
+
+	"github.com/signalfx/splunk-otel-collector/internal/utils"
 )
 
 const defaultIntervalSeconds = 10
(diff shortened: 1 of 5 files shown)


internal/receiver/smartagentreceiver/config.go
```diff
@@ -99,13 +99,7 @@ func (cfg *Config) Unmarshal(componentParser *config.Parser) error {
...
 	// case tag key names and update the keys where applicable.
 	yamlTags := yamlTagsFromStruct(monitorConfigType)
-	for key, val := range allSettings {
-		updatedKey := yamlTags[key]
-		if updatedKey != "" {
-			delete(allSettings, key)
-			allSettings[updatedKey] = val
-		}
-	}
+	recursivelyCapitalizeConfigKeys(allSettings, yamlTags)
```
asuresh4, line 102: I believe this should fix the issue reported in #322, not being able to set `authType` in `kubelet.APIConfig`.

internal/receiver/smartagentreceiver/config.go
```diff
@@ -131,6 +125,19 @@ func (cfg *Config) Unmarshal(componentParser *config.Parser) error {
 	return nil
 }
 
+func recursivelyCapitalizeConfigKeys(settings map[string]interface{}, yamlTags map[string]string) {
```
rmfitzpatrick, line 128: Since this is a viper-centric issue I wonder if `yamlTagsFromStruct` and `recursivelyCapitalizeConfigKeys` should be in a parent function like `respectYamlTagsInAllSettings()`

rmfitzpatrick (approved):
Thanks!

internal/receiver/smartagentreceiver/config.go
asuresh4: @rmfitzpatrick - I've moved these methods to a helper and exposed `RespectYamlTagsInAllSettings` which can be used by both the receiver and the extension. Let me know if you think it's better.

rmfitzpatrick: I do indeed!

rmfitzpatrick (approved):
🎉

PR #348, opened 2021-04-29
Add a config source for env vars

pjanotti:
Although the OTel collector already supports env variables on the config for feature parity with SA we need to add a full config source for env vars. Having env vars as a config source allows injection of YAML fragments, like SA, via environment variables. Added also the following:

- Possibility to define defaults/fallbacks when environment variables used in the config are undefined in the environment.
- A retrieve parameter to indicate that the configuration load should fail if the environment variable is not defined.

### DIFF
--- a/internal/configsource/envvarconfigsource/README.md
+++ b/internal/configsource/envvarconfigsource/README.md
@@ -2,13 +2,15 @@
 
 Use the environmental variable config source instead of direct references to
 environment variables in the config to inject YAML fragments or to have default
-values in case the selected environment variable is undefined.
+values in case the selected environment variable is undefined. For simple environment
+variable expansion without support for YAML fragments or defaults see
+[Collector Configuration Environment Variables](https://opentelemetry.io/docs/collector/configuration/#configuration-environment-variables) 
 
 ## Configuration
 
 Under the `config_sources:` use `env:` or `env/<name>:` to create an
-environment variable config source. The following parameters are available to customize
-environment variable config sources:
+environment variable config source. The following parameters are available to
+customize environment variable config sources:
 
 ```yaml
 config_sources:
@@ -19,8 +21,9 @@ config_sources:
       MY_ENV_VAR: my env var value
 ```
 
-It is possible to fail the configuration load if an environment variable is required
-and must be defined either on the environment or on the `defaults` of the config source.
+By default, the config source will cause an error if it tries to inject an environment variable
+that is not defined or not specified on the `defaults` section. That behavior can be controlled
+via the `optional` parameters when invoking the config source, example:
 
 ```yaml
 config_sources:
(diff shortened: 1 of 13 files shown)


internal/configsource/envvarconfigsource/session.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	defaultValue, ok := e.defaults[selector]
+	if !ok {
+		if actualParams.Required {
+			return nil, &errMissingRequiredEnvVar{fmt.Errorf("env var %q is required but not defined and not present on defaults", selector)}
+		}
+
+		// To keep with default behavior for env vars not defined set the value to empty string
+		defaultValue = ""
```
pjanotti, line 72: This likely can be better handled by the config source manager. I will evaluate that option and will update after merging this PR if appropriate. For now, the desired behavior is enforced here.

internal/configsource/envvarconfigsource/README.md
```diff
@@ -0,0 +1,44 @@
...
+config_sources:
+  env:
+    # defaults is used to create a set of fallbacks in case the original env var is
+    # undefined in the environment.
+    defaults:
+      MY_ENV_VAR: my env var value
+```
+
+It is possible to fail the configuration load if an environment variable is required
+and must be defined either on the environment or on the `defaults` of the config source.
```
owais, line 23: Do you know which case is more common? I don't have any data but I feel it might  be more common to expect an env var to be defined. If so, may be that can be the default behavior and we can add `?optional=true` param instead.

internal/configsource/envvarconfigsource/README.md
```diff
@@ -0,0 +1,44 @@
...
+      BACKED_BY_DEFAULTS_ENV_VAR: some_value
+
+components:
+  component_0:
+    # Not an error if ENV_VAR_NAME is undefined, resulting value is "/data/token".
+    not_required_field: ${env:ENV_VAR_NAME}/data/token 
+
+  component_1:
+    # It will be an error if ENV_VAR_NAME is undefined, the config will fail to load.
+    required_field: ${env:ENV_VAR_NAME?required=true}/data/token
```
owais, line 38: Just confirming: This means if a defaut is provided in config but env var is not found, this will NOT raise an error?

internal/configsource/envvarconfigsource/session.go
```diff
@@ -0,0 +1,90 @@
...
+func (e *envVarSession) Retrieve(_ context.Context, selector string, params interface{}) (configsource.Retrieved, error) {
+	actualParams := retrieveParams{}
+	if params != nil {
+		paramsParser := config.NewParserFromStringMap(cast.ToStringMap(params))
+		if err := paramsParser.UnmarshalExact(&actualParams); err != nil {
+			return nil, &errInvalidRetrieveParams{fmt.Errorf("failed to unmarshall retrieve params: %w", err)}
+		}
+	}
+
+	value, ok := os.LookupEnv(selector)
```
owais, line 59: how about we lookup first and return if found. If not foujnd then we try to parse params and see if error should be raised or not.

internal/configsource/envvarconfigsource/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Under the `config_sources:` use `env:` or `env/<name>:` to create an
+environment variable config source. The following parameters are available to customize
+environment variable config sources:
+
+```yaml
+config_sources:
+  env:
+    # defaults is used to create a set of fallbacks in case the original env var is
+    # undefined in the environment.
```
owais, line 17: would it be helpful to accept a `prefix_` option? When it is defined all env vars will be prefixed with the value before a lookup is executed. Probably not much useful for modern environments though.

internal/configsource/envvarconfigsource/README.md
pjanotti: Correct, but will be switching to `optional` per your suggestion above.

internal/configsource/envvarconfigsource/session.go
pjanotti: What I don't like in this case is that invocation can work even with incorrect parameters. The collector config in general doesn't ignore unknown/incorrect parameters.

internal/configsource/envvarconfigsource/README.md
pjanotti: Interesting idea, let me see how it looks like in a few examples.

internal/configsource/envvarconfigsource/README.md
```diff
@@ -0,0 +1,44 @@
+# Environment Variable Config Source (Alpha)
+
+Use the environmental variable config source instead of direct references to
+environment variables in the config to inject YAML fragments or to have default
+values in case the selected environment variable is undefined.
```
rmfitzpatrick, line 5: It's taking me a few reads to understand what this does.  Would you mind providing an example with/without the config source and demonstrating what it can do for nonscalar yaml types (maybe with a real component)?  It the main benefit is a default value, I'd wonder why we aren't adding `"${ENV_VAR:-defaultVal}"` support to the config in general without needing a new source.

pjanotti: This is about supporting the same capabilities described at https://docs.signalfx.com/en/latest/integrations/agent/remote-config.html#environment-variables

The main point is about injecting YAML fragments that is what is really missing from the current support for env vars in the collector configuration. Adding support to `${ENV_VAR:-defaultVal}` needs to be done in coordination with core but it seems a very natural idea that keeps popping up.

Anyway, if the reading it is not clear to you it is a sign that need some re-work :) I will try to improve it.

rmfitzpatrick: Sorry, rereading my comment I see it wasn't clear that I picked up on the helpful functionality added, but that the readme was a little brief on the details and full capabilities.  Thanks!

internal/configsource/envvarconfigsource/README.md
pjanotti: While convenient I got the impression that it becomes clear if we don't have the option so a text search on the YAML gets to the proper occurrences right away.

internal/configsource/envvarconfigsource/README.md
pjanotti: @rmfitzpatrick take a second read, please. The comment was actually quite good: it made me realize that I hadn't merged the change to support YAML fragments 🤦 that you saw via https://github.com/signalfx/splunk-otel-collector/pull/362

owais (approved):

PR #350, opened 2021-04-30
Unset exclusion and translations by default in gateway config

mstumpfx:
In our other default agent config, we are recommending sending
to the gateway via OTLP. If agents are sending OTLP to the gateway,
we need to do exclusions and translations at the gateway, as these
would only occur if they were using the signalfx exporter.


dmitryax (approved):
LGTM

jrcamp (approved):

PR #362, opened 2021-05-04
Config sources: handle strings and byte slices as YAML

pjanotti:
This keeps parity with SA functionality: config sources can return fragments of YAML to be injected in the configuration. Usage of it will appear on the README for env var config sources.

/cc @owais

### DIFF
--- a/internal/configprovider/manager_test.go
+++ b/internal/configprovider/manager_test.go
@@ -206,6 +206,49 @@ func TestConfigSourceManager_ResolveErrors(t *testing.T) {
 	}
 }
 
+func TestConfigSourceManager_YAMLInjection(t *testing.T) {
+	ctx := context.Background()
+	manager := newManager(map[string]configsource.ConfigSource{
+		"tstcfgsrc": &testConfigSource{
+			ValueMap: map[string]valueEntry{
+				"valid_yaml_str": {Value: `
+bool: true
+int: 42
+source: string
+map:
+  k0: v0
+  k1: v1
+`},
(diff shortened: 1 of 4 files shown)


internal/configprovider/manager_test.go
```diff
@@ -207,6 +207,49 @@ func TestConfigSourceManager_ResolveErrors(t *testing.T) {
...
+	})
+
+	file := path.Join("testdata", "yaml_injection.yaml")
+	cp, err := config.NewParserFromFile(file)
+	require.NoError(t, err)
+
+	expectedFile := path.Join("testdata", "yaml_injection_expected.yaml")
+	expectedParser, err := config.NewParserFromFile(expectedFile)
+	require.NoError(t, err)
+	expectedCfg := expectedParser.Viper().AllSettings()
```
rmfitzpatrick, line 244: If I'm following, the expected `AllSettings()` doesn't seem to be an item under test, so would it make more sense to define it here as a map so as not to allow potential deviations from the expected format?  If `expectedParser.Viper().AllSettings()` is actually being tested as well, then this makes sense as is.

rmfitzpatrick (approved):

internal/configprovider/manager_test.go
pjanotti: What needs to be tested in this condition is that the actual and expected parsed via `config.Parser` are a match.

pjanotti: This was also being used in other tests taking the chance to remove it.

PR #363, opened 2021-05-04
Allow running integration tests against Collector images

rmfitzpatrick:
These changes allow running the existing integration tests against* an arbitrary collector container.  They add a new `CollectorContainer` type that implements the now unified `Collector` interface broken out from the previously lone `CollectorProcess`.   The `make integration-test` target uses this collector type by default to run against `quay.io/signalfx/splunk-otel-collector-dev:latest`.

Also updates the integration test configs to not use the agent extension.  These are breaking changes that require running against a collector binary with a properly configured agent bundle.  An easy way to avoid this and still test local changes is to run:

```bash
SPLUNK_OTEL_COLLECTOR_IMAGE='otelcol:latest' make -e docker-otelcol integration-test
```

Running Collector containers with command line arguments is not currently provided.  This should be added shortly in a follow-up PR.

### DIFF
--- a/tests/testutils/collector_container.go
+++ b/tests/testutils/collector_container.go
@@ -1,0 +1,199 @@
+// Copyright Splunk, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testutils
+
+import (
+	"archive/tar"
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/testcontainers/testcontainers-go"
+	"go.uber.org/zap"
+)
+
+var _ Collector = (*CollectorContainer)(nil)
+var _ testcontainers.LogConsumer = (*collectorLogConsumer)(nil)
+
+type CollectorContainer struct {
+	Image          string
+	ConfigPath     string
+	Args           []string
+	Logger         *zap.Logger
+	LogLevel       string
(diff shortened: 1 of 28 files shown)


dmitryax (approved):

tests/testutils/collector_container.go
```diff
@@ -0,0 +1,199 @@
...
+
+// To be used as a builder whose Build() method provides the actual instance capable of launching the process.
+func NewCollectorContainer() CollectorContainer {
+	return CollectorContainer{Args: []string{}, Container: NewContainer()}
+}
+
+// quay.io/signalfx/splunk-otel-collector:latest by default
+func (collector CollectorContainer) WithImage(image string) CollectorContainer {
+	collector.Image = image
+	return collector
```
pmcollins, line 51: The other builder methods return pointers, but this one returns a value. Should they all be the same, and should the method receivers be references? Also, I wonder if using an [options pattern](https://www.sohamkamani.com/golang/options-pattern/) would be worthwhile.

rmfitzpatrick: Right, it returns a new value so as not to modify the receiver `CollectorContainer` (as a builder method).  Though the others return pointers, they do it because that's what implements the `Collector` interface and the referenced item is still different from the receiver instance.  The mutative, post-build start and stop methods use pointer receivers.

The options pattern is convenient but not when you want to avoid mutating the receiver instance for the builder methods and have a handle on the intermediary instances.

If we were to be testing a multi-collector setup this pattern enables this:

```go
collectorBuilder := NewCollectorContainer().WithImage("my-tested-image")
agentBuilder := collectorBuilder.WithConfigPath("./agent_config")

agentOne := agentBuilder.WithEnv(map[string]string{"MY_ROLE": "SOMETHING"}).Build()
agentTwo := agentBuilder.WithEnv(map[string]string{"MY_ROLE": "SOMETHING_ELSE"}).Build()

gateway := collector.WithConfigPath("./gateway_config").Build()
```

This isn't possible w/ the options pattern as I understand without repeating the image option three times and the agent config option twice.  This may be wonky as is, but it's a style that I liked and used in the container helpers as well: https://github.com/signalfx/splunk-otel-collector/blob/237af7d347ded7f42e834cbb2018f694346cf5dd/tests/testutils/container.go#L47

PR #383, opened 2021-05-07
Fix assign workflow

flands:
Use built in token with limited auth scope


.github/workflows/assign-one-project.yml
```diff
@@ -5,8 +5,12 @@ on:
     types: [opened, labeled]
   pull_request:
     types: [opened, labeled]
+permissions:
+  contents: read
+  issues: write
```
flands, line 10:
```suggestion
  issues: write
  metadata: read
```

PR #389, opened 2021-05-07
Fix assign workflow

flands:
Switch action and limit auth scope


.github/workflows/pr-project-automation.yml
```diff
@@ -0,0 +1,15 @@
+name: Issue Project Automation
+
+on:
+  pull_request:
```
flands, line 4:
```suggestion
  pull_request_target:
```

PR #397, opened 2021-05-12
Add a file config source

pjanotti:
Add a config source that supports the injection of files for feature parity with SA. A separate config source will be used to inject YAML templates.


pjanotti:
test failures: `os.Chtimes` used in Windows to "force file update" (i.e. 'touch') don't work in Linux. Looking for a proper multiplatform way.

internal/configsource/fileconfigsource/README.md
```diff
@@ -0,0 +1,42 @@
+# File Config Source (Alpha)
+
+Use the file config source to inject YAML fragments or scalars into the
+configuration.
+
+## Configuration
+
+Under the `config_sources:` use `file:` or `file/<name>:` to create a
+file config source. The following parameters are available to
+customize file config sources:
```
rmfitzpatrick, line 10: I know this is used elsewhere, but some specificity and removing parameter boilerplate may be helpful (since there is only the optional name):

```suggestion
Under the top-level `config_sources:` component mapping use `file:` or `file/<component_name>:` to create a
file config source capable of reading from subsequently specified files and detecting their changes:
```

```suggestion
Under the `config_sources:` use `file:` or `file/<name>:` to create a
file config source. The following parameters are available to
customize file config sources:
```

internal/configsource/fileconfigsource/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```yaml
+config_sources:
+  file:
+```
+
+By default, the config source will monitor for updates on the used files
+and will trigger a configuration reload when they are updated. Optionally,
+the config source can be configured to delete the injected file (typically
```
rmfitzpatrick, line 19: Mind specifying when the file is deleted?  Appears to be after reading.

internal/configsource/fileconfigsource/README.md
```diff
@@ -0,0 +1,42 @@
...
+to remove secrets from the file system) or to not watch for changes to the
+file.
```
rmfitzpatrick, line 21: Does the reload restart the collector service?  If receivers are temporarily down it would be good to detail this if not done elsewhere.

internal/configsource/fileconfigsource/config_test.go
```diff
@@ -0,0 +1,64 @@
...
+			},
+		},
+	}
+
+	require.Equal(t, expectedSettings, actualSettings)
+
+	params := configprovider.CreateParams{
+		Logger: zap.NewNop(),
+	}
+	_, err = configprovider.Build(context.Background(), actualSettings, params, factories)
```
rmfitzpatrick, line 62: Would evaluating the resulting map of config sources be helpful here?

internal/configsource/fileconfigsource/configsource_test.go
```diff
@@ -0,0 +1,112 @@
...
+		watchErr = m.WatchForUpdate()
+	}()
+	m.WaitForWatcher()
+
+	file = path.Join("testdata", "file_config_source_end_2_end_expected.yaml")
+	expected, err := config.NewParserFromFile(file)
+	require.NoError(t, err)
+	require.NotNil(t, expected)
+
+	assert.Equal(t, expected.ToStringMap(), r.ToStringMap())
```
rmfitzpatrick, line 88: Similar suggestion I've made in another PR, but I think it's often clearer for expected content to be defined in the test, and to better ensure tested components aren't potentially masking issues or regressions.

internal/configsource/fileconfigsource/configsource_test.go
```diff
@@ -0,0 +1,112 @@
...
+	select {
+	case <-watchDone:
+	case <-time.After(3 * time.Second):
+		require.Fail(t, "expected file change notification didn't happen")
+	}
+
+	assert.ErrorIs(t, watchErr, configsource.ErrValueUpdated)
+
+	// Value should not have changed.
+	expected, err = config.NewParserFromFile(file)
```
rmfitzpatrick, line 103: This is now the expected content and not the end_2_end config?

internal/configsource/fileconfigsource/factory_test.go
```diff
@@ -0,0 +1,51 @@
...
+	tests := []struct {
+		config *Config
+		name   string
+	}{
+		{
+			name:   "default",
+			config: &Config{},
+		},
+	}
+	for _, tt := range tests {
```
rmfitzpatrick, line 44: There's only one item here so maybe the extra runner is unnecessary atm

internal/configsource/fileconfigsource/session.go
```diff
@@ -0,0 +1,153 @@
...
+func (fs *fileSession) Retrieve(_ context.Context, selector string, params interface{}) (configsource.Retrieved, error) {
+	actualParams := retrieveParams{}
+	if params != nil {
+		paramsParser := config.NewParserFromStringMap(cast.ToStringMap(params))
+		if err := paramsParser.UnmarshalExact(&actualParams); err != nil {
+			return nil, &errInvalidRetrieveParams{fmt.Errorf("failed to unmarshall retrieve params: %w", err)}
+		}
+	}
+
+	bytes, err := ioutil.ReadFile(selector) // nolint: gosec
```
rmfitzpatrick, line 69: Does gosec complain if the selector is cleaned?

internal/configsource/fileconfigsource/session_test.go
```diff
@@ -0,0 +1,177 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, r)
+	assert.Equal(t, []byte("42"), r.Value())
+
+	assert.Equal(t, configsource.ErrWatcherNotSupported, r.WatchForUpdate())
+}
+
+func TestFileConfigSource_DeleteFileError(t *testing.T) {
+	if runtime.GOOS != "windows" {
+		t.Skip("Windows only test")
```
rmfitzpatrick, line 122: It's not jumping out to me why this is the case, so could the title be updated or a comment provided?  May be overkill, but adding a `session_windows_test.go` file w/ windows build tag could also be a helpful start.

internal/configsource/fileconfigsource/README.md
pjanotti: Added some info about that, likely should add to the other config sources too, will review other README files after adding all planned config sources.

internal/configsource/fileconfigsource/config_test.go
pjanotti: Checking if the expected ones are being created.

internal/configsource/fileconfigsource/configsource_test.go
pjanotti: In this case, the object under test is expected to return a `config.Parser` so it seems reasonable to use it against the expected equivalent YAML loaded via another `config.Parser`.

internal/configsource/fileconfigsource/configsource_test.go
pjanotti: ops, the test was not doing the intended reload. Renamed the var to avoid confusion and corrected the end of the test.

internal/configsource/fileconfigsource/session.go
pjanotti: `filepath.Clean` prevented the gosec warning: feels a bit like cheating :)

internal/configsource/fileconfigsource/session_test.go
pjanotti: Added comment to explain running only on Windows: locking the file is trivial on Windows, but not on *nix given the golang API (require lower-level non-portable calls). For now, I will not add the separate file but if there is time to enable the test on *nix I will separate the files with the "lock" function.

internal/configsource/fileconfigsource/README.md
pjanotti: Added a blurb for that.

rmfitzpatrick (approved):

tigrannajaryan:
Can this and `template` source be the same and one config implementation? Do we need both?

pjanotti:
Per conversation let's unify the config sources after this PR and #402 are merged.

PR #402, opened 2021-05-13
Add YAML config source

pjanotti:
Adds a config source that can be templatized via https://pkg.go.dev/text/template

Since changing the template typically requires also changing the parameters passed to the template this config source doesn't watch for the template files, i.e., changes to the template file won't trigger reload of the configuration.


pjanotti:
After naming it `yaml` I'm starting to think that perhaps it will be better to name it `template` instead. I'm open to other suggestions.

internal/configsource/yamlconfigsource/README.md
```diff
@@ -0,0 +1,53 @@
...
+  yaml:
+
+components:
+  # component_0 is built from the ./templates/component_template file
+  # according to the template parameters and commands. The example below
+  # defines a few parameters to be used by the template.
+  component_0: |
+    $yaml: ./templates/component_template
+    glob_pattern: /var/**/*.log
+    format: json
```
jrcamp, line 37: Not directly related to this PR but could we do multi-param like:

```yaml
    $yaml:
       selector: ./templates/component_template
       glob_pattern: /var/**/*.log
       format: json
```

(Downside of having it within text block is the yaml inside text doesn't get syntax highlighted, etc. when editing)

jrcamp (commented):
Yeah I think template is a better name.

internal/configsource/yamlconfigsource/README.md
```diff
@@ -0,0 +1,53 @@
...
+Under the `config_sources:` use `yaml:` or `yaml/<name>:` to create a
+YAML config source.
+
+```yaml
+config_sources:
+  yaml:
+```
+
+The parameters of an YAML config source are passed to template to be processed.
+For example, assumint that `./templates/component_template` looks like:
```
rmfitzpatrick, line 17:
```suggestion
For example, assuming that `./templates/component_template` looks like:
```

internal/configsource/yamlconfigsource/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+```terminal
+logs_path: {{ .glob_pattern }}
```
rmfitzpatrick, line 20: My immediate impression with this example was that `glob_pattern` and `format` were keywords for the config source renderer and not generic.  Something like `my_globbed_path` and `my_var` may be more obvious imo.

internal/configsource/yamlconfigsource/config_test.go
```diff
@@ -0,0 +1,64 @@
...
+			},
+		},
+	}
+
+	require.Equal(t, expectedSettings, actualSettings)
+
+	params := configprovider.CreateParams{
+		Logger: zap.NewNop(),
+	}
+	_, err = configprovider.Build(context.Background(), actualSettings, params, factories)
```
rmfitzpatrick, line 62: similar built config source request as other PR, if asserting on content seems more desirable than not.

internal/configsource/yamlconfigsource/configsource_test.go
```diff
@@ -0,0 +1,89 @@
...
+	tests := []struct {
+		config *Config
+		name   string
+	}{
+		{
+			name:   "default",
+			config: &Config{},
+		},
+	}
+	for _, tt := range tests {
```
rmfitzpatrick, line 43: nit, unnecessary runner?

internal/configsource/yamlconfigsource/configsource_test.go
```diff
@@ -0,0 +1,89 @@
...
+		assert.ErrorIs(t, watchErr, configsource.ErrSessionClosed)
+	})
+
+	go func() {
+		defer close(watchDone)
+		watchErr = m.WatchForUpdate()
+	}()
+	m.WaitForWatcher()
+
+	file = path.Join("testdata", "templated_expected.yaml")
```
rmfitzpatrick, line 84: similar test-defined expected content request as elsewhere.  Otherwise it seems like the config parser is testing itself.

internal/configsource/yamlconfigsource/README.md
pjanotti: We had some early interactions with it as actual YAML. That ended up being dropped in favor of consistency when used in the single-line format. Actual YAML on single-line didn't look very good and couldn't be concatenated to other values. That said we could have the multiline format being YAML while the single line was our own but that requires a lot more code.

/cc @tigrannajaryan

internal/configsource/yamlconfigsource/configsource_test.go
pjanotti: Instead of reusing `file` added a new local: `fileWithExpectedData` trying to make this clear. This test unlike the one for file config source doesn't need to check for the data after updates since this config source doesn't watch the template.

pjanotti:
Got CI failures that seem unrelated to the change, closing and re-opening to kick another run.

```
ci/circleci: deb-package — Your tests failed on CircleCI

ci/circleci: rpm-package — Your tests failed on CircleCI
```

pjanotti:
close/open doesn't work ...

tigrannajaryan:
> After naming it `yaml` I'm starting to think that perhaps it will be better to name it `template` instead. I'm open to other suggestions.

Would `include` be a better name?
```yaml
config_sources:
  include:
receivers:
  filelog: |
    $include: ./plugins/tomcat.yaml
    log_dir: /var/log/tomcat
```

tigrannajaryan:
BTW, stanza has a ton of plugins which use the go templating and which I want to reuse for Otelcol logs: https://github.com/observIQ/stanza-plugins/tree/master/plugins

I think we can automate conversion of stanza plugins to filelog-based `include`/`template` config sources (or jus tconvert manually, there are not that many).

internal/configsource/templateconfigsource/README.md
```diff
@@ -0,0 +1,50 @@
...
+  template:
+
+components:
+  # component_0 is built from the ./templates/component_template file
+  # according to the template parameters and commands. The example below
+  # defines a few parameters to be used by the template.
+  component_0: |
+    $template: ./templates/component_template
+    my_glob_pattern: /var/**/*.log
+    my_format: json
```
rmfitzpatrick, line 37: Curious if it's possible to use env var expansion here for the template value

rmfitzpatrick (approved):

internal/configsource/templateconfigsource/README.md
pjanotti: Yes, it is.

pjanotti:
@tigrannajaryan 
> Would include be a better name?

I like `template` because all golang template processing is available but no strong opinion in this regard. Let me know if you prefer `include`.

> re: stanza plugins

Interesting stuff: at first glance, it seems that they parse the YAML and extract a template from the comments. It should be doable to convert those to the templates supported here.

internal/configsource/yamlconfigsource/README.md
tigrannajaryan: > @tigrannajaryan
> 
> > Would include be a better name?
> 
> I like `template` because all golang template processing is available but no strong opinion in this regard. Let me know if you prefer `include`.

I think I prefer `include` since it may be easier to understand for end users the purpose of it. Not strongly opposed to `template` but it appears a bit obscure, not self-explanatory what it does.

> 
> > re: stanza plugins
> 
> Interesting stuff: at first glance, it seems that they parse the YAML and extract a template from the comments. It should be doable to convert those to the templates supported here.

Yes, I think we can delete the top portion of their YAMLs, which just declare the parameters (I think we can just replace it by comments, I don't think it enables any functionality) and slightly adjust the `pipeline` portion to match the `operators` expected by `filelog` and it should work.

tigrannajaryan: (Sorry, replied to a wrong comment thread).

pjanotti: np, I will rename this config source to `include`.

PR #408, opened 2021-05-17
Prepare Ansible role for publishing on Galaxy

dmitryax:
- Migrate the standalone role to a role in collection `signalfx.splunk_otel_collector` to be able to publish it from any subdirectory. This is needed anyway since Ansible world is actively moving to collections.
- Apply changes to make sure the role is working on all supported ansible versions 2.9+
- Make sure all the variables are prefixed with `splunk_` to avoid collisions, the vars usually are set on the global scope.
- Move the role vars to `include_role` task in the Readme examples.
- Update documentation for Ansible collection. The links in Readme file have to be with absolute paths since the readme will be published to Galaxy.


deployments/ansible/galaxy.yml
```diff
@@ -0,0 +1,24 @@
...
+description: Ansible collection for Splunk OpenTelemetry Collector role
+version: 0.1.0
+readme: README.md
+authors:
+  - Splunk Inc.
+dependencies:
+  community.general: "*"
+license:
+  - "Apache-2.0"
+tags:
```
rmfitzpatrick, line 13: collector, connector, agent, gateway?

dmitryax: added more tags

rmfitzpatrick (approved):

PR #409, opened 2021-05-18
Automate Ansible Collection release process

dmitryax:


.github/workflows/ansible-push-release-tag.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Push Ansible Release Tag
+        id: check-tag
+        run: >
+          git tag ${RELEASE_TAG} 2>/dev/null 
+            && git push "${REMOTE_REPO}" --tags
+            || echo "tag already exists"
+        env:
+          RELEASE_TAG: ansible-v${{ steps.read-galaxy-yaml.outputs.version }}
+          REMOTE_REPO: https://${GITHUB_ACTOR}:${{ secrets.GITHUB_TOKEN }}@github.com/${REPOSITORY}.git
```
owais, line 39: I think GDI spec (and security guidelines) force us to use GITHUB_TOKEN with "readonly" permissions. 

Which repos are we pushing to here and need write access to as a result?

owais (approved):

owais:
Question about GITHUB_TOKEN. LGTM other than that.

.github/workflows/ansible-push-release-tag.yml
dmitryax: same repo https://github.com/signalfx/splunk-otel-collector. it uses the temporary github action's token

dmitryax: We do the same in https://github.com/signalfx/splunk-otel-collector-chart

.gitlab-ci.yml
```diff
@@ -0,0 +1,35 @@
...
+    - ansible-galaxy collection build ./deployments/ansible --output-path ./dist
+
+ansible-release:
+  extends: .ansible
+  stage: release
+  before_script:
+    - pip3 install ansible==3.4.0 yq==2.12.0
+  script:
+    - export COLLECTION_VERSION=$(cat ./deployments/ansible/galaxy.yml | yq .version -r)
+    - ansible-galaxy collection publish ./dist/signalfx-splunk_otel_collector-${COLLECTION_VERSION}.tar.gz --token=${ANSIBLE_GALAXY_TOKEN}
```
rmfitzpatrick, line 35: Is this visible in the job details?  Would be a good idea to try with an invalid token if you've not already done so.

dmitryax: this will not be visible. the variable set as protected and masked

dmitryax: And this is run on the internal gitlab

PR #414, opened 2021-05-18
Add compile to gitlab ci

jcheng-splunk:


.gitlab-ci.yml
```diff
@@ -2,6 +2,30 @@ stages:
...
+      - .go/pkg/mod
+      - .go/bin
+
+compile:
+  extends: .setup-environment
+  stage: build
+  parallel:
+    matrix:
+      - TARGET: [binaries-darwin_amd64, binaries-linux_amd64, binaries-linux_arm64, binaries-windows_amd64]
+  script: make $TARGET
```
dmitryax, line 24: won't this script override `- make install-tools`? shouldn't `make install-tools` go to `before_script` or another stage prior to build?

jcheng-splunk: I'm new to gitlab, so still learning.  I guess I can try moving `make install-tools` to a separate job in the `.pre` stage?

dmitryax: Yeah, probably `.pre` stage is better. `extends:` is just a way to reuse the same config parts between tasks

PR #418, opened 2021-05-19
Workaround ansible.builtin.apt_key module bug in ansible 4.0.0

dmitryax:
`ansible.builtin.apt_key` module has a know bug in ansible v4.0.0 https://github.com/ansible/ansible/issues/74424 which is fixed in the version 4.0.1 that will be released on May 24. Until that use `ansible.builtin.get_url` module instead


jcheng-splunk:
@dmitryax How about creating conditional task that can handle the version with something like `when: ansible_version.full is version('4.0.0, '==')`?

dmitryax:
> @dmitryax How about creating conditional task that can handle the version with something like when: ansible_version.full is version('4.0.0, '==')?

Thanks Jeffrey! Good point.

jcheng-splunk (approved):

PR #419, opened 2021-05-19
Unify file and include config sources

pjanotti:
Unify files and include config sources. The main change is that the parameters of the file config source are now specified as parameters of the config source itself, see README.md for details.


pjanotti:
Test failure due to git differences on Windows between my machine and CI: on my machine I preserve original line endings on CI uses `\n\a`... thinking on how to properly fix it.

internal/configsource/includeconfigsource/README.md
```diff
@@ -1,19 +1,65 @@
...
 
 Under the `config_sources:` use `include:` or `include/<name>:` to create a
 template config source.
+The following parameters are available to customize the include config sources:
 
 ```yaml
 config_sources:
   include:
+  include/my_name_00:
+    # delete_files can be used to make the "include" config source to delete the
```
rmfitzpatrick, line 16:
```suggestion
    # delete_files can be used to make the "include" config source delete the
```

internal/configsource/includeconfigsource/README.md
```diff
@@ -1,19 +1,65 @@
...
+    # files referenced by it. This is typically used to remove secrets from the
+    # file system. The file is deleted as soon as its value is read by the config
+    # source. The default value is false. It is an invalid configuration to set it
+    # to true together with the watch_files parameter (see below).
+    delete_files: true
+  include/my_name_01:
+    # watch_files can be used to make the "include" config source monitor for
+    # updates on the used files. Setting it to true will trigger a configuration
```
rmfitzpatrick, line 24:
```suggestion
    # updates on the used files (false by default). Setting it to true will trigger a configuration
```

rmfitzpatrick: oop, missed it below.

internal/configsource/includeconfigsource/README.md
```diff
@@ -1,19 +1,65 @@
...
+  include/secret:
+    delete_files: true
+  include/watch_for_updates:
+    watch_files: true
+
+components:
+  component_0:
+    # Default usage: configuration won't be reloaded if the file
+    # '/etc/configs/component_field' is changed.
+    component_field: ${include/default:/etc/configs/component_field}
```
rmfitzpatrick, line 46: If it's possible to use the contents of a yml file for the complete mapping, including the specified components, I think an example below would be helpful for users who want to do something like

```
receivers:
  ${include/default:/etc/receivers.yaml}

pipelines:
  ${include/default:/etc/pipelines.yaml}
```

internal/configsource/includeconfigsource/README.md
```diff
@@ -1,19 +1,65 @@
...
+  component_0:
+    # Default usage: configuration won't be reloaded if the file
+    # '/etc/configs/component_field' is changed.
+    component_field: ${include/default:/etc/configs/component_field} 
+
+  component_1:
+    # 'include/secret' was created with 'delete_files' set to true the
+    # file '/etc/configs/secret' after its value is read. If the deletion
+    # fails the collector won't start.
+    component_field: ${include/secret:/etc/configs/secret}
```
rmfitzpatrick, line 52: Is the lack of space between the `include` and the filename meaningful?  I'm not sure the difference in usage here and in the `|` multiline blocks w/ templating is clear.

rmfitzpatrick (approved):
lgtm, with provided suggestions

internal/configsource/includeconfigsource/README.md
pjanotti: The problem is that as soon as you add a space after `:` the YAML sees that as a tag so we can't use spaces after it. On the multiline is different since we treat the subsequent line as YAML with the parameters.

internal/configsource/includeconfigsource/README.md
pjanotti: Added an example to cover this case.

PR #424, opened 2021-05-20
Improve ansible tests workflow

dmitryax:
- Run tests on any push with changes under deployments/ansible
- Make "push release tag" job dependant on tests
- Do not fail if release tag exists
- Verify that the version is properly fetched from galaxy.yml before pushing new tag


mstumpfx (approved):
cool

PR #436, opened 2021-05-28
Initial version of HTTP sink exporter

owais:
This is a test exporter that allows clients to "watch" for spans they might be interested in. This is very useful for end-to-end tests where we want to test real world applications and ensure data reaches the collector as expected.

The exporter exposes a HTTP endpoint. Clients can make GET requests to this endpoint and specify the characteristics of spans they are interested in using URL query params. Collector will then block the request until it detects the specified spans. Once the spans are detected, they are serialized to JSON (using jaeger proto) and returned back to the client. Multiple clients can watch for the same spans at the same time. The export does not hold anything in memory before a watch is started or after a watch ends. Clients must start the watch first, generate spans and assert the watch result.

All this can be done with the file or logging exporter as well but this dramatically simplifies writing such tests especially when this needs to be done for multiple libraries across different languages.

Example test that leverages this to test a simple but real app instrumented with otel: https://github.com/signalfx/splunk-otel-python/blob/main/tests/integration/test_simple.py#L24-L47 

The implementation is very basic, code is rough and a lot of tests are lacking. I'll improve this over time and if it makes sense, try to migrate it to Otel Contrib but for now this would make life a lot easier for at least a couple of our (instrumentals) projects.

Perhaps down the line, we can permanently host a collector in Splunk lab with this exporter enabled to allow very quick and easy manual and automated testing (even from CI). May be I'll try this for the upcoming hackathon. 

### Example and explanation 
Recorded an example here: https://www.loom.com/share/16256b778d9a40a297a2212692422d0b

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,10 +11,12 @@ require (
 	github.com/fsnotify/fsnotify v1.5.1
 	github.com/go-playground/universal-translator v0.17.0 // indirect
 	github.com/go-zookeeper/zk v1.0.2
+	github.com/gogo/protobuf v1.3.2
 	github.com/hashicorp/vault v1.7.0
 	github.com/hashicorp/vault-plugin-auth-gcp v0.9.0
 	github.com/hashicorp/vault/api v1.1.0
 	github.com/hashicorp/vault/sdk v0.2.1 // indirect
+	github.com/jaegertracing/jaeger v1.26.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/fileexporter v0.36.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter v0.36.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/sapmexporter v0.36.0
@@ -28,6 +30,7 @@ require (
 	github.com/open-telemetry/opentelemetry-collector-contrib/extension/pprofextension v0.36.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/extension/storage v0.36.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/experimentalmetricmetadata v0.36.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/jaeger v0.36.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/zipkin v0.36.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/processor/attributesprocessor v0.36.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/processor/filterprocessor v0.36.0
@@ -149,7 +152,6 @@ require (
 	github.com/go-test/deep v1.0.7 // indirect
 	github.com/gobwas/glob v0.2.4-0.20181002190808-e7a84e9525fe // indirect
 	github.com/gogo/googleapis v1.4.1 // indirect
-	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang-sql/civil v0.0.0-20190719163853-cb61b32ac6fe // indirect
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
@@ -201,7 +203,6 @@ require (
 	github.com/jackc/pgservicefile v0.0.0-20200307190119-3430c5407db8 // indirect
 	github.com/jackc/pgtype v1.3.0 // indirect
 	github.com/jackc/pgx/v4 v4.6.0 // indirect
-	github.com/jaegertracing/jaeger v1.26.0 // indirect
 	github.com/jcmturner/aescts/v2 v2.0.0 // indirect
 	github.com/jcmturner/dnsutils/v2 v2.0.0 // indirect
 	github.com/jcmturner/gofork v1.0.0 // indirect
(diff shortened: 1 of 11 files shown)


internal/exporter/pollingexporter/README.md
```diff
@@ -0,0 +1,28 @@
+# Polling Exporter
+
+This exporter makes span data available via a HTTP endpoint. The endpoint
+accepts requests for spans with specific characteristics and blocks until the exporter
+receives such spans or the request times out. Once the requested spans are detected, they're
+returned back to the client as JSON. 
+
+This exporter return data as [JSON encoding](https://developers.google.com/protocol-buffers/docs/proto3#json)
```
rmfitzpatrick, line 8:
```suggestion
This exporter returns data as [JSON encoding](https://developers.google.com/protocol-buffers/docs/proto3#json)
```

It would be helpful to show an example request w/ params and the response imo.

owais: Updated description with link: https://www.loom.com/share/16256b778d9a40a297a2212692422d0b

jrcamp:
Why jaeger instead of OTLP JSON? Presumably this would be useful for logs and metrics as well. Once https://github.com/open-telemetry/opentelemetry-collector/tree/main/internal/model is more mature can use it to serialize to any protocol.

owais:
It was a lot more work to use OTLP as collector doesn't expose the models. I could re-generate the proto definitions for OTLP have use them but even that was a lot of work at the time. The goal was to allow splunk-otel-python to use this for integration tests and Jaeger allowed me to get there fastest. 

We should definitely support OTLP when we can though. It might even be worth adding `?format=otlp|jaeger|zipkin|...` query param and let the client chose the format it wants.

owais:
@jrcamp @rmfitzpatrick @dmitryax PTAL when you can. Thanks

internal/exporter/pollingexporter/README.md
```diff
@@ -0,0 +1,28 @@
+# Polling Exporter
```
rmfitzpatrick, line 1: Would something like "http testing" or "sanity" exporter be more accurate and denote this is for testing purposes only?

internal/exporter/pollingexporter/client.go
```diff
@@ -0,0 +1,119 @@
...
+
+func newClient(opts options) *client {
+	return &client{
+		ch:   make(chan *model.Batch),
+		opts: opts,
+	}
+}
+
+func (c *client) response() ([]byte, error) {
+	// add support to filter by attributes and names
```
rmfitzpatrick, line 88:
```suggestion
	// TODO: add support to filter by attributes and names
```

internal/exporter/pollingexporter/client.go
```diff
@@ -0,0 +1,119 @@
...
+					return nil, err
+				}
+				spans = append(spans, json)
+				received++
+				if received == c.opts.count {
+					close(done)
+				}
+			}
+		case <-time.After(c.opts.timeout):
+			return nil, fmt.Errorf("timedout while waiting for spans")
```
rmfitzpatrick, line 112:
```suggestion
			return nil, fmt.Errorf("timed out while waiting for spans")
```

rmfitzpatrick: Could be helpful to report how many were received so this isn't all or nothing.

internal/exporter/pollingexporter/README.md
owais: Yes polling isn't really a good name. I don't want to call it testing as that implies automated testing and this can be useful for manual tests as well but for sure that'd be better than polling )

internal/exporter/pollingexporter/polling_exporter.go
```diff
@@ -0,0 +1,110 @@
...
+
+	c := newClient(opts)
+
+	e.mu.Lock()
+	e.clients = append(e.clients, c)
+	e.mu.Unlock()
+
+	result, err := c.response()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
rmfitzpatrick, line 69: 404 makes more sense than 500 to me, so maybe response should also include a status code?

internal/exporter/pollingexporter/polling_exporter.go
```diff
@@ -0,0 +1,110 @@
...
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(result)
```
rmfitzpatrick, line 74: Should a client be unregistered after its response is written?

owais: Yes, I was holding that off for future enhancement but can do that now if you'd like.

rmfitzpatrick: I'd think it'd be preferable to deferring (a loose end to manage)

emaderer:
@owais  can we close this PR for now?

Rauno56:
Can we work towards merging it?
We needed to e2e test JS distro and are using polling exporter for it.

internal/exporter/pollingexporter/README.md
owais: Renamed to `httpsink`.

internal/exporter/pollingexporter/polling_exporter.go
owais: done

internal/exporter/pollingexporter/polling_exporter.go
owais: done

internal/exporter/pollingexporter/client.go
owais: This could be useful but none of the clients (Splunk Otel Python and Splunk Otel JS) need it right now. We can definitely add it if a real user requests it. May be we can even make it a streaming response and returns spans as they arrive without waiting for all.

owais:
I have updated the PR and addressed @rmfitzpatrick's comments. It would be nice to get this in as two projects already depend on this for integration tests (splunk-otel-python and splunk-otel-js).

emaderer (approved):

PR #449, opened 2021-06-03
Add first iteration of translatesfx

pmcollins:
This is a first iteration of a tool to translate a Smart Agent config to an Otelcol config. It supports file config sources in #from directives, including wildcards.

Todo for subsequent PRs:
* Observers
* SignalFx Otelcol extension
* collectd

End-to-end tested using `testdata/sa-complex.yaml`


cmd/translatesfx/README.md
```diff
@@ -0,0 +1,31 @@
...
+directives. The `translatesfx` command uses the working directory to resolve any
+relative paths to files. If you omit the working directory argument, 
+`translatesfx` expands relative files paths using the current working
+directory.
+
+```
+> translatesfx <sfx-file> [<file expansion working directory>]
+```
+
+## Examples
```
rmfitzpatrick, line 19: Mind documenting the output/outcome?  Basic before and after could be helpful to demonstrate what's expected.

cmd/translatesfx/translatesfx/cli.go
```diff
@@ -0,0 +1,55 @@
...
+	case 2:
+		wd, err := os.Getwd()
+		if err != nil {
+			log.Fatalf("error getting working directory: %v", err)
+		}
+		return args[1], wd
+	case 3:
+		return args[1], args[2]
+	default:
+		fmt.Println("usage: translatesacfg <path/to/smart/agent/config.yaml>")
```
rmfitzpatrick, line 30: `log.Fatal()` so that we don't push this down to a failed file read?

cmd/translatesfx/translatesfx/cli.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		log.Fatalf("error reading file %q: %v", fname, err)
+	}
+	var orig interface{}
+	err = yaml.UnmarshalStrict(bytes, &orig)
+	if err != nil {
+		log.Fatalf("error unmarshaling file %q: %v", fname, err)
+	}
+	saExpanded, _ := expandSA(orig, wd)
+	saInfo := saExpandedToCfgInfo(saExpanded.(map[interface{}]interface{}))
```
rmfitzpatrick, line 48: is a tested assertion needed? looks like it can be a `[]interface{}`

cmd/translatesfx/translatesfx/sa.go
```diff
@@ -0,0 +1,226 @@
...
+func processDirective(directive map[interface{}]interface{}, wd string) (interface{}, bool, error) {
+	expanded, err := expandFromSource(directive, wd)
+	if err != nil {
+		return nil, false, err
+	}
+
+	return expanded, flatten(directive), err
+}
+
+func expandFromSource(directive map[interface{}]interface{}, wd string) (interface{}, error) {
```
rmfitzpatrick, line 75: There are a number of existing config sources and I wonder if it'd be better to replace the agent sources with ours rather than reimplement their logic like is done w/ file: https://github.com/signalfx/splunk-otel-collector/tree/main/internal/configsource

rmfitzpatrick: ^ This could require a new agent config source though...

cmd/translatesfx/translatesfx/cli.go
pmcollins: Oh yeah, that's a good point. Will fix.

cmd/translatesfx/translatesfx/cli.go
pmcollins: I think it's unlikely, because the user would have to have a non-functional SA config for it to be unmarshalled as a slice, but I'll add the test to give the user feedback in case things are really sideways.

flands (commented):
Nice! Please add a comment to the top of the outputted YAML file making it clear this is experimental and that manual verification should be done before attempting to use the configuration provided.

cmd/translatesfx/translatesfx/sa.go
pmcollins: Yeah, I'm not sure what to do about that yet. One problem, I think, with using the configsource's file expansion capabilities is that SA supports globbing and merging, and configsource apparently doesn't. I suppose we might want to glob/merge then expand the result into a single file and create a configsource for that? Maybe suitable for a separate PR?

cmd/translatesfx/README.md
pmcollins: Yeah, good point -- I wasn't sure how to handle that. I think I'll add an example input and example output.

pmcollins: Done

cmd/translatesfx/translatesfx/cli.go
pmcollins: Done

cmd/translatesfx/translatesfx/cli.go
pmcollins: Done

cmd/translatesfx/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+> translatesfx path/to/sfx/config.yaml path/to/sfx
+```
+#### Example Input/Output
+
+Given the following input
+```yaml
+signalFxAccessToken: {"#from": "env:SFX_ACCESS_TOKEN"}
+ingestUrl: {"#from": "testdata/ingest_url", default: "https://ingest.signalfx.com"}
```
rmfitzpatrick, line 39: mind showing these files' content?

pmcollins: Sure, done.

cmd/translatesfx/README.md
```diff
@@ -0,0 +1,102 @@
+# SignalFx Smart Agent Configuration Translation Tool (Experimental)
+
+This package provides a command-line tool to translate a SignalFx Smart Agent
+configuration file into an OpenTelemetry Collector configuration.
+
+## Usage
+
+The `translatesfx` command requires one argument, the signalfx configuration
```
rmfitzpatrick, line 8: Not requesting now, but I think this could be a feature for a more generalized config verifier and linter.

`otelcolconfig --from-smart-agent`
`otelcolconfig --verify my_collector_config.yaml`
`otelcolconfig --build --receivers jaeger --exporters otlp` etc.

rmfitzpatrick (approved):

pmcollins:
Thanks for reviewing @rmfitzpatrick

PR #455, opened 2021-06-07
translatesfx: factor out `directive` type

pmcollins:


cmd/translatesfx/translatesfx/directive.go
```diff
@@ -0,0 +1,138 @@
...
+	directiveSourceEtcd2
+	directiveSourceConsul
+	directiveSourceVault
+)
+
+type directive struct {
+	fromPath    string
+	defaultV    string
+	fromType    source
+	isDirective bool
```
rmfitzpatrick, line 38: I'm curious why `isDirective` is preferable to a new interface and checked type assertions.  Is to avoid needing accessors?

pmcollins: Yeah, that was a mistake. I removed `isDirective` in the PR I'm working on now :-)

cmd/translatesfx/translatesfx/directive.go
```diff
@@ -0,0 +1,138 @@
...
+	idx := strings.Index(from, ":")
+	if idx == -1 {
+		return directiveSourceFile, from, nil
+	}
+	s = strToSource(from[:idx])
+	if s == directiveSourceUnknown {
+		return s, target, fmt.Errorf("#from fromType type unknown: %s", from)
+	}
+	target = from[idx+1:]
+	return
```
rmfitzpatrick, line 87: nit, mixing named and unnamed returns makes this more confusing to me than necessary.

cmd/translatesfx/translatesfx/sa_test.go
```diff
@@ -132,53 +120,55 @@ func TestExpand_SAComplex(t *testing.T) {
...
+	d, err := parseDirective(map[interface{}]interface{}{
 		"#from": "testdata/cfgs/map*.yaml",
-	}
-	v, flatten, err := processDirective(d, "")
+	})
+	require.NoError(t, err)
+	v, err := processDirective(d, "")
 	require.NoError(t, err)
-	assert.False(t, flatten)
 	assert.NotNil(t, v)
```
rmfitzpatrick, line 152: not required but in general it could be helpful to have explicit return values even if repeated since the directive is fundamentally different

rmfitzpatrick (approved):

cmd/translatesfx/translatesfx/directive.go
pmcollins: Thanks for the feedback -- fixed.

cmd/translatesfx/translatesfx/sa_test.go
pmcollins: Agreed. This gets reworked in a later PR though so I'll make sure this is addressed there.

pmcollins:
Thanks for the review @rmfitzpatrick 👍

PR #461, opened 2021-06-09
translatesfx: convert `#from` files to configsource includes

pmcollins:
If the #from directive is simple (no default, flatten, or wildcards), this change
converts the #from directive to a configsource include. Otherwise it continues to just
expand/inline the referenced file(s).

Also, did some additional refactoring, making some functions methods on `directive`.


cmd/translatesfx/translatesfx/directive.go
```diff
@@ -136,3 +134,142 @@ func parseDefault(m map[interface{}]interface{}) (string, error) {
...
+}
+
+func (d directive) handleFileType() (interface{}, error) {
+	// configsource doesn't handle flatten, glob, or default values at this time, so
+	// we inline the value if any of those are specified in the #from directive
+	if d.flatten || hasGlob(d.fromPath) || d.defaultV != "" {
+		return d.expandFiles()
+	}
+	// otherwise we replace the #from directive with a configsource one
+	return fmt.Sprintf("${include:%s}", d.fromPath), nil
```
rmfitzpatrick, line 187: Would this be telling the collector's include config source to try to load SA monitor lists?

pmcollins: Excellent point. I think we'll want to flag some parts of the sfx config as: don't translate to configsource but rather just expand/inline. What do you think? I'll work on a solution and will update this PR.

rmfitzpatrick: I think the dynamic/runtime config source use case conflicts with this one and done utility so inlining what we can and throwing up advisories for required changes for zk/vault/etcd if they are nonscalar may be an easy way to start.

Having an intermediary SA config source that uses the actual SA remote config code and the provided connectivity/auth info could be a runtime solution, but it's probably better overall to ask those users to migrate upfront.

pmcollins: Yeah, I think so. I'm thinking we should use configsource when we can, then fall back to inlining if it's simple to do so (local file or env var). If we don't have any recourse (e.g. zk && flatten) either stop and error out with a helpful message, or render the config with a comment inline where we were unable to handle the remote source. But I'm thinking that will be rare -- depending on how esoteric SA configs are in the wild, we should be able to handle the vast majority of cases.

rmfitzpatrick (approved):

PR #462, opened 2021-06-09
Add env var SPLUNK_CONFIG_YAML for storing configuration YAML

bjsignalfx:
Supplying Splunk Otel Collector configuration yaml via environment variable SPLUNK_CONFIG_YAML. 

One advantage is that this allows for having the config YAML as a parameter value in AWS Systems Manager Parameter Store and thus removing the need to have a volume mounted to a Fargate task to hold the configuration file.

### DIFF
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -18,6 +18,7 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"os"
@@ -26,6 +27,7 @@ import (
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/service"
+	"go.opentelemetry.io/collector/service/parserprovider"
 	"go.uber.org/zap"
 
 	"github.com/signalfx/splunk-otel-collector/internal/components"
@@ -37,6 +39,7 @@ import (
 const (
 	ballastEnvVarName         = "SPLUNK_BALLAST_SIZE_MIB"
 	configEnvVarName          = "SPLUNK_CONFIG"
+	configYamlEnvVarName      = "SPLUNK_CONFIG_YAML"
 	configServerEnabledEnvVar = "SPLUNK_DEBUG_CONFIG_SERVER"
 	memLimitMiBEnvVarName     = "SPLUNK_MEMORY_LIMIT_MIB"
 	memTotalEnvVarName        = "SPLUNK_MEMORY_TOTAL_MIB"
@@ -57,8 +60,7 @@ func main() {
 	// TODO: Use same format as the collector
 	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
 
-	args := os.Args[1:]
-	if !contains(args, "-h") && !contains(args, "--help") {
+	if !contains(os.Args[1:], "-h") && !contains(os.Args[1:], "--help") {
 		checkRuntimeParams()
 	}
 
@@ -77,6 +79,7 @@ func main() {
 	}
 
 	parserProvider := configprovider.NewConfigSourceParserProvider(
+		newBaseParserProvider(),
 		zap.NewNop(), // The service logger is not available yet, setting it to NoP.
 		info,
 		configsources.Get()...,
@@ -109,19 +112,21 @@ func contains(arr []string, str string) bool {
 
 // Get the value of a key in an array
 // Support key/value with and with an equal sign
-func getKeyValue(args []string, argName string) string {
-	val := ""
-	for i, arg := range args {
+func getKeyValue(args []string, arg string) (exists bool, value string) {
+	argEq := arg + "="
+	for i := range args {
 		switch {
-		case strings.HasPrefix(arg, argName+"="):
-			s := strings.Split(arg, "=")
-			val = s[1]
-		case arg == argName:
-			i++
-			val = args[i]
+		case strings.HasPrefix(args[i], argEq):
+			return true, strings.SplitN(args[i], "=", 2)[1]
+		case args[i] == arg:
+			exists = true
+			if i < (len(args) - 1) {
+				value = args[i+1]
+			}
+			return
 		}
 	}
-	return val
+	return
 }
 
 // Check runtime parameters
@@ -129,24 +134,7 @@ func getKeyValue(args []string, argName string) string {
 // Config and ballast flags are checked
 // Config and all memory env vars are checked
 func checkRuntimeParams() {
-	args := os.Args[1:]
-	config := ""
-
-	// Check if config flag was passed and its value differs from config env var.
-	// If so, log that it will be used instead of env var value and set env var with that value.
-	// This allows users to set `--config` and have it take priority when running from most contexts.
-	cliConfig := getKeyValue(args, "--config")
-	if cliConfig != "" {
-		config = os.Getenv(configEnvVarName)
-		if config != "" && config != cliConfig {
-			log.Printf(
-				"Both %v and '--config' were specified. Overriding %q environment variable value with %q for this session",
-				configEnvVarName, config, cliConfig,
-			)
-		}
-		os.Setenv(configEnvVarName, cliConfig)
-	}
-	setConfig()
+	checkConfig()
 
 	// Set default total memory
 	memTotalSizeMiB := defaultMemoryTotalMiB
@@ -169,11 +157,10 @@ func checkRuntimeParams() {
 	// Check if memory ballast flag was passed
 	// If so, ensure memory ballast env var is not set
 	// Then set memory ballast and limit properly
-	ballastSize := getKeyValue(args, "--mem-ballast-size-mib")
+	_, ballastSize := getKeyValue(os.Args[1:], "--mem-ballast-size-mib")
 	if ballastSize != "" {
-		config = os.Getenv(ballastEnvVarName)
-		if config != "" {
-			log.Fatalf("Both %v and '--config' were specified, but only one is allowed", ballastEnvVarName)
+		if os.Getenv(ballastEnvVarName) != "" {
+			log.Fatalf("Both %v and '--mem-ballast-size-mib' were specified, but only one is allowed", ballastEnvVarName)
 		}
 		os.Setenv(ballastEnvVarName, ballastSize)
 	}
@@ -181,54 +168,92 @@ func checkRuntimeParams() {
 	setMemoryLimit(memTotalSizeMiB)
 }
 
-// Validate and set the configuration
-func setConfig() {
-	// Check if the config is specified via the env var.
-	config := os.Getenv(configEnvVarName)
-	// If not attempt to use a default config; supports Docker and local
-	if config == "" {
-		_, err := os.Stat(defaultDockerSAPMConfig)
-		if err == nil {
-			config = defaultDockerSAPMConfig
-		}
-		_, err = os.Stat(defaultLocalSAPMConfig)
-		if err == nil {
-			config = defaultLocalSAPMConfig
-		}
-		if config == "" {
-			log.Fatalf("Unable to find the default configuration file, ensure %s environment variable is set properly", configEnvVarName)
-		}
-	} else {
-		// Check if file exists.
-		_, err := os.Stat(config)
-		if err != nil {
-			log.Fatalf("Unable to find the configuration file (%s) ensure %s environment variable is set properly", config, configEnvVarName)
-		}
+// Sets flag '--config' to specified env var SPLUNK_CONFIG, if the flag not specified.
+// Logs a message and returns if env var SPLUNK_CONFIG_YAML specified, and '--config' and SPLUNK_CONFIG no specified.
+// Sets '--config' to default config file path if '--config', SPLUNK_CONFIG and SPLUNK_CONFIG_YAML not specified.
+func checkConfig() {
+	configPathFlagExists, configPathFlag := getKeyValue(os.Args[1:], "--config")
+	configPathVar := os.Getenv(configEnvVarName)
+	configYaml := os.Getenv(configYamlEnvVarName)
+
+	if configPathFlagExists && configPathFlag == "" {
+		log.Fatal("Command line flag --config specified but empty")
 	}
 
-	switch config {
+	if configPathFlag != "" {
+		if _, err := os.Stat(configPathFlag); err != nil {
+			log.Fatalf("Unable to find the configuration file (%s) ensure flag '--config' is set properly", configPathFlag)
+		}
+
+		if configPathVar != "" && configPathVar != configPathFlag {
+			log.Printf("Both environment variable %v and flag '--config' were specified. Using the flag value %s and ignoring the environment variable value %s in this session", configEnvVarName, configPathFlag, configPathVar)
+		}
+
+		if configYaml != "" {
+			log.Printf("Both environment variable %s and flag '--config' were specified. Using the flag value %s and ignoring the environment variable in this session", configYamlEnvVarName, configPathFlag)
+		}
+
+		checkRequiredEnvVars(configPathFlag)
+
+		log.Printf("Set config to %v", configPathFlag)
+		return
+	}
+
+	if configPathVar != "" {
+		if _, err := os.Stat(configPathVar); err != nil {
+			log.Fatalf("Unable to find the configuration file (%s) ensure %s environment variable is set properly", configPathVar, configEnvVarName)
+		}
+
+		os.Args = append(os.Args, "--config="+configPathVar)
+
+		if configYaml != "" {
+			log.Printf("Both %s and %s were specified. Using %s environment variable value %s for this session", configEnvVarName, configYamlEnvVarName, configEnvVarName, configPathVar)
+		}
+
+		checkRequiredEnvVars(configPathVar)
+
+		log.Printf("Set config to %v", configPathVar)
+		return
+	}
+
+	if configYaml != "" {
+		log.Printf("Using environment variable %s for configuration", configYamlEnvVarName)
+		return
+	}
+
+	defaultConfigPath := getExistingDefaultConfigPath()
+	checkRequiredEnvVars(defaultConfigPath)
+	os.Args = append(os.Args, "--config="+defaultConfigPath)
+	log.Printf("Set config to %v", defaultConfigPath)
+}
+
+func getExistingDefaultConfigPath() (path string) {
+	if _, err := os.Stat(defaultLocalSAPMConfig); err == nil {
+		return defaultLocalSAPMConfig
+	}
+	if _, err := os.Stat(defaultDockerSAPMConfig); err == nil {
+		return defaultDockerSAPMConfig
+	}
+	log.Fatalf("Unable to find the default configuration file (%s) or (%s)", defaultLocalSAPMConfig, defaultDockerSAPMConfig)
+	return
+}
+
+func checkRequiredEnvVars(path string) {
+	// Check environment variables required by default configuration.
+	switch path {
 	case
 		defaultDockerSAPMConfig,
 		defaultDockerOTLPConfig,
 		defaultLocalSAPMConfig,
 		defaultLocalOTLPConfig:
-		// The following environment variables are required.
-		// If any are missing stop here.
 		requiredEnvVars := []string{realmEnvVarName, tokenEnvVarName}
 		for _, v := range requiredEnvVars {
 			if len(os.Getenv(v)) == 0 {
 				log.Printf("Usage: %s=12345 %s=us0 %s", tokenEnvVarName, realmEnvVarName, os.Args[0])
-				log.Fatalf("ERROR: Missing required environment variable %s with default config path %s", v, config)
+				log.Fatalf("ERROR: Missing required environment variable %s with default config path %s", v, path)
 			}
 		}
 	}
-
-	args := os.Args[1:]
-	if !contains(args, "--config") {
-		// Inject the command line flag that controls the configuration.
-		os.Args = append(os.Args, "--config="+config)
-	}
-	log.Printf("Set config to %v", config)
 }
 
 // Validate and set the memory ballast
@@ -277,7 +302,7 @@ func setMemoryLimit(memTotalSizeMiB int) {
 
 	// Validate memoryLimit is sane
 	args := os.Args[1:]
-	b := getKeyValue(args, "--mem-ballast-size-mib")
+	_, b := getKeyValue(args, "--mem-ballast-size-mib")
 	ballastSize, _ := strconv.Atoi(b)
 	if (ballastSize * 2) > memLimit {
 		log.Fatalf("Memory limit (%v) is less than 2x ballast (%v). Increase memory limit or decrease ballast size.", memLimit, ballastSize)
@@ -285,6 +310,19 @@ func setMemoryLimit(memTotalSizeMiB int) {
 
 	os.Setenv(memLimitMiBEnvVarName, strconv.Itoa(memLimit))
 	log.Printf("Set memory limit to %d MiB", memLimit)
+}
+
+// Returns a ParserProvider that reads configuration YAML from an environment variable when applicable.
+func newBaseParserProvider() parserprovider.ParserProvider {
+	_, configPathFlag := getKeyValue(os.Args[1:], "--config")
+	configPathVar := os.Getenv(configEnvVarName)
+	configYaml := os.Getenv(configYamlEnvVarName)
+
+	if configPathFlag == "" && configPathVar == "" && configYaml != "" {
+		return parserprovider.NewInMemory(bytes.NewBufferString(configYaml))
+	}
+
+	return parserprovider.Default()
 }
 
 func runInteractive(params service.CollectorSettings) error {
--- a/cmd/otelcol/main_test.go
+++ b/cmd/otelcol/main_test.go
@@ -16,8 +16,14 @@
 package main
 
 import (
+	"bytes"
+	"fmt"
+	"log"
 	"os"
+	"path"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestContains(t *testing.T) {
@@ -50,7 +56,7 @@ func TestGetKeyValue(t *testing.T) {
 		{"foo", "--test", "foo"},
 	}
 	for _, v := range testArgs {
-		result := getKeyValue(v, "--test")
+		_, result := getKeyValue(v, "--test")
 		if result != v[0] {
 			t.Errorf("Expected %v got %v", v[0], v)
 		}
@@ -59,9 +65,8 @@ func TestGetKeyValue(t *testing.T) {
 
 func TestCheckRuntimeParams(*testing.T) {
 	oldArgs := os.Args
-	os.Setenv(configEnvVarName, "../../"+defaultLocalSAPMConfig)
-	setConfig()
-	os.Unsetenv(configEnvVarName)
+	os.Setenv(configEnvVarName, path.Join("../../", defaultLocalSAPMConfig))
+	checkConfig()
 	checkRuntimeParams()
 
 	os.Args = oldArgs
@@ -80,7 +85,7 @@ func TestCheckRuntimeParams(*testing.T) {
 
 func HelperTestSetMemoryBallast(val string, t *testing.T) {
 	args := os.Args[1:]
-	c := getKeyValue(args, "--mem-ballast-size-mib")
+	_, c := getKeyValue(args, "--mem-ballast-size-mib")
 	if c != val {
 		t.Errorf("Expected memory ballast CLI param %v got %v", val, c)
 	}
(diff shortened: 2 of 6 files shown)


rmfitzpatrick:
Would you be comfortable setting https://github.com/signalfx/splunk-otel-collector/pull/450 as the base branch?  They will be stepping on each other otherwise.

Also it would be good to document this in places like https://github.com/signalfx/splunk-otel-collector/blob/main/docs/getting-started/linux-manual.md#custom-configuration

cmd/otelcol/main.go
```diff
@@ -174,31 +169,52 @@ func checkRuntimeParams() {
...
+	// Config YAML from env var SPLUNK_CONFIG_YAML.
+	cfgYAML := os.Getenv(configYamlEnvVarName)
+
+	// Halting if multiple config sources specified.
+	if (pathFlag != "" && pathEnv != "") || (pathFlag != "" && cfgYAML != "") || (pathEnv != "" && cfgYAML != "") {
+		log.Fatalf("'--config', %s and %s must be specified exclusively", configEnvVarName, configYamlEnvVarName)
+	}
+
+	if cfgYAML != "" {
+		log.Printf("Configuring collector using env var %s YAML", configYamlEnvVarName)
```
rmfitzpatrick, line 186: This could log secrets right?

bjsignalfx:
It looks like #450 is ready to be merge. If so, maybe merge then I rebase main? And yes I'll add documentation.

bjsignalfx:
@rmfitzpatrick So, if SPLUNK_CONFIG and SPLUNK_CONFIG_YAML are specified, which should take precedence?

rmfitzpatrick:
> It looks like #450 is ready to be merge. If so, maybe merge then I rebase main? And yes I'll add documentation.

I'm still waiting for a final review and will likely be merging tomorrow if it doesn't come in.



> @rmfitzpatrick So, if SPLUNK_CONFIG and SPLUNK_CONFIG_YAML are specified, which should take precedence? Maybe we should make these mutually exclusive because of equal precedence?

I think it makes sense they should be mutually exclusive (w/ `--config`) always taking priority.

cmd/otelcol/main.go
```diff
@@ -164,10 +156,9 @@ func checkRuntimeParams() {
 	// Check if memory ballast flag was passed
 	// If so, ensure memory ballast env var is not set
 	// Then set memory ballast and limit properly
-	ballastSize := getKeyValue(args, "--mem-ballast-size-mib")
+	ballastSize := getKeyValue(os.Args[1:], "--mem-ballast-size-mib")
 	if ballastSize != "" {
-		config = os.Getenv(ballastEnvVarName)
-		if config != "" {
+		if os.Getenv(ballastEnvVarName) != "" {
 			log.Fatalf("Both %v and '--config' were specified, but only one is allowed", ballastEnvVarName)
```
rmfitzpatrick, line 162:
```suggestion
			log.Fatalf("Both %v and '--mem-ballast-size-mib' were specified, but only one is allowed", ballastEnvVarName)
```

unrelated but currently incorrect

cmd/otelcol/main.go
```diff
@@ -176,32 +167,62 @@ func checkRuntimeParams() {
...
+	// Config file path from cmd flag --config.
+	pathFlag := getKeyValue(os.Args[1:], "--config")
+	// Config file path from env var.
+	pathVar := os.Getenv(configEnvVarName)
+	// Config YAML from env var.
+	yamlVar := os.Getenv(configYamlEnvVarName)
+
+	// Restricting specifying config file path and config YAML env vars simultaneously.
+	if pathVar != "" && yamlVar != "" {
+		log.Fatalf("Specifying env vars %s and %s simultaneously is not allowed", configEnvVarName, configYamlEnvVarName)
```
rmfitzpatrick, line 181: If pathFlag is provided I'm not sure this matters, but believe we'd want to clear both env vars.

cmd/otelcol/main.go
```diff
@@ -176,32 +167,62 @@ func checkRuntimeParams() {
...
+	}
+
+	if pathFlag == "" && yamlVar != "" {
+		log.Printf("Configuring collector using YAML from env var %s", configYamlEnvVarName)
```
rmfitzpatrick, line 185: `--config` should take precedence

rmfitzpatrick: err misread

rmfitzpatrick:
Please add a test in https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/container_test.go

cmd/otelcol/main.go
```diff
@@ -56,8 +59,7 @@ func main() {
 	// TODO: Use same format as the collector
 	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
 
-	args := os.Args[1:]
-	if !contains(args, "-h") && !contains(args, "--help") {
+	if !contains(os.Args[1:], "-h") && !contains(os.Args[1:], "--help") {
```
rmfitzpatrick, line 62: nit, why is this beneficial?

cmd/otelcol/main.go
```diff
@@ -71,11 +73,18 @@ func main() {
 		Version: version.Version,
 	}
 
+	baseParserProvider := parserprovider.Default()
```
rmfitzpatrick, line 76: This functionality belongs in `NewConfigSourceParserProvider()` in my opinion instead of being spread into main.

cmd/otelcol/main.go
```diff
@@ -164,44 +156,72 @@ func checkRuntimeParams() {
...
 }
 
-// Validate and set the configuration
-func setConfig() {
-	// Check if the config is specified via the env var.
-	config := os.Getenv(configEnvVarName)
-	// If not attempt to use a default config; supports Docker and local
-	if config == "" {
+// Validate and equate specified config file path flag to the config file path env var
+func setConfigSource() {
```
rmfitzpatrick, line 171: It would be helpful to describe what the priority and functionality of this function is.  It's pretty hard for me to follow and it's not clear to me that it's doing what's expected.

cmd/otelcol/main.go
```diff
@@ -164,44 +156,72 @@ func checkRuntimeParams() {
...
+	// Config file path from cmd flag --config.
+	pathFlag := getKeyValue(os.Args[1:], "--config")
+	// Config file path from env var.
+	pathVar := os.Getenv(configEnvVarName)
+	// Config YAML from env var.
+	yamlVar := os.Getenv(configYamlEnvVarName)
+
+	// Config file path flag `--config` should take priority when running from most contexts.
```
rmfitzpatrick, line 179: what contexts are outliers?

cmd/otelcol/main.go
```diff
@@ -164,44 +156,72 @@ func checkRuntimeParams() {
...
+	if pathFlag != "" {
+		// Config file path flag takes precedence over config YAML env var.
+		if yamlVar != "" {
+			log.Printf("Both environment variable %v and flag '--config' were specified. Ignoring configuration in the environment variable and using configuration in flag value %q", configYamlEnvVarName, pathFlag)
```
rmfitzpatrick, line 183: nit 
```suggestion
			log.Printf("Both %v and '--config' were specified. Ignoring %v and sourcing configuration from path %q", configYamlEnvVarName, configYamlEnvVarName, pathFlag)
```

cmd/otelcol/main.go
```diff
@@ -164,44 +156,72 @@ func checkRuntimeParams() {
...
+		}
+		// Config file path flag takes precedence over config file path env var.
+		if pathVar != "" && pathVar != pathFlag {
+			log.Printf("Both environment variable %v and flag '--config' were specified. Overriding %q environment variable value with flag value %q for this session", configEnvVarName, pathVar, pathFlag)
```
rmfitzpatrick, line 187:
```suggestion
			log.Printf("Both %v and '--config' were specified. Overriding %q environment variable value with %q for this session", configEnvVarName, pathVar, pathFlag)
```

cmd/otelcol/main.go
```diff
@@ -164,44 +156,72 @@ func checkRuntimeParams() {
...
+		}
+		// Setting the config file path env var to the config file path flag value.
+		pathVar = pathFlag
+		os.Setenv(configEnvVarName, pathVar)
+	} else {
+		// Restricting specifying config file path and config YAML env vars simultaneously.
+		if pathVar != "" && yamlVar != "" {
+			log.Fatalf("Specifying env vars %s and %s simultaneously is not allowed", configEnvVarName, configYamlEnvVarName)
```
rmfitzpatrick, line 195: I think `SPLUNK_CONFIG` should take priority so that `SPLUNK_CONFIG_YAML` will only be used if no path is specified.

cmd/otelcol/main.go
```diff
@@ -164,44 +156,72 @@ func checkRuntimeParams() {
...
-	// If not attempt to use a default config; supports Docker and local
-	if config == "" {
+// Validate and equate specified config file path flag to the config file path env var
+func setConfigSource() {
+	// Config file path from cmd flag --config.
+	pathFlag := getKeyValue(os.Args[1:], "--config")
+	// Config file path from env var.
+	pathVar := os.Getenv(configEnvVarName)
+	// Config YAML from env var.
+	yamlVar := os.Getenv(configYamlEnvVarName)
```
rmfitzpatrick, line 177: I think it would be beneficial to add a corresponding `--config-yaml` flag/option.  We could treat `-` as a marker to read from stdin.

cmd/otelcol/main.go
```diff
@@ -164,44 +156,72 @@ func checkRuntimeParams() {
...
+			log.Fatalf("Specifying env vars %s and %s simultaneously is not allowed", configEnvVarName, configYamlEnvVarName)
+		}
+		if yamlVar != "" {
+			log.Printf("Configuring collector using YAML from env var %s", configYamlEnvVarName)
+			return
+		}
+	}
+
+	// Use a default config if no config given; supports Docker and local
+	if pathVar == "" {
```
rmfitzpatrick, line 204: `config` as a variable was clearer imo since `pathVar` denotes that the flag/env var priorities is still in play.

cmd/otelcol/main_test.go
```diff
@@ -142,3 +148,34 @@ func TestSetMemoryLimit(t *testing.T) {
...
+		setConfigSource()
+	})
+
+	require.NotZero(t, exitCode)
+	assert.Containsf(t, stdErr, expectedErr, "Standard error does not contain the expected error")
+}
+
+// getExecErr should be called once in a test at the beginning and the failing code placed in func failingCode.
+// getExecErr runs the test in a subprocess then returns the output of executing failingCode.
+func getExecErr(t *testing.T, testName string, failingCode func(*testing.T)) (string, int) {
```
rmfitzpatrick, line 167: This doesn't seem like a standard go testing convention.  Why would running a test as a subprocess be necessary, especially one calling itself?

docs/getting-started/linux-manual.md
```diff
@@ -160,11 +162,29 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
 
 > Use of a SemVer tag over `latest` is highly recommended.
 
-In the case of Docker, a volume mount may be required to load custom
-configuration as shown above.
+In the case of Docker, a volume mount may be required to store the custom
```
rmfitzpatrick, line 165: "load" is accurate since the config should already exist (is in storage).

docs/getting-started/linux-manual.md
```diff
@@ -160,11 +162,29 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
...
 `SPLUNK_BALLAST_SIZE_MIB` environment variable. See
 [gateway_config.yaml](../../cmd/otelcol/config/collector/gateway_config.yaml)
 as an example.
+
+Using `SPLUNK_CONFIG_YAML` instead of `SPLUNK_CONFIG` would be something like:
+
+```bash
+$ SPLUNK_CONFIG_YAML="extensions:\n  health_check:\n    endpoint: 0.0.0.0:13133"
+
+$ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
```
rmfitzpatrick, line 179: Why are all these ports mapped if a restricted config is being used?

docs/getting-started/linux-manual.md
```diff
@@ -160,11 +162,29 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
...
+    -e SPLUNK_CONFIG_YAML=${SPLUNK_CONFIG_YAML} -p 13133:13133 -p 14250:14250 \
+    -p 14268:14268 -p 4317:4317 -p 6060:6060 -p 7276:7276 -p 8888:8888 \
+    -p 9080:9080 -p 9411:9411 -p 9943:9943 \
+    --name otelcol quay.io/signalfx/splunk-otel-collector:latest
+```
+Where `SPLUNK_CONFIG_YAML` is the one-line version of the multi-line configuration YAML below:
+```yaml
+extensions:
+  health_check:
+     endpoint: 0.0.0.0:13133
```
rmfitzpatrick, line 189: Is this complete?  A service definition is required to load the extension right?

bjsignalfx: No, it's not complete. It's just an example showing some yaml. Do you think people will confuse that for an actual collector config?

bjsignalfx: I'll add a note that the yaml is not a valid

cmd/otelcol/main.go
bjsignalfx: Oh man, can I submit another PR for this. Also, please elaborate, "We could treat - as a marker to read from stdin". I don't quite follow.

rmfitzpatrick: Yeah, something that can be added later.  `-` is often used as an instruction for the process to read a value from stdin `echo ok | vim -`

docs/getting-started/linux-manual.md
rmfitzpatrick: imo adding the basic service definition is adequate and doesn't leave an exercise for the reader

cmd/otelcol/main_test.go
rmfitzpatrick: Not seeing how this is resolved.  I have only seen equivalent functionality in multiprocessing plugins for test frameworks.  Please write a standard test case without this functionality, and use the integration testing testutils if you want to run a collector process: https://github.com/signalfx/splunk-otel-collector/tree/main/tests

docs/getting-started/linux-manual.md
```diff
@@ -160,11 +162,40 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
...
+```bash
+$ SPLUNK_CONFIG_YAML="receivers:\n   hostmetrics:\n      collection_interval: 10s\n      scrapers:\n         memory:\nexporters:\n   otlp:\n      endpoint: localhost:23456\n      insecure: true\nservice:\n   pipelines:\n      metrics:\n         receivers: [hostmetrics]\n         exporters: [otlp]"
+
+$ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 \
+    -e SPLUNK_REALM=us0 \
+    -e SPLUNK_CONFIG_YAML=${SPLUNK_CONFIG_YAML} \
+    -p 23456:23456 \
+    --name otelcol quay.io/signalfx/splunk-otel-collector:latest
+```
+Where `SPLUNK_CONFIG_YAML` is the single-line version of the multi-line configuration YAML below:
```
rmfitzpatrick, line 185: I feel like if we aren't demonstrating how to actually set this is an env var we create a strong possibility for repeated questions.

cmd/otelcol/main.go
```diff
@@ -169,66 +155,94 @@ func checkRuntimeParams() {
...
-			if len(os.Getenv(v)) == 0 {
-				log.Printf("Usage: %s=12345 %s=us0 %s", tokenEnvVarName, realmEnvVarName, os.Args[0])
-				log.Fatalf("ERROR: Missing required environment variable %s with default config path %s", v, config)
-			}
+func setConfigPathToDefault() {
+	// Checking environment variables required by default config files.
+	for _, v := range []string{realmEnvVarName, tokenEnvVarName} {
+		if len(os.Getenv(v)) == 0 {
+			log.Printf("Usage: %s=12345 %s=us0 %s", tokenEnvVarName, realmEnvVarName, os.Args[0])
+			log.Fatalf("ERROR: Missing required environment variable %s", v)
```
rmfitzpatrick, line 220: It was intentional that only default config paths are applicable for this requirement, even if specifically set by the user.  If this is to change I'd do it in another PR.

bjsignalfx: Yes, checking for this requirement only applies when setting default config. This chunk is in setConfigPathToDefault(). Maybe I should move the checking to the end of the function

docs/getting-started/linux-manual.md
bjsignalfx: I edited this sentence hoping to make it clearer.

cmd/otelcol/main.go
flands: What was the answer to this?

cmd/otelcol/main.go
```diff
@@ -169,44 +155,79 @@ func checkRuntimeParams() {
...
-	if config == "" {
-		_, err := os.Stat(defaultDockerSAPMConfig)
-		if err == nil {
-			config = defaultDockerSAPMConfig
+// Set config file path environment variable to specified config file path flag.
+// Set the config file path flag to the specified config file path environment variable if not specified.
+// Set the config file path flag, config file path environment variable to default config file path if the config
+// file path flag, config file path environment variable and config YAML environment variable are not specified.
+func setConfigPath() {
+	configPathFlag := getKeyValue(os.Args[1:], "--config")
```
flands, line 174: What is `--config` is not first? We do not mandate an order with upstream collector do we?

cmd/otelcol/main.go
```diff
@@ -37,6 +39,7 @@ import (
 const (
 	ballastEnvVarName         = "SPLUNK_BALLAST_SIZE_MIB"
 	configEnvVarName          = "SPLUNK_CONFIG"
```
flands, line 41: We should really create `SPLUNK_CONFIG_FILE` to avoid confusion. Even in beta we may consider deprecating this -- or we could remove it. Perhaps we can do this in a separate PR, but then a GH issue should be filed.

flands (commented):
A few questions -- also I do not see test coverage for the new functionality.

cmd/otelcol/main.go
bjsignalfx: Though the re-slicing is slightly inefficient, I prefer that to creating new variables. Plus, this code runs on starting only.

cmd/otelcol/main.go
bjsignalfx: I don't follow your question, "What is --config is not first?".

flands: Sorry -- what _if_: `otelcol --metrics-addr 0.0.0.0:8123 --config foo.yaml`

cmd/otelcol/main.go
```diff
@@ -109,44 +112,27 @@ func contains(arr []string, str string) bool {
...
+	for i := range args {
 		switch {
-		case strings.HasPrefix(arg, argName+"="):
-			s := strings.Split(arg, "=")
-			val = s[1]
-		case arg == argName:
-			i++
-			val = args[i]
+		case strings.HasPrefix(args[i], argEq):
+			if s := strings.SplitN(args[i], "=", 2); len(s) == 2 {
```
rmfitzpatrick, line 120: When would `len(s)` not be 2?

docs/getting-started/linux-manual.md
```diff
@@ -160,11 +163,44 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
...
+In the case of Docker, a volume mount may be required to load the custom
+configuration file as shown above.
 
 If the custom configuration includes a `memory_limiter` processor then the
 `ballast_size_mib` parameter should be the same as the
 `SPLUNK_BALLAST_SIZE_MIB` environment variable. See
 [gateway_config.yaml](../../cmd/otelcol/config/collector/gateway_config.yaml)
 as an example.
+
+In order to use `SPLUNK_CONFIG_YAML`, the configuration YAML needs to be
```
rmfitzpatrick, line 175: It could be helpful to show something like `cat agent_config.yaml | awk 1 ORS='\\n'` for how they could grab this manually.

cmd/otelcol/main.go
```diff
@@ -169,44 +155,79 @@ func checkRuntimeParams() {
...
+	_, err = os.Stat(defaultLocalSAPMConfig)
+	if err == nil {
+		path = defaultLocalSAPMConfig
+	}
+	if path == "" {
+		log.Fatalf("Unable to find the default configuration file, ensure %s environment variable is set properly", configEnvVarName)
 	}
 
-	switch config {
+	switch path {
```
rmfitzpatrick, line 230: By pushing this to the
```
	if configPathFlag == "" && configPathVar == "" && configYaml == ""
```
case only env var validation won't occur for default installation paths that use the config env var (all of them iirc).  This should occur separately.

tests/general/container_test.go
```diff
@@ -115,3 +115,65 @@ func TestSpecifiedContainerConfigDefaultsToCmdLineArgIfEnvVarConflict(t *testing
...
+	//
+	configYamlStr := "receivers:\n  hostmetrics:\n    collection_interval: 1s\n    scrapers:\n      cpu:\nexporters:\n  logging:\n    logLevel: debug\nservice:\n  pipelines:\n    metrics:\n      receivers: [hostmetrics]\n      exporters: [logging]"
+	configYamlEnv := map[string]string{"SPLUNK_CONFIG_YAML": configYamlStr}
+
+	logCore, logs := observer.New(zap.DebugLevel)
+	logger := zap.New(logCore)
+
+	collector, err := testutils.NewCollectorContainer().
+		WithImage(image).
+		WithEnv(configYamlEnv).
```
rmfitzpatrick, line 149: I could be wrong but if you don't set `SPLUNK_CONFIG: ""` and are seeing the that the yaml is being used this suggests something isn't going as expected in your changes: https://github.com/signalfx/splunk-otel-collector/blob/main/tests/testutils/collector_container.go#L192

bjsignalfx: So, the execution path never gets to line 192 in this case because collector.ConfigPath is zero here https://github.com/signalfx/splunk-otel-collector/blob/main/tests/testutils/collector_container.go#L162

docs/getting-started/linux-manual.md
```diff
@@ -140,12 +140,18 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
...
-environment variable or the `--config` command line argument to provide a
-custom configuration.
+configuration file can be specified and used instead. Use the `SPLUNK_CONFIG`
+environment variable or the `--config` command line argument to provide the 
+path to a file containing custom configuration YAML. Also, environment
+variable `SPLUNK_CONFIG_YAML` is available for specifying custom
+configuration YAML directly. This is useful in environments where access
+to the underlying filesystem is not readily available. For example, in 
+AWS Fargate it is more convenient to pull configuration YAML from a 
+parameter in the AWS Systems Manager Parameter Store.
```
tmorrisonsfx, line 150: Suggested change: "When changes to the default configuration file are needed, create a custom configuration  YAML file. Use the `SPLUNK_CONFIG` environment variable or the `--config` command line argument to provide the path to this file. You can also use the `SPLUNK_CONFIG_YAML` environment variable to specify a custom configuration YAML directly.  This is useful in environments where access to the underlying file system is not readily available. For example, in AWS Fargate, it is more convenient to pull the configuration YAML file from a parameter in the AWS Systems Manager Parameter Store."

It's not exactly clear what this sentence means - "You can also use the `SPLUNK_CONFIG_YAML` environment variable to specify a custom configuration YAML directly. " To specify where a configuration YAML file is pulled from perhaps? Or, to specify a particular configuration YAML file that you want to use?

docs/getting-started/linux-manual.md
```diff
@@ -140,12 +140,18 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
...
 
 > Command line arguments take precedence over environment variables. This
-> applies to `--config` and `--mem-ballast-size-mib`.
+> applies to `--config` and `--mem-ballast-size-mib`. `SPLUNK_CONFIG` 
+> takes precedence over `SPLUNK_CONFIG_YAML`
```
tmorrisonsfx, line 155: "... takes precedence over `SPLUNK_CONFIG_YAML`."

docs/getting-started/linux-manual.md
```diff
@@ -160,11 +166,35 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
 
 > Use of a SemVer tag over `latest` is highly recommended.
 
-In the case of Docker, a volume mount may be required to load custom
-configuration as shown above.
+In the case of Docker, a volume mount may be required to load the custom
+configuration file as shown above.
```
tmorrisonsfx, line 170: "...file, as shown above."

docs/getting-started/linux-manual.md
```diff
@@ -160,11 +166,35 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
...
 
 If the custom configuration includes a `memory_limiter` processor then the
 `ballast_size_mib` parameter should be the same as the
 `SPLUNK_BALLAST_SIZE_MIB` environment variable. See
 [gateway_config.yaml](../../cmd/otelcol/config/collector/gateway_config.yaml)
 as an example.
+
+`SPLUNK_CONFIG_YAML` example:
```
tmorrisonsfx, line 178: Suggested change: "The following example shows the `SPLUNK_CONFIG_YAML` environment variable:"
 
I would also add more detail to say exactly what is happening in the example. For example, what happens when they run `docker run` with all of these options?

tmorrisonsfx:
@bjsignalfx I reviewed the linux-manual.md file. Included a few suggested changes. Thanks!

tests/general/container_test.go
bjsignalfx: @rmfitzpatrick does my response make sense?

docs/getting-started/linux-manual.md
```diff
@@ -139,13 +139,22 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
...
-When changes to the default configuration file is needed, a custom
-configuration file can specified and used instead. Use the `SPLUNK_CONFIG`
-environment variable or the `--config` command line argument to provide a
-custom configuration.
+When changes to the default configuration YAML file are needed, create a
+custom configuration file. Use environment variable `SPLUNK_CONFIG` or
+command line argument `--config` to provide the path to this file.
+
+Also, you can use environment variable `SPLUNK_CONFIG_YAML` to specify 
+your costum configuration YAML at the command line. This is useful in
```
rmfitzpatrick, line 147:
```suggestion
your custom configuration YAML at the command line. This is useful in
```

docs/getting-started/linux-manual.md
```diff
@@ -139,13 +139,22 @@ $ docker run --rm -e SPLUNK_ACCESS_TOKEN=12345 -e SPLUNK_REALM=us0 \
...
+environments where access to the underlying file system is not readily
+available. For example, in AWS Fargate, access to the filesystem is 
+not readily available. There, you can store your custom configuration
```
rmfitzpatrick, line 150:
```suggestion
environments where access to the underlying file system is not readily
available. For example, in AWS Fargate you can store your custom configuration
```
removes some redundancy

cmd/otelcol/main.go
```diff
@@ -169,66 +155,119 @@ func checkRuntimeParams() {
...
-			log.Fatalf("Unable to find the configuration file (%s) ensure %s environment variable is set properly", config, configEnvVarName)
+
+		log.Printf("Set config to %v", configPathFlag)
+		return
+	}
+
+	if configPathVar != "" {
+		validateConfigPath(configPathVar)
+
+		if !contains(os.Args[1:], "--config") {
```
rmfitzpatrick, line 207: This should always be the case since you're returning where `configPathFlag != ""` right?

cmd/otelcol/main.go
```diff
@@ -169,66 +155,119 @@ func checkRuntimeParams() {
...
-// Validate and set the configuration
-func setConfig() {
-	// Check if the config is specified via the env var.
-	config := os.Getenv(configEnvVarName)
-	// If not attempt to use a default config; supports Docker and local
-	if config == "" {
-		_, err := os.Stat(defaultDockerSAPMConfig)
-		if err == nil {
-			config = defaultDockerSAPMConfig
+// Validate then set config file path environment variable to specified config file path flag.
```
rmfitzpatrick, line 169: I don't think this is "necessary" anymore since the function that set the flag from the env var value was removed in these changes.

cmd/otelcol/main.go
```diff
@@ -169,66 +155,119 @@ func checkRuntimeParams() {
...
+
+	if configPathFlag != "" {
+		validateConfigPath(configPathFlag)
+
+		if configPathVar != "" && configPathVar != configPathFlag {
+			log.Printf("Both %v and '--config' were specified. Overriding %q environment variable value with %q for this session", configEnvVarName, configPathVar, configPathFlag)
 		}
-		if config == "" {
-			log.Fatalf("Unable to find the default configuration file, ensure %s environment variable is set properly", configEnvVarName)
+		os.Setenv(configEnvVarName, configPathFlag)
```
rmfitzpatrick, line 194: https://github.com/signalfx/splunk-otel-collector/pull/462/files#r662610037

cmd/otelcol/main.go
```diff
@@ -169,66 +155,119 @@ func checkRuntimeParams() {
...
+			log.Printf("Both %s and %s were specified. Using %s environment variable value %s for this session", configEnvVarName, configYamlEnvVarName, configEnvVarName, configPathVar)
 		}
+
+		log.Printf("Set config to %v", configPathVar)
+	}
+}
+
+// Validate and set the config file path flag and config file path environment variable to the default config file path.
+// Only Docker and local configuration supported.
+func setValidDefaultConfigPath() {
```
rmfitzpatrick, line 222: nit this bombs if not possible so maybe `setDefaultConfigPathIfExists()`?

cmd/otelcol/main.go
```diff
@@ -169,66 +155,119 @@ func checkRuntimeParams() {
...
+		path = defaultLocalSAPMConfig
+	}
+
+	if path == "" {
+		log.Fatalf("Unable to find the default configuration file, ensure %s environment variable is set properly", configEnvVarName)
+	}
+
+	validateConfigPath(path)
+
+	os.Setenv(configEnvVarName, path)
```
rmfitzpatrick, line 240: Don't believe this is necessary anymore

cmd/otelcol/main.go
```diff
@@ -169,66 +155,119 @@ func checkRuntimeParams() {
...
+
+	if !contains(os.Args[1:], "--config") {
```
rmfitzpatrick, line 242: why would this ever not be the case?

cmd/otelcol/main.go
```diff
@@ -169,66 +155,119 @@ func checkRuntimeParams() {
...
+		os.Args = append(os.Args, "--config="+path)
+	}
+
+	log.Printf("Set config to %v", path)
+}
+
+func validateConfigPath(path string) {
+	// Check if the config file exists.
+	if _, err := os.Stat(path); err != nil {
+		log.Fatalf("Unable to find the configuration file (%s) ensure %s environment variable is set properly", path, configEnvVarName)
```
rmfitzpatrick, line 253: They could be using `--config` so this message would be confusing

cmd/otelcol/main.go
bjsignalfx: My understanding is that we always sync flag --config with env var SPLUNK_CONFIG with --config having the highest priority. So, here start by saying SPLUNK_CONFIG is set to "specified" --config value period. And in the next sentence I say --config is set to "specified" SPLUNK_CONFIG value if empty. Then defaults if --config, SPLUNK_CONFIG and SPLUNK_CONFIG_YAML are not specified..

cmd/otelcol/main.go
bjsignalfx: I think we have been overriding the SPLUNK_CONFIG value with the --config value and that's what I'm doing here.

cmd/otelcol/main.go
bjsignalfx: No, this is when configPathFlag == "" and configPathVar != ""

cmd/otelcol/main.go
bjsignalfx: I think it is. I'm setting SPLUNK_CONFIG to the default path

cmd/otelcol/main.go
bjsignalfx: We set the default path when --config, SPLUNK_CONFIG and SPLUNK_CONFIG_YAML are all zero.

cmd/otelcol/main.go
bjsignalfx: I think the name is appropriate because the scope of this function is simply to validate and set --config and SPLUNK_CONFIG to the default path.

cmd/otelcol/main.go
rmfitzpatrick: `configPathFlag == ""` means that `!contains(os.Args[1:], "--config")` will always be true, so why check?

cmd/otelcol/main.go
rmfitzpatrick: You've resolved this without answering or removing the seemingly unnecessary `len(s) == 2` condition

cmd/otelcol/main.go
```diff
@@ -169,66 +161,99 @@ func checkRuntimeParams() {
...
+		return
+	}
+
+	if configYaml != "" {
+		log.Printf("Using environment variable %s for configuration", configYamlEnvVarName)
+		return
+	}
+
+	defaultConfigPath := getExistingDefaultConfigPath()
+	validateDefaultConfig(defaultConfigPath)
```
rmfitzpatrick, line 225: https://github.com/signalfx/splunk-otel-collector/pull/462#discussion_r661566937

required env var validation now only occurs if no config arg or env var are used, which is a functional change.  The collector service sets the env var w/ the default value so `validateDefaultConfig(configPathVar)` should be used above.  I think `validateDefaultConfig(configPathFlag)` should be checked as well to preserve existing functionality.

cmd/otelcol/main.go
bjsignalfx: Sorry, weird! I actually remember giving an example where I think len(s) can be equal to 1 when for instance args[i]="config=".

cmd/otelcol/main.go
bjsignalfx: Yes, I changed the name of the function to checkRequiredEnvVars() and checked both configPathVar and configPathFlag

cmd/otelcol/main.go
rmfitzpatrick: `strings.HasPrefix(args[i], argEq)` would be true in that case though?

bjsignalfx: Yes, I remove the unnecessary check

rmfitzpatrick (approved):

PR #469, opened 2021-06-11
Use internal pipelines for collector prometheus metrics

rmfitzpatrick:
These changes add a new metrics pipeline w/ internal prometheus scraper and a resource detector that overrides attributes (since the `host.name` is 0.0.0.0 as reported by prometheus).

### DIFF
--- a/cmd/otelcol/config/collector/agent_config.yaml
+++ b/cmd/otelcol/config/collector/agent_config.yaml
@@ -65,7 +65,7 @@ receivers:
         endpoint: 0.0.0.0:55681
   # This section is used to collect the OpenTelemetry Collector metrics
   # Even if just a Splunk APM customer, these metrics are included
-  prometheus:
+  prometheus/internal:
     config:
       scrape_configs:
       - job_name: 'otel-collector'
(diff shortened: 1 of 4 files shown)


mstumpfx (approved):

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -100,12 +100,17 @@ processors:
...
-  ## important for creating unique cloud provider dimensions
+  # detect if the collector is running on a cloud system
+  # important for creating unique cloud provider dimensions
   resourcedetection:
     detectors: [system, gce, ecs, ec2, azure]
     override: false
 
+  # Same as above but overrides resource attributes set by receivers
+  resourcedetection/internal:
+    detectors: [system, gce, ecs, ec2, azure]
```
dmitryax, line 111: Why do we need `ecs` at this point?

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -100,12 +100,17 @@ processors:
...
+    override: true
```
dmitryax, line 112: Actually is there a reason to create another receiver ? we already have one with the same config

cmd/otelcol/config/collector/agent_config.yaml
dmitryax: nvm. I see it's the same as another one

cmd/otelcol/config/collector/agent_config.yaml
dmitryax: ok, nvm, it's not the config :)

dmitryax (approved):

PR #470, opened 2021-06-11
translatesfx: translate globalDimensions to otel metricstransform processor

pmcollins:


cmd/translatesfx/translatesfx/cli.go
```diff
@@ -81,3 +72,13 @@ func translateConfig(fname, wd string) {
...
 	fmt.Println(string(bytes))
 }
+
+func loadCfg(fname string) (interface{}, error) {
+	bytes, err := os.ReadFile(fname)
+	if err != nil {
+		return nil, err
+	}
+	var orig interface{}
+	err = yaml.UnmarshalStrict(bytes, &orig)
```
rmfitzpatrick, line 82: Doesn't need to happen in these changes, but it would probably be a good idea to unmarshal to a SA config directly to avoid unnecessary edge case:  https://github.com/signalfx/signalfx-agent/blob/main/pkg/core/config/config.go#L33

We offer this functionality here and it loads all the defaults as well: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/extension/smartagentextension/config.go#L70

rmfitzpatrick (approved):

PR #481, opened 2021-06-16
Add telegraf/sqlserver integration test

rmfitzpatrick:
~~Requires a release with https://github.com/signalfx/signalfx-agent/pull/1802 to verify tls is supported w/ 1.16 builds.~~

These changes verify telegraf/sqlserver SA receiver functionality w/ barebones resource metric definitions.

Also fixes a testutils bug w/ destroyed collector process environment variables and applies formatting.


tests/testutils/collector_process.go
```diff
@@ -43,7 +43,7 @@ type CollectorProcess struct {
 
 // To be used as a builder whose Build() method provides the actual instance capable of launching the process.
```
pmcollins, line 44: Not part of this diff, but my IDE (goland) flags these comments as warnings because they don't start with the name of the function.

pmcollins (approved):
LGTM, just one incidental comment.

tests/testutils/collector_process.go
rmfitzpatrick: Do you know what linter triggers that?  We could add it to the golangci-lint configuration and update as needed in another PR.

pmcollins: Apparently staticcheck will do it https://staticcheck.io/docs/checks#ST1020
staticcheck appears to be enabled but maybe needs to be configured to do this check?

pmcollins: I experimented with telling staticcheck to do all the checks, and it will report these issues elsewhere, but for some reason not in this directory (maybe because it's named test/testutils idk).

PR #497, opened 2021-06-22
Enable collecting in memory config locally by default

flands:
Configured main.go to enable in memory config gathering locally. In
addition, updated support bundle to capture this information.

### DIFF
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -35,12 +35,13 @@ import (
 )
 
 const (
-	ballastEnvVarName     = "SPLUNK_BALLAST_SIZE_MIB"
-	configEnvVarName      = "SPLUNK_CONFIG"
-	memLimitMiBEnvVarName = "SPLUNK_MEMORY_LIMIT_MIB"
-	memTotalEnvVarName    = "SPLUNK_MEMORY_TOTAL_MIB"
-	realmEnvVarName       = "SPLUNK_REALM"
-	tokenEnvVarName       = "SPLUNK_ACCESS_TOKEN"
+	ballastEnvVarName          = "SPLUNK_BALLAST_SIZE_MIB"
+	configEnvVarName           = "SPLUNK_CONFIG"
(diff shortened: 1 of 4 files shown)


cmd/otelcol/main.go
```diff
@@ -35,12 +35,13 @@ import (
...
 const (
-	ballastEnvVarName     = "SPLUNK_BALLAST_SIZE_MIB"
-	configEnvVarName      = "SPLUNK_CONFIG"
-	memLimitMiBEnvVarName = "SPLUNK_MEMORY_LIMIT_MIB"
-	memTotalEnvVarName    = "SPLUNK_MEMORY_TOTAL_MIB"
-	realmEnvVarName       = "SPLUNK_REALM"
-	tokenEnvVarName       = "SPLUNK_ACCESS_TOKEN"
+	ballastEnvVarName          = "SPLUNK_BALLAST_SIZE_MIB"
+	configEnvVarName           = "SPLUNK_CONFIG"
+	configServerEnabledEnvVar = "SPLUNK_DEBUG_CONFIG_SERVER"
```
pjanotti, line 40: nit: spacing 
```suggestion
	configServerEnabledEnvVar  = "SPLUNK_DEBUG_CONFIG_SERVER"
```

cmd/otelcol/main.go
```diff
@@ -61,6 +62,10 @@ func main() {
 		checkRuntimeParams()
 	}
 
+	// Allow dumping configuration locally by default
+	// Used by support bundle script
+	os.Setenv(configServerEnabledEnvVar, "true")
```
pjanotti, line 67: fyi: if the port is not available it would fail to start - I don't recall the exact message in this case.

pjanotti (approved):
LGTM, a minor formatting nit + and FYI.

cmd/otelcol/main.go
flands: bind: port already in use -- already hit it :)

rmfitzpatrick (approved):

PR #499, opened 2021-06-22
translatesfx: add resourcedetection processor

pmcollins:


cmd/translatesfx/translatesfx/otel.go
```diff
@@ -33,28 +33,33 @@ func saInfoToOtelConfig(cfg saCfgInfo) otelCfg {
...
+	const rd = "resourcedetection"
 	out := otelCfg{
 		ConfigSources: map[string]interface{}{
 			// TODO check if should be nil
 			"include": nil,
 		},
 		Receivers: receivers,
+		Processors: map[string]interface{}{
+			rd: map[string]interface{}{
+				"detectors": []string{"system", "env"},
```
mstumpfx, line 45: I wonder if we should add the cloud provider resource detectors for generating their respective unique IDs (`AWSUniqueId`, `gcp_id`, `azure_resource_id`)?

pmcollins: Hmm not sure -- at this point I'm targeting/automating the stuff that [this document](https://github.com/signalfx/splunk-otel-collector/blob/main/docs/signalfx-smart-agent-migration.md) recommends.

Thoughts @rmfitzpatrick?

dmitryax (commented):
Could you change the commit message so it's clear that the change is applied to the translator?

pmcollins:
> Could you change the commit message so it's clear that the change is applied to the translator?

Done

cmd/translatesfx/translatesfx/otel.go
rmfitzpatrick: I think the end goal for the tool is not just 1:1 config content but also all-around agent experience.  Believe the agent detects the cloud provider w/ its hostid utility, which adding all the resource detector targets would effectively do as well.  My concern would be the chattiness and continued noise w/ failed detection attempts though.

dmitryax: I think we should be aligned with the default config where we have all of them added by default https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/agent_config.yaml#L106

pmcollins: Thanks guys for the feedback. I've added "gce", "ecs", "ec2", and "azure" resource detectors.

pmcollins:
Looks like one of the tests is flaky, not sure how because failures only happen in CI and there's no log info, but I'm going to see if I can track it down and fix it.

pmcollins:
Any further comments/questions/concerns?

mstumpfx (approved):
lgtm

PR #547, opened 2021-07-14
Upgrade to otel core/contrib 0.30.0

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>


owais (approved):

internal/receiver/smartagentreceiver/output.go
```diff
@@ -91,11 +91,7 @@ func getMetadataExporters(
 
 	metadataExporters, noClientsSpecified := getDimensionClientsFromMetricsExporters(cfg.DimensionClients, host, nextMetricsConsumer, logger)
 	for _, client := range metadataExporters {
-		if metadataExporter, ok := (*client).(metadata.MetadataExporter); ok {
-			exporters = append(exporters, metadataExporter)
-		} else {
-			logger.Info("cannot send dimension updates to dimension client", zap.Any("client", *client))
-		}
```
jrcamp, line 98: Why removed?

bogdandrutu: Because as you can see I can append without a cast (means the variable was already that type) :))

bogdandrutu: I think I have a different linter (newer version on my machine) which found this issue.

jrcamp: Ah I figured it was returning `[]interface{}` and then checking types but you're right `getDimensionClientsFromMetricsExporters` is already returning `[]metadata.MetadataExporter`

jrcamp (approved):

PR #558, opened 2021-07-16
Fix memory limit calculation for deployments with 20Gi+ of total memory

dmitryax:
Bug: If `SPLUNK_MEMORY_TOTAL_MIB` set to more than 20480, the calculated memory limit drops from 90% to 2048 and collector is failing to start with 
```
main.go:283: Memory limit (2048) is less than 2x ballast (6761). Increase memory limit or decrease ballast size.
```

This change fixes the bug and removes the unnecessary complication for default memory limit calculation. If user wants to adjust memory allocation for a 20Gb+ deployment they can always use `SPLUNK_MEMORY_LIMIT_MIB` env var.

### DIFF
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -52,7 +52,6 @@ const (
 	defaultLocalOTLPConfig         = "cmd/otelcol/config/collector/otlp_config_linux.yaml"
 	defaultMemoryBallastPercentage = 33
 	defaultMemoryLimitPercentage   = 90
-	defaultMemoryLimitMaxMiB       = 2048
 	defaultMemoryTotalMiB          = 512
 )
 
@@ -291,11 +290,6 @@ func setMemoryLimit(memTotalSizeMiB int) {
 	// If not, calculate it from memTotalSizeMiB
 	if memoryLimit == "" {
 		memLimit = memTotalSizeMiB * defaultMemoryLimitPercentage / 100
-		// The memory limit should be set to defaultMemoryLimitPercentage of total memory
-		// while reserving a maximum of defaultMemoryLimitMaxMiB of memory.
-		if (memTotalSizeMiB - memLimit) > defaultMemoryLimitMaxMiB {
-			memLimit = defaultMemoryLimitMaxMiB
-		}
 	} else {
 		memLimit, _ = strconv.Atoi(memoryLimit)
 	}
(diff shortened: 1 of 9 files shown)


pjanotti (approved):

flands:
@dmitryax I cannot reproduce this -- can you please provide the exact command and version to see the described behavior?

```
SPLUNK_MEMORY_TOTAL_MIB=20480 SPLUNK_ACCESS_TOKEN=12345 SPLUNK_API_URL=https://api.us0.signalfx.com SPLUNK_INGEST_URL=https://ingest.us0.signalfx.com SPLUNK_TRACE_URL=https://ingest.us0.signalfx.com/v2/trace SPLUNK_HEC_URL=https://ingest.us0.signalfx.com/v1/log SPLUNK_HEC_TOKEN=12345 SPLUNK_DEBUG_CONFIG_SERVER_PORT=55556 ./otelcol_darwin_amd64 --config ../cmd/otelcol/config/collector/agent_config.yaml
```

Using the above -- which I would assume would repro given the description -- I am able to start the collector and navigating to `http://localhost:55556/debug/configz/effective` shows a properly configured processor. I agree the logic is complicated but this comes directly from the memory processor documentation.

dmitryax:
@flands `20480` may be right on the edge. Please try `SPLUNK_MEMORY_TOTAL_MIB=20490`

cmd/otelcol/main.go
```diff
@@ -291,11 +290,6 @@ func setMemoryLimit(memTotalSizeMiB int) {
 	// If not, calculate it from memTotalSizeMiB
 	if memoryLimit == "" {
 		memLimit = memTotalSizeMiB * defaultMemoryLimitPercentage / 100
-		// The memory limit should be set to defaultMemoryLimitPercentage of total memory
-		// while reserving a maximum of defaultMemoryLimitMaxMiB of memory.
-		if (memTotalSizeMiB - memLimit) > defaultMemoryLimitMaxMiB {
-			memLimit = defaultMemoryLimitMaxMiB
```
flands, line 297: Keeping logic you can just change this to:

```
memLimit = (memTotalSizeMiB - defaultMemoryLimitMaxMiB)
```

flands (commented):
OK, confirmed. So keeping logic can just update line 297 otherwise we can remove logic. Only concern is that as memory increases more and more memory will be "wasted". I am good either way -- note other YAML config will need comments updated.

cmd/otelcol/main.go
dmitryax: Yes, we discussed that option with @pjanotti and agreed on removing the logic.

dmitryax:
>  I am good either way -- note other YAML config will need comments updated.
@flands I think I already updated all the configurations. Please let me know if you see any of them missed

bogdandrutu (approved):

rmfitzpatrick (approved):

PR #572, opened 2021-07-23
Set required URL and TOKEN env vars for agent config

flands:
If you deploy Splunk OTel Connector outside of installer script then the
gateway configuration is used. If the agent configuration was desired,
it was really hard to switch:

```
$ > SPLUNK_ACCESS_TOKEN=1234 SPLUNK_REALM=us0 SPLUNK_CONFIG=../cmd/otelcol/config/collector/agent_config.yaml ./otelcol_darwin_amd64
2021/07/23 11:30:17 main.go:198: Set config to ../cmd/otelcol/config/collector/agent_config.yaml
2021/07/23 11:30:17 main.go:283: Set ballast to 168 MiB
2021/07/23 11:30:17 main.go:307: Set memory limit to 460 MiB
2021-07-23T11:30:17.907-0400	info	service/collector.go:283	Starting otelcol...	{"Version": "v0.29.0-43-g491b2f0", "NumCPU": 12}
2021-07-23T11:30:17.910-0400	info	service/collector.go:343	Using memory ballast	{"MiBs": 168}
2021-07-23T11:30:17.910-0400	info	service/collector.go:188	Setting up own telemetry...
2021-07-23T11:30:17.919-0400	info	service/telemetry.go:99	Serving Prometheus metrics	{"address": ":8888", "level": 0, "service.instance.id": "c928a31c-d214-4287-b7bb-d2b802138d1c"}
2021-07-23T11:30:17.919-0400	info	service/collector.go:224	Loading configuration...
2021-07-23T11:30:17.939-0400	info	service/collector.go:240	Applying configuration...
Error: cannot build extensions: cannot build builtExtensions: failed to create extension http_forwarder: 'egress.endpoint' config option cannot be empty
2021/07/23 11:30:17 main.go:94: application run finished with error: cannot build extensions: cannot build builtExtensions: failed to create extension http_forwarder: 'egress.endpoint' config option cannot be empty
```

To make it work:

```
SPLUNK_ACCESS_TOKEN=1234 SPLUNK_REALM=us0 SPLUNK_CONFIG=../cmd/otelcol/config/collector/agent_config.yaml \
  SPLUNK_API_URL=https://api.us0.signalfx.com SPLUNK_INGEST_URL=https://ingest.us0.signalfx.com \
  SPLUNK_TRACE_URL=https://ingest.us0.signalfx.com/v2/trace SPLUNK_HEC_URL=https://ingest.us0.signalfx.com/v1/log\
  ./otelcol_darwin_amd64
```

This is awful customer experience. The behavior is not surprising as the
agent config needs to optimize for both direct to SaaS and via gateway
data routing. The installer script mitigates this today.

The good news is that this is straightforward to address. Given
`SPLUNK_REALM` is required and the SaaS URLs are known and same for
`SPLUNK_ACCESS_TOKEN` we can set the required environment variable if
not manually specified.

### DIFF
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -289,19 +289,26 @@ func setMemoryLimit(memTotalSizeMiB int) int {
 
 // Sets environment variables expected by agent_config.yaml if missing
 func setDefaultEnvVars() {
-	realm := os.Getenv("SPLUNK_REALM")
-	token := os.Getenv("SPLUNK_ACCESS_TOKEN")
-	testArgs := [][]string{
-		{"SPLUNK_API_URL", "https://api." + realm + ".signalfx.com"},
-		{"SPLUNK_INGEST_URL", "https://ingest." + realm + ".signalfx.com"},
-		{"SPLUNK_TRACE_URL", "https://ingest." + realm + ".signalfx.com/v2/trace"},
-		{"SPLUNK_HEC_URL", "https://ingest." + realm + ".signalfx.com/v1/log"},
-		{"SPLUNK_HEC_TOKEN", token},
+	realm, realmOk := os.LookupEnv("SPLUNK_REALM")
+	if realmOk {
+		testArgs := [][]string{
+			{"SPLUNK_API_URL", "https://api." + realm + ".signalfx.com"},
+			{"SPLUNK_INGEST_URL", "https://ingest." + realm + ".signalfx.com"},
+			{"SPLUNK_TRACE_URL", "https://ingest." + realm + ".signalfx.com/v2/trace"},
+			{"SPLUNK_HEC_URL", "https://ingest." + realm + ".signalfx.com/v1/log"},
+		}
+		for _, v := range testArgs {
+			_, ok := os.LookupEnv(v[0])
+			if !ok {
+				os.Setenv(v[0], v[1])
+			}
+		}
 	}
-	for _, v := range testArgs {
-		_, ok := os.LookupEnv(v[0])
+	token, tokenOk := os.LookupEnv("SPLUNK_ACCESS_TOKEN")
+	if tokenOk {
+		_, ok := os.LookupEnv("SPLUNK_HEC_TOKEN")
 		if !ok {
-			os.Setenv(v[0], v[1])
+			os.Setenv("SPLUNK_HEC_TOKEN", token)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


cmd/otelcol/main.go
```diff
@@ -306,6 +307,26 @@ func setMemoryLimit(memTotalSizeMiB int) {
...
+	token := os.Getenv("SPLUNK_ACCESS_TOKEN")
+	testArgs := [][]string{
+		{"SPLUNK_API_URL", "https://api." + realm + ".signalfx.com"},
+		{"SPLUNK_INGEST_URL", "https://ingest." + realm + ".signalfx.com"},
+		{"SPLUNK_TRACE_URL", "https://ingest." + realm + ".signalfx.com/v2/trace"},
+		{"SPLUNK_HEC_URL", "https://ingest." + realm + ".signalfx.com/v1/log"},
+		{"SPLUNK_HEC_TOKEN", token},
+	}
+	for _, v := range testArgs {
+		tmpVar = os.Getenv(v[0])
```
rmfitzpatrick, line 323: This would prevent setting empty values, which could be desired for testing or other edge cases.  Would it make sense to use `LookupEnv()` and only set if not in the current env?

flands: Hmm, that does seem to make sense -- let me take a look

cmd/otelcol/main_test.go
```diff
@@ -112,6 +112,60 @@ func TestCheckRuntimeParams_LimitAndBallastEnvs(t *testing.T) {
...
+	args := os.Args[1:]
+	_, c := getKeyValue(args, "--mem-ballast-size-mib")
+	if c != val {
+		t.Errorf("Expected memory ballast CLI param %v got %v", val, c)
+	}
+	b := os.Getenv(ballastEnvVarName)
+	if b != val {
+		t.Errorf("Expected memory ballast %v got %v", val, b)
+	}
+}
```
flands, line 167:
```suggestion
```

cmd/otelcol/main.go
```diff
@@ -286,6 +287,25 @@ func setMemoryLimit(memTotalSizeMiB int) int {
 	return memLimit
 }
 
+// Sets environment variables expected by agent_config.yaml if missing
+func setDefaultEnvVars() {
+	realm := os.Getenv("SPLUNK_REALM")
+	token := os.Getenv("SPLUNK_ACCESS_TOKEN")
```
rmfitzpatrick, line 293: realized that if these aren't set invalid urls would be used for the default env vars, which may still be used by a non-default config (the only case where realm and token would be allowed to be unset) and lead to ~hard to track down undesired behavior in edge cases.  Would probably be worth only adding the applicable inner arrays to `testArgs` if realm and then token are detected.

rmfitzpatrick (approved):

PR #590, opened 2021-08-01
Make smartagent/signalfx-forwarder compatible with k8s_tagger processor

owais:
This commit copies the source/reporter client IP from SFX span metadata
to Otel Resource as an attribute. This allows the k8s_tagger processor
to correctly identify k8s pods when they send data to
smartagent/signalfx-forwarder receiver.


internal/receiver/smartagentreceiver/converter/traces.go
```diff
@@ -24,12 +25,44 @@ import (
...
+	}
+
+	traces := pdata.NewTraces()
+	rss := traces.ResourceSpans()
+	for ip, s := range batches {
+		// SFx trace is effectively zipkin, so more convenient to convert to it and then rely
+		// on existing zipkin receiver translator
+		translated, err := zipkinv2Translator.ToTraces(sfxToZipkinSpans(s, logger))
+		if err != nil {
+			return traces, err
```
pjanotti, line 57: Since this returns `traces` with partial results in case of error should it combine all errors and return all spans that it could process? Alternatively it could do the traditional: `return nil, err`.

emaderer:
There are no unit tests that make sure that the source/reporter client IP from SFX span metadata is copied
to Otel Resource as an attribute?

dmitryax (approved):
Thanks, Owais! I tested it out, and it works well.

Please fix the failing tests

owais:
Thanks for testing @dmitryax. Fixed and added tests.

owais:
> There are no unit tests that make sure that the source/reporter client IP from SFX span metadata is copied
to Otel Resource as an attribute?

Have added/updated tests. Was waiting for Dmitri to config this fixes the issue.

internal/receiver/smartagentreceiver/converter/traces_test.go
```diff
@@ -276,9 +348,8 @@ func TestInvalidSFxToPDataConversion(t *testing.T) {
 	logger := zap.New(obs)
 
 	traces, err := sfxSpansToPDataTraces([]*trace.Span{&invalidAsPData}, logger)
-	// error is returned but core conversion persists
 	assert.EqualError(t, err, "invalid length for ID")
-	assert.Equal(t, 1, traces.ResourceSpans().Len())
```
owais, line 281: This behaviour has changed. Given a slice with ALL spans invalid, the function will return pdata.Traces with an empty resource span slice. Earlier it would return a resource span slice with exactly one resource but no spans. I think the net effect is the same so this should be harmless but wanted to point out in case I'm missing something.

internal/receiver/smartagentreceiver/converter/traces.go
```diff
@@ -24,12 +25,51 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	sfxDataSourceKey         = "data_source_ip"
```
dmitryax, line 29: nit: I believe we can use github.com/signalfx/signalfx-agent/pkg/core/common/constants.DataSourceIPKey instead

internal/receiver/smartagentreceiver/converter/traces.go
```diff
@@ -24,12 +25,51 @@ import (
...
+	batches := map[string][]*trace.Span{}
+	for _, span := range spans {
+		if span == nil {
+			continue
+		}
+		var sourceIP string
+		if val, ok := span.Meta[sfxDataSourceKey]; ok {
+			if ip, ok := val.(net.IP); ok {
+				sourceIP = ip.String()
+			}
```
dmitryax, line 48: Should we add couple of debug logs here when sfxDataSourceKey is missing and it doesn't contain an IP?

dmitryax (commented):
LGTM, just couple of nits

internal/receiver/smartagentreceiver/converter/traces.go
owais: For some reason I thought it was internal and not exposed.

internal/receiver/smartagentreceiver/converter/traces.go
owais: Not sure if worth it. logger instance is not available here so probably need to change a few places. I think the it would be obvious that IP was not detected so not sure how much it'd help but can add if you feel strongly about it.

rmfitzpatrick: I believe the `logger` here is the receiver's logger passed through the output's converter.

internal/receiver/smartagentreceiver/converter/traces_test.go
rmfitzpatrick: This seems like more desirable behavior rather than putting something in the pipeline that would never result with an item of value.

internal/receiver/smartagentreceiver/converter/traces.go
```diff
@@ -16,20 +16,58 @@ package converter
...
+	// batch gets translated separately to ensure that spans from sources with different
+	// IPs don't get bundled together under the same resource.
+	batches := map[string][]*trace.Span{}
+	for _, span := range spans {
+		if span == nil {
+			continue
+		}
+		var sourceIP string
+		if val, ok := span.Meta[sfxConstants.DataSourceIPKey]; ok {
+			if ip, ok := val.(net.IP); ok {
```
rmfitzpatrick, line 44: Curious if this process is actually necessary since the tag has been set and we should respect it regardless of its validity.

rmfitzpatrick (approved):
thanks!

internal/receiver/smartagentreceiver/converter/traces.go
owais: Not so much validating data but casting an interface{} to net.IP so we can call ip.String() below.

internal/receiver/smartagentreceiver/converter/traces.go
dmitryax: Empty string doesn't break this functionality, but k8s_tagger won't be able to do its job. Chatty logs on debug level should be fine I guess. Having debug logs makes it easier to find out why correlation doesn't work next time. But i'm ok with keeping it as is if adding logs is not straight-forward.

dmitryax (approved):

dmitryax:
Also please add a line to Changelog

PR #593, opened 2021-08-03
smart agent receiver: set redirected logrus logger level

rmfitzpatrick:
The target logrus standard logger used by the agent monitors will only have its level set to the config-specified value in the agent main, which we do not use.  These changes set the logger from the equivalent receiver zap logger's core.

### DIFF
--- a/internal/receiver/smartagentreceiver/log_test.go
+++ b/internal/receiver/smartagentreceiver/log_test.go
@@ -47,7 +47,7 @@ func logAt(entry *logrus.Entry, level logrus.Level, msg string) {
 }
 
 func TestRedirectMonitorLogs(t *testing.T) {
-	for logrusLevel, zapLevel := range levelsMap {
+	for logrusLevel, zapLevel := range logrusToZapLevel {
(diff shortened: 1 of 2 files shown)


internal/receiver/smartagentreceiver/log_test.go
```diff
@@ -83,6 +83,14 @@ func TestRedirectMonitorLogs(t *testing.T) {
 		logToZap.redirect(monitor2LogrusKey, monitor2ZapLogger)
 		logToZap.redirect(monitor3LogrusKey, monitor3ZapLogger)
 
+		expectedLogrusLevel := logrusLevel
+		if logrusLevel == logrus.TraceLevel {
+			expectedLogrusLevel = logrus.DebugLevel
+		}
+		require.Equal(t, monitor2LogrusKey.Level, expectedLogrusLevel)
```
pmcollins, line 90: Looks like `monitor2LogrusKey.Level, expectedLogrusLevel` args should be swapped, no?

pmcollins (approved):
LGTM, just one comment

PR #604, opened 2021-08-06
translatesfx: only put processlist in logs pipeline if it's specified in SA

pmcollins:
Also, don't build a metrics pipeline if there are no metrics receivers.

### DIFF
--- a/cmd/translatesfx/translatesfx/otel.go
+++ b/cmd/translatesfx/translatesfx/otel.go
@@ -53,17 +53,7 @@ func newOtelCfg() *otelCfg {
 		},
 		Extensions: map[string]map[string]interface{}{},
 		Service: service{
-			Pipelines: map[string]*rpe{
-				"metrics": {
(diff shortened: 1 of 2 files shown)


cmd/translatesfx/translatesfx/cli_test.go
```diff
@@ -33,4 +33,6 @@ func TestTranslateConfig(t *testing.T) {
 	err = yaml.Unmarshal(expected, &expectedV)
 	require.NoError(t, err)
 	assert.Equal(t, expectedV, translatedV)
+
+	println(toYaml(t, translatedV))
```
rmfitzpatrick, line 37: debugging remnant?

rmfitzpatrick (approved):

cmd/translatesfx/translatesfx/cli_test.go
pmcollins: oops! will fix

PR #605, opened 2021-08-06
Add text linking the new release to its Docker image URL

atoulme:
This change adds a paragraph at the top of the draft of release notes for new versions of the Splunk OpenTelemetry Collector that links to the Quay page showing releases of the Splunk OpenTelemetry Collector, and text of the Docker image URL one can use to take advantage of the release.

It has not been tested.

### DIFF
--- a/.github/release-drafter.yaml
+++ b/.github/release-drafter.yaml
@@ -11,6 +11,11 @@ categories:
     label: documentation
 tag-template: - $TITLE @$AUTHOR (#$NUMBER)
 template: |
+  ## [Docker image](https://quay.io/repository/signalfx/splunk-otel-collector)
+    This release is available as a Docker image:
+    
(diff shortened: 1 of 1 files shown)


.github/release-drafter.yaml
```diff
@@ -11,6 +11,11 @@ categories:
     label: documentation
 tag-template: - $TITLE @$AUTHOR (#$NUMBER)
 template: |
+  ## [Docker image](https://quay.io/repository/signalfx/splunk-otel-collector)
```
rmfitzpatrick, line 14: Would you be opposed to putting this after the changes so it's just above the other artifacts?  Definitely helpful to have this direct link in general but if breaking changes are introduced it be helpful to have the link after the disclaimers.

atoulme: No, that would work too. I’ll make the change.

rmfitzpatrick (approved):

PR #609, opened 2021-08-09
translatesfx: handle SA regex filters

pmcollins:
SA allows you to specify a regexp filter by wrapping it in forward slashes (without forward slashes the expression is considered a glob). This change detects that and translates (unwraps) it into an expr expression.

For example the SA directive:

```
metricsToExclude:
  - metricName: /vsphere\.cpu_\w*_percent/
```

should translate into:

```
processors:
  filter:
    metrics:
      exclude:
        match_type: expr
        expressions:
        - MetricName matches "vsphere\\.cpu_\\w*_percent"
```


emaderer (approved):

cmd/translatesfx/translatesfx/otel.go
```diff
@@ -160,16 +160,27 @@ func wrapStatement(stmt string, empty, negated bool) string {
...
 
-func globToRegexpStr(s string) (string, bool) {
+func saToRegexpStr(s string) (string, bool) {
 	negated := false
 	if strings.HasPrefix(s, "!") {
 		s = s[1:]
 		negated = true
 	}
+	if isRegexFilter(s) {
+		return s[1 : len(s)-1], negated
```
rmfitzpatrick, line 170: Are we sure this can't be the empty string? `"/"` would be considered a regex filter as well?

pmcollins: Like this?

```
metricsToExclude:
  - metricName: /
```

rmfitzpatrick: Right, it's an edge case but wouldn't `isRegexFilter("/")` return true when it's not one?  The empty string would lead to an invalid index w/ the check as well.

pmcollins: I see. Fixed, thanks.

PR #620, opened 2021-08-12
translatesfx: translate metricsToInclude filters to expr

pmcollins:
SA's `metricsToInclude` overrides any matching `metricsToExclude`. This change
translates SmartAgent metricsToInclude filters and appends them to filter
processor expr expressions.

For example, the following SA filters:

```
metricsToExclude:
  - metricNames:
      - node_filesystem_*

metricsToInclude:
  - metricNames:
      - node_filesystem_readonly
```

should translate to:

```
filter:
  metrics:
    exclude:
      match_type: expr
      expressions:
      - MetricName matches "^node_filesystem_.*$" and
        (not (MetricName matches "^node_filesystem_readonly$"))
```

### DIFF
--- a/cmd/translatesfx/translatesfx/otel.go
+++ b/cmd/translatesfx/translatesfx/otel.go
@@ -60,64 +60,81 @@ func newOtelCfg() *otelCfg {
 }
 
 func translateFilters(sa saCfgInfo, otel *otelCfg) {
-	if sa.metricsToExclude == nil {
+	if sa.metricsToExclude == nil && sa.metricsToInclude == nil {
 		return
 	}
+
 	otel.Processors[filterProc] = map[string]interface{}{
 		"metrics": map[string]interface{}{
 			"exclude": map[string]interface{}{
 				"match_type":  "expr",
-				"expressions": saFiltersToExpr(sa.metricsToExclude),
+				"expressions": saExcludesToExpr(sa.metricsToExclude, sa.metricsToInclude),
 			},
 		},
 	}
 	otel.Service.Pipelines["metrics"].appendProcessor(filterProc)
 }
 
-func saFiltersToExpr(excludes []interface{}) []string {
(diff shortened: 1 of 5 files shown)


rmfitzpatrick:
Not necessarily affecting these changes, but do you have an idea for what the general expected functionality should be for `monitorType`? https://github.com/signalfx/signalfx-agent/blob/main/docs/config-schema.md#metricstoinclude

rmfitzpatrick:
> What do you think about emitting a warning that translating monitorType is unsupported?

I think in the short term that sounds good, but overall it may make sense to merge with referenced monitor's `datapointsToExclude ` or even create different processors and pipelines.  That may be overkill though.

cmd/translatesfx/translatesfx/otel.go
```diff
@@ -60,64 +60,81 @@ func newOtelCfg() *otelCfg {
...
+			names = []interface{}{nameV}
 		}
-		dimExpr := dimsToExpr(dims)
-		if dimExpr != "" {
-			if line != "" {
-				line += " and "
-			}
-			line += "(" + dimExpr + ")"
+	} else {
+		names = namesV.([]interface{})
```
rmfitzpatrick, line 111: I think I've commented on something like this before, but would it be worthwhile to validate the SA config on the way in w/ `yaml.UnmarshalStrict()` and https://github.com/signalfx/signalfx-agent/blob/main/pkg/core/config/validation/validator.go#L33
 so that untested type assertions are less prone to hit conversion issues?

rmfitzpatrick (approved):

cmd/translatesfx/translatesfx/otel.go
pmcollins: Yeah, I've been thinking that may make things cleaner. I'll try it.

PR #621, opened 2021-08-12
Add Vulnerability Scanning

mwang2016:


.github/workflows/cve-scan.yaml
```diff
@@ -0,0 +1,22 @@
+name: Vulnerability Scanning
```
dmitryax, line 1: Should we call the action `Docker CVE Scanning` since it's scanning the docker image only?

rmfitzpatrick:
@jcheng-splunk would the releaser fork be a better place for this?  We could use the artifact from `build-image`.

jcheng-splunk:
> @jcheng-splunk would the releaser fork be a better place for this? We could use the artifact from `build-image`.

@rmfitzpatrick Only issue with adding this to the gitlab releaser is that github PR checks would not be notified of any issues.  The alternative would be to add this to circleci, which can use the image artifact from the `docker-otelcol` job.

dmitryax:
This is a cron job which supposed to run on daily basis. And since it requires a SNYK token, we will need to move it to gitlab runner. 

Also I'm suggesting that this job creates a github issue (avoiding duplicates) upon finding any CVEs. Not necessary for this PR, but would be nice as a follow-up.

@rmfitzpatrick @jchengsfx thoughts?

rmfitzpatrick:
@dmitryax I'm thinking this does belong in gitlab and am fine with basic functionality for now (~standard failure email).  Ideally we'd be able to use a webhook and check to prevent PRs from landing that introduce vulnerabilities but imagine that would take some investigation if we don't have a precedent.

dmitryax:
@rmfitzpatrick sounds good. Let's start with a regular job notifying about found vulnerabilities through email or slack. Later we can add an integration with github issues to report CVEs and PRs to trigger the scan.

jcheng-splunk:
@dmitryax @rmfitzpatrick Didn't realize we needed a snyk token (the initial commit did not reference the token), so I agree with adding this to gitlab.  LMK if I can help add the token to the gitlab pipeline or with anything else on the gitlab side.

.gitlab-ci.yml
```diff
@@ -316,3 +316,31 @@ puppet-release:
   artifacts:
     paths:
       - deployments/puppet/pkg/*.tar.gz
+
+cve-scan:
+  image: ubuntu:20.04
+  stage: test
```
dmitryax, line 322: I'm not an expert in Gitlab config but shouldn't the `test` stage be added to `stages:` list? Same for the next job...

.gitlab-ci.yml
```diff
@@ -316,3 +316,31 @@ puppet-release:
...
+      - cron 0 8 * * *
+  script:
+    - make docker-otelcol
+    - mkdir -p ~/.docker/cli-plugins && \
+      curl https://github.com/docker/scan-cli-plugin/releases/latest/download/docker-scan_linux_amd64 -L -s -S -o ~/.docker/cli-plugins/docker-scan && \
+      chmod +x ~/.docker/cli-plugins/docker-scan
+    - yes | docker scan --login --token ${SNYK_AUTH_TOKEN}
+    - docker scan otelcol
+
+slack_message:
```
dmitryax, line 335: We probably should call it something specific to CVEs failure like `notify-cve-scan-failure`.

.gitlab-ci.yml
```diff
@@ -316,3 +316,31 @@ puppet-release:
...
+  stage: slack
```
dmitryax, line 336: Maybe `notify` would be better name for the stage?

.gitlab-ci.yml
```diff
@@ -316,3 +316,31 @@ puppet-release:
...
+  when:
+    on-failure: cve-scan
+  script:
+    - echo "Send Slack Messages"
+    - |
+      curl -X POST ${SLACK_WEBHOOK_URL} -H 'Content-Type: application/json' \
+        --data '{\"blocks\": [{"type": "section","text": {"type": "mrkdwn","text": "*@here Gitlab Job #${CI_JOB_ID}*"}},{"type": "section","text": {"type": "mrkdwn","text": "*:ghost: Vulnerability scan failed on splunk-otel-collector*"},\"accessory\": {\"type\": \"button\",\"text\": {\"type\": \"plain_text\",\"text\": \"More Info\",\"emoji\": true},\"style\": \"danger\",\"url\": \"${CI_JOB_URL}",\"action_id\": \"button-action\"}}]}'
+
+
+
```
dmitryax, line 346: nit: too many empty lines :)

.gitlab-ci.yml
mwang2016: sure, I will change that

.gitlab-ci.yml
mwang2016: Yes! I missed that. thank you. Should we call it `test` or `scan` sounds better?

dmitryax: `test` is a good option. It should probably go in between `build` and `sign`

mwang2016: ok, thank you

.gitlab-ci.yml
```diff
@@ -316,3 +317,28 @@ puppet-release:
...
+  script:
+    - make docker-otelcol
+    - mkdir -p ~/.docker/cli-plugins && \
+      curl https://github.com/docker/scan-cli-plugin/releases/latest/download/docker-scan_linux_amd64 -L -s -S -o ~/.docker/cli-plugins/docker-scan && \
+      chmod +x ~/.docker/cli-plugins/docker-scan
+    - yes | docker scan --login --token ${SNYK_AUTH_TOKEN}
+    - docker scan otelcol
+
+notify-cve-scan-failure:
+  stage: notify
```
dmitryax, line 337: I believe this stage should be added to the list as well, probably at the end

dmitryax (approved):

dmitryax (changes requested):
Accidentally approved. Please address the latest comment

dmitryax (approved):

dmitryax:
@jchengsfx PTAL

dmitryax:
@jcheng-splunk PTAL

jcheng-splunk:
@mwang2016 Please rebase as I recently added some new jobs to `.gilab-ci.yml`.

mwang2016:
> @mwang2016 Please rebase as I recently added some new jobs to `.gilab-ci.yml`.

Will do, thanks.

dmitryax (approved):

PR #645, opened 2021-08-19
change the default config file from gateway to agent

emaderer:
Update the default config file from gateway to agent.

### DIFF
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -47,9 +47,9 @@ const (
 	realmEnvVarName           = "SPLUNK_REALM"
 	tokenEnvVarName           = "SPLUNK_ACCESS_TOKEN"
 
-	defaultDockerSAPMConfig        = "/etc/otel/collector/gateway_config.yaml"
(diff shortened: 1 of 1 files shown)


rmfitzpatrick:
I wonder if we should actually provide additional guidance on what's required for this to be valid for most usages.  The manual installation instructions are catered to the gateway config and we'd need to advise on how to mount ro hostfs volumes and using host networking mode for the host metrics receiver.  If we cut over right now I think there'd be fallout.

emaderer:
> I wonder if we should actually provide additional guidance on what's required for this to be valid for most usages. The [manual installation instructions](https://github.com/signalfx/splunk-otel-collector/blob/main/docs/getting-started/linux-manual.md#docker)* are catered to the gateway config and we'd need to advise on how to mount ro hostfs volumes and using host networking mode for the host metrics receiver. If we cut over right now I think there'd be fallout.

That's a good point, I think that we should leave it as is then, for most installation types the config file is set to the agent in any case.

flands:
What's the reason for this change?

PR #690, opened 2021-08-31
Add meta/runtime.yml for ansible

jcheng-splunk:
New galaxy requirement for `meta/runtime.yml` causes publish to fail.


dmitryax (approved):
LGTM. Thanks!

PR #699, opened 2021-09-01
add collector service level proxy for ansible

xp-1000:
Hello,

It is possible to easily use a proxy during ansible deployment by setting right environment variable at playbook level like:

```yaml
---
- name: "Example of proxy usage with splunk otel"
  hosts: all
  become: true
  tasks:
    - name: "Include splunk_otel_collector"
      import_role:
        name: "signalfx.splunk_otel_collector.collector"
      vars:
        splunk_access_token: "{{ lookup('community.hashi_vault.hashi_vault', 'secret/data/signalfx:token') }}"
        splunk_realm: eu0
        install_fluentd: false
      environment:
        https_proxy: "{{ https_proxy }}"
        http_proxy: "{{ https_proxy }}"
        no_proxy: "localhost,127.0.0.1,::1"
```

but it is not possible to configure a proxy to be used by the collector itself.

of course we can add a task after the role import but it seems useful to support this upstream.

### DIFF
--- a/deployments/ansible/roles/collector/defaults/main.yml
+++ b/deployments/ansible/roles/collector/defaults/main.yml
@@ -36,4 +36,4 @@ agent_bundle_dir: /usr/lib/splunk-otel-collector/agent-bundle
 splunk_collectd_dir: /usr/lib/splunk-otel-collector/agent-bundle/run/collectd
 
 # Configure otel collector service to use a proxy
-splunk_service_proxy: ""
+splunk_otel_collector_proxy: ""
--- a/deployments/ansible/roles/collector/defaults/main.yml
+++ b/deployments/ansible/roles/collector/defaults/main.yml
@@ -36,4 +36,5 @@ agent_bundle_dir: /usr/lib/splunk-otel-collector/agent-bundle
 splunk_collectd_dir: /usr/lib/splunk-otel-collector/agent-bundle/run/collectd
 
 # Configure otel collector service to use a proxy
-splunk_otel_collector_proxy: ""
+splunk_otel_collector_proxy_http: ""
+splunk_otel_collector_proxy_https: ""
--- a/deployments/ansible/roles/collector/defaults/main.yml
+++ b/deployments/ansible/roles/collector/defaults/main.yml
@@ -34,3 +34,6 @@ splunk_fluentd_config_source: ""
 
 agent_bundle_dir: /usr/lib/splunk-otel-collector/agent-bundle
 splunk_collectd_dir: /usr/lib/splunk-otel-collector/agent-bundle/run/collectd
+
+# Configure otel collector service to use a proxy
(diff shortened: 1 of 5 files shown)


deployments/ansible/roles/collector/defaults/main.yml
```diff
@@ -34,3 +34,6 @@ splunk_fluentd_config_source: ""
 
 agent_bundle_dir: /usr/lib/splunk-otel-collector/agent-bundle
 splunk_collectd_dir: /usr/lib/splunk-otel-collector/agent-bundle/run/collectd
+
+# Configure otel collector service to use a proxy
+splunk_service_proxy: ""
```
dmitryax, line 39: What do you think about `splunk_otel_collector_proxy` name?

deployments/ansible/roles/collector/templates/collector-service-proxy.conf.j2
```diff
@@ -0,0 +1,4 @@
+[Service] 
+Environment="HTTP_PROXY={{ splunk_service_proxy }}" 
+Environment="HTTPS_PROXY={{ splunk_service_proxy }}"
```
dmitryax, line 3: Do you think that there won't be a use case when these vars need to be set to different values?

xp-1000: I asked myself the same question and I thought it would be very unlikely but I can use 2 variables if you think it could be useful

deployments/ansible/roles/collector/defaults/main.yml
xp-1000: yes it is more accurate

deployments/ansible/roles/collector/templates/collector-service-proxy.conf.j2
xp-1000: we can possibly set the value of one variable using the other one but I am afraid it could lead to undesired configuration by the user.

```suggestion
Environment="HTTP_PROXY={{ splunk_service_proxy_http | default({{ splunk_service_proxy_https }}) }}" 
Environment="HTTPS_PROXY={{ splunk_service_proxy_https | default({{ splunk_service_proxy_http }}) }}" 
```

If you want this so we should change the default values of these variables to `None` and the underlying `when` to check None instead of empty string to allow user to define a proxy for only one protocol

dmitryax: I think explicitly asking for both is a better approach. I like the current state of the PR

dmitryax (approved):
LGTM. Could you please add another line to the Ansible Changelog?

xp-1000:
> LGTM. Could you please add another line to the Ansible Changelog?

my bad, done!

dmitryax (approved):
Thanks!

PR #702, opened 2021-09-01
auto add system.type dimension to all sa receiver datapoints

pmcollins:
Add `system.type` dimension to all datapoints, where `system.type` == "smartagent-&lt;monitor-name&gt;" in support of pivot sidebar. Not sure about "smartagent-&lt;monitor-name&gt;" -- open to other possibilities.

### DIFF
--- a/internal/receiver/smartagentreceiver/receiver.go
+++ b/internal/receiver/smartagentreceiver/receiver.go
@@ -176,8 +176,7 @@ func (r *Receiver) createMonitor(monitorType string, host component.Host) (monit
 		output.AddExtraDimension(k, v)
 	}
 
-	systemTypeVal := fmt.Sprintf("smartagent-%s", monitorType)
-	output.AddExtraDimension(systemTypeKey, systemTypeVal)
+	output.AddExtraDimension(systemTypeKey, monitorType)
 
 	// Configure SmartAgentConfigProvider to gather any global config overrides and
 	// set required envs.
--- a/internal/receiver/smartagentreceiver/receiver.go
+++ b/internal/receiver/smartagentreceiver/receiver.go
@@ -40,6 +40,7 @@ import (
 )
 
 const setOutputErrMsg = "unable to set Output field of monitor"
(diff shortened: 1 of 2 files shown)


internal/receiver/smartagentreceiver/receiver.go
```diff
@@ -175,6 +176,9 @@ func (r *Receiver) createMonitor(monitorType string, host component.Host) (monit
 		output.AddExtraDimension(k, v)
 	}
 
+	systemTypeVal := fmt.Sprintf("smartagent-%s", monitorType)
```
emaderer, line 179: I don't think that we should prefix the type with _**smartagent-**_ the type should be _**kafka**_ for example, when we move to the OTel receivers the type should be the same there.

emaderer: Many of the monitor types start with **collectd** or **telegraf** like **_collectd/elasticsearch_** , we need to make sure that collectd/telegraf is not part of the system.type (took the type from https://docs.signalfx.com/en/latest/integrations/agent/monitors/index.html)

pmcollins: Ah, ok. Will fix.

emaderer: I am not sure if that is indeed the case with the monitorType variable you are using

pmcollins: I ran a test and the `monitorType` variable does have the `collectd/` prefix. Should we strip all `foo/` prefixes or just `collectd/` and `telegraf/`? Stripping all would strip `prometheus/` as well. There would be a little bit of overlap because we have a `prometheus/postgres` monitor as well as a `postgresql` one, but we'd still be able to tell the difference because one is `postgres` and the other is `postgresql`.

emaderer: I think that stripping all makes sense, collectd/telegraf/prometheus are not part of the type.

emaderer (approved):

PR #715, opened 2021-09-02
adding journalctl to image

rockb1017:
need `journalctl` executable to run journald receiver
related: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2332

### DIFF
--- a/cmd/otelcol/Dockerfile
+++ b/cmd/otelcol/Dockerfile
@@ -22,6 +22,10 @@ RUN mkdir -p /usr/lib/splunk-otel-collector && \
           /usr/lib/splunk-otel-collector/agent-bundle/bin/agent-status \
           $SMART_AGENT
 
+FROM debian:10.8 as journal
+RUN apt update
+RUN apt install -y systemd
+
 FROM scratch
 ENV SPLUNK_BUNDLE_DIR=/usr/lib/splunk-otel-collector/agent-bundle
 ENV SPLUNK_COLLECTD_DIR=${SPLUNK_BUNDLE_DIR}/run/collectd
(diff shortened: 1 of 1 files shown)


emaderer:
@rockb1017  there is no journald receiver yet in contrib so what's the reason for adding this to the docker image right now?

rockb1017:
Hello,
we are working on adding journald receiver. we are at testing stage. we discovered journalctl needed to be added to image for it to work. 
So wanted to resolve this dependency for when journald receiver becomes available and wanted to bring it up to community early on to see if there is better approach or preferred way of doing it. Thanks!

emaderer:
> Hello,
> we are working on adding journald receiver. we are at testing stage. we discovered journalctl needed to be added to image for it to work.
> So wanted to resolve this dependency for when journald receiver becomes available and wanted to bring it up to community early on to see if there is better approach or preferred way of doing it. Thanks!

I don't think that we want to add these dependencies into our Docker image before there is an actual receiver that needs it.

rockb1017:
yea, it's fine. let's wait. we will add journald receiver once it is available on contrib

emaderer:
@rockb1017  do you mind closing this PR for now?

dmitryax:
> That is the very reason we are shipping journalctl in the image, then the OS version it is running on is irrelevant. As long as there are journald log files on file system, it will just read using journalctl from image.
> So that it wouldn't fail and break the collector, so it can be enabled by default.

Sounds good. Can we at least test it manually on an OS without systemsd?

luckyj5:
I have tested this on an OS with and without systemd, and it runs as expected. (No failures).

luckyj5:
Moving this PR to another PR [912](https://github.com/signalfx/splunk-otel-collector/pull/912) which is created to add and register journald.

PR #720, opened 2021-09-03
Add suse support to installer script (collector only)

jcheng-splunk:


jcheng-splunk:
cc @tmorrisonsfx

docs/getting-started/linux-manual.md
```diff
@@ -35,27 +35,57 @@ installed on x86_64/amd64 platforms.
...
-enabled=1
-EOH
-
-yum install -y splunk-otel-collector
-```
+  ```sh
+  $ curl -sSL https://splunk.jfrog.io/splunk/otel-collector-deb/splunk-B3CD4420.gpg > /etc/apt/trusted.gpg.d/splunk.gpg
+  $ echo 'deb https://splunk.jfrog.io/splunk/otel-collector-deb release main' > /etc/apt/sources.list.d/splunk-otel-collector.list
+  $ apt-get update
+  $ apt-get install -y splunk-otel-collector
```
rmfitzpatrick, line 42: I'm not familiar with doc standards and am pretty inconsistent with this myself, but it can be nice to copy multiple commands for pasting at once and using `$` prevents this.

internal/buildscripts/packaging/tests/installer_test.py
```diff
@@ -73,53 +116,31 @@ def test_installer_mode(distro, version, mode):
...
+            if "opensuse" not in distro:
+                # the td-agent service should only be running when installing
+                # collector packages that have our custom fluent config
+                if container.exec_run("test -f /etc/otel/collector/fluentd/fluent.conf").exit_code == 0:
+                    assert container.exec_run("systemctl status td-agent").exit_code == 0
+                else:
+                    assert container.exec_run("systemctl status td-agent").exit_code != 0
 
             # test support bundle script
             if container.exec_run("test -f /etc/otel/collector/splunk-support-bundle.sh").exit_code == 0:
```
rmfitzpatrick, line 134: not introduced here but should we fail if the file is missing or are some cases expected to not have it?  If the latter it seems like we should restrict to where it's expected and fail otherwise.

internal/buildscripts/packaging/tests/package_test.py
```diff
@@ -43,6 +43,15 @@ def get_package(distro, name, path):
         return None
 
 
+def get_rpm_command(container):
```
rmfitzpatrick, line 46: Could this be something like `get_libcap_command` and return the whole thing instead of using an additional zypper check?

internal/buildscripts/packaging/tests/installer_test.py
jcheng-splunk: I think I added that initially since we had older packages that didn't include the fluentd configs.  But should be fine to remove that check now.

docs/getting-started/linux-manual.md
jcheng-splunk: I can remove the `$`.  Should we also chain these commands?

rmfitzpatrick: I think removing the `$` is good enough but defer to @tmorrisonsfx

internal/buildscripts/packaging/tests/installer_test.py
rmfitzpatrick: was referring to `splunk-support-bundle.sh` but isn't necessary to update here

rmfitzpatrick (approved):

internal/buildscripts/packaging/tests/installer_test.py
jcheng-splunk: I ended up removing both checks. :)

docs/getting-started/linux-manual.md
tmorrisonsfx: Yes, you can remove the `$`. The copy function on the site will include the `$` by default.

PR #727, opened 2021-09-08
translatesfx: kubernetes-events as logs receiver

rmfitzpatrick:
These changes add the `kubernetes-events` monitor to the logs pipeline cases that are* only currently applicable to the processlist monitor.

### DIFF
--- a/cmd/translatesfx/translatesfx/otel.go
+++ b/cmd/translatesfx/translatesfx/otel.go
@@ -23,6 +23,7 @@ import (
 
 const (
 	processlist       = "smartagent/processlist"
+	kubernetesEvents  = "smartagent/kubernetes-events"
 	sfxFwder          = "smartagent/signalfx-forwarder"
 	resourceDetection = "resourcedetection"
 	sfx               = "signalfx"
@@ -31,6 +32,12 @@ const (
 	k8sObserver       = "k8s_observer"
 	hostObserver      = "host_observer"
 )
+
+// monitors that should be converted to both metrics and traces
+var metricsAndTracesReceiverMonitorTypes = map[string]bool{sfxFwder: true}
+
+// monitors that should be converted to logs receivers only
+var exclusivelyLogsReceiverMonitorTypes = map[string]bool{processlist: true, kubernetesEvents: true}
 
 func saInfoToOtelConfig(sa saCfgInfo, vaultPaths []string) (otel *otelCfg, warnings []error) {
 	otel = newOtelCfg()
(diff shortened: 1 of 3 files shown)


cmd/translatesfx/translatesfx/otel.go
```diff
@@ -284,11 +285,17 @@ func translateMonitors(sa saCfgInfo, cfg *otelCfg) (warnings []error) {
...
 
-	if _, ok := cfg.Receivers[processlist]; ok {
-		cfg.Service.Pipelines["logs"] = &rpe{
-			Receivers:  []string{processlist},
-			Processors: []string{resourceDetection},
-			Exporters:  []string{sfx},
+	for _, logReceiver := range []string{processlist, kubernetesEvents} {
+		if _, ok := cfg.Receivers[logReceiver]; ok {
+			if existingRPE, ok := cfg.Service.Pipelines["logs"]; ok {
+				existingRPE.Receivers = append(existingRPE.Receivers, logReceiver)
```
pmcollins, line 291: You can make this `existingRPE.appendReceiver` so the receivers are sorted

cmd/translatesfx/translatesfx/otel.go
```diff
@@ -410,7 +417,7 @@ func dimsToMetricsTransformProcessor(m map[interface{}]interface{}) map[string]i
 func receiverList(receivers map[string]map[string]interface{}) []string {
 	var keys []string
 	for k := range receivers {
-		if k == processlist {
```
pmcollins, line 413: I wonder if it would make things more understandable to readers if `k == processlist || k == kubernetesEvents` were factored into an `isLogsReceiver` function.

pmcollins (approved):
Thanks for doing this. LGTM, just a couple of comments.

cmd/translatesfx/translatesfx/otel.go
rmfitzpatrick: I tend toward the "third time's a charm" approach for breaking out helpers but makes sense either way.

When I was updating this it became apparent the receiver creator logic isn't monitor type-dependent, which could be problematic.  It's not immediately clear to me when discovery rules would be used for sfx forward, processlist, or k8s events monitors so I didn't attempt to resolve it in these changes but did make the `receiverList` function more specific.  This ok?

cmd/translatesfx/translatesfx/otel.go
rmfitzpatrick: adopted this, but changed it a bit in the subsequent update to where isn't no longer needed.

cmd/translatesfx/translatesfx/otel.go
pmcollins: Ok no worries -- I wasn't really thinking of this as a helper but rather as a way to document by breaking this into its own function with a descriptive name.

I'm not quite sure I understand the receiver creator comment -- but basically discovery rules for sfx forwarder, processlist, or k8s events are not supported (at least not yet). And `receiverList` should be renamed to `metricsReceiverList` (this code could use a cleanup). We can discuss separately if you want, but your change looks fine to me.

PR #734, opened 2021-09-09
translatesfx: rename metricsToExclude to datapointsToExclude on SA monitors

pmcollins:
If metricsToExclude is specified for a monitor (not the agent -- this is
apparently an invalid configuration according to the docs, but it actually
works and some customers are using it) rename it to the supported
datapointsToExclude.

### DIFF
--- a/cmd/translatesfx/translatesfx/testdata/sa-metrics-to-exclude-monitor.yaml
+++ b/cmd/translatesfx/translatesfx/testdata/sa-metrics-to-exclude-monitor.yaml
@@ -1,0 +1,14 @@
+---
+signalFxAccessToken: "abc123"
+signalFxRealm: us1
+
+intervalSeconds: 10
+
+logging:
+  level: info
+
+monitors:
+  - type: cpu
+    metricsToExclude:
+      - metricNames:
+          - foo*
--- a/cmd/translatesfx/translatesfx/otel.go
+++ b/cmd/translatesfx/translatesfx/otel.go
@@ -30,6 +30,8 @@ const (
 	filterProc        = "filter"
 	k8sObserver       = "k8s_observer"
 	hostObserver      = "host_observer"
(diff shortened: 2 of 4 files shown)


rmfitzpatrick:
I'm not seeing how this is possible in the agent after https://github.com/signalfx/signalfx-agent/pull/1163 so guess we're targeting much older agent versions?  The approach in these changes is described in that PR though so this is does seem helpful.

cmd/translatesfx/translatesfx/testdata/sa-metrics-to-exclude-monitor.yaml
```diff
@@ -0,0 +1,14 @@
...
+signalFxRealm: us1
+
+intervalSeconds: 10
+
+logging:
+  level: info
+
+monitors:
+  - type: cpu
+    metricsToExclude:
```
rmfitzpatrick, line 12: If we do start to use the agent config type directly in translation this will be a problem since it's not valid afaict.

rmfitzpatrick (approved):

cmd/translatesfx/translatesfx/testdata/sa-metrics-to-exclude-monitor.yaml
pmcollins: Noted, will look into it (it has worked in the tests I have run).

pmcollins:
> I'm not seeing how this is possible in the agent after [signalfx/signalfx-agent#1163](https://github.com/signalfx/signalfx-agent/pull/1163) so guess we're targeting much older agent versions? The approach in these changes is described in that PR though so this is does seem helpful.

I suppose the idea is that if a customer is running an older version of SA and they have a valid config, it's fair game for the translation tool, to the extent that it's possible to support the older SA configs without breaking translation for newer ones.

PR #755, opened 2021-09-19
Move Heroku buildpack to dedicated repo

flands:


dmitryax:
Why do we want to move heroku to a dedicated repo? Last time we discussed the same for Ansible and agreed to stay here...

flands:
Because it does not work :) The way you push buildpacks is that the required directories need to be at the root level of the repo. I am not aware of a workaround -- appears to be how everyone else does it too.

rmfitzpatrick:
Another option we tried in the past when galaxy had the same requirements was to keep a ~mirror of the deployment at a new project and manually refresh its contents (https://github.com/signalfx/ansible-signalfx-agent).  I think the two projects diverged almost immediately due to lack of clear coordination, so it probably would have been better to make the deployment directory its own repo and clone it from the relative path at the standalone project.

Unless there are clear reasons to keep the deployment in this project like shared, relative deps, updater scripts, or test infrastructure simply migrating is the easiest way.

flands:
No dependencies on this repo :)

rmfitzpatrick (approved):

dmitryax:
Sounds good! Let's move it away

dmitryax (approved):

PR #759, opened 2021-09-20
Handle backwards compatibility of internal ballast removal

pmcollins:
This change adds a `ballastParserProvider` which wraps a
`ConfigSourceParserProvider` and adds a `memory_ballast` extension to the
Otel config if one isn't already present. Doing so ensures that users
have memory ballast regardless of whether they explicitly added the
newly-developed `memory_ballast` extension, which may not have existed
when they set up their collector.


emaderer:
I think that the main issue to address is what happens if the deprecated ballast field will get removed from the  [memory limiter processor](https://github.com/open-telemetry/opentelemetry-collector/blob/16c681ea0d17125d9af373397b74cc73090f111e/processor/memorylimiterprocessor/config.go#L46)
Collectors with old config files will fail to start in that case.
I am not sure if we want to add the memory_ballast extension even if it's not present, maybe just in cases where it's not present but the memory limiter processor contains ballast_size_mib

pmcollins:
> I think that the main issue to address is what happens if the deprecated ballast field will get removed from the [memory limiter processor](https://github.com/open-telemetry/opentelemetry-collector/blob/16c681ea0d17125d9af373397b74cc73090f111e/processor/memorylimiterprocessor/config.go#L46)
> Collectors with old config files will fail to start in that case.
> I am not sure if we want to add the memory_ballast extension even if it's not present, maybe just in cases where it's not present but the memory limiter processor contains ballast_size_mib

So, to my knowledge, the ballast field in the memory limiter processor never actually allocated a ballast. The memory limiter just used the ballast size (set manually) by subtracting it from the reported memory usage so the ballast wouldn't be counted when it calculated how much memory was being used. The ballast was actually allocated in the core of the collector (collector.go -- and it no longer is).

The Splunk Otel Collector at this point just sets the environment variable SPLUNK_BALLAST_SIZE_MIB either based on the `--mem-ballast-size-mib` command line arg or based on a calculation, and provides default configs that use this environment variable to set up a collector with a memory_ballast extension.  This works fine for new installs, but previous installs of older versions of the collector, where the config has been modified (and possibly in other cases as well), won't get the new config and therefore after upgrading will no longer have a memory ballast allocated. That was my understanding of this issue (from Jay's comment).

My thought was that the memory limiter processor ballast field going away is an issue, but it's perhaps a separate one. In any case, it could be handled in this PR since we're updating the collector config anyway -- we could just delete that key if we find it.

Anyway, having said all that 😄 I'm happy to remove this PR's ballast extension addition code, and just remove the memory limiter ballast key (if you think that's the best way to handle its upcoming removal).

And I realize that enhancing a Collector config maybe feels a bit odd at this point, but my personal opinion is that in the long term we may want to take this route to make setup easier for users. I am interested in what other folks think about this as well.

emaderer:
> Here's another option @emaderer -- I just pushed the change. This one just removes the ballast key from the memory limiter processor, if one exists, and doesn't add a ballast extension. Alternatively we could do both 😄 .

Can we take the approach of - if the ballast key exists in the memory limiter processor remove it and make sure we enable the ballast extension with the same value, otherwise there is no need to do anything?

pmcollins:
Ok, discussed separately with Eyal, and we're going to just remove the ballast key if it exists in the memory limiter processor, which is what this PR does in its current form. I've tested these changes against a local copy of the core collector, removed the ballast field and saw that the collector started up (verified that it didn't without these changes).

internal/configconverter/parser_provider.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+
+	"go.opentelemetry.io/collector/config/configparser"
+	"go.opentelemetry.io/collector/service/parserprovider"
+)
+
+// converterProvider wraps a ParserProvider and accepts a list of functions that
+// convert ConfigMaps. The idea is for this type to conform to the open-closed
+// principle.
+type converterProvider struct {
```
pmcollins, line 27: We may want to make use of https://github.com/open-telemetry/opentelemetry-collector/pull/40 when it's ready.

emaderer:
This looks great now I think! can we add a kill switch to disable any kind of config conversion just in case?

cmd/otelcol/main.go
```diff
@@ -67,7 +68,7 @@ func main() {
 
 	// Allow dumping configuration locally by default
 	// Used by support bundle script
-	os.Setenv(configServerEnabledEnvVar, "true")
+	_ = os.Setenv(configServerEnabledEnvVar, "true")
```
rmfitzpatrick, line 71: Since we're now explicitly ignoring all these errors I wonder if it's better to `log.Fatal`?

internal/configconverter/ballast_remover.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"log"
+	"regexp"
+
+	"go.opentelemetry.io/collector/config/configparser"
+)
+
+var ballastKeyRegexp *regexp.Regexp
+
+func init() {
```
rmfitzpatrick, line 26: If `RemoveBallastKey` is only called once is there a benefit to doing this in an init if there's only one consumer of `ballastKeyRegexp`?

pmcollins:
> This looks great now I think! can we add a kill switch to disable any kind of config conversion just in case?

Thanks -- yes sure. Should it be a command line switch?

cmd/otelcol/main.go
pmcollins: (Just trying to get rid of Goland warnings while I'm in here) It looks like `Setenv` only returns an error if the key is either empty, contains an equals character, or a has a null byte, which I don't think we're in danger of since it's hard coded. But happy to do so if you'd prefer.

internal/configconverter/ballast_remover.go
pmcollins: Yeah, there's no benefit. I'll move it to the function. (I hate `init()`s anyway 😛)

emaderer:
> > This looks great now I think! can we add a kill switch to disable any kind of config conversion just in case?
> 
> Thanks -- yes sure. Should it be a command line switch?

command line switch sounds good.

cmd/otelcol/main.go
rmfitzpatrick: Appears the remaining error condition in the linux syscall is oom, so logging would be impossible anyway.  no need to change here I guess.

rmfitzpatrick (approved):

Issue #700, opened 2021-09-01
[ansible] simple way to add new receivers


xp-1000:
Hello,

I am new on otel collector so may be I miss something.

I deploy it using ansible and I look for a way to keep the official/default configuration file but only add some components to it, especially receivers.

For now the only way I see is to copy the original `/etc/otel/collector/agent_config.yaml` next to my playbook, adapt its configuration and set it to  `splunk_otel_collector_config_source` variable.

the problem I see is my deployment will not be synced anymore with this official config and I could miss future fixes, improvements or best practices you could add.

with the smart agent we had the `monitors` variables allowing to add multiple monitors on the fly and there is also the `from` directive to dynamically include set of external configurations files which could be managed by ansible.

Here is an example of a custom ansible role played in addition to the official for smart agent deployment:

1. `defaults/main.yml`:

```yaml
---
paths: 
  monitors: "{{ config_path | default('/etc/signalfx') }}/monitors"
  scripts: "{{ config_path | default('/etc/signalfx') }}/scripts"
user: "signalfx-agent"
```

2. `tasks/main.yml`:

```yaml
---
- name: Check and create signalfx directories
  file: 
    path: "{{ item.value }}"
    owner: "{{ user }}"
    group: "{{ user }}"
    state: directory
  loop: "{{ paths | dict2items }}"

- name: Copy scripts files
  template:
    src: "{{ item }}"
    dest: "{{ paths['scripts'] }}/{{ (item | basename | splitext)[0] }}"
    owner: "{{ user }}"
    group: "{{ user }}"
    mode: 0775
  with_fileglob: "{{ role_path }}/templates/scripts/*.j2"
  notify: Agent restart

- name: Copy monitors configuration files
  template:
    src: "{{ item }}"
    dest: "{{ paths['monitors'] }}/{{ (item | basename | splitext)[0] }}"
    owner: "{{ user }}"
    group: "{{ user }}"
    mode: 0664
  with_fileglob: "{{ role_path }}/templates/monitors/*.j2"
  notify: Agent restart
```

3. then we just have to add template files in `templates/monitors` and/or `templates/scripts` which will be automatically deployed by ansible and read by the agent.

4. an example of playbook supposing this role is named `signalfx.smart_agent_config`:

```yaml
---
- hosts: all
  roles:
    - { role: signalfx.smart_agent, become: yes }
    - { role: signalfx.smart_agent_config, become: yes }
  vars:
    - sfx_agent_config:
[...]

PR #770, opened 2021-09-22
custom configuration file templating support

xp-1000:
waiting for feature or answer addressing https://github.com/signalfx/splunk-otel-collector/issues/700, using `template` module instead of `copy` will, at least, allow user to inject some variables from ansible to the otel collector config file.

For example if we need to define secret values in otel collector file, the only way is to hardcode them in the static file.
another way could be to provide variable to change src of `splunk-otel-collector.conf.j2`. this will allow users to define his own custom variable and use them in the otel collector yaml config file which could so remain static but this way seems less powerful.


xp-1000:
Hi @dmitryax, please tell me if I miss a border case where changing module could break something

Issue #767, opened 2021-09-21
splunk-otel-collector RPM conflicting with signalfx-agent


badnetmask:
While installing `splunk-otel-collector` via RPM on a Red Hat system, a message is shown indicating some of the files contained in the RPM conflict with files contained in the `signalfx-agent` RPM:
```
Transaction check error:
  file /usr/lib/.build-id/02/ba7db5d16044f210ebfca15e6012cb68dc4e74 from install of splunk-otel-collector-0.35.0-1.x86_64 conflicts with file from package signalfx-agent-5.13.0-1.x86_64
  file /usr/lib/.build-id/2f/368cd4b23ce7a802cb27bddacfc126777b4080 from install of splunk-otel-collector-0.35.0-1.x86_64 conflicts with file from package signalfx-agent-5.13.0-1.x86_64
  file /usr/lib/.build-id/30/89bd65d6fb05423a60bde2c6868a506b4c456b from install of splunk-otel-collector-0.35.0-1.x86_64 conflicts with file from package signalfx-agent-5.13.0-1.x86_64
  file /usr/lib/.build-id/96/652dd4786fcd6016a4c0f82e3072bf8cbdfab6 from install of splunk-otel-collector-0.35.0-1.x86_64 conflicts with file from package signalfx-agent-5.13.0-1.x86_64
  file /usr/lib/.build-id/97/7c39fe87abfa426d3043f6c8e21f7be3f0e876 from install of splunk-otel-collector-0.35.0-1.x86_64 conflicts with file from package signalfx-agent-5.13.0-1.x86_64
  file /usr/lib/.build-id/d6/4c191ddc16923d5e46ac8227e3bfc0016b2b2d from install of splunk-otel-collector-0.35.0-1.x86_64 conflicts with file from package signalfx-agent-5.13.0-1.x86_64
```

There are multiple ways to solve this conflict, but I would let the project owners to decide:
- You can remove the conflicting files from either packages.
- You can `Conflicts` in either packages.

badnetmask:
SignalFX Agent source -> `https://splunk.jfrog.io/splunk/signalfx-agent-rpm/release`
OTel Collector source -> `https://splunk.jfrog.io/splunk/otel-collector-rpm/release/x86_64/`

jcheng-splunk:
Thanks @badnetmask.  The conflicts are due to the `splunk-otel-collector` package containing many of the same files from `signalfx-agent`.  We'll take a look into how best address this.

[...]

PR #773, opened 2021-09-23
Exclude build-id links from rpm

jcheng-splunk:
For #767.  Resolves installation conflicts for `/usr/lib/.build-id` links if the signalfx-agent rpm is already installed.


mstumpfx (approved):

rmfitzpatrick:
Curious why the build ids collide, but if these aren't necessary and it works that seems good enough.

jcheng-splunk:
I think that since the collector and signalfx-agent may contain the same ELF files, the same build-id symlinks are generated.  But from what I understand, these symlinks are primarily for debugging, so excluding them shouldn't be an issue.

PR #775, opened 2021-09-24
Use sapm for trace exporter

pmcollins:


rmfitzpatrick:
Should we detect if `sendTraceHostCorrelationMetrics` is false and apply the sfx exporter otherwise?

pmcollins:
Oh, I wasn't familiar with this flag. If that's your recommendation then yes 😄

rmfitzpatrick (approved):

PR #810, opened 2021-10-07
[Ansible] Fix `splunk_bundle_dir` parameter inconsistency

dmitryax:
Documentation mentions `splunk_bundle_dir` parameter, but in reality `agent_bundle_dir` is used instead. This change makes the parameter consistent under `splunk_bundle_dir` name.


dmitryax:
@mstumpfx added a breaking change line to Changelog

PR #813, opened 2021-10-07
update ansible changelog after #770

xp-1000:
following up https://github.com/signalfx/splunk-otel-collector/pull/770 I forgot this


dmitryax:
@xp-1000 can you please rebase? I did another change to the CHANGELOG so it'll be conflicting I believe

PR #822, opened 2021-10-08
Put system detector last so cloud detectors can win

pmcollins:
Addresses https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/5313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 - Add Chocolately package support. (#724)
 
+## 🛑 Breaking changes 🛑
+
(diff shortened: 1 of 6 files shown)


mstumpfx:
I would call this out in the changelog as a potentially breaking change. EC2 and Azure both attempt to set host.name, so if someone decides to also follow our lead in this config change and make it in their own config change, this will change the host.name on all those MTSs. This could probably cause some detectors to fire. From a billing perspective this is an OK change since we will look for cloud resource IDs before host.name, which will remain the same.

mstumpfx (approved):

PR #824, opened 2021-10-08
Core/Contrib dependency test workflow

jeffreyc-splunk:
- Basic workflow to update the core/contrib deps to main, compile, and run tests once per weekday to check for upstream breaking changes
- See #823 for an example of an issue created from failure


emaderer:
This is awesome! what will the **Create Issue** do in case of a failure? will it create a GitHub issue?

PR #832, opened 2021-10-11
Add config converter function to unsquash Splunk HEC exporter tls fields

pmcollins:
Not enabled yet but available for when contrib dependency is updated.


dmitryax:
I would say let's add everything needed for the migration in this PR (including enabling the converter) and merge it right before the release? core/contrib releases are already under way, so it can be ready tomorrow

PR #847, opened 2021-10-15
Document windows service debug logging

rmfitzpatrick:
This is a common question without a clear approach.  Suggestions welcome (@jcheng-splunk).


rmfitzpatrick:
> Should this also be added to https://github.com/signalfx/splunk-otel-collector/blob/main/docs/getting-started/windows-installer.md?

I'm not sure doubling up is ideal, but how about the link I added from the installer to the manual doc?

Issue #860, opened 2021-10-20
Vulnerabilities Detected During Security Scan (v0.36.1)


asomensari-splunk:
Two issues were identified during a CIS/twistlock scan:
- (CIS_Docker_v1.2.0 - 4.1) Image should be created with a non-root user
- Private keys stored in image (Found: /usr/lib/splunk-otel-collector/agent-bundle/collectd-python/etcd/integration-test/certs/client.key, /usr/lib/splunk-otel-collector/agent-bundle/collectd-python/etcd/integration-test/certs/server.key, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/site-packages/novaclient/tests/unit/idfake.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/allsans.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/badcert.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/badkey.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/idnsans.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/keycert.passwd.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/keycert.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/keycert2.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/keycert3.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/keycert4.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/keycertecc.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/pycakey.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/ssl_key.passwd.pem, /usr/lib/splunk-otel-collector/agent-bundle/lib/python3.8/test/ssl_key.pem)

dloucasfx:
The private keys are test certs that were bundled with the python distribution and collectd plugins, and are not used by the collector or smart agent.
This PR will address https://github.com/signalfx/splunk-otel-collector/pull/861 

For the non-root, this PR will address it https://github.com/signalfx/splunk-otel-collector/pull/864

PR #864, opened 2021-10-21
create and run the otel container process as non-root user

dloucasfx:
verified that the default k8s helm config worked with no errors.
 

dloucasfx:
cc: https://github.com/signalfx/splunk-otel-collector/issues/860

cmd/otelcol/Dockerfile
```diff
@@ -10,6 +10,7 @@ COPY migratecheckpoint /
 RUN chmod 755 /otelcol
 RUN chmod 755 /translatesfx
 RUN chmod 755 /migratecheckpoint
+RUN echo "splunk-otel:x:999:999::/:" > /etc_passwd
```
jcheng-splunk, line 13: Just a nit, but the deb/rpm creates and runs the collector service as the `splunk-otel-collector` user/group.  Should we use the same name for consistency?

PR #866, opened 2021-10-21
Build/push arm64 image and multiarch manifest

jcheng-splunk:
- Update `docker-otelcol` make target for `ARCH` var and `docker buildx`
- Update Dockerfile to only download/extract smart agent bundle for amd64
- Update circleci to build/run images for both archs 
- Build/tag/push `quay.io/signalfx/splunk-otel-collector:<version>-[amd64|arm64]` images from gitlab
- Create/tag/push `quay.io/signalfx/splunk-otel-collector:<version>` manifest with both archs from gitlab

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,17 @@ executors:
     docker:
       - image: cimg/go:1.17.2
 
+  docker-amd64:
+    machine:
+      image: ubuntu-2004:202101-01
+      docker_layer_caching: true
+
+  docker-arm64:
+    machine:
+      image: ubuntu-2004:202101-01
+      docker_layer_caching: true
+    resource_class: arm.medium
+
 commands:
   attach_to_workspace:
     steps:
@@ -89,7 +100,16 @@ workflows:
       - docker-otelcol:
           requires:
             - cross-compile
-      - windows-docker-otelcol:
+          name: docker-otelcol-amd64
+          os: docker-amd64
+          arch: amd64
+      - docker-otelcol:
+          requires:
+            - cross-compile
+          name: docker-otelcol-arm64
+          os: docker-arm64
+          arch: arm64
+      - docker-otelcol-windows:
           requires:
             - cross-compile
       - test:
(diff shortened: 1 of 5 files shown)


rmfitzpatrick:
What the implications of manifests being experimental and of these changes when using an older docker client and trying to run `quay.io/signalfx/splunk-otel-collector:latest`?

jcheng-splunk:
> What the implications of manifests being experimental and of these changes when using an older docker client and trying to run `quay.io/signalfx/splunk-otel-collector:latest`?

According to https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility:

> When pulling images, clients indicate support for this new version of the manifest format by sending the application/vnd.docker.distribution.manifest.v2+json and application/vnd.docker.distribution.manifest.list.v2+json media types in an Accept header when making a request to the manifests endpoint. Updated clients should check the Content-Type header to see whether the manifest returned from the endpoint is in the old format, or is an image manifest or manifest list in the new format.
If the manifest being requested uses the new format, and the appropriate media type is not present in an Accept header, the registry will assume that the client cannot handle the manifest as-is, and rewrite it on the fly into the old format. If the object that would otherwise be returned is a manifest list, the registry will look up the appropriate manifest for the amd64 platform and linux OS, rewrite that manifest into the old format if necessary, and return the result to the client. If no suitable manifest is found in the manifest list, the registry will return a 404 error.

Also, since our current images use the v2 manifest schema (same as the multiarch manifest), it should not be an issue.

jcheng-splunk:
A test multiarch manifest was pushed to `quay.io/signalfx/splunk-otel-collector-dev:bc4b79dd130cbd179a21b61f728cb3c20d3f5a9a` (based on the v0.37.1 release):
```
$ docker manifest inspect quay.io/signalfx/splunk-otel-collector-dev:bc4b79dd130cbd179a21b61f728cb3c20d3f5a9a
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 3034,
         "digest": "sha256:e1e9a78fe54b9c953671332d2567f754667ea74829a06f9296e401bf7b296466",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 3028,
         "digest": "sha256:ffa8af1cc4219ba6b140c579c4f54059a2a80a0beac1843230a88d6b9bbfd346",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}
```

rmfitzpatrick (approved):

PR #902, opened 2021-10-29
Migrate windows-test, puppet-test, and installer-script-test jobs from CircleCi to GitHub Action.

nitaliya:
**Migrate windows-test, puppet-test, and installer-script-test jobs**
Ticket: [OTL-1256](https://signalfuse.atlassian.net/browse/OTL-1256).

- Adding GitHub Action jobs of `windows-test`, `puppet-test`, and `installer-script-test`.
- Removing CircleCi jobs of `windows-test`, `puppet-test`, and `installer-script-test`.


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,233 @@
...
+    runs-on: windows-latest
+    steps:
+      - name: Check out the codebase.
+        uses: actions/checkout@v2
+
+      - name: Caching dependency
+        uses: actions/cache@v2
+        with:
+          path: |
+            %LocalAppData%\go-build
```
jcheng-splunk, line 21: Should this match `$env:GOPATH` below?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,233 @@
...
+      PIP_VERSION: 20.2.4
+      REQUIREMENTS_PATH: "internal/buildscripts/packaging/tests/requirements.txt"
+      RESULT_PATH: "~/testresults"
+    steps:
+      - name: Check out the codebase.
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Check for relevant changes
```
jcheng-splunk, line 60: Can we try https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths instead?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,233 @@
...
+            export BASE_BRANCH_NAME=${gitRef[2]}
+          fi
+          export HEAD_BRANCH_NAME=${{ github.sha }}
+          if [[ "${BASE_BRANCH_NAME}" != "main" ]] && [[ "${gitRef[1]}" != "tags" ]] && ! .github/scripts/changes-include internal/buildscripts/packaging/installer internal/buildscripts/packaging/tests ; then
+            echo "SKIP_TESTS=yes" >> $GITHUB_ENV
+          else
+            echo "SKIP_TESTS=no" >> $GITHUB_ENV
+          fi
+
+      - name: Setup pyenv
```
jcheng-splunk, line 77: Can we try https://github.com/actions/setup-python instead of pyenv?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,233 @@
+name: build-and-test
+
+on: [push, pull_request]
+
+concurrency:
+  group: build-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  windows-test:
```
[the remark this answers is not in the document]
nitaliya: Yes, we can create but we need to specify all dependent jobs in one workflow file. for example in upcoming PRs for `setup-environment`, `lint`, `test` we can not create separate files for it. for simplicity, I have kept all jobs in one file.

if you fill that we need to have separate workflow files for jobs that are not dependant then we can do that thing here.

@jcheng-splunk What's your suggestion?

.github/workflows/build-and-test.yml
nitaliya: All cache data are stored on `%LocalAppData%\go-build` so I have taken it in cashing section. `$env:GOPATH` will contain all data that are statics between build, Those files are not "cashe" files actually.

.github/workflows/build-and-test.yml
nitaliya: If we follow the way that you have suggested then we need to create different workflow files for each job.
Can we do that?

.github/workflows/build-and-test.yml
nitaliya: Yes, the same thing we can do using setup-python. In CircleCi we are using `pyenv` so I have kept it over here.
If you suggest then we can use `setup-python` also.

.github/workflows/build-and-test.yml
jcheng-splunk: Yes.  These jobs should be able to run independently and only when there are relevant changes.

.github/workflows/build-and-test.yml
nitaliya: Change workflow to use `setup-python`.

.github/workflows/build-and-test.yml
nitaliya: Changes made.

.github/workflows/build-and-test.yml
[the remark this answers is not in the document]
nitaliya: Creating a separate workflow for jobs that are not dependant.

.github/workflows/puppet-test.yml
```diff
@@ -0,0 +1,85 @@
+name: puppet-test
+
+# Only run installer tests, Rake spec, lint for main branch, tags, or if the branch has relevant changes
+on: 
+  push:
+    tags:
+      - '*'
+    paths:
+      - 'internal/buildscripts/packaging/installer/**'
```
jcheng-splunk, line 9:
```suggestion
      - 'deployments/puppet/**'
```

.github/workflows/installer-script-test.yml
```diff
@@ -0,0 +1,71 @@
+name: installer-script-test
+
+on: 
+  push:
+    tags:
```
jcheng-splunk, line 5: Let's skip for tags.

.github/workflows/puppet-test.yml
```diff
@@ -0,0 +1,85 @@
+name: puppet-test
+
+# Only run installer tests, Rake spec, lint for main branch, tags, or if the branch has relevant changes
+on: 
+  push:
+    tags:
```
jcheng-splunk, line 6: Let's skip for tags.

.github/workflows/installer-script-test.yml
```diff
@@ -0,0 +1,71 @@
...
+      - '*'
+    paths:
```
jcheng-splunk, line 7: Let's also include this workflow file itself.

.github/workflows/puppet-test.yml
```diff
@@ -0,0 +1,85 @@
...
+      - '*'
+    paths:
```
jcheng-splunk, line 8: Let's also include this workflow file itself.

.github/workflows/installer-script-test.yml
```diff
@@ -0,0 +1,71 @@
...
+  pull_request:
+    branches:
+      - 'main'
+
+concurrency:
+  group: installer-script-test-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  installer-script-test:
```
jcheng-splunk, line 19:
```suggestion
  linux-installer-script-test:
```

.github/workflows/installer-script-test.yml
```diff
@@ -0,0 +1,71 @@
...
+    name: installer-script-test
```
jcheng-splunk, line 20:
```suggestion
    name: linux-installer-script-test
```

.github/workflows/puppet-test.yml
```diff
@@ -0,0 +1,85 @@
...
+
+      - name: Lint
+        run: |
+          make -C deployments/puppet lint
+
+      - name: Rake spec
+        run: |
+          make -C deployments/puppet rake-spec
+
+      - name: Setup pyenv
```
jcheng-splunk, line 45: Let's use `actions/setup-python@v2` like the installer script test.

.github/workflows/installer-script-test.yml
nitaliya: Changes made.

.github/workflows/installer-script-test.yml
nitaliya: Changes made.

.github/workflows/puppet-test.yml
nitaliya: Changes made.

.github/workflows/puppet-test.yml
nitaliya: Changes made.

.github/workflows/puppet-test.yml
nitaliya: Changes made.

.github/workflows/puppet-test.yml
```diff
@@ -0,0 +1,81 @@
+name: puppet-test
+
+# Only run installer tests, Rake spec, lint for main branch, tags, or if the branch has relevant changes
+on: 
+  push:
+    paths:
+      - '.github/workflows/**'
```
jcheng-splunk, line 7:
```suggestion
      - '.github/workflows/puppet-test.yml'
```

.github/workflows/installer-script-test.yml
```diff
@@ -0,0 +1,70 @@
+name: installer-script-test
+
+on: 
+  push:
+    paths:
+      - '.github/workflows/**'
```
jcheng-splunk, line 6:
```suggestion
      - '.github/workflows/installer-script-test.yml'
```

jcheng-splunk (approved):

nitaliya:
@rmfitzpatrick can you give your review on this PR. or if no changes are required then I would suggest go ahead and get the PR merged.

PR #939, opened 2021-11-12
Support set flag and update log-level example usage

rmfitzpatrick:
https://github.com/open-telemetry/opentelemetry-collector/commit/a51bf5a2753263aa0284f76548847bea46e75d07 broke `--set` support for our distribution and https://github.com/open-telemetry/opentelemetry-collector/commit/48a2e01652fa679c89259866210473fc0d42ca95 removed `--log-level` support from the collector so that setting non-info log level requires manual config changes (read: inconvenient).

These changes ~port the `--set` properties parser and use it in our default config provider.  Also updates `--log-level` usage throughout project to the required `--set service.telemetry.logs.level` value.

### DIFF
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -19,7 +19,9 @@ package main
 
 import (
 	"bytes"
+	"flag"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"strconv"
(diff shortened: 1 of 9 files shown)


tmorrisonsfx (approved):

cmd/otelcol/main.go
```diff
@@ -113,6 +115,31 @@ func main() {
 	}
 }
 
+// required to support --set functionality no longer directly parsed by the core config loader.
+// taken from https://github.com/open-telemetry/opentelemetry-collector/blob/48a2e01652fa679c89259866210473fc0d42ca95/service/flags.go#L39
+type stringArrayValue struct {
```
mstumpfx, line 120: is it worth adding the interface check for flag.Value for this?

mstumpfx (approved):

cmd/otelcol/main.go
rmfitzpatrick: I think there's a compile time check for line 137 and don't think the stdlib will be updating their type anytime soon.  If we were using this w/ indirection/another interface I would though.

Issue #981, opened 2021-11-24
Please add contrib routingprocessor to Splunk distro


billg-splunk:
Context: Customer wants to deploy OTel to a kubernetes cluster with multiple org destinations. They will need some metrics and traces to go to one org and some going to another. The routing processor appears to support this use case.

https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/routingprocessor

PR #982, opened 2021-11-24
Add routing processor to components

mstumpfx:
Fixes https://github.com/signalfx/splunk-otel-collector/issues/981


go.sum
```diff
@@ -1880,6 +1880,8 @@ github.com/onsi/gomega v1.16.0/go.mod h1:HnhC7FXeEQY45zxNK3PPoIUhzk/80Xly9PcubAl
 github.com/op/go-logging v0.0.0-20160315200505-970db520ece7/go.mod h1:HzydrMdWErDVzsI23lYNej1Htcns9BCg93Dk0bBINWk=
 github.com/open-telemetry/opentelemetry-collector-contrib/exporter/fileexporter v0.39.0 h1:PB7zyHwV6p+Z8z4MDcFzgBBHPMEd0rasNKK8c0bMC6o=
 github.com/open-telemetry/opentelemetry-collector-contrib/exporter/fileexporter v0.39.0/go.mod h1:wJmDmp9OeDbaWWg18jhfjzvWSIxJfJWYKZ9R0izyiz0=
+github.com/open-telemetry/opentelemetry-collector-contrib/exporter/jaegerexporter v0.39.0 h1:Rlo1LtOqL82WLvO1DbD4oxkGfHjpEba0qtFdphCHP1I=
+github.com/open-telemetry/opentelemetry-collector-contrib/exporter/jaegerexporter v0.39.0/go.mod h1:rtWpnSnfT06AdQMPnqGDQP+X8qc8lkF6pHGPNsAW8/4=
```
dmitryax, line 1884: Sad that we are bringing it because the routing processor use in one test case

dmitryax (approved):

go.sum
billg-splunk: Not sure what the downside is, but there will surely be more needs here. And the alternative, recommending they build their own distro, isn't better for them or Splunk.

I do appreciate the quick action here!

dmitryax: I'm talking just about `jaegerexporter` dependency brought by the routing processer where it's being used just in one test. Not a big deal but makes our build unnecessary bigger. 

Adding the processor itself makes sense and LGTM

PR #996, opened 2021-11-30
Update exclude list for semgrep

jcheng-splunk:
The splunk semgrep policy was recently updated and now fails if Dockerfiles do not define or run as `USER` .  This change excludes the windows and ansible test Dockerfiles from semgrep scans.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -27,7 +27,7 @@ semgrep:
   retry: 2
   variables:
     SAST_SCANNER: "Semgrep"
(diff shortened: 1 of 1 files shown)


rmfitzpatrick:
Should we be updating the windows dockerfile to run as a splunk-otel-collector?

jcheng-splunk:
> Should we be updating the windows dockerfile to run as splunk-otel-collector?*

Not sure... I assumed that the collector process in the windows image needs to run as admin, similar to the standalone collector when running directly on windows.

rmfitzpatrick (approved):

PR #1018, opened 2021-12-08
Update docs and comments to use Collector term instead of Connector

dmitryax:


rmfitzpatrick:
@tmorrisonsfx mind taking a look?

rmfitzpatrick (approved):

docs/experimental/using-fluentd-only.md
```diff
@@ -8,7 +8,7 @@
 
 Use-case: If you have an existing signalfx-agent collecting metrics and/or 
 traces and want to try Splunk Log Observer without migrating to Splunk 
-OpenTelemetry Connector, you can use fluentd to send logs directly to Splunk Log
+OpenTelemetry Collector, you can use fluentd to send logs directly to Splunk Log
```
rpille-sfx, line 11: Correct capitalization would be "Fluentd" per https://www.fluentd.org/. This could be a find/replace thing throughout as well, since I see us using the term a few different ways, but is not a blocker for this PR.

rpille-sfx (approved):
Took a look since Traci is out of the office today. LGTM. Left one nit about Fluentd capitalization that can be handled later if needed.

docs/experimental/using-fluentd-only.md
dmitryax: Thank, we will update usages of Fluentd in other PR in order to keep the purpose of this this PR for Collector renaming only

PR #1024, opened 2021-12-09
Upgrade lychee-action with required changes

dmitryax:
Recent [change to the actions](https://github.com/lycheeverse/lychee-action/pull/64) added quotes to arguments, so we need to remove our own.

Replaces https://github.com/signalfx/splunk-otel-collector/pull/1023


dmitryax:
Closing the PR because "*.md */**.md" part is being translated to the list of matching file and lychee doesn't pick all the files from the list for some reason. May be will take a look later

PR #1026, opened 2021-12-09
Add journalctl to docker image

jcheng-splunk:
`collect-libs.sh` is based on https://github.com/signalfx/signalfx-agent/blob/main/scripts/collect-libs which we use for bundling collectd dependencies in the smart agent.


dmitryax (approved):
This is great! Thanks Jeff!

PR #1047, opened 2021-12-15
add note to troubleshooting doc for updating metrics address

mstumpfx:
Specifying that users will need to change their agent config
when changing the default metrics address since we default to
:8888 in all configs. 
Also changed links point at a tagged version for specific lines of code since the `main` branch frequently changes and makes links to it inconsistent.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -67,10 +67,13 @@ typically requires modifying either:
 - [Metrics
   address](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/troubleshooting.md#metrics)
   (if port `8888`)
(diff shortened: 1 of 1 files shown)


rmfitzpatrick (approved):

rmfitzpatrick:
This is helpful info but I wonder if we should add a new env var to set this port to avoid manual config changes.

mstumpfx:
> This is helpful info but I wonder if we should add a new env var to set this port to avoid manual config changes.
There's some work in progress to move this to the service config already - https://github.com/open-telemetry/opentelemetry-collector/pull/4069
I'm in favor of waiting for that to land until we make that type of change

PR #1051, opened 2021-12-17
New example showing routing techniques with the collector

atoulme:


examples/otel-logs-routing/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+      interval: 5s
+      timeout: 5s
+      retries: 20
+    volumes:
+      - ./splunk.yml:/tmp/defaults/default.yml
+      - /opt/splunk/var
+      - /opt/splunk/etc
+  # OpenTelemetry Collector
+  otelcollector:
+    image:  quay.io/signalfx/splunk-otel-collector:0.29.0
```
rmfitzpatrick, line 31: Any reason not to use latest or a more recent build?

examples/otel-logs-routing/otel-collector-config.yml
```diff
@@ -0,0 +1,87 @@
...
+    zpages:
+        endpoint: :55679
+
+service:
+    extensions: [ pprof, zpages, health_check ]
+    pipelines:
+        logs:
+            receivers: [ filelog ]
+            processors: [ batch ]
+            exporters: [ splunk_hec/logs, splunk_hec/logs2, splunk_hec/routing ]
```
rmfitzpatrick, line 83: If logs2 were moved to the routing pipeline it may be more informative than using a logging exporter?

examples/otel-logs-routing/docker-compose.yml
atoulme: Copied from another sample. I assumed there was a job updating all samples or something in place around that and therefore a good reason for the version used.

examples/otel-logs-routing/otel-collector-config.yml
atoulme: A big ask has been to send data to 2 different Splunk instance, so logs and logs2 show this. The routing is extra to show we can route to another exporter. I debated creating another collector instead of sending to the same collector.

atoulme: OK I made the example even more complex with 2 otel collectors talking to each other.

rmfitzpatrick (approved):

PR #1058, opened 2021-12-21
Fix bug requiring $$ when using config source variables

pmcollins:
Previously, if a user wanted to use a config source variable (which have the
form `${env:MY_ENV_VAR}`) they would have to use two dollar signs, like
`$${env:MY_ENV_VAR}`. This was because the `expandMapProvider`, which lives in
core and expands `${MY_ENV_VAR}`-style variables, would run before the config
source providers, and would replace any `${env:MY_ENV_VAR}`s with "" but
convert any `$${env:MY_ENV_VAR}`s to `${env:MY_ENV_VAR}`. The fix proposed here
is to reverse the order in which these providers run: now the config source
providers run first, then `expandMapProvider` runs. In addition, this change
fixes up any `$${env:MY_ENV_VAR}` workarounds users may have put in place.

### DIFF
--- a/internal/configconverter/dollar_dollar.go
+++ b/internal/configconverter/dollar_dollar.go
@@ -26,14 +26,21 @@ import (
 // in how the Collector expanded these variables.
 func ReplaceDollarDollar(m *config.Map) *config.Map {
 	re := dollarDollarRegex()
+	replace := func(s string) string {
+		return replaceDollarDollar(re, s)
+	}
 	for _, k := range m.AllKeys() {
-		v := m.Get(k)
-		if s, ok := v.(string); ok {
-			replaced := replaceDollarDollar(re, s)
-			if replaced != s {
-				log.Printf("[WARNING] the notation %q is no longer recommended. Please replace with %q.\n", v, replaced)
+		switch v := m.Get(k).(type) {
+		case string:
+			replaced := replace(v)
+			if replaced != v {
+				format := "[WARNING] the notation %q is no longer recommended. Please replace with %q.\n"
+				log.Printf(format, v, replaced)
 				m.Set(k, replaced)
 			}
+		case []interface{}:
+			replaced := replaceArray(v, replace)
+			m.Set(k, replaced)
 		}
 	}
 	return m
--- a/internal/configconverter/dollar_dollar.go
+++ b/internal/configconverter/dollar_dollar.go
@@ -1,0 +1,48 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configconverter
+
+import (
+	"log"
+	"regexp"
+
+	"go.opentelemetry.io/collector/config"
+)
+
+// ReplaceDollarDollar replaces any $${foo:MY_VAR} config source variables with
+// ${foo:MY_VAR}. These might exist because of customers working around a bug
+// in how the Collector expanded these variables.
+func ReplaceDollarDollar(m *config.Map) *config.Map {
+	re := dollarDollarRegex()
+	for _, k := range m.AllKeys() {
+		v := m.Get(k)
+		if s, ok := v.(string); ok {
+			replaced := replaceDollarDollar(re, s)
+			if replaced != s {
+				log.Printf("[WARNING] the notation %q is no longer recommended. Please replace with %q.\n", v, replaced)
+				m.Set(k, replaced)
+			}
+		}
+	}
+	return m
+}
+
+func dollarDollarRegex() *regexp.Regexp {
+	return regexp.MustCompile(`\$\${(.+?:.+?)}`)
+}
+
+func replaceDollarDollar(re *regexp.Regexp, s string) string {
+	return re.ReplaceAllString(s, "${$1}")
+}
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -18,7 +18,6 @@
 package main
 
 import (
-	"bytes"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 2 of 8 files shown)


rmfitzpatrick:
There's an existing integration test that would need to be updated/could prove useful for this: https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/envvar_config_source_test.go#L33

internal/configconverter/dollar_dollar.go
```diff
@@ -0,0 +1,48 @@
...
+				log.Printf("[WARNING] the notation %q is no longer recommended. Please replace with %q.\n", v, replaced)
+				m.Set(k, replaced)
+			}
+		}
+	}
+	return m
+}
+
+func dollarDollarRegex() *regexp.Regexp {
+	return regexp.MustCompile(`\$\${(.+?:.+?)}`)
```
rmfitzpatrick, line 43: Is it preferable to obtain via a function and not a global var?

rmfitzpatrick (approved):

internal/configconverter/dollar_dollar.go
```diff
@@ -0,0 +1,48 @@
...
+	"log"
+	"regexp"
+
+	"go.opentelemetry.io/collector/config"
+)
+
+// ReplaceDollarDollar replaces any $${foo:MY_VAR} config source variables with
+// ${foo:MY_VAR}. These might exist because of customers working around a bug
+// in how the Collector expanded these variables.
+func ReplaceDollarDollar(m *config.Map) *config.Map {
```
dmitryax, line 27: It's not only `${foo:MY_VAR}` that is affected. Any env variables added to the config has to have the following `$${ENV_VAR}` format to work currently, and if user wants to have a dollar signed it must be escaped as `$$$$`.

I'm not sure if we want to implicitly translate any of the use cases.

I would suggest just adding a warning message for any encountered value with `$${` and `$$$$` but do not change them implicitly. Also we should make a clear statement about fixing this bug as a breaking change, add a line to changelog and to the upgrade guideline. 

@rmfitzpatrick WDYT?

dmitryax (commented):
Thanks for fixing it!

internal/configconverter/dollar_dollar.go
dmitryax: Added comment assuming that the use cases I outlined are fixed as well

rmfitzpatrick: > It's not only `${foo:MY_VAR}` that is affected. Any env variables added to the config currently has to have the following `$${ENV_VAR}` format to work properly, and, if user wants to have a dollar sign, it must be escaped as `$$$$`.

I don't think this is accurate and `$ENVVAR` and `${ENVVAR}` should be functional as is and after these changes (for example https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/testdata/config_with_envvars.yaml#L9 and https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/agent_config.yaml#L25).

If users want to have a `$` literal I think it makes sense that they'd need to escape it somehow and `$$` makes the most sense to me. This should be clearly documented and ideally vetted with unit/integration tests.

dmitryax: > I don't think this is accurate and $ENVVAR and ${ENVVAR} should be functional as is and after these changes (for example https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/testdata/config_with_envvars.yaml#L9 and https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/agent_config.yaml#L25).

Maybe only user provided config is affected, but I'm pretty sure I was the case. Let me verify it

> If users want to have a $ literal I think it makes sense that they'd need to escape it somehow and $$ makes the most sense to me. This should be clearly documented and ideally vetted with unit/integration tests.

Yes, it should be `$$` not `$$$$` which we have to set because of the but in [the helm chart](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L142) but it's not needed when contrib image is used

dmitryax: Probably we can keep the implicit translation for some time, it shouldn't break existing configs, but in order to cover all the use cases we probably should just convert `$$` -> `$`. 

I would still clearly mark it in the changelog and translation guidelines and set expectation when this translation is removed. Let's say in 0.44.0?

rmfitzpatrick: local testing with these changes:

```
config_sources:
  env:
    defaults:
      HOST_METRICS_SCRAPERS_TO_EXPAND: {}
      HOST_METRICS_SCRAPERS_DEFAULT_TO_USE: { cpu: null }

receivers:
  hostmetrics:
    collection_interval: $$HOST_METRICS_COLLECTION_INTERVAL
    scrapers: ${env:HOST_METRICS_SCRAPERS_TO_EXPAND}
  hostmetrics/default-env-config-source:
    collection_interval: $${HOST_METRICS_COLLECTION_INTERVAL}
    scrapers: $${env:HOST_METRICS_SCRAPERS_DEFAULT_TO_USE}
```

http://localhost:55554/debug/configz/effective:

```
receivers:
  hostmetrics:
    collection_interval: $HOST_METRICS_COLLECTION_INTERVAL
    scrapers:
      cpu: {}
      disk: {}
      filesystem: {}
      load: {}
      memory: {}
      network: {}
      paging: {}
      processes: {}
  hostmetrics/default-env-config-source:
    collection_interval: ${HOST_METRICS_COLLECTION_INTERVAL}
    scrapers:
      cpu: null
```

What should the expected behavior be if not this?

dmitryax: AFAIR from my testing sometime ago http://localhost:55554/debug/configz/effective doesn't show results of the second evaluation

dmitryax: collector should not be able to start with the following part using contrib build
```
receivers:
  hostmetrics:
    collection_interval: $$HOST_METRICS_COLLECTION_INTERVAL
```

dmitryax (changes requested):
I would like to have make sure we solve the problem uniformly not just for "env" configsource and decide if we want to keep backward compatibility and for how long.

dmitryax:
Given the double evaluation is fixed in https://github.com/signalfx/splunk-otel-collector/pull/1099, i'm ok if we keep keep only `$${<source>:<val>}` -> `${<source>:<val>}` backward compatibility translation. This PR needs to go after 1099

dmitryax:
Superseded by https://github.com/signalfx/splunk-otel-collector/pull/1099

PR #1079, opened 2022-01-05
Update the version of all the examples to use the latest release of the collector

atoulme:


dmitryax:
I don't think we should use `latest` given that otel-collector is still in beta and configuration interface can be changed

atoulme:
OK, I'll use latest release instead.

rmfitzpatrick (approved):

PR #1107, opened 2022-01-11
check if monitor accepts endpoints before setting host/port

mstumpfx:
Only select smart-agent monitors can be created via endpoints. This is
defined as a struct tag on the monitor configs "acceptsEndpoints". In
the case where we attempt to set Host/Port on a monitor that has it's
"acceptsEndpoints" value set to false, an error occurs and we do not
create the monitor.

This change would create the monitor but avoid setting the
Host/Port fields. This allows the receiver creator to
create any smart-agent monitor with receiver creator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ This Splunk OpenTelemetry Collector release includes changes from the [opentelem
 - [`docker_observer`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/observer/dockerobserver) to detect and create container endpoints, to be used with the [`receiver_creator`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/receivercreator) (#1044)
 - [`ecs_task_observer`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/observer/ecstaskobserver) to detect and create ECS task container endpoints, to be used with the [`receiver_creator`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/receivercreator) (#1125)
 
+### 🧰 Bug fixes 🧰
+
+- [`smartagent` receiver](https://github.com/signalfx/splunk-otel-collector/tree/main/internal/receiver/smartagentreceiver) will now attempt to create _any_ monitor from a Receiver Creator instance, disregarding its provided `endpoint`. Previously would error out if a monitor did not accept endpoints ([#1107](https://github.com/signalfx/splunk-otel-collector/pull/1107))
+
 ## v0.41.0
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🧰 Bug fixes 🧰
+
+- [`smartagent` receiver](https://github.com/signalfx/splunk-otel-collector/tree/main/internal/receiver/smartagentreceiver) will now attempt to create _any_ monitor. Previously would error out if a monitor did not accept endpoints ([#1107](https://github.com/signalfx/splunk-otel-collector/pull/1107))
```
rmfitzpatrick, line 7:
```suggestion
- [`smartagent` receiver](https://github.com/signalfx/splunk-otel-collector/tree/main/internal/receiver/smartagentreceiver) will now attempt to create _any_ monitor from a Receiver Creator instance, disregarding its provided `endpoint`. Previously would error out if a monitor did not accept endpoints ([#1107](https://github.com/signalfx/splunk-otel-collector/pull/1107))
```

internal/receiver/smartagentreceiver/config_test.go
```diff
@@ -342,19 +343,35 @@ func TestLoadInvalidConfigWithInvalidEndpoint(t *testing.T) {
 	require.Nil(t, cfg)
 }
 
-func TestLoadInvalidConfigWithUnsupportedEndpoint(t *testing.T) {
+// Even though this smart-agent monitor does not accept endpoints,
+// we can create it without setting Host/Port fields.
+func TestLoadConfigMonitorNoHostPort(t *testing.T) {
```
rmfitzpatrick, line 348: nit, since this test is about the endpoint value I'd maybe update to something like `TestLoadConfigWithUnsupportedEndpoint `

internal/receiver/smartagentreceiver/config_test.go
```diff
@@ -342,19 +343,35 @@ func TestLoadInvalidConfigWithInvalidEndpoint(t *testing.T) {
...
 	factories, err := componenttest.NopFactories()
 	assert.Nil(t, err)
 
 	factory := NewFactory()
 	factories.Receivers[typeStr] = factory
 	cfg, err := configtest.LoadConfig(
-		path.Join(".", "testdata", "unsupported_endpoint.yaml"), factories,
+		path.Join(".", "testdata", "monitor_no_hostport.yaml"), factories,
```
rmfitzpatrick, line 355: similar request (no rename?)

internal/receiver/smartagentreceiver/config.go
```diff
@@ -110,11 +110,18 @@ func (cfg *Config) Unmarshal(componentParser *config.Map) error {
...
 	}
 
-	err = setHostAndPortViaEndpoint(cfg.Endpoint, monitorConfig)
+	acceptsEndpoints, err := monitorAcceptsEndpoints(monitorConfig)
 	if err != nil {
 		return err
 	}
 
+	if acceptsEndpoints {
+		err = setHostAndPortViaEndpoint(cfg.Endpoint, monitorConfig)
```
rmfitzpatrick, line 119: following up from slack, we probably want to log that the endpoint is being ignored to prevent unnecessary debugging/confusion.

rmfitzpatrick (approved):

PR #1099, opened 2022-01-10
update deps to 0.42.0 and provide double dollar config source parsing compatibility

rmfitzpatrick:
initial adoption of config (map) provider helpers (https://github.com/open-telemetry/opentelemetry-collector/pull/4637) to address breaking changes (https://github.com/open-telemetry/opentelemetry-collector/pull/4600).

Probably shouldn't land until it is known to accomplish what https://github.com/signalfx/splunk-otel-collector/pull/1058 does and has been vetted w/ helm chart and additional integration test(s).

PR #1108, opened 2022-01-11
Add env var and env config source integration tests

rmfitzpatrick:
In reconciling https://github.com/signalfx/splunk-otel-collector/pull/1058 and https://github.com/signalfx/splunk-otel-collector/pull/1099 it is helpful to understand the current behavior. These changes add two integration tests (thanks to @pmcollins for the approach) that demonstrate what the behavior from 0.37.0+ has been. Deletes the previous version of this test since the config server is [unreliable for core config parsing](https://github.com/signalfx/splunk-otel-collector/pull/1058#discussion_r773439013).

Also including testutils resource metrics string improvements since their output is unfortunate to say the least and yaml marshalling is repeatable. Without these changes debugging these added tests is difficult.


dmitryax (approved):
LGTM. Thanks for adding these!

PR #1109, opened 2022-01-11
instrumentation: initial revision of libsplunk.so

pmcollins:
Going to use this repo instead of https://github.com/signalfx/linux-autoinstrumentation/ which I will close.

This change has been partially reviewed here:

https://github.com/signalfx/linux-autoinstrumentation/pull/1

before I moved it into this repo.

### DIFF
--- a/instrumentation/Makefile
+++ b/instrumentation/Makefile
@@ -1,17 +1,23 @@
 .PHONY: all
 all: so/libsplunk.so
 
+so:
+	mkdir so
+
+obj:
+	mkdir obj
+
 .PHONY: clean
 clean:
 	rm tests so/* obj/*
 
-obj/logger.o: src/logger.c
+obj/logger.o: obj src/logger.c
 	gcc -c -Wall -Werror -fpic -o obj/logger.o src/logger.c
 
-obj/splunk.o: src/splunk.c
+obj/splunk.o: obj src/splunk.c
 	gcc -c -Wall -Werror -fpic -o obj/splunk.o src/splunk.c
 
-so/libsplunk.so: obj/splunk.o obj/logger.o
+so/libsplunk.so: obj so obj/splunk.o obj/logger.o
 	gcc -shared -o so/libsplunk.so obj/splunk.o obj/logger.o
 
 tests: src/splunk.c src/test_main.c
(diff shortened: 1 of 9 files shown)


instrumentation/Makefile
```diff
@@ -0,0 +1,22 @@
+.PHONY: all
+all: so/libsplunk.so
```
jcheng-splunk, line 2: The targets should also create the `obj` and `so` directories.

instrumentation/README.md
```diff
@@ -0,0 +1,11 @@
+# linux-autoinstrumentation
+
+This directory contains functionality to auto instrument Java programs that run on a Linux host.
```
jcheng-splunk, line 3: Will this need to support both amd64 and arm64?

jcheng-splunk: Are we packaging this only in the deb and rpm packages, or in the linux docker images as well?

rmfitzpatrick:
It could be quite helpful moving forward for a build container so we have a shared dev/test environment. It would also be easier that way to intuitively tie into CI.

instrumentation/README.md
pmcollins: Haven't tested this in an arm environment, but my guess is that this will just work. However, my preference is that this PR just be for amd64 and we separate out arm support/testing.

I don't think this will work in a docker environment, so just deb/rpm. For security reasons, I think we should go with separate packages for this functionality.

cc: @aunshc

instrumentation/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The installation process will be done automatically in a later release, but for now, you can build a `.so` file for
+Linux with `make all`, then place the `.so` file on the host you want to auto instrument and put the full path to the
+`.so` file in a text file at the location `/etc/ld.so.preload`. Make sure you have a
+[Splunk OTel Java jar](https://github.com/signalfx/splunk-otel-java) at the location
+`/usr/lib/splunk-instrumentation/splunk-otel-javaagent.jar`. Once done, the `.so` will set the environment variable
```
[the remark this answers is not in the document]
pmcollins: I think the package should install it.

rmfitzpatrick (approved):

PR #1131, opened 2022-01-14
Add upgrade guideline for 0.42.0 version

dmitryax:


README.md
```diff
@@ -166,6 +166,13 @@ manually before the backward compatibility is dropped. For every configuration u
 [the default agent config](https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/agent_config.yaml)
 as a reference.
 
+### From 0.41.0 to 0.42.0
+
+- The Splunk OpenTelemetry Collector used to [evaluate user configuration 
+  twice](https://github.com/signalfx/splunk-otel-collector/issues/628) and this required escaping of 
+  each `$` symbol with `$$` to prevent unwanted environment variable expansion. The issue was fixed in 
+  0.42.0 version. Any occurrences of `$$` in your configuration should be replaced with `$`.
```
rmfitzpatrick, line 174: I'm not sure this correct since `$$` is still required for `$` literals: https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/testdata/resource_metrics/envvar_labels.yaml. I think the issue was that it had to be doubly escaped?

dmitryax: Which literals? If user have any of these: `$${ENV_VAR}`, `$$ENV_VAR`, `$${<source>:<val>}`, `$$<source>:<val>`, they should be replaced with one dollar sign `$`, because now we have only one expansion, right?

dmitryax: Even if they have just `$$` literal, it can be replaced with `$`, because expanded `$` still gives `$`.

rmfitzpatrick: > Even if they have just `$$` literal, it can be replaced with `$`, because expanded `$` still gives `$`.

It depends on its context (subsequent characters): https://github.com/signalfx/splunk-otel-collector/pull/1140/files. I think the current behavior is relatively surprising.

dmitryax: > edit: If a $ literal is wanted in rendered config it needs to be escaped with another $ if it proceeds another character.*

If we a talking about standalone `$` literals, it's not necessary expanded `$` still gives `$`.

To get `$something` in the resulted config, user had to set `$$$$something` prior to 0.42.0, now they have to set `$$something`. So the recommendation to replace every occurrence of `$$` with `$` still holds true.

dmitryax: I believe that keeping the ask as simple as `$$`->`$` is valid and easy to do, but we can also extend the recommendation to all the possible scenarios if you think it's reasonable.

rmfitzpatrick: Right, the issue was that `$$$$` needed to be used. Saying that `$$` should be replaced with `$` in accurate unless there were four signs.  If a user has `$$something` I believe you are telling them to replace it with `$something` and that will evaluate to an empty string.

dmitryax: > If a user has $$something I believe you are telling them to replace it with $something and that will evaluate to an empty string.

If user has `$$something` they already have been getting an empty string because it's expanded twice, so asking then to change it now to `$something` will give the same result.

rmfitzpatrick (approved):

PR #1145, opened 2022-01-21
Adopt windows perf counter receiver

rrupapara:
Adding the windowsPerfCounter receiver as a component.

Made required changes in the testutils to run Windows test case

Skipreaper change in testutils:
While running the integration test as a docker container, the Reaper container is started and it mounts /var/run/docker.sock so the container can communicate with the Docker daemon on the host. However, if you are running Docker on a system that is not Linux, this may not be the case (Windows, for example). While running it on windows the test fails and throws “creating reaper failed” error. The workaround is to set the skipReaper variable to true which will skip creating the Reaper container and works fine at the moment. 
Reference link - https://github.com/testcontainers/testcontainers-go/issues/156

The test case in the pipeline would run with subprocess only as to run with docker it needs to have following items which seem to be not possible to configure in the pipeline:

- `Docker Desktop for Windows` installed
- `Windows Subsystem for Linux` enabled

All the test cases have been passed


atoulme:
Closed as inactive. Feel free to reopen if this PR is still being worked on.

PR #1151, opened 2022-01-24
Enable integration-test workflow for PRs

jcheng-splunk:


.github/workflows/integration-test.yml
```diff
@@ -4,6 +4,20 @@ on:
...
+      - 'go.mod'
+      - 'go.sum'
+      - 'cmd/otelcol/**'
+      - 'internal/**'
+      - '!internal/buildscripts/packaging/choco/**'
+      - '!internal/buildscripts/packaging/fpm/**'
+      - '!internal/buildscripts/packaging/installer/**'
+      - '!internal/buildscripts/packaging/msi/**'
+      - '!internal/buildscripts/packaging/release/**'
+      - '!internal/buildscripts/packaging/tests/**'
```
rmfitzpatrick, line 19: Would a more general `'!internal/buildscripts/**'` be wanted? I'm not sure if `update-deps` is worth running these for instance.

jcheng-splunk: Unfortunately, I think we need to include `internal/buildscripts/packaging/collect-libs.sh` and `internal/buildscripts/packaging/smart-agent-release.txt` for the docker image build.  We can move these files to another dir and update the build scripts in a future PR?

rmfitzpatrick: More a question than a suggestion, but if we find this runs unnecessarily we can update as you suggest.

rmfitzpatrick (approved):
🤞

.github/workflows/integration-test.yml
jcheng-splunk: Maybe we can just use:
```suggestion
      - '!internal/buildscripts/*'
      - '!internal/buildscripts/packaging/*/**'
```
This should ignore everything currently in `internal/buildscripts/` except those 2 files.

PR #1154, opened 2022-01-25
Add CI job to build libsplunk.so

jcheng-splunk:
- Initial CI job to build `libsplunk.so` for amd64 and arm64
- Add basic dockerfile with build dependencies
- Add `dist` makefile target to build docker image and test/compile for `$ARCH`


pmcollins (approved):
Very cool. Thanks for doing this.

PR #1174, opened 2022-01-27
Run windows tests on 2019 and 2022

jcheng-splunk:
- For https://github.com/actions/virtual-environments/issues/4856: Pin the docker and choco build jobs to `windows-2019` since that is what we use for release builds from gitlab
- Update the msi, choco, and windows test jobs to run for both `windows-2019` and `windows-2022`
- Update choco test job to run with and without fluentd
- Move powershell test steps to scripts

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -315,10 +315,11 @@ jobs:
 
   windows-msi-validation:
     name: windows-msi-validation
-    runs-on: windows-latest
+    runs-on: ${{ matrix.OS }}
     needs: [windows-msi]
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


dmitryax (approved):
LGTM. 

The changes are independent, right? 

I think going forward we should try to submit separate PRs in these cases. It would make PR review process much easier and reduces chance of overlooking some issues.

jcheng-splunk:
> LGTM.
> 
> The changes are independent, right?
> 
> I think going forward we should try to submit separate PRs in these cases. It would make PR review process much easier and reduce chance of overlooking some issues.

Thanks, makes sense. Will try to split it up next time, but just wanted to confirm that the installation and services are not affected on windows 2022.

PR #1181, opened 2022-01-31
Enabling shared pipeline for profiling by default

aunshc:
Since the pipeline for logs and profiling is the same, this is a PR to enable it for ec2/ecs and fargate by default to avoid any configuration changes to do so.  

Context: https://docs.google.com/document/d/1KABqbnRCi--JGc6ZHIADMeyccTlTrUJZnPVQZk7F4Cs/edit?disco=AAAATAk4C5Y

### DIFF
--- a/cmd/otelcol/config/collector/ecs_ec2_config.yaml
+++ b/cmd/otelcol/config/collector/ecs_ec2_config.yaml
@@ -153,11 +153,11 @@ service:
       exporters: [signalfx]
     logs:
       receivers: [otlp]
-#      processors:
-#        - memory_limiter
-#        - batch
-#        - resourcedetection
-#      #- resource/add_environment
+      processors:
+        - memory_limiter
+        - batch
+        - resourcedetection
+      #- resource/add_environment
       exporters: [splunk_hec]
 #      # Use instead when sending to gateway
 #      #exporters: [otlp]
--- a/cmd/otelcol/config/collector/ecs_ec2_config.yaml
+++ b/cmd/otelcol/config/collector/ecs_ec2_config.yaml
@@ -125,12 +125,12 @@ exporters:
     access_token: "${SPLUNK_ACCESS_TOKEN}"
     realm: "${SPLUNK_REALM}"
     correlation:
-#  # Logs
-#  splunk_hec:
-#    token: "${SPLUNK_HEC_TOKEN}"
-#    endpoint: "${SPLUNK_HEC_URL}"
-#    source: "otel"
-#    sourcetype: "otel"
+  # Logs
+  splunk_hec:
+    token: "${SPLUNK_HEC_TOKEN}"
+    endpoint: "${SPLUNK_HEC_URL}"
+    source: "otel"
+    sourcetype: "otel"
 
 service:
   extensions: [health_check, http_forwarder, zpages, memory_ballast]
@@ -151,15 +151,13 @@ service:
       receivers: [prometheus/internal]
       processors: [memory_limiter, batch, resourcedetection/internal]
       exporters: [signalfx]
-#    logs:
-#      receivers: [otlp]
+    logs:
+      receivers: [otlp]
 #      processors:
 #        - memory_limiter
 #        - batch
 #        - resourcedetection
 #      #- resource/add_environment
-#      exporters: [splunk_hec]
+      exporters: [splunk_hec]
 #      # Use instead when sending to gateway
 #      #exporters: [otlp]
-
-
--- a/cmd/otelcol/config/collector/ecs_ec2_config.yaml
+++ b/cmd/otelcol/config/collector/ecs_ec2_config.yaml
@@ -125,7 +125,7 @@ exporters:
     access_token: "${SPLUNK_ACCESS_TOKEN}"
     realm: "${SPLUNK_REALM}"
     correlation:
-  # Logs
+  # Logs + Profiling
   splunk_hec:
     token: "${SPLUNK_HEC_TOKEN}"
     endpoint: "${SPLUNK_HEC_URL}"
@@ -159,5 +159,5 @@ service:
         - resourcedetection
       #- resource/add_environment
       exporters: [splunk_hec]
-#      # Use instead when sending to gateway
-#      #exporters: [otlp]
+      # Use instead when sending to gateway
+      #exporters: [otlp]
--- a/cmd/otelcol/config/collector/fargate_config.yaml
+++ b/cmd/otelcol/config/collector/fargate_config.yaml
@@ -127,10 +127,10 @@ service:
       exporters: [signalfx]
     logs:
       receivers: [otlp]
-#      processors:
-#        - memory_limiter
-#        - batch
-#        - resourcedetection
+      processors:
+        - memory_limiter
+        - batch
+        - resourcedetection
 #      #- resource/add_environment
       exporters: [splunk_hec]
 #      # Use instead when sending to gateway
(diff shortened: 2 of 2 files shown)


cmd/otelcol/config/collector/ecs_ec2_config.yaml
```diff
@@ -151,15 +151,13 @@ service:
...
       processors: [memory_limiter, batch, resourcedetection/internal]
       exporters: [signalfx]
-#    logs:
-#      receivers: [otlp]
+    logs:
+      receivers: [otlp]
 #      processors:
 #        - memory_limiter
 #        - batch
 #        - resourcedetection
```
dmitryax, line 159: We also need to enable these processors, they are enabled by default for any other data types

aunshc: Got it, makes sense. Just updated the PR with that change.

cmd/otelcol/config/collector/fargate_config.yaml
```diff
@@ -103,12 +103,12 @@ exporters:
     access_token: "${SPLUNK_ACCESS_TOKEN}"
     realm: "${SPLUNK_REALM}"
     correlation:
-#  # Logs
-#  splunk_hec:
-#    token: "${SPLUNK_HEC_TOKEN}"
-#    endpoint: "${SPLUNK_HEC_URL}"
-#    source: "otel"
-#    sourcetype: "otel"
+  # Logs
```
dmitryax, line 106:
```suggestion
  # Logs + Profiling
```

cmd/otelcol/config/collector/ecs_ec2_config.yaml
```diff
@@ -125,12 +125,12 @@ exporters:
     access_token: "${SPLUNK_ACCESS_TOKEN}"
     realm: "${SPLUNK_REALM}"
     correlation:
-#  # Logs
-#  splunk_hec:
-#    token: "${SPLUNK_HEC_TOKEN}"
-#    endpoint: "${SPLUNK_HEC_URL}"
-#    source: "otel"
-#    sourcetype: "otel"
+  # Logs
```
dmitryax, line 128:
```suggestion
  # Logs + Profiling
```

cmd/otelcol/config/collector/ecs_ec2_config.yaml
```diff
@@ -151,15 +151,13 @@ service:
...
+    logs:
+      receivers: [otlp]
+      processors:
+        - memory_limiter
+        - batch
+        - resourcedetection
+      #- resource/add_environment
+      exporters: [splunk_hec]
 #      # Use instead when sending to gateway
 #      #exporters: [otlp]
```
dmitryax, line 163:
```suggestion
      # Use instead when sending to gateway
      #exporters: [otlp]
```

cmd/otelcol/config/collector/fargate_config.yaml
```diff
@@ -125,13 +125,13 @@ service:
...
+      receivers: [otlp]
+      processors:
+        - memory_limiter
+        - batch
+        - resourcedetection
 #      #- resource/add_environment
-#      exporters: [splunk_hec]
+      exporters: [splunk_hec]
 #      # Use instead when sending to gateway
 #      #exporters: [otlp]
```
dmitryax, line 137:
```suggestion
      # Use instead when sending to gateway
      #exporters: [otlp]
```

cmd/otelcol/config/collector/fargate_config.yaml
```diff
@@ -125,13 +125,13 @@ service:
...
-#        - memory_limiter
-#        - batch
-#        - resourcedetection
+    logs:
+      receivers: [otlp]
+      processors:
+        - memory_limiter
+        - batch
+        - resourcedetection
 #      #- resource/add_environment
```
dmitryax, line 134:
```suggestion
      #- resource/add_environment
```

dmitryax (commented):
Few nits otherwise LGTM

cmd/otelcol/config/collector/ecs_ec2_config.yaml
aunshc: Should we add this comment above the pipelines configuration instead instead of just the exporter?

aunshc: Should we add this comment above the pipelines configuration instead of just the exporter?

dmitryax: It's just for consistency with comments for other exporters. We can move all of them to pipelines as well, maybe as follow up PR?

aunshc: Sounds good

cmd/otelcol/config/collector/ecs_ec2_config.yaml
```diff
@@ -125,12 +125,12 @@ exporters:
...
-#  # Logs
-#  splunk_hec:
-#    token: "${SPLUNK_HEC_TOKEN}"
-#    endpoint: "${SPLUNK_HEC_URL}"
-#    source: "otel"
-#    sourcetype: "otel"
+  # Logs + Profiling
+  splunk_hec:
+    token: "${SPLUNK_HEC_TOKEN}"
+    endpoint: "${SPLUNK_HEC_URL}"
```
rmfitzpatrick, line 131: these env vars aren't documented in the deployment instructions, and if they are now intended to be used for profiling we probably want to have a templated endpoint like we do w/ the sapm exporter.

dmitryax: > these env vars aren't documented in the deployment instructions, and if they are now intended to be used for profiling we probably want to have a templated endpoint like we do w/ the sapm exporter and use SPLUNK_ACCESS_TOKEN*

We [compose these env vars](https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/main.go#L365-L388) from `SPLUNK_ACCESS_TOKEN` and `SPLUNK_REALM` by default. Using these env vars leaves an option for users to send logs to another Splunk endpoint instead of o11y without providing a custom config. Also it's consistent with other configs that we have

> imo we should have a dedicated splunk_hec/profiling exporter.

The profiling data shares the same pipeline including otlp receiver. Profiling libraries use the same OTLP port by default. So we cannot split it and create another exporter.

rmfitzpatrick: I think the autosetting pass is ok where users aren't setting these env vars for splunk cloud usage. Otherwise wouldn't profiling data be sent directly to Splunk Cloud?

dmitryax: > Otherwise wouldn't profiling data be sent directly to Splunk Cloud?
Yes they would, and Splunk Cloud backend can digest and show them is some way, not as good as Splunk O11y tho.

Use case when user wants logs flowing to Splunk Cloud and Profiling data to Splunk O11y will require separate pipelines and changing logs and profiling libraries to send to different ports. So it using `SPLUNK_ACCESS_TOKEN` instead of `SPLUNK_HEC_TOKEN` doesn't help. 

I think we should use `SPLUNK_HEC_TOKEN` at least for consistency with other default configs.

dmitryax: @rmfitzpatrick please let us know if you're ok to merge it

PR #1191, opened 2022-02-02
Initial implementation of Databricks receiver

pmcollins:


rmfitzpatrick:
Is there an api server container that can be used for an integration test? didn't appear their standard runtime has is but I know nothing about databricks

pmcollins:
> Is there an api server container that can be used for an integration test? didn't appear their standard runtime has it* but I know nothing about databricks

Looks like Databricks is only available as a cloud service -- I'm not seeing a way to run it locally.

internal/receiver/databricksreceiver/scraper.go
```diff
@@ -0,0 +1,64 @@
...
+	jmp          metricsProvider
+	instanceName string
+}
+
+func (s scraper) scrape(_ context.Context) (pdata.Metrics, error) {
+	out := pdata.NewMetrics()
+	rms := out.ResourceMetrics()
+	rm := rms.AppendEmpty()
+	rm.Resource().Attributes().Insert(
+		"instance_name",
```
rmfitzpatrick, line 38: I think the otel semantic conventions (mostly from the tracing side) would using something like `spark.instance.name` instead.

pmcollins: Ah good to know. Will fix.

internal/receiver/databricksreceiver/README.md
```diff
@@ -0,0 +1,33 @@
...
+Supported pipeline types: `metrics`
+
+> :construction: This receiver is in **ALPHA**. Behavior, configuration fields, and metric data model are subject to change.
+
+## Configuration
+
+The following fields are required:
+
+- `instance_name`: A string representing the name of the instance. This value gets set as an `instance_name` resource attribute.
+- `base_url`: The protocol (http or https), hostname, and port for the Databricks API, without a trailing slash.
```
rmfitzpatrick, line 16: There's been some effort in adopting [`HTTPClientSettings`](https://github.com/open-telemetry/opentelemetry-collector/blob/a7ef0790a0a59aae6f1cd1508b146f42ea8a24c0/config/confighttp/confighttp.go#L38) in different components, and it has a `ToClient()` method that can be used over the default.  This way `endpoint` would also be available for receiver creator instances to use.

internal/receiver/databricksreceiver/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Supported pipeline types: `metrics`
+
+> :construction: This receiver is in **ALPHA**. Behavior, configuration fields, and metric data model are subject to change.
+
+## Configuration
+
+The following fields are required:
+
+- `instance_name`: A string representing the name of the instance. This value gets set as an `instance_name` resource attribute.
```
rmfitzpatrick, line 15: my bad, meant something like 

```suggestion
- `instance_name`: A string representing the name of the instance. This value gets set as an `databricks.instance.name` resource attribute.
```

internal/receiver/databricksreceiver/json_types.go
```diff
@@ -0,0 +1,152 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package databricksreceiver
+
+// This file contains structs into which responses from the databricks API are
+// unmarshalled. There are two top-level types: jobsList and jobRun.
+
+// jobsList is a top level type
+type jobsList struct {
```
rmfitzpatrick, line 21: Would you mind linking to the api docs for these response structures?

internal/receiver/databricksreceiver/paginator.go
```diff
@@ -0,0 +1,83 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package databricksreceiver
+
+import "fmt"
+
+// paginator is extracted from apiPaginator for swapping out in unit tests
+type paginator interface {
```
rmfitzpatrick, line 20: nit, I think `paginator` is a bit of an implementation detail and wonder if `databricksClient` is preferable considering it's used by the metrics providers.

internal/receiver/databricksreceiver/paginator.go
```diff
@@ -0,0 +1,83 @@
...
+	jobsList() (out []job, err error)
```
rmfitzpatrick, line 21: since this is an outlier:
```suggestion
	jobs() (out []job, err error)
```

internal/receiver/databricksreceiver/paginator.go
```diff
@@ -0,0 +1,83 @@
...
+// combines the returned objects into one array.
+type apiPaginator struct {
+	unmarshaller unmarshaller
+	limit        int
+}
+
+func newPaginator(api databricksAPI) apiPaginator {
+	return apiPaginator{
+		unmarshaller: unmarshaller{api: api},
+		limit:        25, // max is 25
```
rmfitzpatrick, line 36: Do you imaging this ever needing to change and being worth a config option?

rmfitzpatrick:
Is `metadata.yaml` usage an option/worth the time? Could be a later pr for sure.

internal/receiver/databricksreceiver/factory.go
```diff
@@ -0,0 +1,85 @@
...
+	scraperhelper.ScraperControllerSettings `mapstructure:",squash"`
+}
+
+func createDefaultConfig() config.Receiver {
+	return &Config{
+		ScraperControllerSettings: scraperhelper.ScraperControllerSettings{
+			ReceiverSettings:   config.NewReceiverSettings(config.NewComponentID(typeStr)),
+			CollectionInterval: time.Second * 30,
+		},
+	}
```
mstumpfx, line 52:
```suggestion
	scs := scraperhelper.DefaultScraperControllerSettings(typeStr)
	scs.CollectionInterval = 30 * time.Second
	
	return &Config{
		ScraperControllerSettings: scs,
	}
```
This could maybe help if scraper settings change in the future. Also, is the default of `30s` something specific to databricks or just an arbitrary number? I would probably leave as default for consistency with other scrapers otherwise.

mstumpfx (commented):
+1 to adding metadata.yaml and generated metrics. Otherwise LGTM

internal/receiver/databricksreceiver/factory.go
pmcollins: Thanks, fixed. I set it to 30s because that's double the highest job frequency of 1/minute -- added a comment.

internal/receiver/databricksreceiver/json_types.go
pmcollins: Sure -- I've added a link to the API page which has pointers to the OpenAPI spec.

internal/receiver/databricksreceiver/paginator.go
pmcollins: Yeah, I see what you're saying. The thing is that if we call it databricksClient, I would probably get confused between it, apiClient, and http.Client. I kind of like paginator because it tells you exactly what it does :)

internal/receiver/databricksreceiver/paginator.go
pmcollins: Oh yeah -- I was just using the names of the api methods. Fixed.

internal/receiver/databricksreceiver/paginator.go
pmcollins: I was thinking it wouldn't since the max is only 25, but I've made it configurable anyway since it does help with testing.

pmcollins:
Will do the metadata.yaml stuff in the next PR.

internal/receiver/databricksreceiver/README.md
pmcollins: Done.

internal/receiver/databricksreceiver/paginator.go
rmfitzpatrick: I had similar thoughts where authClient < apiClient < paginator and "paginator" doesn't seem to reflect its role as the actual client handle. I'd think it's presumed a databricks client would correctly deal w/ paginated results.

internal/receiver/databricksreceiver/databricks_api.go
```diff
@@ -0,0 +1,61 @@
...
+func newAPIClient(endpoint string, tok string, httpClient *http.Client) databricksAPI {
+	return &apiClient{authClient{
+		httpClient: httpClient,
+		endpoint:   endpoint,
+		tok:        tok,
+	}}
+}
+
+func (c apiClient) jobsList(limit int, offset int) (out []byte, err error) {
+	return c.authClient.get(fmt.Sprintf(jobsListPath, limit, offset))
```
rmfitzpatrick, line 52: Would debug logs for all requests be helpful in diagnosing potential performance issues?

internal/receiver/databricksreceiver/paginator.go
pmcollins: Makes sense, fixed. Wasn't sure what to name the interfaces so I just appended `Interface` which at least makes things pretty clear 😄

rmfitzpatrick: Thanks, this is much easier for me to follow.

internal/receiver/databricksreceiver/metrics_provider.go
```diff
@@ -0,0 +1,104 @@
...
+		jobPt.Attributes().Insert("job_id", jobIDAttr)
+		for _, task := range j.Settings.Tasks {
+			taskPt := taskPts.AppendEmpty()
+			taskPt.SetIntVal(pauseStatus)
+			taskAttrs := taskPt.Attributes()
+			taskAttrs.Insert("job_id", jobIDAttr)
+			taskAttrs.Insert("task_id", pdata.NewAttributeValueString(task.TaskKey))
+		}
+	}
+	return jobIDs, nil
```
dloucasfx, line 67: Are you planning to add the task and job properties in a future release?
Example: https://github.com/dloucasfx/signalfx-agent/commit/287456eabe2ed01ec351a878d1ceec16b1a0405d#diff-84aefb0f500a3fb8b2b091d108e98045485ce47b27c662fb3932b8161f716f4bR147-R162

internal/receiver/databricksreceiver/databricks_client.go
```diff
@@ -0,0 +1,83 @@
...
+func (c databricksClient) completedJobRuns(jobID int, prevStartTime int64) (out []jobRun, err error) {
+	hasMore := true
+	for i := 0; hasMore; i++ {
+		resp, err := c.unmarshaller.completedJobRuns(jobID, c.limit, c.limit*i)
+		if err != nil {
+			return nil, fmt.Errorf("databricksClient.completedJobRuns(): %w", err)
+		}
+		out = append(out, resp.Runs...)
+		if prevStartTime == 0 || resp.Runs == nil || resp.Runs[len(resp.Runs)-1].StartTime < prevStartTime {
+			// Don't do another api request if this is the first time through (time == 0) or
```
dloucasfx, line 75: I might have overlooked it, but what happens on the first run, when the `prevStartTime` is 0?

dloucasfx (changes requested):

internal/receiver/databricksreceiver/metrics_provider.go
pmcollins: We don't have a concept of properties in otel so I punted on this. Do you think task type should be an attribute (which becomes a dimension)?

internal/receiver/databricksreceiver/databricks_client.go
pmcollins: On the first run, we assume that the latest run we receive is not current so we discard it. I have a comment here about it: https://github.com/signalfx/splunk-otel-collector/pull/1191/files#diff-da54499d3e4797f461511ae53fa6259cc2fc98ee03b43edc2a2b7461ed1fa222R38
Otherwise, we'd have to backfill.

internal/receiver/databricksreceiver/metrics_provider.go
dloucasfx: oh, I was not aware that properties are not implemented in the SFX exporter, well, that's a bummer.
I think task type don't change frequently, so yes, we can add them as dimensions

rmfitzpatrick: ftr these are available in otel https://github.com/signalfx/splunk-otel-collector/blob/1d289f68e9896eb58fdea3ef290e83a50774a5b0/internal/receiver/smartagentreceiver/dimension.go#L28. Considering the size of this PR already my 2c would be to add metadata exporter adoption in a later pass.

pmcollins: Ok, sounds good, I'll add task type as an attribute. BTW I meant that there isn't a concept of properties in Otel (receivers are exporter agnostic), but there is some support for properties in the SFx exporter, but it's using a metadata exporter which is apparently experimental, and I'm not quite sure it's the right approach here. Can always do this later if we decide to.

dloucasfx (approved):

PR #1216, opened 2022-02-08
Update core/contrib deps to v0.44.0

jcheng-splunk:


dmitryax (approved):
Thanks for taking care of it

PR #1229, opened 2022-02-10
Add ansible.windows as a galaxy dependency

jcheng-splunk:


deployments/ansible/molecule/default/Dockerfile.j2
```diff
@@ -1,9 +1,5 @@
 {% if item.image not in ["opensuse12", "opensuse15"] %}
 FROM geerlingguy/docker-{{ item.image }}-ansible:latest
-{% if item.image == "centos8" %}
-RUN sed -i 's|\$releasever|8-stream|g' /etc/yum.repos.d/CentOS*.repo
-RUN dnf install -y --allowerasing centos-stream-release
-{% endif %}
 {% else %}
```
dloucasfx, line 3: Was this change on purpose?

dloucasfx (changes requested):

deployments/ansible/molecule/default/Dockerfile.j2
jcheng-splunk: It was a workaround until the [centos8 image](https://hub.docker.com/r/geerlingguy/docker-centos8-ansible) was updated for the new `8-stream` repo.  Otherwise the docker build would fail.

PR #1236, opened 2022-02-11
Add instrumentation deb/rpm to gitlab release

jeffreyc-splunk:
- Build `libsplunk.so` for amd64 and arm64
- Build `splunk-otel-auto-instrumentation` deb/rpm packages for amd64 and arm64
- Sign and release deb/rpm packages to `otel-collector` repos in artifactory
- Include `libsplunk_amd64.so`, `libsplunk_arm64.so`, and deb/rpm packages in github release assets
- Updated `dist` make target since we can't mount writable volumes in gitlab runner containers

PR #1255, opened 2022-02-16
zc: read jar file path and service name from a config file

pmcollins:
This change provides incremental functionality for the .so to read a config file at a known location at startup. To start, the config file just contains the path to the auto-instrumentation jar and the service.name. In a later PR, the service name string in the config file will be a pattern/template into which the arguments to the executable (most likely just the jar name) will be inserted.


instrumentation/src/splunk.h
```diff
@@ -3,15 +3,22 @@
 
 #include <stdbool.h>
 #include "logger.h"
+#include "config.h"
 
-static char *const DISABLE_ENV_VAR_NAME = "DISABLE_SPLUNK_AUTOINSTRUMENTATION";
+static char *const conf_file = "/usr/lib/splunk-instrumentation/splunk.conf";
```
jcheng-splunk, line 8: Do we need to include a default or sample `splunk.conf` in the deb/rpm packages?

Also, we should include some documentation about this file and the supported keys/values.

pmcollins: Yeah, we'll likely want to include a config file in the packages and for sure document everything -- didn't want to do that in this PR as I wanted to wait until how this is going to work gets firmed up. This change is incremental to keep PRs small and work easier to track on the way towards the full solution, which will have the service name set properly by reading the CLI args, and possibly have the install script set the service name pattern. Do you think we should put this into a feature branch instead of main?

jcheng-splunk: I think it should be fine to merge this to main since we're not releasing this yet.  I'm working on adding this to the release pipeline (#1236), but will wait until until we are ready.

emaderer (approved):

Issue #1289, opened 2022-02-23
Ansible Template Don't Expand Variables Properly


fvaliquette:
Ansible template version: 0.5.0

Impact: Traces and Logs cannot be sent to Splunk o11y (traces work)

Tested work around: manually edit /etc/otel/collector/splunk-otel-collector.conf to set proper values and restart agent

Details:
After installation, we can see variable names that should have been expended in /etc/otel/collector/splunk-otel-collector.conf

What we got:
SPLUNK_INGEST_URL=https://ingest.us1.signalfx.com
SPLUNK_TRACE_URL=${SPLUNK_INGEST_URL}/v2/trace
SPLUNK_HEC_URL=${SPLUNK_INGEST_URL}/v1/log

What is expected:
SPLUNK_INGEST_URL=https://ingest.us1.signalfx.com
SPLUNK_TRACE_URL=https://ingest.us1.signalfx.com/v2/trace
SPLUNK_HEC_URL=https://ingest.us1.signalfx.com/v1/log

jcheng-splunk:
@fvaliquette Thanks for reporting this.  We have released version 0.6.0 to [galaxy](https://galaxy.ansible.com/signalfx/splunk_otel_collector) to hopefully address this issue.  Please test it out and let us know if this is resolved.

fvaliquette:
Thank you. We tested the new version and it now works.

Francois Valiquette

he/him/his what's this?
<https://hrc-prod-requests.s3-us-west-2.amazonaws.com/files/assets/resources/TalkingAboutPronouns_onesheet_FINAL.pdf?mtime=20200713142932&focal=none>


Professional Services Consultant  |  Professional Services, East

Splunk Inc.

+1 (917) 254-6256  |  ***@***.***

New York | AMER | APAC | EMEA

<https://www.splunk.com/en_us/enterprise-data-platform.html>

This message is intended only for the personal, confidential, and
authorized use of the recipient(s) named above. If you are not that person,
you are not authorized to review, use, copy, forward, distribute or
otherwise disclose the information contained in the message.

PR #1290, opened 2022-02-23
Fix SPLUNK_TRACE_URL and SPLUNK_HEC_URL env vars

jcheng-splunk:
For #1289


dmitryax (approved):
Thanks for fixing this!

PR #1294, opened 2022-02-24
update to Go 1.17.7

emaderer:


examples/splunk-hec-traces/tracing/go.mod
```diff
@@ -1,6 +1,6 @@
 module app
 
-go 1.17
+go 1.17.7
```
jcheng-splunk, line 3: I think this needs to stay at `1.17`.

examples/splunk-hec-traces/tracing/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.17-stretch
+FROM golang:1.17.7-stretch
```
jcheng-splunk, line 1: I think we can leave this as `1.17` since it gets re-tagged with the latest patch version.

examples/splunk-hec-traces/tracing/go.mod
emaderer: thanks! my bad

dloucasfx (approved):

PR #1307, opened 2022-03-01
Override host.name attribute by default

dmitryax:
Instrumentation libraries running in container environments cannot always correctly detect host attributes while OTel Collector Agent has all required access to underlying host and can be considered as a source of truth for all host and cloud metadata. This change reconfigures Resource Detection processor on the agent to override all host and cloud attributes by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### 🧰 Bug fixes 🧰
+
+- As a bug fix for hosts number miscalculation in Splunk Observability Cloud, Splunk OpenTelemetry Collector running in 
+  agent mode now is configured to override `host.name` attribute of all signals sent from instrumentation libraries by 
+  default (#1307)
+
 ## v0.45.0
 
 This Splunk OpenTelemetry Collector release includes changes from the [opentelemetry-collector v0.45.0](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.45.0) and the [opentelemetry-collector-contrib v0.45.1](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.45.1) releases.
--- a/cmd/otelcol/config/collector/agent_config.yaml
+++ b/cmd/otelcol/config/collector/agent_config.yaml
@@ -104,13 +104,10 @@ processors:
 
   # Detect if the collector is running on a cloud system, which is important for creating unique cloud provider dimensions.
   # Detector order is important: the `system` detector goes last so it can't preclude cloud detectors from setting host/os info.
+  # Resource detection processor is configured to override all host and cloud attributes because instrumentation
+  # libraries can send wrong values from container environments.
   # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor#ordering
   resourcedetection:
-    detectors: [gce, ecs, ec2, azure, system]
-    override: false
-
-  # Same as above but overrides resource attributes set by receivers
-  resourcedetection/internal:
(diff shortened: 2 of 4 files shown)


mstumpfx:
I think there's some cases where a receiver may set a `host.name` attribute and we should honor it, e.g. a receiver is monitoring an external host. I know the prometheus receiver sets this attribute (it also sets an `instance` attribute to the same value though), I'm not aware of others and not sure if there's any guidelines around setting or not setting it in otel. 

My point is this can also cause billing issues, but in the opposite direction of under counting hosts if we're not careful. I think the code changes are fine but we should talk about how to document this. Besides docs, perhaps a separate pipeline again with `override:false` for those types of receivers?

mstumpfx (approved):

dmitryax:
@mstumpfx thanks for bringing this. We need to decide before merging this PR if we want to do this uniformly for metrics and traces or just for traces. cc @emaderer @aunshc

dmitryax:
I think if prometheus is configured in agent mode it supposed to scrape metrics from the the same machine. Otherwise, if all the agents are configured to scrape metrics from different machines, the metrics will be duplicated, and seems like a misuse to me. So I would prefer to override all the telemetry uniformly for consistency. Otherwise we can mess up the correlation.

emaderer:
> @mstumpfx thanks for bringing this. We need to decide before merging this PR if we want to do this uniformly for metrics and traces or just for traces. cc @emaderer @aunshc

I think that it's okay to have this just traces for now

mstumpfx:
@dmitryax 
> I think if prometheus is configured in agent mode it supposed to scrape metrics from the the same machine. Otherwise, if all the agents are configured to scrape metrics from different machines, the metrics will be duplicated, and seems like a misuse to me. So I would prefer to override all the telemetry uniformly for consistency. Otherwise we can mess up the correlation.

There's nothing stopping someone from scraping multiple hosts - [scrape-config](https://github.com/prometheus/prometheus/blob/v2.28.1/docs/configuration/configuration.md#scrape_config) supports scraping any machine. It seems like a valid use case. There should be enough cardinality in the dimensions to distinguish different machines (`instance`, `endpoint`, `host.name`). For billing purposes, in my opinion, if 1 collector is monitoring N number of hosts, we should bill for N number of hosts, not per collector. I think whatever we decide we can update docs and explain why a separate pipeline may be needed for these cases.

---

@emaderer 
> What about receivers like MongoDB/Atlas that monitor a saas product? is that going to create issues?

Yeah same issue - [you nailed the example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/66f91ea79034c41b7c6864352413ecc9cb2f3294/receiver/mongodbatlasreceiver/receiver.go#L101). `mongodbatlas` receiver does set the `host.name` to the host is it monitoring. I'm not aware of any other beside that one and `prometheus` right now.

emaderer (approved):

dmitryax:
> There's nothing stopping someone from scraping multiple hosts - [scrape-config](https://github.com/prometheus/prometheus/blob/v2.28.1/docs/configuration/configuration.md#scrape_config) supports scraping any machine. It seems like a valid use case. There should be enough cardinality in the dimensions to distinguish different machines (instance, endpoint, host.name). For billing purposes, in my opinion, if 1 collector is monitoring N number of hosts, we should bill for N number of hosts, not per collector. I think whatever we decide we can update docs and explain why a separate pipeline may be needed for these cases.

I understand, my idea is that this would be a wrong usage of the agent default config. I believe in these cases gateway mode must be used. I think we should clearly document that.

dmitryax:
Looks like nomad test is always failing. submitted an issue: https://github.com/signalfx/splunk-otel-collector/issues/1314

PR #1354, opened 2022-03-15
Add installer script support for debian 11 and remove debian 8

jcheng-splunk:


crobert-1 (commented):
Is this change explicitly limited to the installer scripts and tests?

I'm seeing quite a few more `jessie` references in the repository, so I'm not sure if those should be removed now, or later.

jeffreyc-splunk:
> Is this change explicitly limited to the installer scripts and tests?
> 
> I'm seeing quite a few more `jessie` references in the repository, so I'm not sure if those should be removed now, or later.

Yep, I am planning on removing those in separate PRs.

crobert-1 (approved):

PR #1368, opened 2022-03-17
Upgrade to latest instead of main for deps test

jcheng-splunk:
Upgrading core/contrib deps to `main` has been too unstable, so just try testing with the latest release instead (only if there are changes).

Sample run: https://github.com/signalfx/splunk-otel-collector/runs/5605814999?check_suite_focus=true

### DIFF
--- a/.github/workflows/deps-test.yml
+++ b/.github/workflows/deps-test.yml
@@ -1,7 +1,13 @@
 name: Dependency Test
 on:
+  pull_request:
+    paths:
+      - .github/workflows/deps-test.yml
   schedule:
     - cron: '0 12 * * 1-5'
+
+env:
+  OTEL_VERSION: latest
 
 jobs:
   deps-test:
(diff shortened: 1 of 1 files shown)


jvoravong (approved):

.github/workflows/deps-test.yml
```diff
@@ -26,39 +34,50 @@ jobs:
...
-        with:
-          name: gomods
-          path: |
-            **/go.mod
-            !**/examples/**/go.mod
+          run_url="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+          echo "[$run_url]($run_url)" > deps-test.md
+          if [ -f /tmp/output.txt ]; then
+            echo "\`\`\`" >> deps-test.md
+            tail -n50 /tmp/output.txt >> deps-test.md
```
rmfitzpatrick, line 75: we've not truncated before so was the rest noise?

rmfitzpatrick (approved):

.github/workflows/deps-test.yml
jcheng-splunk: The message in the issue could get pretty long, so I just thought truncating would be best.  Also, the link to the run will be included in the issue, so we can see the full output there.

PR #1376, opened 2022-03-18
Include the TCP log receiver in our distribution

emaderer:
Include the TCP log receiver in our distribution


jvoravong:
Most of this looks good. components_test.go needs to get updated too, there is at least one test failure.

jvoravong (approved):
Please also add an entry to the CHANGELOG.md.

PR #1379, opened 2022-03-21
Update core version during build

emaderer:
During build we need to also update the core version file, otherwise, the internal metrics show "latest" instead of the real version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,8 +37,8 @@ LINT=golangci-lint
 MISSPELL=misspell -error
 MISSPELL_CORRECTION=misspell -w
 
-BUILD_INFO_IMPORT_PATH=go.opentelemetry.io/collector/internal/version
-BUILD_INFO_IMPORT_PATH_CORE=github.com/open-telemetry/opentelemetry-collector/internal/version
+BUILD_INFO_IMPORT_PATH=github.com/signalfx/splunk-otel-collector/internal/version
+BUILD_INFO_IMPORT_PATH_CORE=go.opentelemetry.io/collector/internal/version
 VERSION=$(shell git describe --match "v[0-9]*" HEAD)
 BUILD_X1=-X $(BUILD_INFO_IMPORT_PATH).Version=$(VERSION)
 BUILD_X2=-X $(BUILD_INFO_IMPORT_PATH_CORE).Version=$(VERSION)
--- a/Makefile
+++ b/Makefile
@@ -37,10 +37,12 @@ LINT=golangci-lint
 MISSPELL=misspell -error
 MISSPELL_CORRECTION=misspell -w
 
-BUILD_INFO_IMPORT_PATH=github.com/signalfx/splunk-otel-collector/internal/version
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -37,10 +37,12 @@ LINT=golangci-lint
 MISSPELL=misspell -error
 MISSPELL_CORRECTION=misspell -w
 
-BUILD_INFO_IMPORT_PATH=github.com/signalfx/splunk-otel-collector/internal/version
+BUILD_INFO_IMPORT_PATH=go.opentelemetry.io/collector/internal/version
+BUILD_INFO_IMPORT_PATH_CORE=github.com/open-telemetry/opentelemetry-collector/internal/version
```
dmitryax, line 41: Does it really make it work?

go.opentelemetry.io/collector is an alias for github.com/open-telemetry/opentelemetry-collector

I think it currently doesn't work because of `internal` part

dmitryax (approved):

dmitryax:
Approved by accident. Not sure this solution will solve the problem

emaderer:
> Approved by accident. Not sure this solution will solve the problem

I tested it, it works.

Makefile
emaderer: tested it, it works

Makefile
```diff
@@ -37,10 +37,12 @@ LINT=golangci-lint
 MISSPELL=misspell -error
 MISSPELL_CORRECTION=misspell -w
 
-BUILD_INFO_IMPORT_PATH=github.com/signalfx/splunk-otel-collector/internal/version
+BUILD_INFO_IMPORT_PATH=go.opentelemetry.io/collector/internal/version
```
jcheng-splunk, line 40: Should we keep `github.com/signalfx/splunk-otel-collector/internal/version`, otherwise how will the version get set for https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/main.go#L83?

dmitryax: Right. I think it should be 

```
BUILD_INFO_IMPORT_PATH=github.com/signalfx/splunk-otel-collector/internal/version
BUILD_INFO_IMPORT_PATH_CORE=go.opentelemetry.io/collector/internal/version
```

I didn't notice that it was replacement for splunk distro. Now it makes sense why it didn't work before

emaderer: > Right. I think it should be
> 
> ```
> BUILD_INFO_IMPORT_PATH=github.com/signalfx/splunk-otel-collector/internal/version
> BUILD_INFO_IMPORT_PATH_CORE=go.opentelemetry.io/collector/internal/version
> ```
> 
> I didn't notice that it was replacement for splunk distro. Now it makes sense why it didn't work before

Sorry, my bad, it was actually pointing to the version in our distro when I tested it but for the PR somehow I replaced that line, thanks for catching that!

jcheng-splunk (approved):

dmitryax (approved):

PR #1383, opened 2022-03-22
zc: add support for resource_attributes

pmcollins:


instrumentation/README.md
```diff
@@ -20,22 +20,39 @@ Once instrumented, Java executables send traces to a locally running
...
 #### OTEL_SERVICE_NAME
 
 This environment variable sets the service name for the soon-to-be running Java application, usually derived from the
 arguments of the Java executable (can be overridden via the config).
 
+This variable is set directly from the `service_name` attribute in the config.
+
 _Meta: link to docs about how service name is used and why it's required._
 
+#### OTEL_EXPORTER_OTLP_ENDPOINT
```
aunshc, line 43: Since we are providing this feature with the collector and the customer will go through the collector wizard, we need not provide the option of overriding the endpoint. The traces should be sent to the local collector endpoint itself, users cannot provide the ingest url for traces to be sent directly to the backend if they want zc autoinstrumentation.

pmcollins: Ok -- will delete.

instrumentation/install/instrumentation.conf
```diff
@@ -1,2 +1,3 @@
 # service_name=default.service
 java_agent_jar=/usr/lib/splunk-instrumentation/splunk-otel-javaagent.jar
+exporter_otlp_endpoint=http://localhost:4317
```
jcheng-splunk, line 3: Should we include `resource_attributes=deployment.environment=<some default value>` in the default conf file, commented or uncommented?

pmcollins: I was thinking this value would come from the installation script and be appended to this file at install time. Does that sound good? cc: @aunshc

jcheng-splunk: Makes sense to me.  If that's the case, should we have a default value for the installer script, maybe something like `prod`?  Or would it be a required argument?

pmcollins: hmm not sure. @aunshc do you have an opinion on this?

aunshc: I do believe we should stay consistent with the current Java APM wizard in the first iteration; where the environment field is optional and we should only pass it to the installer script if it is set by the customer.

jvoravong (approved):
These changes look good to me, but since I am newer you may want to get another approval from another team member.

PR #1385, opened 2022-03-22
Update SA event type to fix processlist

dloucasfx:
After this change https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/e794616815bd9a145503f87078a623d6e2908e2d the SA processlist monitor stopped working.

This PR uses the new event type attribute in the event to log conversion .
 
### DIFF
--- a/internal/receiver/smartagentreceiver/converter/event.go
+++ b/internal/receiver/smartagentreceiver/converter/event.go
@@ -27,14 +27,14 @@ const (
 	SFxEventCategoryKey = "com.splunk.signalfx.event_category"
 	// SFxEventPropertiesKey key for splunk event properties.
 	SFxEventPropertiesKey = "com.splunk.signalfx.event_properties"
+	// SFxEventType key for splunk event type
+	SFxEventType = "com.splunk.signalfx.event_type"
 )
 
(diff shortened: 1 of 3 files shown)


internal/receiver/smartagentreceiver/converter/event.go
```diff
@@ -58,6 +58,8 @@ func sfxEventToPDataLogs(event *event.Event, logger *zap.Logger) pdata.Logs {
 		attrs.InsertInt(SFxEventCategoryKey, int64(event.Category))
 	}
 
+	attrs.InsertString(SFxEventType, event.EventType)
```
tigrannajaryan, line 61: I think it will be useful to check `if event.EventType != "" {` to avoid creating an empty attribute if the EventType is not defined.

tigrannajaryan: Nit: also maybe change line 46 to `attrsCapacity := 2 + len(event.Dimensions)` to reserve one additional space for the new attribute (which likely is going to be present in the majority of cases).

tigrannajaryan (approved):

tigrannajaryan:
It does not let me merge it. Probably needs someone else's approval.

emaderer (approved):

PR #1393, opened 2022-03-23
gateway config: Remove signalfx exporter from traces pipeline

asuresh4:
Remove `signalfx` exporter from traces pipeline. `signalfx` exporter in the traces pipeline is used only for correlation when the Collector runs as an agent. It shouldn't be enabled for a Collector running as the gateway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


dmitryax (approved):

dmitryax:
@asuresh4 can you please add a Changelog entry?

PR #1404, opened 2022-03-25
Cloud Foundry Buildpack for Collector

crobert-1:
This is a deployment of the OpenTelemetry Collecter in the format of a
Cloud Foundry buildpack. This buildpack supplies the Collector to the
app it is used for. When the app is deployed it can run and configure
the Collector as a sidecar (as described in the README). This will allow
the Collector to observe the given app as well as the whole environment's
metrics.

### DIFF
--- a/deployments/cloudfoundry/README.md
+++ b/deployments/cloudfoundry/README.md
@@ -1,0 +1,6 @@
+# OpenTelemetry Collector Pivotal Cloud Foundry (PCF) Integrations
+
+### Cloud Foundry Buildpack
+
+This integration can be used to install and run the Collector as a sidecar to your app.
+In this configuration, the collector will run in the same container as the app.
--- a/deployments/cloudfoundry/README.md
+++ b/deployments/cloudfoundry/README.md
@@ -1,4 +1,4 @@
-# OpenTelemetry Collector Pivotal Cloud Foundry (PCF) Integrations
+# Splunk OpenTelemetry Collector Pivotal Cloud Foundry (PCF) Integrations
 
 ### Cloud Foundry Buildpack
 
--- a/deployments/cloudfoundry/buildpack/README.md
+++ b/deployments/cloudfoundry/buildpack/README.md
@@ -64,6 +64,11 @@ Required:
 - `UAA_USERNAME` - Name of the UAA user.
 - `UAA_PASSWORD` - Password for the UAA user.
 - `SPLUNK_ACCESS_TOKEN` - Your Splunk organization access token.
+- The Splunk Observability Suite requires an endpoint, so one of the two options below must be specified:
+    1) - `SPLUNK_INGEST_URL` - The ingest base URL for Splunk. This option takes precedence over SPLUNK_REALM.
+       - `SPLUNK_API_URL` - The API server base URL for Splunk. This option takes precedence over SPLUNK_REALM.
+    2) - `SPLUNK_REALM` - The Splunk realm in which your organization resides. Used to derive SPLUNK_INGEST_URL
+       and SPLUNK_API_URL.
 
 Optional:
 - `OS` - Operating system that Cloud Foundry is running. Must match format of Otel Collector executable name.
@@ -72,18 +77,12 @@ Optional:
 - `OTEL_VERSION` - Executable version of OpenTelemetry Collector (contrib) to use. The buildpack depends on features present in version
     0.47.0+. Default: `0.47.0`
 - `OTEL_BINARY` - OpenTelemetry Collector executable file name. Default: `otelcontribcol_$OS-v$OTEL_VERSION`
-- `OTELCOLLECTOR_DOWNLOAD_URL` - URL to download the OpenTelemetry Collector from.
+- `OTEL_BINARY_DOWNLOAD_URL` - URL to download the OpenTelemetry Collector from.
 - `RLP_GATEWAY_SHARD_ID` - Metrics are load balanced between receivers that use the same shard ID.
    Only use if multiple receivers must receive all metrics instead of
    balancing metrics between them. Default: `otelcol`
 - `RLP_GATEWAY_TLS_INSECURE` - Whether to skip TLS verify for the RLP gateway endpoint. Default: `false`
 - `UAA_TLS_INSECURE` - Whether to skip TLS verify for the UAA endpoint. Default: `false`
-- `SPLUNK_INGEST_URL` - The ingest base URL for Splunk. If specified, SPLUNK_API_URL is also required.
-   This option takes precedence over SPLUNK_REALM.
-- `SPLUNK_API_URL` - The API server base URL for Splunk. If specified, SPLUNK_INGEST_URL is also required.
-   This option take precedence over SPLUNK_REALM.
-- `SPLUNK_REALM` - The Splunk realm in which your organization resides. Used to derive SPLUNK_INGEST_URL and
-   SPLUNK_API_URL.
 
 ## Sidecar Configuration
 
--- a/deployments/cloudfoundry/buildpack/README.md
+++ b/deployments/cloudfoundry/buildpack/README.md
@@ -63,9 +63,9 @@ include them in the `manifest.yml` file, as shown in the [included example](#sid
 
 Required:
 - `RLP_GATEWAY_ENDPOINT` - The URL of the RLP gateway that acts as the proxy for the firehose,
-    e.g. `https://log-stream.sys.(TAS environment name).cf-app.com`
+    e.g. `https://log-stream.sys.TAS_ENVIRONMENT_NAME.cf-app.com`
 - `UAA_ENDPOINT` - The URL of UAA provider,
-    e.g. `https://uaa.sys.(TAS environment name).cf-app.com`
+    e.g. `https://uaa.sys.TAS_ENVIRONMENT_NAME.cf-app.com`
 - `UAA_USERNAME` - Name of the UAA user.
 - `UAA_PASSWORD` - Password for the UAA user.
 - `SPLUNK_ACCESS_TOKEN` - Your Splunk organization access token.
--- a/deployments/cloudfoundry/buildpack/README.md
+++ b/deployments/cloudfoundry/buildpack/README.md
@@ -20,7 +20,10 @@ apps and deployments that emit metrics and logs to the Loggregator Firehose as l
 $ cf create-buildpack otel_collector_buildpack . 99 --enable
 ```
 
-Note: `wget` is used to download the Splunk OpenTelemetry Collector.
+### Dependencies in Cloud Foundry Environment
+
+- `wget`
+- `jq`
 
 ### Using PCF Buildpack With an Application
 This section covers basic cf CLI (Cloud Foundry Command Line Interface) commands to use the buildpack. 
@@ -74,7 +77,8 @@ Required:
 
 Optional:
 - `OS` - Operating system that Cloud Foundry is running. Must match format of Otel Collector executable name.
-    Default: `linux_amd64`
+    Default: `linux_amd64`. This is the only officially supported OS, so any changes to this variable should
+    be done at the user's own risk.
 - `OTEL_CONFIG` - Local name of Splunk OpenTelemetry config file. Default: `otelconfig.yaml`
 - `OTEL_VERSION` - Executable version of Splunk OpenTelemetry Collector to use. The buildpack depends on features present in version
     v0.48.0+. Default: `latest`. Example valid value: `v0.48.0`.
@@ -89,6 +93,10 @@ Optional:
    balancing metrics between them. Default: `otelcol`
 - `RLP_GATEWAY_TLS_INSECURE` - Whether to skip TLS verify for the RLP gateway endpoint. Default: `false`
 - `UAA_TLS_INSECURE` - Whether to skip TLS verify for the UAA endpoint. Default: `false`
+- `SMART_AGENT_VERSION` - Version of the Smart Agent that should be downloaded. This is a dependency of
+    the collector's `signalfx` receiver. Default: `latest`. Example valid value: `v5.19.1`.
+    Note that if left the default value, the latest version will be found and later variable references will be
+    to a valid version number, not simply the word "latest".
 
 ## Sidecar Configuration
 
--- a/deployments/cloudfoundry/buildpack/README.md
+++ b/deployments/cloudfoundry/buildpack/README.md
@@ -120,8 +120,8 @@ applications:
     memory: 256M
     random-route: true
     env:
-      RLP_GATEWAY_ENDPOINT: "https://log-stream.sys.<TAS environment name>.cf-app.com"
-      UAA_ENDPOINT: "https://uaa.sys.<TAS environment name>.cf-app.com"
+      RLP_GATEWAY_ENDPOINT: "https://log-stream.sys.TAS_ENVIRONMENT_NAME.cf-app.com"
+      UAA_ENDPOINT: "https://uaa.sys.TAS_ENVIRONMENT_NAME.cf-app.com"
       UAA_USERNAME: "..."
       UAA_PASSWORD: "..."
       SPLUNK_ACCESS_TOKEN: "..."
--- a/deployments/cloudfoundry/buildpack/README.md
+++ b/deployments/cloudfoundry/buildpack/README.md
@@ -90,7 +90,7 @@ Optional:
     Default: `https://github.com/signalfx/splunk-otel-collector/releases/download/${OTEL_VERSION}/otelcol_${OS}`
 - `RLP_GATEWAY_SHARD_ID` - Metrics are load balanced between receivers that use the same shard ID.
    Only use if multiple receivers must receive all metrics instead of
-   balancing metrics between them. Default: `otelcol`
+   balancing metrics between them. Default: `opentelemetry`
 - `RLP_GATEWAY_TLS_INSECURE` - Whether to skip TLS verify for the RLP gateway endpoint. Default: `false`
 - `UAA_TLS_INSECURE` - Whether to skip TLS verify for the UAA endpoint. Default: `false`
 - `SMART_AGENT_VERSION` - Version of the Smart Agent that should be downloaded. This is a dependency of
--- a/deployments/cloudfoundry/buildpack/README.md
+++ b/deployments/cloudfoundry/buildpack/README.md
@@ -63,9 +63,9 @@ include them in the `manifest.yml` file, as shown in the [included example](#sid
 
 Required:
 - `RLP_GATEWAY_ENDPOINT` - The URL of the RLP gateway that acts as the proxy for the firehose,
-    e.g. `https://log-stream.sys.<TAS environment name>.cf-app.com`
+    e.g. ```https://log-stream.sys.<TAS environment name>.cf-app.com```
 - `UAA_ENDPOINT` - The URL of UAA provider,
-    e.g. `https://uaa.sys.<TAS environment name>.cf-app.com`
+    e.g. ```https://uaa.sys.<TAS environment name>.cf-app.com```
 - `UAA_USERNAME` - Name of the UAA user.
 - `UAA_PASSWORD` - Password for the UAA user.
 - `SPLUNK_ACCESS_TOKEN` - Your Splunk organization access token.
--- a/deployments/cloudfoundry/buildpack/README.md
+++ b/deployments/cloudfoundry/buildpack/README.md
@@ -1,14 +1,14 @@
-# OpenTelemetry Collector Pivotal Cloud Foundry (PCF) Buildpack
+# Splunk OpenTelemetry Collector Pivotal Cloud Foundry (PCF) Buildpack
 
 A [Cloud Foundry buildpack](https://docs.pivotal.io/application-service/2-11/buildpacks/) to install
-the OpenTelemetry Collector for use with PCF apps.
+the Splunk OpenTelemetry Collector for use with PCF apps.
 
 The buildpack's default functionality, as described in this document, is to deploy the OpenTelemetry Collector
 as a sidecar for the given app that's being deployed. The Collector is able to observe the app as a 
 [nozzle](https://docs.pivotal.io/tiledev/2-10/nozzle.html#nozzle) to
 the [Loggregator Firehose](https://docs.cloudfoundry.org/loggregator/architecture.html).
 The Loggregator Firehose is one of the architectures Cloud Foundry
-uses to emit logs and metrics. This means that the OpenTelemetry Collector will be observing all
+uses to emit logs and metrics. This means that the Splunk OpenTelemetry Collector will be observing all
 apps and deployments that emit metrics and logs to the Loggregator Firehose as long as it's running.
 
 ## Installation
@@ -16,11 +16,11 @@ apps and deployments that emit metrics and logs to the Loggregator Firehose as l
 - Change to this directory.
 - Run the following command:
 ```sh
-# Add buildpack for OpenTelemetry Collector
+# Add buildpack for Splunk OpenTelemetry Collector
 $ cf create-buildpack otel_collector_buildpack . 99 --enable
 ```
 
-Note: `wget` is used to download the OpenTelemetry Collector.
+Note: `wget` is used to download the Splunk OpenTelemetry Collector.
 
 ### Using PCF Buildpack With an Application
 This section covers basic cf CLI (Cloud Foundry Command Line Interface) commands to use the buildpack. 
@@ -44,12 +44,12 @@ $ cf push <app-name> -b otel_collector_buildpack -b <main_buildpack> -f manifest
 
 Note: This buildpack requires another buildpack to be supplied after it, it is not allowed to
 be the last one for an app. Also, the manifest.yml file will need to provide the
-command to run the OpenTelemetry Collector as a sidecar for the application.
+command to run the Splunk OpenTelemetry Collector as a sidecar for the application.
 
 ## Configuration
 
 The following only applies if you are using the `otelconfig.yaml` config
-provided by the buildpack.  If you provide a custom configuration file for the OpenTelemetry Collector 
+provided by the buildpack.  If you provide a custom configuration file for the Splunk OpenTelemetry Collector
 in your application (and refer to it in the sidecar configuration), these might not
 work unless you have preserved the references to the environment variables in the config file.
 For proper functionality, the `OTEL_CONFIG` environment variable must point to
@@ -75,13 +75,13 @@ Required:
 Optional:
 - `OS` - Operating system that Cloud Foundry is running. Must match format of Otel Collector executable name.
     Default: `linux_amd64`
-- `OTEL_CONFIG` - Local name of OpenTelemetry config file. Default: `otelconfig.yaml`
-- `OTEL_VERSION` - Executable version of OpenTelemetry Collector (contrib) to use. The buildpack depends on features present in version
+- `OTEL_CONFIG` - Local name of Splunk OpenTelemetry config file. Default: `otelconfig.yaml`
+- `OTEL_VERSION` - Executable version of Splunk OpenTelemetry Collector to use. The buildpack depends on features present in version
     v0.48.0+. Default: `latest`. Example valid value: `v0.48.0`.
     Note that if left the default value, the latest version will be found and later variable references will be
     to a valid version number, not simply the word "latest".
-- `OTEL_BINARY` - OpenTelemetry Collector executable file name. Default: `otelcontribcol_$OS-v$OTEL_VERSION`
-- `OTEL_BINARY_DOWNLOAD_URL` - URL to download the OpenTelemetry Collector from. This takes precedence over other
+- `OTEL_BINARY` - Splunk OpenTelemetry Collector executable file name. Default: `otelcol_$OS-v$OTEL_VERSION`
+- `OTEL_BINARY_DOWNLOAD_URL` - URL to download the Splunk OpenTelemetry Collector from. This takes precedence over other
     version variables. Only the Splunk distribution is supported.
     Default: `https://github.com/signalfx/splunk-otel-collector/releases/download/${OTEL_VERSION}/otelcol_${OS}`
 - `RLP_GATEWAY_SHARD_ID` - Metrics are load balanced between receivers that use the same shard ID.
@@ -122,7 +122,7 @@ applications:
       - name: otel-collector
         process_types:
           - web
-        command: "$HOME/../deps/otelcontribcol_${OS:-linux_amd64}-v${OTEL_VERSION:-0.47.0} --config=$HOME/../deps/${OTEL_CONFIG:-otelconfig.yaml}"
+        command: "$HOME/.otelcollector/otelcol_${OS:-linux_amd64}-v${OTEL_VERSION:-0.48.0} --config=$HOME/.otelcollector/${OTEL_CONFIG:-otelconfig.yaml}"
         memory: 100MB
 ```
 If using a `manifest.yaml` file, you may push your app simply with the following command:
@@ -133,11 +133,12 @@ $ cf push
 # If you are using cf CLI v6
 $ cf v3-push <app-name>
 ```
-This will deploy the app with the proper buildpacks, and the OpenTelemetry Collector running in the sidecar configuration.
+This will deploy the app with the proper buildpacks, and the Splunk OpenTelemetry Collector running in the
+sidecar configuration.
 
 ## Troubleshooting
 
-* If the app is running but the OpenTelemetry Collector is not, it may be that the sidecar configuration is not
+* If the app is running but the Splunk OpenTelemetry Collector is not, it may be that the sidecar configuration is not
 being picked up properly from the manifest file. Try running the following commands:
 
 ```sh
@@ -156,7 +157,7 @@ sidecars:
   - name: otel-collector
     process_types:
       - web
-    command: "$HOME/../deps/otelcontribcol_${OS:-linux_amd64}-v${OTEL_VERSION:-0.47.0} --config=$HOME/../deps/${OTEL_CONFIG:-otelconfig.yaml}"
+    command: "$HOME/.otelcollector/otelcol_${OS:-linux_amd64}-v${OTEL_VERSION:-0.48.0} --config=$HOME/.otelcollector/${OTEL_CONFIG:-otelconfig.yaml}"
 ```
 
 ### Useful CF CLI debugging commands
--- a/deployments/cloudfoundry/buildpack/README.md
+++ b/deployments/cloudfoundry/buildpack/README.md
@@ -1,8 +1,7 @@
-# OpenTelemetry Pivotal Cloud Foundry (PCF) Buildpack
+# OpenTelemetry Collector Pivotal Cloud Foundry (PCF) Buildpack
 
 A [Cloud Foundry buildpack](https://docs.pivotal.io/application-service/2-11/buildpacks/) to install
-the OpenTelemetry Collector for use with PCF apps.  This will probably work for generic Cloud Foundry
-apps as well, but it is only tested and supported on Pivotal Platform.
+the OpenTelemetry Collector for use with PCF apps.
 
 The buildpack's default functionality, as described in this document, is to deploy the OpenTelemetry Collector
 as a sidecar for the given app that's being deployed. The Collector is able to observe the app as a nozzle through
@@ -19,6 +18,8 @@ apps and deployments that emit metrics and logs to the Loggregator Firehose as l
 $ cf create-buildpack otel_collector_buildpack . 99 --enable
 ```
 
+Note: `wget` is used to download the OpenTelemetry Collector.
+
 ### Using PCF Buildpack With an Application
 This section covers basic cf CLI (Cloud Foundry Command Line Interface) commands to use the buildpack. 
 ```sh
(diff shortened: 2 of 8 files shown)


deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,155 @@
+# OpenTelemetry Pivotal Cloud Foundry (PCF) Buildpack
```
rmfitzpatrick, line 1:
```suggestion
# OpenTelemetry Collector Pivotal Cloud Foundry (PCF) Buildpack
```

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+A [Cloud Foundry buildpack](https://docs.pivotal.io/application-service/2-11/buildpacks/) to install
+the OpenTelemetry Collector for use with PCF apps.  This will probably work for generic Cloud Foundry
```
rmfitzpatrick, line 4: It may be clearer to tighten this compatibility a bit and not carry over the agent phrasing about generic functionality.

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,155 @@
...
+apps as well, but it is only tested and supported on Pivotal Platform.
+
+The buildpack's default functionality, as described in this document, is to deploy the OpenTelemetry Collector
+as a sidecar for the given app that's being deployed. The Collector is able to observe the app as a nozzle through
+the Loggregator Firehose. The Loggregator Firehose is one of the architectures Cloud Foundry
```
rmfitzpatrick, line 9: Is it worth linking to an official doc?

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+### Using PCF Buildpack With an Application
+This section covers basic cf CLI (Cloud Foundry Command Line Interface) commands to use the buildpack. 
+```sh
+# Basic setup, see the configuration section for more envvars that can be set
+$ cf set-env <app-name> OTEL_CONFIG <config_file_name>
+$ cf set-env <app-name> OTELCOL <desired_collector_executable_name>
+
+$ cd <my-application-direcory>/
```
rmfitzpatrick, line 29:
```suggestion
$ cd <my-application-directory>/
```

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,155 @@
...
+include them in the `manifest.yml` file, as shown in the [included example](#sidecar-configuration).
+
+Required:
+- `RLP_GATEWAY_ENDPOINT` - The URL of the RLP gateway that acts as the proxy for the firehose,
+    e.g. `https://log-stream.sys.<TAS environment name>.cf-app.com`
+- `UAA_ENDPOINT` - The URL of UAA provider,
+    e.g. `https://uaa.sys.<TAS environment name>.cf-app.com`
+- `UAA_USERNAME` - Name of the UAA user.
+- `UAA_PASSWORD` - Password for the UAA user.
+- `SIGNALFX_ACCESS_TOKEN` - Your SignalFx organization access token.
```
rmfitzpatrick, line 65: I think we've completely moved to the `SPLUNK_` prefix for new offerings. Also SignalFx references should probably be updated to Splunk Observability Suite.

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,155 @@
...
+- `OTEL_CONFIG` - Local name of OpenTelemetry config file. Default: `otelconfig.yaml`
+- `OTEL_VERSION` - Executable version of OpenTelemetry Collector (contrib) to use. The buildpack depends on features present in version
+    0.47.0+. Default: `0.47.0`
+- `OTELCOL` - OpenTelemetry Collector executable file name. Default: `otelcontribcol_$OS-v$OTEL_VERSION`
+- `OTELCOLLECTOR_DOWNLOAD_URL` - URL to download the OpenTelemetry Collector from.
+
+- `RLP_GATEWAY_SHARD_ID` - Metrics are load balanced between receivers that use the same shard ID.
+   Only use if multiple receivers must receive all metrics instead of
+   balancing metrics between them. Default: `otelcol`
+- `RLP_GATEWAY_TLS_INSECURE` - Whether to skip TLS verify for the RLP gateway endpoint. Default: `true`
```
rmfitzpatrick, line 79: is it better to default to false to recommend validation?

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,155 @@
...
+$ cf push
+
+# If you are using cf CLI v6
+$ cf v3-push <app-name>
+```
+This will deploy the app with the proper buildpacks, and the OpenTelemetry Collector running in the sidecar configuration.
+
+## Troubleshooting
+
+* If the app is running but the OpenTelemetry Collector is not, it may be that the sidecar configuration is not
```
rmfitzpatrick, line 135: are there example errors and error sources that could be provided?

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,155 @@
...
+being picked up properly from the manifest file. Try running the following commands:
+
+```sh
+# This will apply the manifest file to an existing app
+$ cf v3-apply-manifest -f manifest.yml
+# This will re-load the app with the sidecar configuration included
+$ cf push
+```
+
+* Another possibility is the sidecar was not allocated memory. The `memory` option
```
rmfitzpatrick, line 145: possible for an example?

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,68 @@
...
+CACHE_DIR=$2
+DEPS_DIR=$3
+IDX=$4
+
+BUILDPACK_DIR="$(readlink -f ${BASH_SOURCE%/*})"
+TARGET_DIR="$BUILD_DIR/.otelcollector"
+
+OTEL_CONFIG="${OTEL_CONFIG:-otelconfig.yaml}"
+# Version must be >=0.47.0, the first release included the enabled Cloud Foundry receiver
+OTEL_VERSION="${OTEL_VERSION:-0.47.0}"
```
rmfitzpatrick, line 15: Not sure it needs to be addressed here, but a self updating mechanism is preferable imo (maybe made part of https://github.com/signalfx/splunk-otel-collector/blob/main/internal/buildscripts/update-deps or something else?) cc @jcheng-splunk

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,68 @@
...
+# File name for given release version
+OTELCOL="${OTELCOL:-otelcontribcol_${OS}-v${OTEL_VERSION}}"
+OTELCOL_TAR="$OTELCOL.tar.gz"
+CACHED_OTELCOL_TAR="$CACHE_DIR/$OTELCOL_TAR"
+
+if [[ -f "$CACHED_OTELCOL_TAR" ]]; then
+  echo "-----> Using cached Otel Collector install: $CACHED_OTELCOL_TAR"
+  cp $CACHED_OTELCOL_TAR $TARGET_DIR
+else
+  OTELCOL_DOWNLOAD_URL="${OTELCOL_DOWNLOAD_URL:-https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v${OTEL_VERSION}/otelcol-contrib_${OTEL_VERSION}_${OS}.tar.gz}"
```
rmfitzpatrick, line 42: We may want to make this configurable since some systems will have access restrictions.

deployments/cloudfoundry/buildpack/bin/supply
jcheng-splunk: Could `OTEL_VERSION` default to `latest` and just use the github api to automatically determine the latest version to download? Something like https://github.com/signalfx/splunk-otel-collector/blob/main/internal/buildscripts/packaging/fpm/common.sh#L77?

deployments/cloudfoundry/buildpack/README.md
crobert-1: I'm sure it would be helpful! 

Is there a specific part of this paragraph that you're referring to as far as an official doc? Or should every statement have a reference?

As far as collector functionality I can link the README, but I'm not aware of any other official source of information. Please let me know if I'm missing something though.

deployments/cloudfoundry/buildpack/README.md
crobert-1: Will update. I wasn't aware of the proper terminology, thanks for letting me know!

crobert-1: One question, I assume in the actual config file we'll still have the exporter as `signalfx`, right? My assumption is that should be the default exporter for the collector here, but if there's some other endpoint we should be sending this too, let me know.

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,155 @@
...
+- `UAA_PASSWORD` - Password for the UAA user.
+- `SIGNALFX_ACCESS_TOKEN` - Your SignalFx organization access token.
+
+Optional:
+- `OS` - Operating system that Cloud Foundry is running. Must match format of Otel Collector executable name.
+    Default: `linux_amd64`
+- `OTEL_CONFIG` - Local name of OpenTelemetry config file. Default: `otelconfig.yaml`
+- `OTEL_VERSION` - Executable version of OpenTelemetry Collector (contrib) to use. The buildpack depends on features present in version
+    0.47.0+. Default: `0.47.0`
+- `OTELCOL` - OpenTelemetry Collector executable file name. Default: `otelcontribcol_$OS-v$OTEL_VERSION`
```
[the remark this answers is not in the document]
crobert-1: I'll update to `OTEL_BINARY`

deployments/cloudfoundry/buildpack/README.md
crobert-1: Fixed, thanks!

deployments/cloudfoundry/buildpack/README.md
crobert-1: Removed the extra sentence.

deployments/cloudfoundry/buildpack/README.md
crobert-1: Fixed, thanks!

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,68 @@
...
+  echo "-----> Downloading OpenTelemetry Collector v$OTEL_VERSION ($OTELCOL_DOWNLOAD_URL)"
+  wget -O "$TARGET_DIR/$OTELCOL_TAR" $OTELCOL_DOWNLOAD_URL > /dev/null 2>&1
```
jcheng-splunk, line 44: Should we document in the readme that `wget` is a requirement?

Also, since you're redirecting all output to `/dev/null`, download failures would be silent.  Should there be a message printed for failures, or remove the redirection?

deployments/cloudfoundry/buildpack/README.md
crobert-1: Yes, I'll update. The receiver's defaults are `false` as well, so that's the right way to do it.

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,68 @@
...
+# Move Collector executable and config to dependencies directory to enable app access
+tar -xf "$TARGET_DIR/$OTELCOL_TAR" -C "$TARGET_DIR"
+mv "$TARGET_DIR/otelcol-contrib" "$TARGET_DIR/$OTELCOL"
+cp -f "$TARGET_DIR/$OTELCOL" "$TARGET_DIR/$OTEL_CONFIG" "$DEPS_DIR"
+chmod 755 "$DEPS_DIR/$OTELCOL"
+
+# Define Collector's default environment variables by adding bash script to .profile.d directory in root of app
+mkdir $DEPS_DIR/$IDX/profile.d
+echo "export RLP_GATEWAY_SHARD_ID=\"otelcol\"
+export RLP_GATEWAY_TLS_INSECURE=true
```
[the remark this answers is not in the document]
crobert-1: Yes, I tested it, and the values provided in the manifest file will override these defaults.

deployments/cloudfoundry/buildpack/README.md
crobert-1: Unfortunately both errors are silent failures. The push commands are successful, and the logs don't include any relevant information. The only way I figured these out is from reading documentation and trial and error.

deployments/cloudfoundry/buildpack/README.md
crobert-1: I can include an example yaml block for the sidecar without the memory specified explicitly, but that's all I can really think of as an example here.

deployments/cloudfoundry/buildpack/bin/detect
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+if [ -n "$DISABLE_OTEL_COLLECTOR" ]; then
```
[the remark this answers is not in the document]
crobert-1: No, this isn't used. The sole purpose of this buildpack is to run the collector, so if a user doesn't want the collector, all they have to do is not include the buildpack. I'll remove it.

deployments/cloudfoundry/buildpack/bin/supply
crobert-1: I asked about this in our team's slack, and @emaderer and I had thought it may be better to stick with a specific version in case of breaking changes. Another concern I had at the time was that I didn't know how to download latest. Now that I have a way to do it that's not as much of a concern.

Do either of you have concerns about possible breaking changes? I'm not sure how much effort it would take to update a hardcoded version, but it would be more tedious than always setting it to latest.

deployments/cloudfoundry/buildpack/bin/supply
crobert-1: I'm not sure exactly what you're requesting to be made configurable. Could you expand your what configurability you'd want here?

deployments/cloudfoundry/buildpack/bin/supply
crobert-1: I've added a note in the installation section of the README that wget is used.

There's a check directly after this that prints out an error message and stops the deployment if the download fails, but I've added the `--no-verbose` option to get rid of spam, but still print error messages if any occur.

deployments/cloudfoundry/buildpack/bin/supply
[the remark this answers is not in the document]
crobert-1: Are there any plans to remove the executable from releases? I'm thinking of moving over to directly downloading the executable, instead of the current process of the `executable.tar.gz` (especially since there is no `executable.tar.gz` in our collector distribution). I just want to make sure my method is supported, and will continue to be since the upstream collector no longer directly includes executables.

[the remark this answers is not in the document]
crobert-1: Alright, I'll work on moving over to use the Splunk distro release, and the executable binary itself rather than the upstream version.

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,68 @@
...
+if [[ -f "$CACHED_OTEL_BINARY_TAR" ]]; then
+  echo "-----> Using cached Otel Collector install: $CACHED_OTEL_BINARY_TAR"
+  cp $CACHED_OTEL_BINARY_TAR $TARGET_DIR
+else
+  OTEL_BINARY_DOWNLOAD_URL="${OTEL_BINARY_DOWNLOAD_URL:-https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v${OTEL_VERSION}/otelcol-contrib_${OTEL_VERSION}_${OS}.tar.gz}"
+  echo "-----> Downloading OpenTelemetry Collector v$OTEL_VERSION ($OTEL_BINARY_DOWNLOAD_URL)"
+  wget -nv -O "$TARGET_DIR/$OTEL_BINARY_TAR" $OTEL_BINARY_DOWNLOAD_URL
+fi
+
+if [ $? -ne 0 ]; then
```
jcheng-splunk, line 47: Since you have the `set -e` option at the beginning, the script will exit immediately before reaching this line if `wget` or any command before this line fails, and the message below will not be displayed.

jcheng-splunk:
Maybe outside of the scope of this PR, but is there anything we can add for CI testing?

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,163 @@
...
+   Only use if multiple receivers must receive all metrics instead of
+   balancing metrics between them. Default: `otelcol`
+- `RLP_GATEWAY_TLS_INSECURE` - Whether to skip TLS verify for the RLP gateway endpoint. Default: `false`
+- `UAA_TLS_INSECURE` - Whether to skip TLS verify for the UAA endpoint. Default: `false`
+- `SPLUNK_INGEST_URL` - The ingest base URL for Splunk. If specified, SPLUNK_API_URL is also required.
+   This option takes precedence over SPLUNK_REALM.
+- `SPLUNK_API_URL` - The API server base URL for Splunk. If specified, SPLUNK_INGEST_URL is also required.
+   This option take precedence over SPLUNK_REALM.
+- `SPLUNK_REALM` - The Splunk realm in which your organization resides. Used to derive SPLUNK_INGEST_URL and
+   SPLUNK_API_URL.
```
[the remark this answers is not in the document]
crobert-1: I'm going to move these options to required to make it more clear. Either the INGEST and API variables are required, or REALM is. There are no defaults, so hopefully it'll be clearer when I move it to required.

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,163 @@
...
+- `SPLUNK_ACCESS_TOKEN` - Your Splunk organization access token.
+
+Optional:
+- `OS` - Operating system that Cloud Foundry is running. Must match format of Otel Collector executable name.
+    Default: `linux_amd64`
+- `OTEL_CONFIG` - Local name of OpenTelemetry config file. Default: `otelconfig.yaml`
+- `OTEL_VERSION` - Executable version of OpenTelemetry Collector (contrib) to use. The buildpack depends on features present in version
+    0.47.0+. Default: `0.47.0`
+- `OTEL_BINARY` - OpenTelemetry Collector executable file name. Default: `otelcontribcol_$OS-v$OTEL_VERSION`
+- `OTELCOLLECTOR_DOWNLOAD_URL` - URL to download the OpenTelemetry Collector from.
```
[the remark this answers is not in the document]
crobert-1: I've updated to reference the same variable, and once I get this working with the Splunk distribution of the collector, I'll add the default.

deployments/cloudfoundry/buildpack/README.md
rmfitzpatrick: The Loggregator Firehose (maybe https://docs.cloudfoundry.org/loggregator/architecture.html#firehose-architecture?)

deployments/cloudfoundry/buildpack/README.md
rmfitzpatrick: right, the signalfx exporter sends metrics in that format so its component name will likely always be that.

deployments/cloudfoundry/buildpack/README.md
crobert-1: I've added a couple links, one to the firehose and one for nozzles.

deployments/cloudfoundry/buildpack/README.md
crobert-1: Thanks for the clarification! I've removed SignalFx references.

deployments/cloudfoundry/buildpack/bin/supply
crobert-1: Okay, as long as this is a common and supported functionality it should be fine. There's a buildpack variable `OTEL_VERSION`, if that's set we can use the version specified, otherwise use the latest.

deployments/cloudfoundry/buildpack/bin/supply
crobert-1: This variable can be set by the user and passed in, but I also made some adjustments so that we're not using wget at all if the user specifies the download url.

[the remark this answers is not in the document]
crobert-1: Latest commit implements using the Splunk distribution

crobert-1:
@jeffreyc-splunk In regards to adding CI testing, I don't think the actual functionality can be automated at this point. The problem is that this requires a Tanzu Application Service (TAS) instance, which currently needs to be created by specific users through a VMware dashboard. I don't believe the current usage agreement would allow for spinning up instances for testing. I'll check to make sure my understanding is correct, but that's why I hadn't implemented anything at this point.

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,70 @@
...
+OTEL_BINARY="${OTEL_BINARY:-otelcontribcol_${OS}-${OTEL_VERSION}}"
+CACHED_OTEL_BINARY="$CACHE_DIR/$OTEL_BINARY"
+
+if [[ -f "$CACHED_OTEL_BINARY" ]]; then
+  echo "-----> Using cached Otel Collector install: $CACHED_OTEL_BINARY"
+  cp $CACHED_OTEL_BINARY $TARGET_DIR
+else
+  OTEL_BINARY_DOWNLOAD_URL="${OTEL_BINARY_DOWNLOAD_URL:-${OTEL_BASE_URL}/download/${OTEL_VERSION}/otelcol_${OS}}"
+  echo "-----> Downloading OpenTelemetry Collector $OTEL_VERSION ($OTEL_BINARY_DOWNLOAD_URL)"
+  wget -nv -O "$TARGET_DIR/$OTEL_BINARY" $OTEL_BINARY_DOWNLOAD_URL
```
rmfitzpatrick, line 53: Can you also download and patch the smart agent bundle like the dockerfile does? https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/Dockerfile#L27 `SPLUNK_BUNDLE_DIR` and `SPLUNK_COLLECTD_DIR` env vars should also be set for these to be detected.

crobert-1: I'm not very familiar with the smart agent, or with the SignalFx receiver, so I just have some clarifying questions:

1. From your description, this requirement is because of the SignalFx receiver, so is this executable bundle required with every collector that wants to use the SignalFx receiver?
2. This is specifically a SignalFx receiver dependency, and not a Smart Agent dependency right? I originally assumed this functionality was relevant only for the Smart Agent, so I want to make sure that my assumption was wrong here.

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,168 @@
+# OpenTelemetry Collector Pivotal Cloud Foundry (PCF) Buildpack
+
+A [Cloud Foundry buildpack](https://docs.pivotal.io/application-service/2-11/buildpacks/) to install
+the OpenTelemetry Collector for use with PCF apps.
```
jeffreyc-splunk, line 4:
```suggestion
the Splunk OpenTelemetry Collector for use with PCF apps.
```
I think we should append `Splunk` for all other references to `OpenTelemetry Collector`.

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+Optional:
+- `OS` - Operating system that Cloud Foundry is running. Must match format of Otel Collector executable name.
+    Default: `linux_amd64`
+- `OTEL_CONFIG` - Local name of OpenTelemetry config file. Default: `otelconfig.yaml`
+- `OTEL_VERSION` - Executable version of OpenTelemetry Collector (contrib) to use. The buildpack depends on features present in version
+    v0.48.0+. Default: `latest`. Example valid value: `v0.48.0`.
+    Note that if left the default value, the latest version will be found and later variable references will be
+    to a valid version number, not simply the word "latest".
+- `OTEL_BINARY` - OpenTelemetry Collector executable file name. Default: `otelcontribcol_$OS-v$OTEL_VERSION`
```
jeffreyc-splunk, line 83: I think we should remove references to `contrib`.
```suggestion
- `OTEL_BINARY` - OpenTelemetry Collector executable file name. Default: `otelcol_$OS-v$OTEL_VERSION`
```

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,168 @@
...
+      UAA_ENDPOINT: "https://uaa.sys.<TAS environment name>.cf-app.com"
+      UAA_USERNAME: "..."
+      UAA_PASSWORD: "..."
+      SPLUNK_ACCESS_TOKEN: "..."
+      SPLUNK_REALM: "..."
+    sidecars:
+      - name: otel-collector
+        process_types:
+          - web
+        command: "$HOME/../deps/otelcontribcol_${OS:-linux_amd64}-v${OTEL_VERSION:-0.47.0} --config=$HOME/../deps/${OTEL_CONFIG:-otelconfig.yaml}"
```
jeffreyc-splunk, line 125:
```suggestion
        command: "$HOME/../deps/otelcontribcol_${OS:-linux_amd64}-v${OTEL_VERSION:-0.48.0} --config=$HOME/../deps/${OTEL_CONFIG:-otelconfig.yaml}"
```

jeffreyc-splunk:
> @jeffreyc-splunk In regards to adding CI testing, I don't think the actual functionality can be automated at this point. The problem is that this requires a Tanzu Application Service (TAS) instance, which currently needs to be created by specific users through a VMware dashboard. I don't believe the current usage agreement would allow for spinning up instances for testing. I'll check to make sure my understanding is correct, but that's why I hadn't implemented anything at this point.

Maybe at a minimum, would it be possible to add a job to just create the buildpack and ensure that the `supply` script runs correctly?

deployments/cloudfoundry/buildpack/README.md
crobert-1: Updated, all references should now be `Splunk OpenTelemetry Collector`.

deployments/cloudfoundry/buildpack/README.md
crobert-1: I've removed contrib references.

deployments/cloudfoundry/buildpack/README.md
crobert-1: Good find, thanks! I've updated to be 0.48.0 now.

crobert-1:
> > @jeffreyc-splunk In regards to adding CI testing, I don't think the actual functionality can be automated at this point. The problem is that this requires a Tanzu Application Service (TAS) instance, which currently needs to be created by specific users through a VMware dashboard. I don't believe the current usage agreement would allow for spinning up instances for testing. I'll check to make sure my understanding is correct, but that's why I hadn't implemented anything at this point.
> 
> Maybe at a minimum, would it be possible to add a job to just create the buildpack and ensure that the `supply` script runs correctly?

That should be doable, I'll work on adding that.

crobert-1:
@jeffreyc-splunk I've added a basic CI test that just makes sure the supply script runs successfully. I wanted to also make sure the buildpack is built successfully, but unfortunately that also requires a cloudfoundry instance to be running, so it isn't feasible.

I'm not very familiar with GitHub automation, so please let me know if there's anything to fix!

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,108 @@
...
+echo "export RLP_GATEWAY_SHARD_ID=\"otelcol\"
+export RLP_GATEWAY_TLS_INSECURE=false
+export UAA_TLS_INSECURE=false
+export OTEL_VERSION=$OTEL_VERSION" >> $DEPS_DIR/$IDX/profile.d/otel_config_env_vars.sh
+
+echo "-----> Successfully installed OpenTelemetry Collector $OTEL_VERSION and set default environment variables"
+
+# Need to run SignalFx patch interpreter for the SignalFx receiver to work properly.
+# Get proper version number if we're getting latest release
+SMART_AGENT_VERSION="${SMART_AGENT_VERSION:-latest}"
```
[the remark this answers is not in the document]
crobert-1: I've added a reference for the `SMART_AGENT_VERSION` variable in the README.

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,169 @@
...
+- `UAA_PASSWORD` - Password for the UAA user.
+- `SPLUNK_ACCESS_TOKEN` - Your Splunk organization access token.
+- The Splunk Observability Suite requires an endpoint, so one of the two options below must be specified:
+    1) - `SPLUNK_INGEST_URL` - The ingest base URL for Splunk. This option takes precedence over SPLUNK_REALM.
+       - `SPLUNK_API_URL` - The API server base URL for Splunk. This option takes precedence over SPLUNK_REALM.
+    2) - `SPLUNK_REALM` - The Splunk realm in which your organization resides. Used to derive SPLUNK_INGEST_URL
+       and SPLUNK_API_URL.
+
+Optional:
+- `OS` - Operating system that Cloud Foundry is running. Must match format of Otel Collector executable name.
```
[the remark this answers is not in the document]
crobert-1: No, not officially. It's all bash scripts though, so theoretically it should work with other amd64 architectures,  at least. I'll make a note though.

deployments/cloudfoundry/README.md
```diff
@@ -0,0 +1,6 @@
+# OpenTelemetry Collector Pivotal Cloud Foundry (PCF) Integrations
```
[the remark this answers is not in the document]
crobert-1: Updated, thanks!

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,108 @@
...
+echo "       TARGET_DIR: $TARGET_DIR"
+
+mkdir -p $TARGET_DIR
+
+# Copy over default OpenTelemetry configuration file for sidecar configuration use
+cp "$BUILDPACK_DIR/../$OTEL_CONFIG" "$TARGET_DIR/"
+
+# Get proper version number if we're getting latest release
+if [ $OTEL_VERSION = "latest" ] && [ -z "${OTEL_BINARY_DOWNLOAD_URL:-}" ]; then
+        OTEL_VERSION=$( wget -qO- --header="Accept: application/json" "$OTEL_BASE_URL/latest" | jq -r '.tag_name' )
```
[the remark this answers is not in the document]
crobert-1: Added

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,108 @@
...
+mkdir -p $SA_TARGET_DIR
+
+echo "-----> Downloading SignalFx Smart Agent $SMART_AGENT_VERSION ($SMART_AGENT_DOWNLOAD_URL)"
+wget -nv -O "$SA_TARGET_DIR/$SMART_AGENT" $SMART_AGENT_DOWNLOAD_URL
+tar -xf "$SA_TARGET_DIR/$SMART_AGENT" -C "$SA_TARGET_DIR"
+mv "${SA_TARGET_DIR}/signalfx-agent" "${SA_TARGET_DIR}/agent-bundle"
+
+# Absolute path of interpreter in smart agent dir is set in dependent binaries
+# requiring the interpreter location not to change.
+SPLUNK_BUNDLE_DIR="/home/vcap/app/.signalfx/agent-bundle"
```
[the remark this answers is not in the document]
crobert-1: The `SA_TARGET_DIR` is actually a temporary sub-directory of the $BUILD_DIR variable passed into the script. These directories are wiped after the supply script is finished. The `SPLUNK_BUNDLE_DIR` variable requires an absolute path to the final application directory, so it shouldn't be dependent on the temp directory.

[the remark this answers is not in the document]
jeffreyc-splunk: In that case, do the contents in `SA_TARGET_DIR` need to be explicitly copied/moved to `SPLUNK_BUNDLE_DIR`?

deployments/cloudfoundry/buildpack/bin/supply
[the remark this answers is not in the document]
jeffreyc-splunk: If `OS` is anything other than `linux_amd64`, the smart agent download/install steps below should be skipped.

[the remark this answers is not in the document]
crobert-1: Good point, I'll update!

deployments/cloudfoundry/buildpack/bin/supply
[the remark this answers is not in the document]
crobert-1: No, everything under `BUILD_DIR` is copied over automatically to `/home/vcap/app`. Theoretically we could just use `SPLUNK_BUNDLE_DIR/.signalfx`for `SA_TARGET_DIR`, but my understanding is that it's not best practice.

Just for reference, this is the output when installing the buildpack for an app:
```
 -----> Installing Otel Collector v0.48.0
          BUILD_DIR: /tmp/app
          CACHE_DIR: /tmp/cache/c04188b8834b0c8b00afe22af7cdf13b
          DEPS_DIR: /tmp/contents1438789478/deps
          BUILDPACK_INDEX: 0
          BUILDPACK_DIR: /tmp/buildpacks/c04188b8834b0c8b00afe22af7cdf13b/bin
          TARGET_DIR: /tmp/app/.otelcollector
```
So all of the original paths get deleted, but some get copied over to the `/home/vcap` (like `BUILD_DIR` and `DEPS_DIR`. So `/tmp/app` becomes `/home/vcap/app` and `/tmp/.../deps` becomes `/home/vcap/deps`

Hopefully this generally makes sense, feel free to let me know if you have any questions or if I need to update it.

deployments/cloudfoundry/buildpack/bin/supply
```diff
@@ -0,0 +1,113 @@
...
+${SA_TARGET_DIR}/agent-bundle/bin/patch-interpreter ${SPLUNK_BUNDLE_DIR}
+rm -f $SPLUNK_BUNDLE_DIR/bin/signalfx-agent \
+            $SPLUNK_BUNDLE_DIR/bin/agent-status \
+            $SA_TARGET_DIR/$SMART_AGENT;
+
+echo "export SPLUNK_BUNDLE_DIR=${SPLUNK_BUNDLE_DIR}
+export SPLUNK_COLLECTD_DIR=${SPLUNK_COLLECTD_DIR}
+export PATH=${PATH}:${SPLUNK_BUNDLE_DIR}/bin" >> $DEPS_DIR/$IDX/profile.d/signalfx_agent_env_vars.sh
+
+echo "-----> Successfully installed SignalFx patch interpreter"
```
rmfitzpatrick, line 113:
```suggestion
echo "-----> Successfully installed SignalFx Smart Agent bundle"
```

rmfitzpatrick: `patch-interpreter` uses [patchelf](https://github.com/NixOS/patchelf) to set the elf interpreter to ld-linux for the bundled agent executables.

rmfitzpatrick (approved):

deployments/cloudfoundry/buildpack/bin/supply
crobert-1: I've updated the message, thanks for the extra context!

deployments/cloudfoundry/buildpack/bin/supply
crobert-1: I talked to Ryan so I'll just include answers here.
1. This is required for every collector that wants to use the [SignalFx SmartAgent receiver](https://github.com/signalfx/splunk-otel-collector/blob/main/internal/receiver/smartagentreceiver/README.md#signalfx-smart-agent-receiver). The Smart Agent receiver is a way to allow use of all Smart Agent monitors, so the bundle is really just the release of the Smart Agent.
2. This is a Smart Agent dependency, but since the receiver relies on the Smart Agent, it also relies on the bundle as well.
Note: This is regarding the smartagent receiver, not the signalfx receiver.

deployments/cloudfoundry/buildpack/README.md
```diff
@@ -0,0 +1,177 @@
...
+the configuration file, whether using the default or a customer version.
+
+Set the following environment variables with `cf set-env` as applicable to configure this buildpack, or
+include them in the `manifest.yml` file, as shown in the [included example](#sidecar-configuration).
+
+Required:
+- `RLP_GATEWAY_ENDPOINT` - The URL of the RLP gateway that acts as the proxy for the firehose,
+    e.g. `https://log-stream.sys.<TAS environment name>.cf-app.com`
+- `UAA_ENDPOINT` - The URL of UAA provider,
+    e.g. `https://uaa.sys.<TAS environment name>.cf-app.com`
```
jeffreyc-splunk, line 68: @crobert-1 Can you update https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/lychee.yml#L24 to exclude `*.cf-app.com`?

crobert-1: Will do, thanks for pointing this out!

jeffreyc-splunk (approved):

Issue #1413, opened 2022-03-29
Enable cloud foundry receiver


crobert-1:
The upstream collector has the cloud foundry receiver enabled and working now, as part of the contrib release 0.47.0. This receiver should also be enabled in our distribution.

PR #1414, opened 2022-03-29
Enable Cloud Foundry receiver

crobert-1:
This PR is to resolve issue #1413 

The Cloud Foundry receiver has been enabled in the OpenTelementry Collector Contrib repository, and is now supported. This receiver should also be supported in the Splunk distribution. This change adds it as a dependency, adds it to the list of valid receivers, and adds a reference to documentation.


docs/components.md
```diff
@@ -36,15 +36,15 @@ support at this time). While in alpha, breaking changes may be introduced in a
 new release. In addition, any of these components may be removed prior to the
 1.0 release.
 
-| Receivers                                                                                                                 | Processors | Exporters                                                                                           | Extensions |
```
emaderer, line 39: do you know why it's marking all these lines as deleted and then added?

crobert-1: I think it's because I changed some of the spacing to try to line the columns up in the plain-text. If we don't want whitespace changes in this commit I can remove my changes.

crobert-1: I've removed unnecessary whitespace changes.

emaderer (approved):

PR #1532, opened 2022-05-04
zc: fix bugs, make service name generation more resilient

pmcollins:
**Use a thorough algorithm to detect a main class**
The auto-instrumentation logic uses the Java main class (if one is provided as one of the arguments to the `java` command) to generate the service name, transforming for example `my.package.path.MyApp` to a service name of `my-package-path-myapp`. Previous to this change, the main class was detected from the command line arguments by whether it was the first argument (from left to right) that was not a command line flag, nor a path to a jar, which contained dots. This logic has been improved to ensure that the argument has legal Java package names separated by dots and that the main class (short name) is both a legal package name and starts with a capital letter.

**Deduplicate redundant parts of a service name**
If no main class is detected on the command line and the service name is instead generated by one or more colon-separated paths to jars, redundant tokens in between hyphens are removed since they don't provide additional identifying information. For example, if two jars are provided: `myapp/lib/foolib.jar:myapp/lib/webapp.jar`, the generated service name would be `myapp-foolib-webapp` -- the `lib` directories would be stripped out because they are generic, and the second `myapp` would be removed because it would be redundant.

**Improve handling of very large commands**
In the unlikely event that the length of a single argument is longer than 8k, it is split at the 8k boundary. If the entire CLI command is longer than 16k, the .so truncates the command at the 16k boundary.

**Bail out if java_tool_options env var is already set**
If the JAVA_TOOL_OPTIONS environment variable is already set, the .so exits early and cedes control to the main program, without altering the environment.


emaderer:
@pmcollins  can you add a couple of examples of service names before and after this change to the description?

emaderer (approved):

PR #1533, opened 2022-05-04
zc: add telemetry/metrics

pmcollins:
This change causes the zero config .so to send a metric to the local collector each time it runs.

### DIFF
--- a/instrumentation/src/metrics_client.c
+++ b/instrumentation/src/metrics_client.c
@@ -1,0 +1,119 @@
+#include "metrics_client.h"
+
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <string.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <stdbool.h>
+
+static const int RECV_BUF_LEN = 1024;
+
+static char *const expected = "HTTP/1.1 200 OK";
+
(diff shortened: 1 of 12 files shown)


instrumentation/src/metrics_client.c
```diff
@@ -0,0 +1,108 @@
...
+#include <stdio.h>
+#include <stdbool.h>
+
+static const int RECV_BUF_LEN = 1024;
+
+static char *const expected = "HTTP/1.1 200 OK";
+
+bool http_post(char *host, int port, char *method, char *path, char *postData);
+
+void send_otlp_metric() {
```
emaderer, line 17: can we also add the service name as a tag?

instrumentation/src/metrics_client.c
```diff
@@ -0,0 +1,108 @@
...
+
+bool connect_http(const char *host, int port, int socket_descriptor);
+
+bool post(int socket_descriptor, char *host, int port, char *method, char *path, char *postData);
+
+bool receive(int socket_descriptor);
+
+bool http_post(char *host, int port, char *method, char *path, char *postData) {
+    int socket_descriptor = make_socket(1);
+    if (socket_descriptor == -1) {
```
jvoravong, line 38: The if statements on lines 38 and 45 use {} while the rest don't.  I'm not requesting a change, just wondering if there is a reason for it?

PR #1541, opened 2022-05-06
initial version of the pulsar exporter (metrics). Tests yet to be added.

sparam:
Initial version of the pulsar exporter . This currently support metrics export. This uses a similar pattern as the existing kafka exporter https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/kafkaexporter

**NOTE:** The go-client is non-GA. However as per Pulsar documentation, the ask is to use go-client, as the existing CGo is deprecated
https://pulsar.apache.org/docs/en/client-libraries-go/
https://github.com/apache/pulsar-client-go


internal/exporter/pulsarexporter/factor_test.go
```diff
@@ -25,16 +26,13 @@ func TestCreateDefaultConfig(t *testing.T) {
 	cfg := createDefaultConfig().(*Config)
 	assert.NotNil(t, cfg, "failed to create default config")
 	assert.NoError(t, configtest.CheckConfigStruct(cfg))
-	assert.Equal(t, defaultBroker, cfg.Brokers)
-	assert.Equal(t, "", cfg.Topic)
+	assert.Equal(t, defaultBroker, cfg.Broker)
+	assert.Equal(t, defaultMetricsTopic, cfg.Topic)
 }
 
 func TestCreateMetricsExport(t *testing.T) {
```
pmcollins, line 33: I tried to run these tests locally but they tried to connect to a Pulsar instance and were unable. It's doable to create and run tests that connect to a Pulsar instance, but the tests should probably live in a separate file e.g. `integration_test.go` which is flagged as `integration` at the top, and they should start the containers they need. 

Here's a good example:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/apachereceiver/integration_test.go

I think that in order to exercise the code in factory.go without creating containers, the production code will have to be reworked a bit so that creating an exporter doesn't try to do any I/O. This will be a good change in my opinion, because we generally don't want to do I/O (e.g. connecting to Pulsar) until we Start() the exporter.

internal/exporter/pulsarexporter/pulsar_exporter.go
```diff
@@ -17,91 +17,94 @@ package pulsarexporter
...
-	"go.uber.org/zap"
 )
 
 var errUnrecognizedEncoding = fmt.Errorf("unrecognized encoding")
 
-// pulsarMetricsProducer uses sarama to produce metrics messages to pulsar
-type pulsarMetricsProducer struct {
+// pulsarMetricsExporter produce metrics messages to pulsar
+type pulsarMetricsExporter struct {
+	client    pulsar.Client
```
pmcollins, line 33: Is this field necessary? I'm not seeing it being used.

internal/exporter/pulsarexporter/pulsar_exporter.go
```diff
@@ -17,91 +17,94 @@ package pulsarexporter
...
+	if producerErr != nil {
+		return nil, nil, producerErr
 	}
-	return producer, nil
+
+	return client, producer, nil
 }
 
-func newMetricsExporter(config Config, set component.ExporterCreateSettings, marshalers map[string]MetricsMarshaler) (*pulsarMetricsProducer, error) {
+func newMetricsExporter(config Config, set component.ExporterCreateSettings, marshalers map[string]MetricsMarshaler) (*pulsarMetricsExporter, error) {
```
pmcollins, line 92: Can we move this constructor to right below its corresponding struct?

internal/exporter/pulsarexporter/pulsar_exporter.go
```diff
@@ -17,91 +17,94 @@ package pulsarexporter
...
+	return fmt.Errorf("pulsar producer failed to send metric data due to error: %w", errors)
+}
+
+func (e *pulsarMetricsExporter) Close(context.Context) error {
+	e.producer.Close()
 	return nil
 }
 
-func newPulsarProducer(config Config) (pulsar.Producer, error) {
+func newPulsarProducer(config Config) (pulsar.Client, pulsar.Producer, error) {
```
pmcollins, line 64: Does this need to return a client and a producer? It doesn't seem like the client is used.

internal/exporter/pulsarexporter/pulsar_exporter.go
```diff
@@ -17,91 +17,94 @@ package pulsarexporter
...
 )
 
 var errUnrecognizedEncoding = fmt.Errorf("unrecognized encoding")
 
-// pulsarMetricsProducer uses sarama to produce metrics messages to pulsar
-type pulsarMetricsProducer struct {
+// pulsarMetricsExporter produce metrics messages to pulsar
+type pulsarMetricsExporter struct {
+	client    pulsar.Client
 	producer  pulsar.Producer
```
pmcollins, line 34: This is nice because `pulsar.Producer` is actually an interface. That means that we'll be able to swap out the implementation with our own fake one for unit tests. (The current unit tests in pulsar_exporter_test.go try to connect to a real Pulsar instance.)

PR #1576, opened 2022-05-18
Add support for multiple config arguments

crobert-1:
The core distribution of the collector now supports multiple "--config" input arguments, so this change updates the Splunk
distro to also support this. This change also adds new flag parsing functionality to use golang's flag library, instead of manual parsing. Testing has also been added to ensure the config server works for multiple files, flag parsing works as expected, and the collector can properly use multiple input config files.


dmitryax:
@crobert-1 can you please take a look at the failing integration test `TestSpecifiedContainerConfigDefaultsToCmdLineArgIfEnvVarConflict`? I seems like it can be related

cmd/otelcol/flags.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"flag"
+	"io"
+	"strings"
+
+	"go.opentelemetry.io/collector/service/featuregate"
+)
+
+var (
```
pmcollins, line 25: What do you think about putting these variables into a struct and returning that from a function that takes the command line args? Like this:
```
func main() {
	flags, err := parseFlags(os.Args[1:])
	// ...
}

type flags struct {
	a bool
	b int
	c string
}

func parseFlags(args []string) (flags, error) {
	flagSet := new(flag.FlagSet)
	out := flags{}
	flagSet.BoolVar(&out.a, "a", false, "usage")
	flagSet.IntVar(&out.b, "b", 42, "usage")
	flagSet.StringVar(&out.c, "c", "default", "usage")
	err := flagSet.Parse(args)
	return out, err
}
```
Then we can avoid using package (global) variables and we have a testable, self-contained function.

pmcollins: Awesome. Yeah I think the Flag suffixes can go away now since the struct is called flags.

cmd/otelcol/flags.go
```diff
@@ -0,0 +1,109 @@
...
+	configFlags           *stringArrayValue
+	setFlags              *stringArrayValue
+	gatesList             featuregate.FlagValue
+	helpFlag              bool
+	noConvertConfigFlag   bool
+	versionFlag           bool
+	memBallastSizeMibFlag int
+}
+
+var defaultUndeclaredFlag = -1
```
pmcollins, line 36: I think this can be `const` and moved up, no?

cmd/otelcol/flags.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (f *flags) getConfigFlags() []string {
```
pmcollins, line 38: I don't have a strong opinion on this, but I'm wondering if these two getters can go away and we just use direct field access, since we're using direct field access elsewhere.

cmd/otelcol/main.go
```diff
@@ -239,63 +167,67 @@ func checkRuntimeParams() {
...
-		if _, err := os.Stat(configPathVar); err != nil {
-			log.Fatalf("Unable to find the configuration file (%s) ensure %s environment variable is set properly", configPathVar, configEnvVarName)
-		}
+	if configYaml != "" {
+		log.Printf("Both %s and %s were specified. Using %s environment variable value %s for this session", configEnvVarName, configYamlEnvVarName, configEnvVarName, configPathVar)
+	}
 
-		os.Args = append(os.Args, "--config="+configPathVar)
+	if !inputFlags.configFlags.contains(configPathVar) {
+		inputFlags.configFlags.Set(configPathVar)
```
pmcollins, line 204: GoLand complains about these as unhandled error (even though the implementation never returns an error). Can we ignore explicitly?

```
_ = inputFlags.configFlags.Set(configPathVar)
```

cmd/otelcol/main.go
```diff
@@ -65,8 +62,11 @@ func main() {
 	// TODO: Use same format as the collector
 	log.SetFlags(log.Ldate | log.Ltime | log.Lshortfile)
 
-	if !hasFlag("-h") && !hasFlag("--help") && !hasFlag("-v") && !hasFlag("--version") {
-		checkRuntimeParams()
+	// Core flag parser will handle errors, we don't have to handle them here.
+	inputFlags, _ := parseFlags(os.Args[1:])
```
dmitryax, line 66: Why ignoring error? Shouldn't the program exit and show usage on invalid arguments?

internal/configprovider/config_source_provider.go
```diff
@@ -109,6 +118,12 @@ func (c *configSourceConfigMapProvider) Get(context.Context) (*config.Map, error
 		return nil, err
 	}
 
+	// If we've already initiated and started a server, shut it down and create a new one for
+	// updated config sources.
+	if c.configServer != nil {
```
pmcollins, line 123: Do you think it would be possible and desirable to move the server start (and stop) code out of this function? It seems like starting a server is an unfortunate side effect of this Get function.

crobert-1: I originally wanted to be able to do that, but there isn't any clean and simple way to make it work. The config server is a Splunk distribution-specific thing, and the only call into the Splunk distro is a `Retrieve` call made on a single config location. The Splunk distro doesn't have the high level view of how many config sources there are, or what the configs are. These methods know if a config has been seen one before, but it don't know if the current one is the last, which is really what is needed to know to make it a simple "start once" operation.

The clean option would be to have a single call to start the server after we've retrieved all the config sources. However, this would require a change in the core repository, but that doesn't work since there's no concept of the config server there.

cmd/otelcol/flags.go
crobert-1: I'll remove the `Flag` suffixes, thanks!

cmd/otelcol/flags.go
crobert-1: Sure, we can probably remove them since they're not needed at this point.

cmd/otelcol/main.go
crobert-1: Fixed, thanks!

cmd/otelcol/flags.go
crobert-1: Will do, thanks!

cmd/otelcol/main.go
crobert-1: Good point. I was originally following existing functionality, but it's much clearer to just directly raise errors.

This is now the message on failure, any feedback is welcome!

```
crobert@ ~/dev/splunk-otel-collector $ ./bin/otelcol --invalid_flag
2022/05/20 15:06:32 main.go:68: Error: flag provided but not defined: -invalid_flag
Use "--help" to show valid usage
```

crobert-1:
(I'm investigating the failing test to see how it's related)

tests/general/container_test.go
```diff
@@ -90,7 +90,7 @@ func TestSpecifiedContainerConfigDefaultsToCmdLineArgIfEnvVarConflict(t *testing
 			if strings.Contains(
 				log.Message,
 				`Both environment variable SPLUNK_CONFIG and flag '--config' were specified. `+
-					`Using the flag value /etc/config.yaml and ignoring the environment variable value `+
+					`Using the flag values and ignoring the environment variable value `+
```
rmfitzpatrick, line 93: Is logging the values no longer desirable?

rmfitzpatrick:
I'd strongly recommend adding an integration test that uses multiple configs to be able to better vet this over time. Maybe one file for a host metrics receiver, one for a processor, and another for an exporter?

cmd/otelcol/flags.go
```diff
@@ -0,0 +1,101 @@
...
+		configs:   new(stringArrayValue),
+		sets:      new(stringArrayValue),
+		gatesList: featuregate.FlagValue{},
+	}
+
+	// This is an internal flag parser, it shouldn't give any output to user.
+	flagSet.SetOutput(io.Discard)
+
+	// Need to account for full flag names and abbreviations
+	flagSet.BoolVar(&out.help, "h", false, "")
```
rmfitzpatrick, line 75: May be helpful to note

```suggestion
        // usage str isn't necessary for these since it's provided by actual collector cobra command flags from core
	flagSet.BoolVar(&out.help, "h", false, "")
```

cmd/otelcol/flags_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFlagParseFailure(t *testing.T) {
+	oldArgs := os.Args
+
+	// Parsing should stop once an unspecified flag is found
+	os.Args = []string{"otelcol", "--invalid-flag", "100", "--version", "true"}
+	inputFlags, _ := parseFlags(os.Args[1:])
```
rmfitzpatrick, line 29: This can introduce some maintenance overhead but it's helpful to capture what we expect to be the case for possible errors. Similarly w/ `assert.NoError(t, err)`:

```suggestion
	inputFlags, err := parseFlags(os.Args[1:])
	assert.EqualError(t, err, "expected content")
```

internal/configprovider/config_source_provider.go
```diff
@@ -62,7 +62,16 @@ func (c *configSourceConfigMapProvider) Retrieve(
 	if err != nil {
 		return config.Retrieved{}, err
 	}
-	c.wrappedRetrieved = wr
+
+	existingMap, _ := c.wrappedRetrieved.AsMap()
```
rmfitzpatrick, line 66: Even though hardcoded as nil atm, it's probably best to handle and expose the error.

go.mod
```diff
@@ -474,3 +474,5 @@ replace github.com/googleapis/gnostic v0.5.6 => github.com/googleapis/gnostic v0
 
 // required to drop dependency on deprecated git.apache.org/thrift.git
 exclude go.opencensus.io v0.19.1
+
+replace go.opentelemetry.io/collector v0.0.0-00010101000000-000000000000 => ../opentelemetry-collector
```
rmfitzpatrick, line 478: Mind removing?

internal/configprovider/config_source_provider.go
```diff
@@ -109,6 +118,12 @@ func (c *configSourceConfigMapProvider) Get(context.Context) (*config.Map, error
 		return nil, err
 	}
 
+	// If we've already initiated and started a server, shut it down and create a new one for
+	// updated config sources.
+	if c.configServer != nil {
+		c.configServer.shutdown()
```
rmfitzpatrick, line 124: Rather than cycling like this can we slightly alter the config server's handlers to render the initial and effective by reference at call time instead of marshaling beforehand?

cmd/otelcol/flags.go
crobert-1: Updated, thanks!

cmd/otelcol/flags_test.go
crobert-1: I've updated.

go.mod
crobert-1: Removed, good catch!

internal/configprovider/config_source_provider.go
crobert-1: Updated

internal/configprovider/config_source_provider.go
crobert-1: Sorry, but I'm not sure I understand your request here. I'm not following your thought as far as ordering of events. Could you expand on your thought `to render the initial and effective by reference at call time`?

tests/general/container_test.go
crobert-1: There's another log message that shows up directly after this that includes the values that `config` is set to. Here's the shell output:

```
crobert@ ~/dev/splunk-otel-collector $ ./bin/otelcol_darwin_amd64 --config processlistconfig.yaml --config services.yml 
2022/05/23 15:26:44 main.go:184: Both environment variable SPLUNK_CONFIG and flag '--config' were specified. Using the flag values and ignoring the environment variable value /etc/fake_config.yaml in this session
2022/05/23 15:26:44 main.go:222: Set config to [processlistconfig.yaml,services.yml]
```
It seemed unnecessary to have the flag values for both log messages, and more consistent if we just log what the config sources are every time, regardless if it's coming from an environment variable or command line flag.

Let me know if you think it's still valuable to have both though!

internal/configprovider/config_source_provider.go
rmfitzpatrick: The initial and effective map references would be updated for each `Retrieve` call and the mux handlers would be updated to not use closures but instead marshal initial and redacted effective dynamically. I think the added overhead per request would be negligible.

```go
if c.configServer == nil {
    c.configServer = newConfigServer(c.logger)
    if err = c.configServer.start(); err != nil {
        return nil, err
    }
}

// these may need locks, as well as the updated mux handlers (not shown)*
c.configServer.setInitial(c.wrappedRetrieved.Map.ToStringMap())
c.configServer.setEffective(effectiveMap.ToStringMap())
```

crobert-1: Updated to use locking to serve initial and effective configs by reference.

crobert-1:
> I'd strongly recommend adding an integration test that uses multiple configs to be able to better vet this over time. Maybe one file for a host metrics receiver, one for a processor, and another for an exporter?

Added an integration test to cover this case, thanks!

tests/testutils/collector_process_test.go
```diff
@@ -134,3 +136,48 @@ func TestCollectorProcessWithInvalidPaths(t *testing.T) {
 	err = collector.Shutdown()
 	require.NoError(t, err)
 }
+
+func TestCollectorProcessWithMultipleConfigs(t *testing.T) {
```
rmfitzpatrick, line 140: These tests are for the actual testutils helper functionality. Mind moving to a new ~`tests/general/process_test.go`? It may make sense to add a new `tc.SkipIfNotProcess()` check: https://github.com/signalfx/splunk-otel-collector/blob/ed553cbcaecc3f3fa86884b1e0a986a9e8ba8518/tests/testutils/testcase.go#L70

crobert-1: I'll move the test, thanks for the pointer!

I actually don't think it would make sense to create a `SkipIfNotProcess` check here. `SkipIfNotContainer` is required to make sure the environment is properly setup with a config path to use by default. However, for the `CollectorProcess` utility, the collector default path is set automatically internally, so no default is required.  Let me know if I'm missing something though, happy to add one if there's some extra functionality you think is necessary.

rmfitzpatrick: `SkipIfNotContainer` checks to see if `SPLUNK_OTEL_COLLECTOR_IMAGE` is set, which is the mechanism used to run the integration tests against the target image as opposed to a built executable, and doesn't really have to do w/ the collector configuration: https://github.com/signalfx/splunk-otel-collector/tree/main/tests/testutils#testcase

https://github.com/signalfx/splunk-otel-collector/blob/main/tests/testutils/testcase.go#L119

By not skipping the added test if the env var is set (which could later be turned into the requested helper) this makes a built executable required for the integration tests to pass which isn't necessarily a requirement otherwise.

tests/general/process_test.go
```diff
@@ -0,0 +1,59 @@
...
+				return true
+			}
+		}
+		return false
+	}, 20*time.Second, time.Second)
+
+	require.Eventually(t, func() bool {
+		for _, log := range logs.All() {
+			// Confirm collector starts and runs successfully
+			if strings.Contains(log.Message, "Everything is ready. Begin running and processing data.") {
```
rmfitzpatrick, line 50: The check seems ok for now since it's used as a measure of collector container readiness, but I think ideally we'd get the result of the config server's effective config or receive metrics w/ the metric receiver sink to know that all the desired components are working.

crobert-1: I've added checks to make sure the config server is showing the proper configuration contents, thanks for the suggestion!

tests/general/process_test.go
```diff
@@ -0,0 +1,100 @@
...
+				return true
+			}
+		}
+		return false
+	}, 20*time.Second, time.Second)
+
+	require.Eventually(t, func() bool {
+		// Need expected configs for comparison to config server contents. Sections in the config server can be in
```
rmfitzpatrick, line 61: This seems ok for now since an unrelated issue but ideally we'd have the config server behave more reliably.

crobert-1: More than anything else this eventually loop is to just brush over the timing of starting the collector and the config server. The config server is expected to be up and running by the time we see the "Everything is ready" message, but since these are asynchronous methods, we'd have to wait for the first to finish successfully and then run this one a single time. I'm sure it could be done but I didn't think it was important enough to spend more time trying to figure it out for a test.

tests/testutils/collector_process_test.go
crobert-1: To make sure I'm understanding properly, you're saying it would be useful to add this check to ensure that the executable is already built successfully for the process tests to run, right?

rmfitzpatrick: This is something that we can address later as the divergence between test suites grows.

Issue #1340, opened 2022-03-09
Unable to configure hostmetrics receiver from container environment


dmitryax:
Since hostmetrics receiver uses gopsutil library, the only way to configure host metrics collection from container is to setup environment variables as it's done in [the helm chart](https://github.com/signalfx/splunk-otel-collector-chart/blob/f5b19188bdac80928e11a8a373e7c7abfd021436/helm-charts/splunk-otel-collector/templates/daemonset.yaml#L288-L299). However appeared that signalfx-agent dependency [overrides all the env variables](https://github.com/signalfx/signalfx-agent/blob/72362e5f82b42bd5a4060af2f28500d7df8a75ab/pkg/utils/gopsutilhelper/gopsutilhelper.go) making hostmetrics receiver collect metrics from container itself, not from the host machine.

dmitryax:
This one is not resolved yet

PR #1615, opened 2022-05-26
Don't override gopsutil env vars in SA receiver

rmfitzpatrick:
Resolves https://github.com/signalfx/splunk-otel-collector/issues/1340 * but is a breaking change to prevent an existing compatibility issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🛑 Breaking changes 🛑
+
+- Respect set [gopsutil environment variables](https://github.com/shirou/gopsutil#usage) in the `smartagent` receiver to provide better experience with the `hostmetrics` receiver, even if set in the `smartagent` extension configuration ([#1615](https://github.com/signalfx/splunk-otel-collector/pull/1615))
+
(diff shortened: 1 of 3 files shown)


rmfitzpatrick:
@dmitryax I also considered using the https://github.com/signalfx/splunk-otel-collector/tree/main/internal/extension/smartagentextension#signalfx-smart-agent-extension in the helm chart instead of these changes but either way the behavior is a little unintuitive.

dmitryax:
I don't think we should solve this issue until https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/5879 is resolved. Once this is merge collector will be spamming errors in any k8s environment

atoulme:
I note https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/5879 was just marked resolved.

rmfitzpatrick:
With https://github.com/signalfx/splunk-otel-collector/pull/3509 I think this PR needs to be updated to have the SA config paths respect the gopsutil common env vars but implementing the `defaults.Setter` interface. Doing that when able.

rmfitzpatrick:
Not going to move forward with this approach for now. Manually specifying in the SA config remains the support path.

PR #1626, opened 2022-06-01
Add transform processor to our distro

emaderer:


pmcollins (approved):

jeffreyc-splunk:
We should add it to the list in https://github.com/signalfx/splunk-otel-collector/blob/main/docs/components.md

PR #1694, opened 2022-06-17
Upgrade metricstransform processor to the latest main commit

dmitryax:
to pick up https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/10817 that fixes a few issues.

This change also updates tests that were relying on the following bugs in metricstransform processor that are fixed now:
- non-cumulative sums used to be translated to gauges
- scope info (name + version) used to be dropped


Makefile
```diff
@@ -38,11 +38,13 @@ MISSPELL=misspell -error
 MISSPELL_CORRECTION=misspell -w
 
 BUILD_INFO_IMPORT_PATH=github.com/signalfx/splunk-otel-collector/internal/version
+BUILD_INFO_IMPORT_PATH_TESTS=github.com/signalfx/splunk-otel-collector/tests/internal/version
```
dmitryax, line 41: Need to use introduce another `version` package to avoid cross modules dependency. The root module already depends on `tests` module, so we cannot make `tests` module depend on the root module. 

I don't think we need `tests` module at all, but that can be handled separately.

tests/testutils/resource_metrics.go
```diff
@@ -43,6 +45,8 @@ const (
 	IntNonmonotonicUnspecifiedSum    MetricType = "IntNonmonotonicUnspecifiedSum"
 )
 
+const buildVersionPlaceholder = "<BUILD VERSION TO BE SET DYNAMICALLY>"
```
rmfitzpatrick, line 48: nit - this is a little verbose imo

```suggestion
const buildVersionPlaceholder = "<FROM_BUILD>"
```

Makefile
rmfitzpatrick: We could break out testutils since I think only the available port finder is shared but that can be done in another change that separates its tests from the actual integration runs.

dmitryax: The `testutils` has to be put to another new module in that case.  Do we need `tests` module to reduce the build side?

rmfitzpatrick (approved):

PR #1696, opened 2022-06-17
translatesfx: add support for multiple SA monitors of the same type

pmcollins:
There was a bug in translatesfx where if the SA config had multiple monitors of the same type, only one corresponding Otel receiver would be created in the translated config. This change intends to fix that bug.


cmd/translatesfx/translatesfx/otel.go
```diff
@@ -524,21 +524,59 @@ func saMonitorToRCReceiver(monitor map[string]interface{}, observers []interface
...
+		attrs: map[string]interface{}{
 			"rule":   rcr,
 			"config": monitor,
 		},
-	}, warnings
+	}
+	return
+}
+
+type component struct {
```
rmfitzpatrick, line 538: preferential but it may be clearer to have these declarations before their use

cmd/translatesfx/translatesfx/otel.go
```diff
@@ -524,21 +524,59 @@ func saMonitorToRCReceiver(monitor map[string]interface{}, observers []interface
...
+	attrs          map[string]interface{}
+	provisionalKey string
```
rmfitzpatrick, line 540: nit - how is this provisional? isn't it more of a ~`baseType` or `baseName`?

cmd/translatesfx/translatesfx/otel_test.go
```diff
@@ -318,6 +318,44 @@ func TestInfoToOtelConfig_MetricsToExclude_Monitor(t *testing.T) {
...
+	}
+	sort.Strings(keys)
+	assert.Equal(t, []string{"mycomponent/0", "mycomponent/1"}, keys)
+}
+
+func TestInfoToOtelConfig_DuplicateMonitors(t *testing.T) {
+	cfg, _ := yamlToOtelConfig(t, "testdata/sa-duplicate-monitors.yaml")
+	assert.Equal(t, 2, len(cfg.Receivers))
+	metrics := cfg.Service.Pipelines["metrics"]
+	assert.Equal(t, []string{"smartagent/sql/0", "smartagent/sql/1"}, metrics.Receivers)
```
rmfitzpatrick, line 356: Do we have any coverage that the different receiver attributes resolve as desired in the final config?

cmd/translatesfx/translatesfx/otel.go
pmcollins: Sure -- I've moved them to another file since this one is getting rather large.

cmd/translatesfx/translatesfx/otel.go
pmcollins: Sure -- changed to `baseName` and added a comment.

cmd/translatesfx/translatesfx/otel_test.go
pmcollins: I had only tested that manually -- added coverage.-

rmfitzpatrick (approved):

PR #1757, opened 2022-07-07
Add log enrichment example

atoulme:


examples/log-enrichment/otel-collector-config.yml
```diff
@@ -0,0 +1,49 @@
...
+        # Whether to disable gzip compression over HTTP. Defaults to false.
+        disable_compression: false
+        # HTTP timeout when sending data. Defaults to 10s.
+        timeout: 10s
+        # Whether to skip checking the certificate of the HEC endpoint when sending data over HTTPS. Defaults to false.
+        # For this demo, we use a self-signed certificate on the Splunk docker instance, so this flag is set to true.
+        insecure_skip_verify: true
+processors:
+    batch:
+    attributes/add_metadata:
```
rmfitzpatrick, line 26: Since resource attributes are converted to fields would it be preferable to use the resource processor since we don't need to modify each record? If record attributes are the point of the demo then that makes sense as well.

rmfitzpatrick (approved):

PR #1812, opened 2022-07-27
Initial discovery receiver readme

rmfitzpatrick:
This readme describes the initial desired functionality of the proposed discovery receiver. The actual receiver implementation will be added in a series of commits, ideally based on some feedback from this doc content. There are a few limitations to unified log record content based on reported telemetry and I'd be happy to describe or attempt to reconcile them w/ suggestions/requests.

### DIFF
--- a/internal/receiver/discoveryreceiver/README.md
+++ b/internal/receiver/discoveryreceiver/README.md
@@ -1,0 +1,334 @@
+# Discovery Receiver
+
+| Status                   |                  |
+| ------------------------ | ---------------- |
+| Stability                | [in-development] |
+| Supported pipeline types | logs             |
+| Distributions            | [Splunk]         |
+
+The Discovery receiver is a receiver compatible with logs pipelines that allows you to test the functional
+status of any receiver type whose target is reported by an Observer. It provides configurable `status`
+match rules that evaluate the generated receiver's emitted metrics (if any), or component-level log statements
+via the instance's [zap.Logger](https://pkg.go.dev/go.uber.org/zap). It works similarly to the
+[Receiver Creator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/receivercreator/README.md)
+(it actually wraps an internal instance of one), but the resulting dynamically-instantiated receivers don't actually
+report their metric content to your metrics pipelines. Instead, the metrics are intercepted by an internal metrics
+consumer capable of translating desired metrics to log records based on the `status: metrics` rules you define. All
+component-level log statements are similarly intercepted by a log evaluator, and can be translated to emitted log
+records based on the `status: statements` rules you define.
+
+The receiver also allows you to emit log records for all
+[Endpoint](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/observer/endpoints.go)
+events from the specified `watch_observers`. This way you can report your environment as observed by platform-specific
+observers in real time, with or without discovering receiver statuses:
+
+```yaml
+extensions:
+  docker_observer:
+receivers:
+  discovery:
+    watch_observers: [docker_observer]
+    log_endpoints: true
+exporters:
+  logging:
+    logLevel: debug
+service:
+  extensions: [docker_observer]
+  pipelines:
+    logs:
+      receivers: [discovery]
+      exporters: [logging]
+```
+
+```
+2022-07-27T19:00:32.305Z	info	LogsExporter	{"kind": "exporter", "data_type": "logs", "name": "logging", "#logs": 3}
+2022-07-27T19:00:32.306Z	info	ResourceLog #0
+Resource SchemaURL:
+Resource labels:
+     -> event.type: STRING(endpoint.added)
+ScopeLogs #0
+ScopeLogs SchemaURL:
+InstrumentationScope
+LogRecord #0
+ObservedTimestamp: 2022-07-27 19:00:32.30572809 +0000 UTC
+Timestamp: 2022-07-27 19:00:32.305727862 +0000 UTC
+Severity: info
+Body: {"alternate_port":5432,"command":"postgres -c shared_preload_libraries=pg_stat_statements -c wal_level=logical -c max_replication_slots=2","container_id":"58e71612910cd3e2a89d809f1b53fee779c4c81d9bbf36ef37f4c05b04037353","endpoint":"172.17.0.4:5432","host":"172.17.0.4","id":"58e71612910cd3e2a89d809f1b53fee779c4c81d9bbf36ef37f4c05b04037353:5432","image":"postgres","labels":{},"name":"naughty_heisenberg","port":5432,"tag":"latest","transport":"TCP","type":"container"}
+Attributes:
+     -> type: STRING(endpoint)
+     -> image: STRING(postgres)
+     -> tag: STRING(latest)
+     -> port: STRING(5432)
+     -> alternate_port: STRING(5432)
+     -> host: STRING(172.17.0.4)
+     -> transport: STRING(TCP)
+     -> labels: STRING(map[])
+     -> endpoint: STRING(172.17.0.4:5432)
+     -> name: STRING(naughty_heisenberg)
+     -> command: STRING(postgres -c shared_preload_libraries=pg_stat_statements -c wal_level=logical -c max_replication_slots=2)
+     -> container_id: STRING(58e71612910cd3e2a89d809f1b53fee779c4c81d9bbf36ef37f4c05b04037353)
+     -> id: STRING(58e71612910cd3e2a89d809f1b53fee779c4c81d9bbf36ef37f4c05b04037353:5432)
+Trace ID:
+Span ID:
+Flags: 0
+LogRecord #1
+ObservedTimestamp: 2022-07-27 19:00:32.305771365 +0000 UTC
+Timestamp: 2022-07-27 19:00:32.305771316 +0000 UTC
+Severity: info
+Body: {"alternate_port":0,"command":"nginx -g daemon off;","container_id":"2567fdbc764706d29120b01efcc3a310d87e9e121ec9debbc977d66f5497cdda","endpoint":"172.17.0.2:80","host":"172.17.0.2","id":"2567fdbc764706d29120b01efcc3a310d87e9e121ec9debbc977d66f5497cdda:80","image":"nginx","labels":{"maintainer":"NGINX Docker Maintainers \u003cdocker-maint@nginx.com\u003e"},"name":"ecstatic_davinci","port":80,"tag":"latest","transport":"TCP","type":"container"}
+Attributes:
+     -> type: STRING(endpoint)
+     -> labels: STRING(map[maintainer:NGINX Docker Maintainers <docker-maint@nginx.com>])
+     -> id: STRING(2567fdbc764706d29120b01efcc3a310d87e9e121ec9debbc977d66f5497cdda:80)
+     -> tag: STRING(latest)
+     -> port: STRING(80)
+     -> alternate_port: STRING(0)
+     -> command: STRING(nginx -g daemon off;)
+     -> container_id: STRING(2567fdbc764706d29120b01efcc3a310d87e9e121ec9debbc977d66f5497cdda)
+     -> transport: STRING(TCP)
+     -> name: STRING(ecstatic_davinci)
+     -> image: STRING(nginx)
+     -> host: STRING(172.17.0.2)
+     -> endpoint: STRING(172.17.0.2:80)
+Trace ID:
+Span ID:
+Flags: 0
+LogRecord #2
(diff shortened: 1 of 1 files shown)


internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
+# Discovery Receiver
+
+| Status                   |                  |
+| ------------------------ | ---------------- |
+| Stability                | [in-development] |
+| Supported pipeline types | logs             |
+| Distributions            | [Splunk]         |
+
+The Discovery receiver is a logs pipeline compatible receiver that provides the ability to test the functional
```
aurbiztondo-splunk, line 9: How about: "The Discovery receiver is a receiver compatible with logs pipeline which allows you to test the functional status..."

internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
...
+status of any receiver type whose target is reported by an Observer. It does this by providing configurable `status`
+match rules capable of evaluating the generated receiver's emitted metrics, if any, or component-level log statements
+(via the instance's [zap.Logger](https://pkg.go.dev/go.uber.org/zap)). It works similarly to the
+[Receiver Creator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/receivercreator/README.md)
+(in fact it wraps an internal instance of one) but the resulting dynamically-instantiated receivers don't actually
+report their metric content to your metrics pipelines. Instead the metrics are intercepted by an internal metrics
+consumer capable of translating desired metrics to log records based on `status: metrics` rules you define. All
+component-level log statements are similarly intercepted by a log evaluator and can be translated to emitted log records
+based on `status: statements` rules you define.
+
```
aurbiztondo-splunk, line 19: Suggestion:

The Discovery receiver is a receiver compatible with logs pipeline which allows you to test the functional status of any receiver type whose target is reported by an Observer. It provides configurable `status`
match rules which evaluate the generated receiver's emitted metrics (if any), or component-level log statements, via the instance's [zap.Logger](https://pkg.go.dev/go.uber.org/zap). 

It works similarly to the [Receiver Creator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/receivercreator/README.md)
(in fact, it wraps an internal instance of one), but the resulting dynamically-instantiated receivers don't actually
report their metric content to your metrics pipelines. Instead, the metrics are intercepted by an internal metrics
consumer capable of translating desired metrics to log records based on the `status: metrics` rules you define. All
component-level log statements are similarly intercepted by a log evaluator, and can be translated to emitted log records based on the `status: statements` rules you define.

internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
...
+The receiver also provides the ability to emit log records for all
+[Endpoint](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/observer/endpoints.go)
+events from the specified `watch_observers` so that you can report your environment as observed by platform specific
+observers in real time, even without trying to discover receiver statuses:
+
```
aurbiztondo-splunk, line 24: The receiver also allows you to emit log records for all
[Endpoint](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/observer/endpoints.go)
events from the specified `watch_observers`. Thus you can report your environment as observed by platform specific
observers in real time, even without trying to discover receiver statuses:

internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
...
+The following Collector configuration will create a Discovery receiver instance that receives
+endpoints from a [Docker Observer](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/observer/dockerobserver/README.md)
+that reports log records denoting the status of a [Smart Agent receiver](https://github.com/signalfx/splunk-otel-collector/blob/main/internal/receiver/smartagentreceiver/README.md)
+configured to use the `collectd/redis` monitor. The `status` mapping comprises entries that signal the Smart Agent
+receiver has been instantiated with a `successful`, `partial`, or `failed` status, based on reported `metrics` or
+recorded application log `statements`.
+
+The Redis receiver will warrant a status log record from the Discovery receiver of...
+
```
aurbiztondo-splunk, line 130: The Redis receiver warrants the following a status log record from the Discovery receiver:

internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
...
+      receivers:
+        - discovery
+      exporters:
+        - logging
+```
+
+Given this configuration, if the Discovery receiver's Docker observer instance reports an active Redis container and
+the Smart Agent receiver's associated `collectd/redis` monitor is able to generate metrics for the container, something
+like the following log record would be emitted by the receiver:
+
```
aurbiztondo-splunk, line 192: Suggestion:

Given this configuration, if the Discovery receiver's Docker observer instance reports an active Redis container, and the Smart Agent receiver's associated `collectd/redis` monitor is able to generate metrics for the container, the receiver will emit something similar to the following log record:

internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
...
+     -> metric.name: STRING(counter.expired_keys)
+     -> discovery.status: STRING(successful)
+Trace ID:
+Span ID:
+Flags: 0
+```
+
+If instead the Docker observer reports an active Redis container but the `collectd/redis` authentication information is
+incorrect, the following log record would be emitted by the Discovery receiver:
+
```
aurbiztondo-splunk, line 222: Suggestion: 

Instead, if the Docker observer reports an active Redis container but the `collectd/redis` authentication information is incorrect, the Discovery receiver will emit the following log record:

internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
...
+### Match
+
+**One of `regexp`, `strict`, or `expr` is required.**
+
+| Name | Type | Default | Docs |
+| ---- | ---- | ------- | ---- |
+| `regexp` | string | <no value> | The regexp pattern to evaluate reported received metric names or component log statements |
+| `strict` | string | <no value> | The string literal to compare equivalence against reported received metric names or component log statements |
+| `expr` | string | <no value> | The expr program run with the reported received metric names or component log statements (vm env TBD) |
+| `first_only` | bool | false | Whether to emit only one log record for the first matching metric or log statement, ignoring all subsequent matches. |
```
aurbiztondo-splunk, line 318: Remove . at the end, or add dots everywhere

aurbiztondo-splunk (approved):
I've left a few suggestions, LMK if you have any questions!

internal/receiver/discoveryreceiver/README.md
rmfitzpatrick: I don't know why, but I'm a fan of passive voice (updated).

internal/receiver/discoveryreceiver/README.md
rmfitzpatrick: updated

internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
...
+Redis server is available at the endpoint.
+
+```yaml
+extensions:
+  docker_observer:
+receivers:
+   discovery:
+     watch_observers: [docker_observer]
+     receivers:
+       smartagent/redis:
```
aunshc, line 145: Thinking about the next steps, as we add receivers to the discovery receiver subconfiguration in phases - is there a way to enable or disable specific receiver instrumentation?

Context: Based on recent feedback from customers, it would be helpful if we can detect the observer endpoints (discovery) and allow them to then enable the already injected receiver specific configuration for instrumentation.

internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
...
+```
+
+## Config
+
+### Main
+
+| Name | Type | Default | Docs |
+| ---- | ---- | ------- | ---- |
+| `watch_observers` (required) | []string | <no value> | The array of Observer extensions to receive Endpoint events from |
+| `log_endpoints` | bool | false | Whether to emit log records for Observer Endpoint events |
```
aunshc, line 296: Correct me if I am wrong, with the log_endpoints option, a new receiver type (eg.  observer_events receiver) would not be needed. Does that mean we are following the inventory events [representation](https://docs.google.com/document/d/17gc7i9_zRJn_ztKNbk-jvyXMhiXXn4tjvmojW0NNf0U/edit#) to help make it easier to integrate with the inventory service once it is available?

internal/receiver/discoveryreceiver/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+| Name | Type | Default | Docs |
+| ---- | ---- | ------- | ---- |
+| `regexp` | string | <no value> | The regexp pattern to evaluate reported received metric names or component log statements |
+| `strict` | string | <no value> | The string literal to compare equivalence against reported received metric names or component log statements |
+| `expr` | string | <no value> | The expr program run with the reported received metric names or component log statements (vm env TBD) |
+| `first_only` | bool | false | Whether to emit only one log record for the first matching metric or log statement, ignoring all subsequent matches |
+| `record` | LogRecord | <no value> | The emitted log record content |
+
+### LogRecord
```
aunshc, line 321: Are these the log records that could be consumed by LO and then visualized on data setup views? 

Thinking aloud, (this is not specific to this PR; moreover a question we on the product side need to answer), if a customer does not have a LO entitlement (review inventory on the data setup wizard only visualizes logs if they do), what is the best way for us to guide the customer for visibility?

internal/receiver/discoveryreceiver/README.md
rmfitzpatrick: I think you're referring to the custom config map provider that is being worked on that uses this receiver internally so it's not part of the design for this component directly. All metrics produced by receivers created within the discovery receiver are consumed in the component itself and not forwarded to a metric pipeline since it's logs compatible only.

For the map provider there would be something like `discovery.receivers.<receiver type>.enabled=false` to prevent the discovery config from forwarding that receiver's config.

internal/receiver/discoveryreceiver/README.md
rmfitzpatrick: Right, since this receiver provides an observer.Notify implementation and emits log records it's straightforward for it to also log all endpoint events from configured observers. My thought was to emit log records that contain endpoints themselves since they're an established model in the collector: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/observer/endpoints.go (content documented here: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/receivercreator/README.md#rule-expressions). We could provide conversion in the signalfx/hec exporters if their form isn't suitable for backend requirements.

internal/receiver/discoveryreceiver/README.md
rmfitzpatrick: This seems applicable to log records from the collector in general and without LO. One way would be to establish practices for using attributes or transform processor to insert the required record attributes to convert them to metric events (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/splunk/common.go#L27, https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/8e4a062fa9aef1543b477a4b3469085d4b1c95db/exporter/signalfxexporter/internal/translation/logdata_to_signalfxv2.go#L49). The other would be to make the discovery receiver a traces pipeline compatible component and emit traces for each endpoint event with endpoints as spans.

internal/receiver/discoveryreceiver/README.md
aunshc: Yes, that's right was referring to the config provider. Thanks!

internal/receiver/discoveryreceiver/README.md
aunshc: Great ideas! I'll set up some time to chat about our direction moving forward based on these recommendations.

dmitryax: I believe we should use SignalFx metric events since the discovered receivers are mostly for metrics. APM also can be disabled on Splunk O11y side

dmitryax:
There is ongoing work in OTel spec to establish API and semantic conventions for events using Log Record representation https://github.com/open-telemetry/opentelemetry-specification/pull/2676. I'd suggest taking a look and see if we can align with that for future compatibility

rmfitzpatrick:
> Another Q: why it has to be added in Splunk distro, not contrib?

It doesn't have to be but my proposed path is to house it in our distribution for initial versions and development (especially when used by our upcoming map provider), then offering it to contrib once it's more established and things like component status reporting are available.

jvoravong (approved):

PR #1830, opened 2022-08-03
Initial discovery receiver skeleton

rmfitzpatrick:
These changes introduce the initial Discovery receiver component methods and config validation (proposed readme: https://github.com/signalfx/splunk-otel-collector/pull/1812).

### DIFF
--- a/internal/receiver/discoveryreceiver/config.go
+++ b/internal/receiver/discoveryreceiver/config.go
@@ -1,0 +1,162 @@
+// Copyright Splunk, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package discoveryreceiver
+
+import (
+	"fmt"
+
+	"go.opentelemetry.io/collector/config"
+	"go.uber.org/multierr"
+)
+
+var (
+	_ config.Receiver = (*Config)(nil)
+
+	allowedStatusTypes = []string{"successful", "partial", "failed"}
+	allowedStatuses    = func() map[string]struct{} {
+		sm := map[string]struct{}{}
+		for _, status := range allowedStatusTypes {
+			sm[status] = struct{}{}
+		}
+		return sm
+	}()
+
+	allowedMatchTypes = []string{"regexp", "strict", "expr"}
+)
+
+type Config struct {
+	// Receivers is a mapping of receivers to discover to their receiver creator configs
+	// and evaluated metrics and application statements, which are used to determine component status.
+	Receivers               map[config.ComponentID]ReceiverEntry `mapstructure:"receivers"`
+	config.ReceiverSettings `mapstructure:",squash"`
+	// The configured Observer extensions from which to receive Endpoint events.
+	// Must implement the observer.Observable interface.
+	WatchObservers []config.ComponentID `mapstructure:"watch_observers"`
+	// Whether to emit log records for all endpoint activity, consisting of Endpoint
+	// content as record attributes.
+	LogEndpoints bool `mapstructure:"log_endpoints"`
+	// Whether to include the receiver config as a base64-encoded "discovery.receiver.config"
+	// resource attribute string value. Will also contain the configured observer that
+	// produced the endpoint leading to receiver creation in `watch_observers`.
+	// Warning: these values will include the literal receiver subconfig from the parent Collector config.
+	// The feature provides no secret redaction and its output is easily decodable into plaintext.
+	EmbedReceiverConfig bool `mapstructure:"embed_receiver_config"`
+}
+
+// ReceiverEntry is a definition for a receiver instance to instantiate for each Endpoint matching
+// the defined rule. Its Config, ResourceAttributes, and Rule will be marshaled to the internal
+// Receiver Creator config.
+type ReceiverEntry struct {
+	Config             map[string]any    `mapstructure:"config"`
+	Status             *Status           `mapstructure:"status"`
+	ResourceAttributes map[string]string `mapstructure:"resource_attributes"`
+	Rule               string            `mapstructure:"rule"`
+}
+
+// Status defines the Match rules for applicable app and telemetry sources.
+// At this time only Metrics and zap logger Statements status source types are supported.
+type Status struct {
+	Metrics    map[string][]Match `mapstructure:"metrics"`
+	Statements map[string][]Match `mapstructure:"statements"`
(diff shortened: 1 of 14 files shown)


internal/receiver/discoveryreceiver/config.go
```diff
@@ -0,0 +1,162 @@
...
+					err = multierr.Combine(err, fmt.Errorf(" %q log record validation failure: %w", statusType, e))
+				}
+			}
+		}
+	}
+	return err
+}
+
+func (lr *LogRecord) validate() error {
+	// TODO: supported severity text validation
```
atoulme, line 160: I don't mind if you want to use your own impl of LogRecord, but it might be easier to use plog.LogRecord if you want to access all the fields. For example, your LogRecord struct doesn't have SeverityText or SeverityNum. You would get that for free if you use plog.LogRecord.

atoulme:
Changes look ok:
- the receiver is not mapped to components, so this is not going to affect the collector yet.
- it would be good to have a README

internal/receiver/discoveryreceiver/receiver.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (d *discoveryReceiver) ConsumeMetrics(ctx context.Context, md pmetric.Metrics) error {
+	// TODO: md -> plog.Logs and evaluation
+	return nil
+}
+
+// observablesFromHost finds configured `watch_observers` extension instances from the host
+// by their ComponentID. It is based on the equivalent logic in the Receiver Creator:
+// https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d6042eda45ec9d8a5df1ae553388eaca67d9d16c/receiver/receivercreator/receiver.go#L79
+func (d *discoveryReceiver) observablesFromHost(host component.Host) (map[config.ComponentID]observer.Observable, error) {
```
pmcollins, line 88: Would it be worthwhile to add a unit test for this method?

rmfitzpatrick:
> * the receiver is not mapped to components, so this is not going to affect the collector yet.

I don't plan on adding this as a component until its initial functionality has been merged (observer event emitting and metrics/statement evaluation are next PRs). There are also some required upstream PRs for this to be fully functional:

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/12801
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/12670

> * it would be good to have a README

This was the first PR and has landed: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/receiver/discoveryreceiver/README.md

atoulme:
> > * the receiver is not mapped to components, so this is not going to affect the collector yet.
> 
> I don't plan on adding this as a component until its initial functionality has been merged (observer event emitting and metrics/statement evaluation are next PRs). There are also some required upstream PRs for this to be fully functional:
No worries.
> > * it would be good to have a README
> 
> This was the first PR and has landed: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/receiver/discoveryreceiver/README.md
I looked for it and missed it. Apologies!

internal/receiver/discoveryreceiver/receiver.go
rmfitzpatrick: Good call, updated.

internal/receiver/discoveryreceiver/config.go
rmfitzpatrick: The config log record is used to express the values that are desired in the actual `plog.LogRecord` item that's going to be produced by the receiver. I don't know if it's possible to use a `plog.LogRecord` in its place given the implementation is in "go.opentelemetry.io/collector/pdata/internal/data/protogen/logs/v1" and doesn't use mapstructure, but considering we only care about specific fields with more convenient accessors I don't think it's a concern.

`Severity` becomes `SeverityText` but considering we may want to specify num as well I can update it.

internal/receiver/discoveryreceiver/config.go
```diff
@@ -0,0 +1,162 @@
...
+		return sm
+	}()
+
+	allowedMatchTypes = []string{"regexp", "strict", "expr"}
+)
+
+type Config struct {
+	// Receivers is a mapping of all receiver types to discovery and whose metrics
+	// and application statements to evaluate for assessing component status.
+	Receivers               map[config.ComponentID]ReceiverEntry `mapstructure:"receivers"`
```
aurbiztondo-splunk, line 42: I don't understand this explanation. "Whose metrics" are those? Receivers? If so, I'd go with:
"Receivers is a mapping of all receiver types. Receiver metrics and application statements are used to evaluate and assess component status."

internal/receiver/discoveryreceiver/config.go
```diff
@@ -0,0 +1,162 @@
...
+	// Must implement the observer.Observable interface.
+	WatchObservers []config.ComponentID `mapstructure:"watch_observers"`
+	// Whether to emit log records for all endpoint activity, consisting of Endpoint
+	// content as record attributes.
+	LogEndpoints bool `mapstructure:"log_endpoints"`
+	// Whether to include the receiver config as a base64-encoded "discovery.receiver.config"
+	// resource attribute string value. Will also contain the configured observer that
+	// produced the endpoint leading to receiver creation.
+	// Warning: these values will include the literal receiver subconfig from the parent Collector config
+	// and provides no secret redaction and is easily decodable into plaintext.
```
aurbiztondo-splunk, line 54: Warning: these values will include the literal receiver subconfig from the parent Collector config. 
They provide no secret redaction, and are easily decodable into plaintext.

aurbiztondo-splunk (approved):
Left a couple of comments!

internal/receiver/discoveryreceiver/config.go
rmfitzpatrick: I meant "receiver types to discover" but clarified a bit.

aurbiztondo-splunk: I'd still break it in two sentences: 
Receivers is a mapping of all receivers to discover. Receiver metrics and application statements are evaluated to assess component status.

rmfitzpatrick: The mapping is from the receivers (their names) to their evaluated metrics and statements (the `ReceiverEntry` and that contains `Status`).

internal/receiver/discoveryreceiver/config.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"go.opentelemetry.io/collector/config"
+	"go.uber.org/multierr"
+)
+
+var (
+	_ config.Receiver = (*Config)(nil)
+
+	allowedStatusTypes = []string{"successful", "partial", "failed"}
+	allowedStatuses    = func() map[string]struct{} {
```
pmcollins, line 28: Would it be appropriate to put this logic an `init` block?

rmfitzpatrick: Do you think so? I think this is evaluated before the init function call but is still deferred until importing at runtime. Determinism doesn't seem important to me in this case since it's only used in functions.

pmcollins: Just thought it might be more readable that way (I think functionally it's the same), but admittedly this is a nit 😄

pmcollins (approved):

PR #1863, opened 2022-08-11
Add tar.gz distribution of the Splunk Otel Collector

atoulme:
Adds a tar distro of the Splunk Otel Collector. It doesn't interact with pytest much, as there is no path to upgrade and install is a `tar xzf`.

### DIFF
--- a/docs/getting-started/linux-manual.md
+++ b/docs/getting-started/linux-manual.md
@@ -443,9 +443,9 @@ This ensures that the binaries in the bundle have the right loader set on them s
 
 The tar archive contains the default agent and gateway configuration files.
 Both refer to environment variables described in the [Other](#Other) section above.
-Additionally, the agent configuration file requires that you set two environment variables:
-- `SPLUNK_BUNDLE_DIR` (no default): The path to the Smart Agent bundle, e.g. /usr/lib/splunk-otel-collector/agent-bundle
-- `SPLUNK_COLLECTD_DIR` (no default): The path to the collectd config directory for the Smart Agent, e.g. /usr/lib/splunk-otel-collector/agent-bundle/run/collectd
+If you are running the Collector from a non-default location, the Smart Agent receiver and agent configuration file require that you set two environment variables currently used in the Smart Agent extension:
+- `SPLUNK_BUNDLE_DIR` (`/usr/lib/splunk-otel-collector/agent-bundle` default): The path to the Smart Agent bundle, e.g. `/opt/my/environment/splunk-otel-collector/agent-bundle`
+- `SPLUNK_COLLECTD_DIR` (`/usr/lib/splunk-otel-collector/agent-bundle/run/collectd` default): The path to the collectd config directory for the Smart Agent, e.g. `/opt/my/environment/splunk-otel-collector/agent-bundle/run/collectd`
 
 ## Advanced Configuration
 
--- a/docs/getting-started/linux-manual.md
+++ b/docs/getting-started/linux-manual.md
@@ -14,6 +14,7 @@ The following deployment options are supported:
 - [DEB and RPM Packages](#deb-and-rpm-packages)
 - [Docker](#docker)
 - [Binary](#binary)
+- [Tar archive](#tar)
 
 ## Getting Started
 
@@ -425,6 +426,26 @@ make install-tools
 make otelcol
 SPLUNK_ACCESS_TOKEN=12345 SPLUNK_REALM=us0 ./bin/otelcol
 ```
+
+### Tar
+
+We offer for convenience a tar.gz archive of the distribution.
+
+To use the archive:
+
+1. Unarchive it to a directory of your choice on the target system.
+```bash
+tar xzf splunk-otel-collector_<version>_<arch>.tar.gz
+```
+
+2. On amd64 systems, go into the unarchived `agent-bundle` directory and run `bin/patch-interpreter $(pwd)`. 
+This ensures that the binaries in the bundle have the right loader set on them since your host's loader may not be compatible.
+
+The tar archive contains the default agent and gateway configuration files.
+Both refer to environment variables described in the [Other](#Other) section above.
+Additionally, the agent configuration file requires that you set two environment variables:
+- `SPLUNK_BUNDLE_DIR` (no default): The path to the Smart Agent bundle, e.g. /usr/lib/splunk-otel-collector/agent-bundle
+- `SPLUNK_COLLECTD_DIR` (no default): The path to the collectd config directory for the Smart Agent, e.g. /usr/lib/splunk-otel-collector/agent-bundle/run/collectd
 
 ## Advanced Configuration
 
--- a/internal/buildscripts/packaging/fpm/tar/build.sh
+++ b/internal/buildscripts/packaging/fpm/tar/build.sh
@@ -56,6 +56,8 @@ tar_download_smart_agent() {
     tar -xzf "$buildroot/signalfx-agent.tar.gz" -C "$buildroot/"
     mv "$buildroot/signalfx-agent" "$buildroot/$AGENT_BUNDLE_INSTALL_DIR"
     find "$buildroot/$AGENT_BUNDLE_INSTALL_DIR" -wholename "*test*.key" -delete -or -wholename "*test*.pem" -delete
+    rm -f "$buildroot/$AGENT_BUNDLE_INSTALL_DIR/bin/signalfx-agent"
+    rm -f "$buildroot/$AGENT_BUNDLE_INSTALL_DIR/bin/agent-status"
     rm -f "$buildroot/signalfx-agent.tar.gz"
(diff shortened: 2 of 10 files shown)


jeffreyc-splunk:
Should we exclude any files/directories from the archive, such as:
- `etc/otel/collector/splunk-otel-collector.conf.example` and `lib/systemd/system/splunk-otel-collector.service` (if service management is not supported)
- `etc/otel/collector/fluentd/` (if fluentd log collection is not supported)

jeffreyc-splunk:
One required post-install step that we should document is that the `patch-interpreter` script needs to be run manually after extracting the `amd64` archive (not required for the `arm64` archive since the smart agent bundle would not be included/supported):
```
$ cd usr/lib/splunk-otel-collector/agent-bundle
$ bin/patch-interpreter $(pwd)
```
This is currently performed automatically in deb/rpm post-install scripts.  See step 2 in https://github.com/signalfx/signalfx-agent#linux-standalone-targz.

aurbiztondo-splunk (approved):

atoulme:
I have simplified the archive. Here is a screenshot of what it looks like:
Screen Shot 2022-08-12 at 3 27 23 PM
Is it too whittled down now?

internal/buildscripts/packaging/tests/package_test.py
```diff
@@ -62,6 +64,23 @@ def get_libcap_command(container):
...
+    pkg_path = get_package(distro, PKG_NAME, PKG_DIR)
+    assert pkg_path, f"{PKG_NAME} package not found in {PKG_DIR}"
+    pkg_base = os.path.basename(pkg_path)
+
+    with run_distro_container(distro) as container:
+
+        copy_file_into_container(container, pkg_path, f"/test/{pkg_base}")
+        run_container_cmd(container, f"tar xzf /test/{pkg_base} -C /tmp")
+        run_container_cmd(container, f"test -d /tmp/bin")
+        run_container_cmd(container, f"test -d /tmp/signalfx-agent")
```
rmfitzpatrick, line 81: To work w/ the SA extension defaults I think it'd be preferable to rename this dir to `agent-bundle` like the other installers do so that users can extract to `/usr/lib/splunk-otel-collector` and run the patch-interpreter script @jeffreyc-splunk mentioned.

https://github.com/signalfx/splunk-otel-collector/blob/main/internal/extension/smartagentextension/factory.go#L59

rmfitzpatrick: It may be a better experience to add something like a `configure.sh` that does the patch step for users, but that can be in a later PR.

atoulme: OK, I have changed the packaging to follow that direction.

rmfitzpatrick: Thanks, validated with integration tests.

rmfitzpatrick:
> One required post-install step that we should document is that the `patch-interpreter` script needs to be run manually after extracting the `amd64` archive (not required for the `arm64` archive since the smart agent bundle would not be included/supported)
👍  for this being added to https://github.com/signalfx/splunk-otel-collector/blob/main/docs/getting-started/linux-manual.md#binary or a new subsection since the bundle isn't usable without doing so.

> Is it too whittled down now?
I think we should also provide the agent and gateway default configs.

jeffreyc-splunk:
I wonder if the archive should be created within a `splunk-otel-collector` root directory?  For comparison, the signalfx-agent archive has a root `signalfx-agent` directory with all files under it.

atoulme:
Indeed - added root folder.

internal/buildscripts/packaging/tests/package_test.py
```diff
@@ -62,6 +64,23 @@ def get_libcap_command(container):
...
+    pkg_base = os.path.basename(pkg_path)
+
+    with run_distro_container(distro) as container:
+
+        copy_file_into_container(container, pkg_path, f"/test/{pkg_base}")
+        run_container_cmd(container, f"tar xzf /test/{pkg_base} -C /tmp")
+        run_container_cmd(container, f"test -d /tmp/bin")
+        run_container_cmd(container, f"test -d /tmp/agent-bundle")
+        run_container_cmd(container, f"test -f /tmp/bin/otelcol")
+        run_container_cmd(container, f"test -f /tmp/bin/translatesfx")
```
jeffreyc-splunk, line 83: Do these paths need to be updated for the new `splunk-otel-collector` base dir?

docs/getting-started/linux-manual.md
```diff
@@ -426,6 +427,26 @@ make otelcol
...
+We offer for convenience a tar.gz archive of the distribution.
+
+To use the archive:
+
+1. Unarchive it to a directory of your choice on the target system.
+```bash
+tar xzf splunk-otel-collector_<version>_<arch>.tar.gz
+```
+
+2. Go into the unarchived `agent-bundle` directory and run `bin/patch-interpreter $(pwd)`.
```
jeffreyc-splunk, line 441: This step is only applicable for the `amd64` archive.

internal/buildscripts/packaging/tests/package_test.py
atoulme: yes they do. Fixing...

docs/getting-started/linux-manual.md
```diff
@@ -426,6 +427,26 @@ make otelcol
...
+```
+
+2. On amd64 systems, go into the unarchived `agent-bundle` directory and run `bin/patch-interpreter $(pwd)`. 
+This ensures that the binaries in the bundle have the right loader set on them since your host's loader may not be compatible.
+
+The tar archive contains the default agent and gateway configuration files.
+Both refer to environment variables described in the [Other](#Other) section above.
+Additionally, the agent configuration file requires that you set two environment variables:
+- `SPLUNK_BUNDLE_DIR` (no default): The path to the Smart Agent bundle, e.g. /usr/lib/splunk-otel-collector/agent-bundle
+- `SPLUNK_COLLECTD_DIR` (no default): The path to the collectd config directory for the Smart Agent, e.g. /usr/lib/splunk-otel-collector/agent-bundle/run/collectd
```
rmfitzpatrick, line 448: The smart agent extension sets these defaults:
https://github.com/signalfx/splunk-otel-collector/blob/main/internal/extension/smartagentextension/factory.go#L60
https://github.com/signalfx/splunk-otel-collector/blob/main/internal/extension/smartagentextension/factory.go#L75

```suggestion
The tar archive contains the default agent and gateway configuration files.
Both refer to environment variables described in the [Other](#Other) section above.
If you are running the Collector from a non-default location, the Smart Agent receiver and agent configuration file require that you set two environment variables currently used in the Smart Agent extension:
- `SPLUNK_BUNDLE_DIR` (`/usr/lib/splunk-otel-collector/agent-bundle` default): The path to the Smart Agent bundle, e.g. `/opt/my/environment/splunk-otel-collector/agent-bundle`
- `SPLUNK_COLLECTD_DIR` (`/usr/lib/splunk-otel-collector/agent-bundle/run/collectd` default): The path to the collectd config directory for the Smart Agent, e.g. `/opt/my/environment/splunk-otel-collector/agent-bundle/run/collectd`
```

rmfitzpatrick: Actually may be best just to set the bundle dir since the collectd dir defaults based on it. (updated)

rmfitzpatrick (approved):

docs/getting-started/linux-manual.md
atoulme: OK I will remove them then. Less is more.

aurbiztondo-splunk (approved):

internal/buildscripts/packaging/fpm/tar/build.sh
```diff
@@ -0,0 +1,117 @@
...
+        exit 1
+    fi
+
+    echo "Downloading $dl_url ..."
+    curl -sL "$dl_url" -o "$buildroot/signalfx-agent.tar.gz"
+
+    mkdir -p "$buildroot/$BUNDLE_BASE_DIR"
+    tar -xzf "$buildroot/signalfx-agent.tar.gz" -C "$buildroot/"
+    mv "$buildroot/signalfx-agent" "$buildroot/$AGENT_BUNDLE_INSTALL_DIR"
+    find "$buildroot/$AGENT_BUNDLE_INSTALL_DIR" -wholename "*test*.key" -delete -or -wholename "*test*.pem" -delete
```
jeffreyc-splunk, line 58: Should we also exclude the `signalfx-agent` binary and `agent-status` symlink, similar to https://github.com/signalfx/splunk-otel-collector/blob/main/internal/buildscripts/packaging/fpm/common.sh#L98-L99 with the deb/rpm packages?

atoulme: Good catch!

PR #1864, opened 2022-08-11
Update core/contrib deps to v0.58.0

jeffreyc-splunk:


Makefile
```diff
@@ -168,14 +180,14 @@ impi:
...
-	go install github.com/client9/misspell/cmd/misspell@v0.3.4
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.45.0
-	go install github.com/google/addlicense@v0.0.0-20200906110928-a0294312aa76
-	go install github.com/jstemmer/go-junit-report@v0.9.1
-	go install github.com/ory/go-acc@v0.2.8
-	go install github.com/pavius/impi/cmd/impi@v0.0.3
-	go install github.com/tcnksm/ghr@v0.14.0
-	go install golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@latest
+	cd ./internal/tools && go install github.com/client9/misspell/cmd/misspell
+	cd ./internal/tools && go install github.com/golangci/golangci-lint/cmd/golangci-lint
```
jvoravong, line 184: Just double checking, we don't need the pinned versions anymore right?

jvoravong (approved):

Makefile
rmfitzpatrick: I'm not sure we did across the board, or at least we don't appear to now based on the checks.

PR #1890, opened 2022-08-17
fmt/lint against 1.19

rmfitzpatrick:
With latest go release builds will fail due to fmt differences and deprecated ioutil functions.


atoulme (approved):
LGTM

PR #1900, opened 2022-08-19
discovery receiver: add endpoint event logging

rmfitzpatrick:
These changes add the initial logging functionality to the discovery receiver when `log_endpoints` is configured. It also makes some basic signature changes to the receiver in preparation of* future metric consumption/evaluation of dynamically generated receivers.


internal/receiver/discoveryreceiver/endpoint_tracker.go
```diff
@@ -0,0 +1,203 @@
...
+func newEndpointTracker(observables map[config.ComponentID]observer.Observable, config *Config, logger *zap.Logger, pLogs chan plog.Logs) *endpointTracker {
+	return &endpointTracker{
+		logEndpoints: config.LogEndpoints,
+		observables:  observables,
+		logger:       logger,
+		pLogs:        pLogs,
+	}
+}
+
+func (el *endpointTracker) start() {
```
jvoravong, line 65: Nit: Should el be et?

internal/receiver/discoveryreceiver/receiver.go
```diff
@@ -57,29 +66,62 @@ func newDiscoveryReceiver(
...
+	d.endpointTracker = newEndpointTracker(d.observables, d.config, d.logger, d.pLogs)
+	d.endpointTracker.start()
+
+	loopStarted := &sync.WaitGroup{}
+	loopStarted.Add(1)
+	d.loopFinished.Add(1)
+	go d.consumerLoop(loopStarted)
+	// wait until we know consumer loop is running before starting receiver creator
+	// so as not to miss any resulting telemetry
+	loopStarted.Wait()
```
jvoravong, line 91: I might suggest putting a debug message around here about starting and waiting.

jvoravong:
Just two small things from me. LGTM. You might want a second set of eyes on this.

jvoravong (approved):

internal/receiver/discoveryreceiver/receiver.go
rmfitzpatrick: good call, updated.

internal/receiver/discoveryreceiver/endpoint_tracker.go
rmfitzpatrick: This was originally an "endpoint logger" but it has another responsibility to be added later so I renamed.

PR #1902, opened 2022-08-22
Update tar signing

jeffreyc-splunk:


atoulme (approved):
This makes sense to me 
- I forgot deb signing would also leave a stray tar.gz in the folder as well
- When we sign the tar.gz, since it's signing as a detached .asc file, we must move the tar.gz file.

rmfitzpatrick (approved):

PR #1903, opened 2022-08-22
sync host metadata for internal metrics

dloucasfx:
The current GW config does not sync host metadata by default, which created confusion when users enable hostmetrics (and other GW metrics) and not seeing the host metadata in the backend.

### DIFF
--- a/cmd/otelcol/config/collector/gateway_config.yaml
+++ b/cmd/otelcol/config/collector/gateway_config.yaml
@@ -99,6 +99,10 @@ exporters:
     ## to avoid double translations and exclusions.
     #translation_rules: []
     #exclude_metrics: []
+  signalfx/internal:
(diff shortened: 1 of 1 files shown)


cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -102,7 +102,10 @@ exporters:
   # Debug
   #logging:
     #loglevel: debug
-
+  signalfx/internal:
+    access_token: "${SPLUNK_ACCESS_TOKEN}"
+    realm: "${SPLUNK_REALM}"
+    sync_host_metadata: true
```
dmitryax, line 108: Can you please move it up a bit closer to the other `signalfx` exporter?

dloucasfx: done

dmitryax (approved):

PR #1918, opened 2022-08-30
Update config server to support multiple confmap providers

rmfitzpatrick:
The current config server doesn't support multiple config providers for initial or effective values but instead has a "last wins" implementation. These changes move the config server to:

1. act as the final config converter such that it always has access to the final resolved config.
2. index* the initial config view by the provider scheme so that each source content is accessible.


aurbiztondo-splunk (approved):

cmd/otelcol/main.go
```diff
@@ -108,26 +108,32 @@ func main() {
...
-					zap.NewNop(), // The service logger is not available yet, setting it to NoP.
-					info,
-					configsources.Get()...,
-				),
+			ResolverSettings: confmap.ResolverSettings{
+				URIs: configLocations(inputFlags),
+				Providers: map[string]confmap.Provider{
+					envProvider.Scheme(): configprovider.NewConfigSourceConfigMapProvider(
+						envProvider,
+						zap.NewNop(), // The service logger is not available yet, setting it to NoP.
```
atoulme, line 123: this gives me a bit of a pause - is it set later? Is it important? It looks like at minimum we might want to fix this in a subsequent PR.

internal/configconverter/config_server.go
```diff
@@ -43,117 +46,134 @@ const (
...
+				endpoint = "localhost:" + portOverride
+			}
 
-	listener, err := net.Listen("tcp", endpoint)
-	if err != nil {
-		return err
-	}
+			listener, err := net.Listen("tcp", endpoint)
+			if err != nil {
+				log.Print(fmt.Errorf("error starting config server: %w", err).Error())
```
atoulme, line 146: Should it crash the collector startup instead?

cmd/otelcol/main.go
rmfitzpatrick: This inherits the existing functionality, but deferred service logger creation is a problem in our distribution as a whole: https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/service.go#L67.

internal/configconverter/config_server.go
atoulme: I see, so until we have a better way to manage those conflicts, we should instead ignore this error. Understood.

atoulme (approved):
LGTM.

internal/configconverter/config_server.go
rmfitzpatrick: Updated to log port-setting guidance on conflicts.

Issue #1876, opened 2022-08-14
Installation script production use clarification


mukundbabbar:
The installation script mentions that it is "not recommended for production use". Is this still valid? 

<img width="533" alt="Screen Shot 2022-08-14 at 3 06 54 am" src="https://user-images.githubusercontent.com/5012739/184542866-e735cdbb-b830-4910-ab65-d650561ef086.png">

<img width="763" alt="Screen Shot 2022-08-15 at 1 00 15 am" src="https://user-images.githubusercontent.com/5012739/184542882-0ae5a5ca-4cd2-45a5-bc53-c14d630823f6.png">

Thanks

jeffreyc-splunk:
Thanks @mukundbabbar.  The doc and script has been updated.

PR #1921, opened 2022-08-31
Update recommended installation methods

jeffreyc-splunk:
For #1876.  Also add link to chef cookbook.

### DIFF
--- a/internal/buildscripts/packaging/installer/install.sh
+++ b/internal/buildscripts/packaging/installer/install.sh
@@ -15,7 +15,10 @@
 # limitations under the License.
 
 # A convenience script to install the collector package on any of our supported
-# distros.  NOT recommended for production use.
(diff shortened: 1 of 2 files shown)


internal/buildscripts/packaging/installer/install.sh
```diff
@@ -15,7 +15,10 @@
 # limitations under the License.
 
 # A convenience script to install the collector package on any of our supported
-# distros.  NOT recommended for production use.
+# distros.  NOT recommended for production use.  Please refer to
```
rmfitzpatrick, line 18: This is our suggested path so it's unclear why this would be the case.

```suggestion
# distros.  Please refer to
```

aurbiztondo-splunk (approved):

atoulme (approved):

PR #1930, opened 2022-09-01
Support installing with ansible and skipping restart of services

### LINKED ISSUE #460: [ansible] ability to not start the agent
Hello,

thanks for providing the role and resolving https://github.com/signalfx/splunk-otel-collector/issues/352 

one common use case of ansible is to use it in combination with packer to build image.

In this case we want to avoid to start the agent for different reasons:
- this is useless to monitor a temporary instance used to build image
- it can lead to undesired billed host
- it can lead to undesired alert / oncall (e.g. for heartbeat detector)

of course the default behavior must remain to start the agent but an option to let him enabled at boot but do not start it at the end of role execution would be nice.

thanks

### LINKED ISSUE #465: [ansible] flag to disable logs
Hello,

Similar to the helm chart https://github.com/signalfx/splunk-otel-collector-chart/blob/01d60f316bc003c473b55f33dcfe19742a2a85c6/helm-charts/splunk-otel-collector/values.yaml#L58

we need a way to install the otel collector without enabling logging.

For now there is an option to disable fluentd installation (which does not uninstall it when we change it to false for your information).

But the splunk hec exporter is still configured which will lead to a fail:

``` 
Jun 10 13:04:17 lveulnx4053q otelcol[385619]: 2021-06-10T13:04:17.241Z#011info#011exporterhelper/queued_retry.go:314#011Exporting failed. Will retry the request after interval.#011{"kind": "exporter", "name": "splunk_hec", "error": "HTTP 503 \"Service Unavailable\"", "interval": "7.032856404s"}
```

we can use a custom configuration for otel collector but it seems not straightforward enough.

thanks

atoulme:
Fixes #460

### DIFF
--- a/deployments/ansible/molecule/default/verify.yml
+++ b/deployments/ansible/molecule/default/verify.yml
@@ -3,8 +3,6 @@
   hosts: all
   gather_facts: true
   tasks:
-    - name: Import shared verification tasks
-      ansible.builtin.import_tasks: ../shared/verify_tasks.yml
 
     - name: Assert SPLUNK_ACCESS_TOKEN env var is set
       ansible.builtin.lineinfile:
@@ -47,3 +45,14 @@
         dest: /etc/otel/collector/splunk-otel-collector.conf
         state: present
       check_mode: yes
+
+    - name: Populate service facts
+      ansible.builtin.service_facts:
+
+    - name: Assert splunk-otel-collector service is stopped
+      assert:
+        that: ansible_facts.services['splunk-otel-collector.service'].state == 'stopped'
+
+    - name: Assert td-agent service is stopped
+      assert:
+        that: ansible_facts.services['td-agent.service'].state == 'stopped'
--- a/deployments/ansible/molecule/default/verify.yml
+++ b/deployments/ansible/molecule/default/verify.yml
@@ -56,3 +56,4 @@
     - name: Assert td-agent service is stopped
       assert:
         that: ansible_facts.services['td-agent.service'].state == 'stopped'
+      when: ansible_os_family != "Suse"
--- a/deployments/ansible/molecule/default/verify.yml
+++ b/deployments/ansible/molecule/default/verify.yml
@@ -4,6 +4,13 @@
   gather_facts: true
   tasks:
 
+    - name: Populate service facts
+      ansible.builtin.service_facts:
+
+    - name: Assert splunk-otel-collector service is stopped
+      assert:
+        that: ansible_facts.services['splunk-otel-collector.service'].state == 'stopped'
+
     - name: Assert SPLUNK_ACCESS_TOKEN env var is set
       ansible.builtin.lineinfile:
         line: SPLUNK_ACCESS_TOKEN=fake-token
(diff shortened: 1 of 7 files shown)


deployments/ansible/roles/collector/handlers/main.yml
```diff
@@ -7,28 +7,38 @@
   listen:
     - restart splunk-otel-collector
     - restart td-agent
+  when:
+    - skip_restart | default("false") == "false"
```
rmfitzpatrick, line 11: nit, vars that require negatives for enabling actions can be confusing where `start_service` w/ true default is clearer and doesn't center the restarting detail imo.

deployments/ansible/roles/collector/handlers/main.yml
```diff
@@ -7,28 +7,38 @@
...
 
 - name: Restart td-agent
   ansible.builtin.service:
     name: td-agent
     state: restarted
-  when: install_fluentd
+  when:
+    - install_fluentd
+    - skip_restart | default("false") == "false"
```
rmfitzpatrick, line 19: I think this would then expect `skip_restart` to be the string value `"false"` where bools are probably wanted?

```suggestion
    - (skip_restart | default(false) | bool) == false
```

jeffreyc-splunk:
Should there also be new handlers/tasks to stop the collector and fluentd running services if `start_service` is false?

deployments/ansible/roles/collector/README.md
```diff
@@ -34,6 +34,13 @@ how to use the role in a playbook with minimal required configuration:
         splunk_realm: SPLUNK_REALM
 ```
 
+You can disable starting the collector, smart agent, and fluentd services by setting 
+the argument `start_service` to `false`:
+
+```
+$> ansible-playbook playbook.yaml -e start_service=false
```
jeffreyc-splunk, line 41: Should `start_service` also be available as a role variable?

atoulme: yes ok, let's do that.

deployments/ansible/molecule/default/verify.yml
```diff
@@ -3,8 +3,13 @@
...
   tasks:
-    - name: Import shared verification tasks
-      ansible.builtin.import_tasks: ../shared/verify_tasks.yml
+
+    - name: Populate service facts
+      ansible.builtin.service_facts:
+
+    - name: Assert splunk-otel-collector service is stopped
+      assert:
+        that: ansible_facts.services['splunk-otel-collector.service'].state == 'stopped'
```
jeffreyc-splunk, line 12: Should this task be moved to `custom_vars` or a new test suite?  Otherwise, issues with starting the collector service may not get caught.

atoulme: this default test suite wasn't doing much with the collector. It wasn't starting it or using it, just checking config. There are other tests that run against the collector to make sure it can ingest logs. I figured I'd use this test.

jeffreyc-splunk: Ah, I guess we check the service status in the other tests, so hopefully any issues would get caught there.

aurbiztondo-splunk (approved):

rmfitzpatrick (approved):

PR #1937, opened 2022-09-01
add processlist to default config

dloucasfx:
The OOTB content in IMM uses the data emitted from processlist monitor, it makes sense to add this by default.
If we agreed to add processlist to the default config, then I will make the same update to the remaining OOTB configs we ship with the splunk distro


jvoravong (approved):
LGTM. We should probably get 1 more approval on this.

atoulme (approved):
LGTM.

PR #1940, opened 2022-09-02
Add resourcedetection to make the processlist events register with the backend

dloucasfx:
This is related to https://github.com/signalfx/splunk-otel-collector/pull/1937
Without resourcedetection, the event will be missing the `host.name` label which is used to store the event in the memcache, if host.name is missing then the event is dropped.


atoulme (approved):
LGTM

jvoravong (approved):

aurbiztondo-splunk (approved):

PR #1952, opened 2022-09-07
Refactor main flags as settings.Settings

rmfitzpatrick:
These changes move existing main and `flags` option parsing, env var evaluation/setting, and default logic to a new internal `settings` package where a more minimal `Settings` interface is used to populate the underlying config provider resolver settings and command line argument adjustment. They are largely attempted to be noops/transparent readability and logic changes with some test practices improvements except for:

1. changing the std flag library to spf13/pflag so we can use the `VarP` helpers for version and help shorthands and not have to track with all collector flag names (something we're most likely failing to do already).
2. Removing the feature flag value implementation and parsing since we don't use them or the feature registry at this time (which is ok now w/ the above).
3. Remove the project-local default gateway and otlp linux config paths since they are only used by tests and none of the installers apply a `cmd/`-rooted location

I didn't change the bulk of the config and memory ballast logic to keep the change set more manageable and the likelihood of introducing bugs down, but think it could benefit from an overhaul in subsequent changes.*


atoulme (approved):
The refactoring makes sense to me and makes main.go a lot easier to read. LGTM!

crobert-1 (approved):

internal/settings/settings.go
```diff
@@ -0,0 +1,446 @@
...
+	memBallastSizeMiB int
+	helpFlag          bool
+	versionFlag       bool
+	noConvertConfig   bool
+}
+
+func (f *flags) ResolverURIs() []string {
+	var configPaths []string
+	if configPaths = f.configPaths.value; len(configPaths) == 0 {
+		if configEnvVal := os.Getenv(ConfigEnvVarName); len(configEnvVal) != 0 {
```
pmcollins, line 106: Long term, might be nice to separate Getenv and Setenv from the settings code since reading/writing them is effectively I/O, and instead pass this information around in some kind of abstraction that could be swappable for testing. Removes the need to deal with global (env) variables in tests.

pmcollins (approved):
Big ups for removing code from main and making things testable 👍
LGTM

PR #1956, opened 2022-09-08
sign-exe job: Use updated signing template

crobert-1:
The signing service has moved to a new template for signing files and archives. This update changes the sign-exe gitlab CI job to use this new signing functionality as the old functionality is no longer working.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -167,7 +167,7 @@ instrumentation-rpm:
 sign-exe:
   extends:
     - .trigger-filter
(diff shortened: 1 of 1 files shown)


atoulme (approved):

rmfitzpatrick (changes requested):
Let's wait until we're sure the signing service updates are correct in ci.

rmfitzpatrick:
opting for https://github.com/signalfx/splunk-otel-collector/pull/1959 since source is synced w/ releaser mirror.

PR #1972, opened 2022-09-13
includeconfigsource: add tests for watch file and template files

ccordi:
This PR fixes a bug in the retrieve function when using templated files. 

It also adds unit tests to cover templated and watched files.


internal/configsource/includeconfigsource/source.go
```diff
@@ -60,7 +60,13 @@ func (is *includeConfigSource) Retrieve(_ context.Context, selector string, para
 	}
 
 	var buf bytes.Buffer
-	if err = tmpl.Execute(&buf, paramsConfigMap); err != nil {
+	var params map[string]interface{}
```
rmfitzpatrick, line 63: nit, we've moved to `any` but don't have a lint check for it atm.

rmfitzpatrick:
Do you mind adding at least one integration test similar to https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/envvar_expansion_test.go#L83 that expands yaml ~mappings from multiple files (maybe in a new `general/config` suite? It'd be helpful to vet the config source complying with the actual service config resolution system with scalar and blobs: https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/main.go#L77

tests/general/template_config_test.go
```diff
@@ -0,0 +1,114 @@
+package tests
```
rmfitzpatrick, line 1: mind renaming the file to something like `include_config_source_test.go`?

atoulme (approved):

tests/general/include_config_source_test.go
```diff
@@ -0,0 +1,114 @@
...
+		Build()
+
+	require.NotNil(t, collector)
+	require.NoError(t, err)
+
+	err = collector.Start()
+	require.NoError(t, err)
+
+	_, ok := collector.(*testutils.CollectorProcess)
+	require.True(t, ok)
```
rmfitzpatrick, line 36:
```suggestion
```

I don't think these are necessary when not using the `CollectorProcess` directly.

rmfitzpatrick (approved):

tests/general/template_config_test.go
ccordi: done

PR #1980, opened 2022-09-15
discovery receiver: add metrics evaluation

rmfitzpatrick:
These changes add metric consumption, evaluation, and log record emission to the discovery receiver. They require https://github.com/signalfx/splunk-otel-collector/pull/1971.

### DIFF
--- a/internal/receiver/discoveryreceiver/metric_evaluator.go
+++ b/internal/receiver/discoveryreceiver/metric_evaluator.go
@@ -16,26 +16,198 @@ package discoveryreceiver
 
 import (
 	"context"
+	"fmt"
+	"sync"
+	"time"
 
 	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/plog"
 	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
+	"github.com/signalfx/splunk-otel-collector/internal/receiver/discoveryreceiver/statussources"
 )
 
 var _ consumer.Metrics = (*metricEvaluator)(nil)
 
+const (
+	metricMatch = "metric.match"
+)
+
+var (
+	jsonMarshaler = pmetric.NewJSONMarshaler()
+)
+
 // metricEvaluator conforms to a consumer.Metrics to receive any metrics from
 // receiver creator-created receivers and determine if they match any configured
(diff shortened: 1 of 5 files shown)


internal/receiver/discoveryreceiver/metric_evaluator.go
```diff
@@ -16,26 +16,164 @@ package discoveryreceiver
...
+func (m *metricEvaluator) ConsumeMetrics(_ context.Context, md pmetric.Metrics) error {
+	if pLogs := m.evaluateMetrics(md); pLogs.LogRecordCount() > 0 {
+		m.pLogs <- pLogs
+	}
 	return nil
 }
+
+func (m *metricEvaluator) evaluateMetrics(md pmetric.Metrics) plog.Logs {
+	if ce := m.logger.Check(zapcore.DebugLevel, "evaluating metrics"); ce != nil {
+		if mbytes, err := jsonMarshaler.MarshalMetrics(md); err == nil {
```
atoulme, line 80: Would it be possible to log the error if not nil?

internal/receiver/discoveryreceiver/metric_evaluator.go
```diff
@@ -16,26 +16,164 @@ package discoveryreceiver
...
+					logRecord.SetSeverityText(severityText)
+					logRecord.Attributes().InsertString(metricNameAttr, metricName)
+					logRecord.Attributes().UpsertString(statusAttr, status)
+					var timestamp pcommon.Timestamp
+					switch metric.DataType() {
+					case pmetric.MetricDataTypeGauge:
+						timestamp = metric.Gauge().DataPoints().At(0).Timestamp()
+					case pmetric.MetricDataTypeSum:
+						timestamp = metric.Sum().DataPoints().At(0).Timestamp()
+					case pmetric.MetricDataTypeHistogram:
```
atoulme, line 166: I think you are missing a few metric types: MetricDataTypeExponentialHistogram, MetricDataTypeSummary.
You might need to support MetricDataTypeNone as well. Maybe a default case to log a debug statement would help.

atoulme (approved):
Approved. Two small nits.

internal/receiver/discoveryreceiver/metric_evaluator.go
rmfitzpatrick: updated

internal/receiver/discoveryreceiver/metric_evaluator.go
rmfitzpatrick: Thanks, updated w/ missing types and added more tests for the timestamp extraction.

PR #2011, opened 2022-09-20
Add an Oracle database receiver

atoulme:


receiver/oracledbreceiver/db_client.go
```diff
@@ -0,0 +1,99 @@
...
+
+func newDbClient(db *sql.DB, sql string, logger *zap.Logger) dbClient {
+	return dbSQLClient{
+		db:     db,
+		sql:    sql,
+		logger: logger,
+	}
+}
+
+func (cl dbSQLClient) metricRows(ctx context.Context) ([]metricRow, error) {
```
rmfitzpatrick, line 50: This looks like https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/sqlqueryreceiver/db_client.go was c/p instead of wrapping the sqlquery receiver and configuring w/ default queries. Creating diverging implementations out of the gate seems undesirable.

aurbiztondo-splunk (approved):

receiver/oracledbreceiver/scraper.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	return nil
+}
+
+func (s *scraper) Scrape(ctx context.Context) (pmetric.Metrics, error) {
+	s.logger.Debug("Begin scrape")
+	now := pcommon.NewTimestampFromTime(time.Now())
+	if s.metricsSettings.OracledbQueryCPUTime.Enabled {
+		if err := s.executeOneQueryWithFullText(ctx, now, queryCPUTimeSQL, s.metricsBuilder.RecordOracledbQueryCPUTimeDataPoint); err != nil {
+			return pmetric.Metrics{}, fmt.Errorf("error executing %s: %w", queryCPUTimeSQL, err)
```
rmfitzpatrick, line 105: Is it better to track with a multierr instead of early exiting?

rmfitzpatrick: Could it be a bit more maintainable ranging over a slice of structs w/ the query, enabled, and recorder instead of each individual item?

receiver/oracledbreceiver/scraper.go
```diff
@@ -0,0 +1,306 @@
...
+	s.db, err = s.dbProviderFunc()
+	if err != nil {
+		return fmt.Errorf("failed to open db connection: %w", err)
+	}
+	return nil
+}
+
+func (s *scraper) Scrape(ctx context.Context) (pmetric.Metrics, error) {
+	s.logger.Debug("Begin scrape")
+	now := pcommon.NewTimestampFromTime(time.Now())
```
rmfitzpatrick, line 102: Using a static ts and unreliable query performance could introduce some unintuitive drift where updated w/ the query would be more accurate.

atoulme: I wasn't sure about this. I used the approach taken by sqlqueryreceiver, but I had the same gut reaction as you have here.
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/sqlqueryreceiver/scraper.go#L65

I'd feel comfortable changing to your suggestion.

receiver/oracledbreceiver/scraper.go
atoulme: multierr - certainly. Do we want to collect metrics we can then and still error out? How is that best handled? Do we log the multierr or use the obsreceiver approach?

atoulme: I can refactor this code to make it look less jarring as well.

rmfitzpatrick: Appears `scrapererror.NewPartialScrapeError()` will cause the correct metric count to be used in the obs report: https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/scraperhelper/scrapercontroller.go#L198

rmfitzpatrick: I think manually logging may be required though to surface to the user

receiver/oracledbreceiver/db_client.go
rmfitzpatrick: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14321 could be adopted after the fact.

receiver/oracledbreceiver/scraper.go
rmfitzpatrick: unless you want to keep it transparent, addressing this in your sqlquery lib PR could make sense

atoulme: Good call, I'll do it in both places.

receiver/oracledbreceiver/config.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/collector/receiver/scraperhelper"
+
+	"github.com/signalfx/splunk-otel-collector/receiver/oracledbreceiver/internal/metadata"
+)
+
+type Config struct {
+	config.ReceiverSettings                 `mapstructure:",squash"`
+	DataSource                              string `mapstructure:"datasource"`
```
rmfitzpatrick, line 31: Not necessary in first pass but I can see also providing other options and using [`BuildUrl()`](https://github.com/sijms/go-ora/blob/master/v2/connection_string.go#L123) as convenient alternative for the user.

receiver/oracledbreceiver/config.go
```diff
@@ -0,0 +1,54 @@
...
+	scraperhelper.ScraperControllerSettings `mapstructure:",squash"`
+	MetricsSettings                         metadata.MetricsSettings `mapstructure:"metrics"`
+}
+
+func (c Config) Validate() error {
+	if c.DataSource == "" {
+		return errors.New("'datasource' cannot be empty")
+	}
+	if _, err := url.Parse(c.DataSource); err != nil {
+		return errors.New("'datasource' is invalid")
```
rmfitzpatrick, line 41: to provide failing info:

```suggestion
		return fmt.Errorf("'datasource' is invalid: %w", err)
```

receiver/oracledbreceiver/db_client.go
rmfitzpatrick, line 50: do you mind having an attributing comment link and maybe a TODO w/ a link to your contrib PR?

receiver/oracledbreceiver/integration_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
+	"github.com/testcontainers/testcontainers-go/wait"
+	"go.opentelemetry.io/collector/component/componenttest"
+	"go.opentelemetry.io/collector/consumer/consumertest"
+)
+
+// This test ensures the collector can connect to an Oracle DB, and properly get metrics. It's not intended to
+// test the receiver itself.
+func TestOracleDBIntegration(t *testing.T) {
```
rmfitzpatrick, line 36: Do you mind moving this to a new `tests/oracledb` directory? https://github.com/signalfx/splunk-otel-collector/tree/main/tests/receivers

Makefile.Common
```diff
@@ -0,0 +1,140 @@
+# SRC_ROOT is the top of the source tree.
+SRC_ROOT := $(realpath $(dir $(lastword $(MAKEFILE_LIST))))
```
rmfitzpatrick, line 2: I think establishing this in a separate PR may be better than associating it w/ these changes directly since they just need to include it.

atoulme: Good idea, opening #2040

receiver/oracledbreceiver/integration_test.go
atoulme: Done, though this is not all working quite yet.

receiver/oracledbreceiver/db_client.go
pmcollins: Yeah, I think a comment about where this code came from and what the plan is would be helpful.

aurbiztondo-splunk (approved):

receiver/oracledbreceiver/scraper.go
atoulme: OK, implemented the partial scrape error.

receiver/oracledbreceiver/db_client.go
atoulme: Done.

pkg/receiver/oracledbreceiver/scraper.go
```diff
@@ -0,0 +1,316 @@
...
+	s.startTime = pcommon.NewTimestampFromTime(time.Now())
+	var err error
+	s.db, err = s.dbProviderFunc()
+	if err != nil {
+		return fmt.Errorf("failed to open db connection: %w", err)
+	}
+	return nil
+}
+
+func (s *scraper) Scrape(ctx context.Context) (pmetric.Metrics, error) {
```
pmcollins, line 98: What do you think about the number of lines of code in this method? Would it be better to break it up and see if boilerplate can be factored out? (Can be done in a separate PR.) Also, more generally, I think a lot of this logic could be done once, at Start() time, rather than on each scrape. The sqlquery receiver, for example, configures and registers the `dbClient`s* and scrapers at start time, one for each query. In any case, doesn't have to be done now IMO since I think this provides functionality that customers are looking for.

*terrible name (my mistake); should be called statement runners or something

pmcollins (approved):
I think there's opportunity for refactoring, but can be done separately if we want to get this into customers' hands sooner.

pkg/receiver/oracledbreceiver/scraper.go
atoulme: No worries - changed to initialize in Start, good call.

dmitryax (approved):

PR #2027, opened 2022-09-22
Fixes pre release v0.60.0

jvoravong:
These updates help unblock the v0.60.0 release.


Makefile
```diff
@@ -161,9 +161,9 @@ misspell-correction:
 
 .PHONY: tidy
 tidy:
-	go mod tidy -compat=1.17
-	cd tests && go mod tidy -compat=1.17
-	cd internal/tools && go mod tidy -compat=1.17
```
rmfitzpatrick, line 166: Should we just bump to `-compat=1.18` like in contrib?

jvoravong: Sorry I just barely missed this before merging. 

Our collector project is at 1.19, I thought we should go with what our collector go.mod version has. 

Thoughts?

rmfitzpatrick: from the docs the default behavior is to use 1 version below the module version so they're the same behavior as stripping I think.  Curious why the compat needed to be removed?

jvoravong: That is a useful bit of information to know. The -compat=1.17 was only out of date so I removed it, now I'll set -compat=1.18 in my next PR.

PR #2040, opened 2022-09-23
Component Makefile changes

atoulme:
Create a common makefile for components to be added.
Use the same approach as taken by otel contrib for the main Makefile to identify sources to avoid touching component code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,16 +6,17 @@ ADDLICENSE= addlicense
 ALL_DOC := $(shell find . \( -name "*.md" -o -name "*.yaml" \) \
                                 -type f | sort)
 
-# All source code excluding any third party code and excluding the testbed.
-# This is the code that we want to run tests for and lint, etc.
-ALL_SRC := $(shell find . -name '*.go' \
-							-not -path './examples/*' \
-							-not -path './tests/*' \
-							-not -path './internal/tools/*' \
-							-type f | sort)
+ALL_PKG_DIRS := $(shell go list -f '{{ .Dir }}' ./... | sort)
 
-# ALL_PKGS is the list of all packages where ALL_SRC files reside.
-ALL_PKGS := $(shell go list $(sort $(dir $(ALL_SRC))))
+ALL_SRC := $(shell find $(ALL_PKG_DIRS) -name '*.go' \
+                                -not -path '*/third_party/*' \
+                                -not -path '*/local/*' \
+                                -type f | sort)
+
+# All source code and documents. Used in spell check.
+ALL_SRC_AND_DOC := $(shell find $(ALL_PKG_DIRS) -name "*.md" -o -name "*.go" -o -name "*.yaml" \
+                                -not -path '*/third_party/*' \
+                                -type f | sort)
 
 ALL_TESTS_DIRS := $(shell find tests -name '*_test.go' | xargs -L 1 dirname | uniq | sort -r)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -6,16 +6,17 @@ ADDLICENSE= addlicense
...
                                 -type f | sort)
 
-# All source code excluding any third party code and excluding the testbed.
-# This is the code that we want to run tests for and lint, etc.
-ALL_SRC := $(shell find . -name '*.go' \
-							-not -path './examples/*' \
-							-not -path './tests/*' \
-							-not -path './internal/tools/*' \
-							-type f | sort)
+ALL_PKG_DIRS := $(shell go list -f '{{ .Dir }}' ./... | sort)
```
rmfitzpatrick, line 9: Should we include the common one and remove repeated vars and targets?

Makefile.Common
```diff
@@ -0,0 +1,143 @@
...
+.PHONY: do-integration-tests-with-cover
+do-integration-tests-with-cover:
+	@echo "running $(GOCMD) integration test ./... + coverage in `pwd`"
+	@$(GOTEST) $(GOTEST_OPT_WITH_INTEGRATION) ./...
+	@if [ -e integration-coverage.txt ]; then \
+  		$(GOCMD) tool cover -html=integration-coverage.txt -o integration-coverage.html; \
+  	fi
+
+.PHONY: benchmark
+benchmark:
```
rmfitzpatrick, line 76: I'm not sure it's desirable to introduce these tasks in these changes since I think we moved away from codecov requirements, don't have benchmark tests, and don't use the `integration` build tag. These are potentially great things overall but probably when addressed directly?

atoulme: I am not sure and will follow your guidance.

Makefile.Common
```diff
@@ -0,0 +1,117 @@
...
+
+.PHONY: checklinks
+checklinks:
+	command -v $(MDLINKCHECK) >/dev/null 2>&1 || { echo >&2 "$(MDLINKCHECK) not installed. Run 'npm install -g markdown-link-check'"; exit 1; }
+	find . -name \*.md -print0 | xargs -0 -n1 \
+		$(MDLINKCHECK) -q -c $(SRC_ROOT)/.github/workflows/check_links_config.json || true
+
+.PHONY: fmt
+fmt: addlicense misspell-correction
+	go fmt  -w -s ./...
```
rmfitzpatrick, line 89:
```suggestion
	gofmt -w -s .
```

or `go fmt ./...`

Makefile
```diff
@@ -60,36 +42,28 @@ ARCH=amd64
...
 
 .PHONY: all
 all: checklicense impi lint misspell test otelcol
 
-.PHONY: test
-test: integration-vet
-	$(GOTEST) $(GOTEST_OPT) $(ALL_PKGS)
-
-.PHONY: integration-vet
-integration-vet:
```
rmfitzpatrick, line 91: mind not deleting this check?

rmfitzpatrick:
I'm seeing `main module (github.com/signalfx/splunk-otel-collector) does not contain package github.com/signalfx/splunk-otel-collector/internal/tools` as a warning when running commands w/ these changes. I had to exclude the internal tools module from `ALL_SRC` to prevent this when introducing the lib: https://github.com/signalfx/splunk-otel-collector/commit/b392b5b734a7c8ef68f3896c58a4aee52f1226de#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R14

Makefile
atoulme: done

atoulme:
I'll chase that one down.

atoulme:
OK, that warning is gone - and the makefile should be a bit faster.

rmfitzpatrick (approved):

PR #2042, opened 2022-09-24
Offer a wait strategy longer than one minute

atoulme:


tests/testutils/container.go
```diff
@@ -151,7 +158,7 @@ func (container Container) Build() *Container {
 		Name:           container.ContainerName,
 		Networks:       container.ContainerNetworks,
 		NetworkMode:    networkMode,
-		WaitingFor:     wait.ForAll(container.WaitingFor...),
+		WaitingFor:     waitStrategy,
```
rmfitzpatrick, line 161: If the aim is to increase the startup timeout of the multistrategy I wonder if adding a new ~`StartupTimeout` field to the container and doing the following would be more direct:

```suggestion
		WaitingFor:     wait.ForAll(container.WaitingFor...).WithStartupTimeout(container.StartupTimeout),
```

atoulme: No jira or anything for this, it's a peel off the work to add the oracledb receiver. The integration tests won't run well with it, here is what's happening:
* We define a wait strategy of waiting for the oracle db container to be healthy to run tests, with a 15 minutes timeout
* The oracle db container takes more than a minute to come up.
* The tests fail.

This happens because the wait.ForAll method here wraps our wait health strategy with:
```
func ForAll(strategies ...Strategy) *MultiStrategy {
	return &MultiStrategy{
		startupTimeout: defaultStartupTimeout(),
		Strategies:     strategies,
	}
}
```
This `defaultStartupTimeout()` is set to 1 minute by default.

Here is the ideal fix:
The MultiStrategy is able to introspect all strategies passed to it to find the longest timeout and accepts it.
However, strategies do not all have a timeout and do not expose their timeout value anyway.
Another ideal fix: remove the timeout from MultiStrategy, it has no business there.

My take is to reduce the risk somewhat and not change anything to existing by getting rid of the MultiStrategy wrapper if only one strategy is used.

Ryan's approach is to enforce that we define the timeout in 2 places, one at the container level and one at the level of each wait strategy. In that case, we duplicate a bit and there's still room for misses.

We could also set an arbitrarily high value for multistrategy's default timeout.

atoulme: I changed my code to take a bit of everything:
* It is now possible to set a container startup timeout
* The default value is raised from 1 minute to 5 by default

rmfitzpatrick (approved):

tests/testutils/container.go
rmfitzpatrick: thanks!

PR #2053, opened 2022-09-28
Different sourcetypes example

atoulme:
new example showing how the collector can assign different sourcetypes to different log sources.


examples/otel-logs-with-sourcetypes-splunk/README.md
```diff
@@ -0,0 +1,18 @@
+# Splunk HEC Example
+
+This example showcases how the collector can collect data from files and send it to Splunk Enterprise, associating with each source a different sourcetype.
```
pmcollins, line 3: Would it be worthwhile to touch on or link to what 'sourcetype' means or do we expect users to already know if they're reading this?

examples/otel-logs-with-sourcetypes-splunk/otel-collector-config.yml
```diff
@@ -0,0 +1,72 @@
...
+        # Whether to disable gzip compression over HTTP. Defaults to false.
+        disable_compression: false
+        # HTTP timeout when sending data. Defaults to 10s.
+        timeout: 10s
+        tls:
+          # Whether to skip checking the certificate of the HEC endpoint when sending data over HTTPS. Defaults to false.
+          # For this demo, we use a self-signed certificate on the Splunk docker instance, so this flag is set to true.
+          insecure_skip_verify: true
+
+processors:
```
pmcollins, line 32: Wondering if it might make things easier to understand for readers if the processors block were above the exporters block.

examples/otel-logs-with-sourcetypes-splunk/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+The example runs as a Docker Compose deployment. The collector can be configured to send logs to Splunk Enterprise.
```
pmcollins, line 5: Would it be worthwhile to add a little explanation about what this example does? e.g. it creates three pipelines, each with its own filelog receiver and resource processor. Each resource processor sets a `com.splunk.sourcetype` resource attribute to a different value...

examples/otel-logs-with-sourcetypes-splunk/otel-collector-config.yml
```diff
@@ -0,0 +1,72 @@
...
+exporters:
+    splunk_hec/logs:
+        # Splunk HTTP Event Collector token.
+        token: "00000000-0000-0000-0000-0000000000000"
+        # URL to a Splunk instance to send data to.
+        endpoint: "https://splunk:8088/services/collector"
+        # Optional Splunk source: https://docs.splunk.com/Splexicon:Source
+        source: "output"
+        # Default sourcetype. This example overrides sourcetype, so it will never be used.
+        sourcetype: "otel:shouldnotshow"
```
pmcollins, line 18: Does this mean that sourcetype is a required field? Perhaps we should add a comment indicating either that it is, or why we are setting it here.

pmcollins (approved):
LGTM, just added a few comments/questions.

aurbiztondo-splunk (approved):

examples/otel-logs-with-sourcetypes-splunk/otel-collector-config.yml
atoulme: If we don't set it, Splunk Enterprise sets a default value of its own instead based on the token configuration. We can remove it I think to avoid going into that conversation.

examples/otel-logs-with-sourcetypes-splunk/README.md
atoulme: copied verbatim and finished the sentence, thanks!

examples/otel-logs-with-sourcetypes-splunk/otel-collector-config.yml
atoulme: moved up

examples/otel-logs-with-sourcetypes-splunk/README.md
atoulme: done

PR #2068, opened 2022-09-30
add support for installing dotnet instrumentation

pmcollins:
This change adds a `with_dotnet_tracing` parameter to the collector install script for windows. When set, the script downloads and installs the dotnet tracing library, and it sets environment variables so that dotnet applications that run on the local machine will be auto-instrumented.

### DIFF
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -68,6 +68,9 @@
     (OPTIONAL) Whether to install and configure .NET tracing to forward .NET application traces to the local collector (default: $false)
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true
+.PARAMETER instrumentation_exclude_processes
+    (OPTIONAL) A semicolon-delimited list of process names to be excluded from auto-instrumentation (default: Powershell.exe;dotnet.exe)
+    .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true -instrumentation_exclude_processes "Powershell.exe;dotnet.exe;myprogram.exe"
 .PARAMETER signalfx_service_name
     (OPTIONAL) A system-wide SignalFx service name override for .NET tracing. Sets the SIGNALFX_SERVICE_NAME environment variable. Ignored if -with_dotnet_instrumentation is false.
     .EXAMPLE
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -567,7 +567,7 @@ if ($with_dotnet_instrumentation) {
     echo "Installing SignalFx Instrumentation for .NET ..."
     # signalfx-dotnet-tracing github repository API
     $api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
-    $module_name = "install.psm1"
+    $module_name = "OpenTelemetry.DotNet.Auto.psm1"
     echo "Downloading .NET Instrumentation installer ..."
     $download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $module_name } | Select-Object -Property browser_download_url,name
     $dotnet_auto_path = Join-Path $env:temp $download.name
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -147,12 +147,14 @@ $old_config_path = "$program_data_path\config.yaml"
 $agent_config_path = "$program_data_path\agent_config.yaml"
 $gateway_config_path = "$program_data_path\gateway_config.yaml"
 $config_path = ""
+
 try {
     Resolve-Path $env:TEMP
     $tempdir = "${env:TEMP}\Splunk\OpenTelemetry Collector"
 } catch {
     $tempdir = "\tmp\Splunk\OpenTelemetry Collector"
 }
+
 $regkey = "HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment"
 
 $fluentd_msi_name = "td-agent-4.3.2-x64.msi"
@@ -385,6 +387,9 @@ if ($with_fluentd -And (Test-Path -Path "$fluentd_base_dir\bin\fluentd")) {
     throw "$fluentd_base_dir\bin\fluentd is already installed. Remove/Uninstall fluentd and re-run this script."
 }
 
+# create a temporary directory
+$tempdir = create_temp_dir -tempdir $tempdir
+
 if ($with_dotnet_instrumentation) {
     echo "Installing SignalFx Instrumentation for .NET ..."
     $api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
@@ -433,9 +438,6 @@ if ("$env:VERIFY_ACCESS_TOKEN" -ne "false") {
         echo '- Verified Access Token'
     }
 }
-
-# set up a temporary directory
-$tempdir = create_temp_dir -tempdir $tempdir
 
 if ($collector_msi_url) {
     $collector_msi_name = "splunk-otel-collector.msi"
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -394,8 +394,18 @@ if ($with_fluentd -And (Test-Path -Path "$fluentd_base_dir\bin\fluentd")) {
     throw "$fluentd_base_dir\bin\fluentd is already installed. Remove/Uninstall fluentd and re-run this script."
 }
 
-if ($with_dotnet_instrumentation -and Get-IsSignalFxInstalled) {
-    throw "SignalFx Instrumentation for .NET is already installed. Remove/Uninstall SignalFx Instrumentation for .NET and re-run this script."
+if ($with_dotnet_instrumentation) {
+    echo "Installing SignalFx Instrumentation for .NET ..."
+    $api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
+    $module_name = "Splunk.SignalFx.DotNet.psm1"
+    echo "Downloading .NET Instrumentation installer ..."
+    $download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $module_name } | Select-Object -Property browser_download_url,name
+    $dotnet_autoinstr_path = Join-Path $tempdir $download.name
+    Invoke-WebRequest -Uri $download.browser_download_url -OutFile $dotnet_autoinstr_path
+    Import-Module $dotnet_autoinstr_path
+    if (Get-IsSignalFxInstalled) {
+        throw "SignalFx Instrumentation for .NET is already installed. Remove/Uninstall SignalFx Instrumentation for .NET and re-run this script."
+    }
 }
 
 if ($ingest_url -eq "") {
@@ -559,14 +569,6 @@ if ($with_fluentd) {
 }
 
 if ($with_dotnet_instrumentation) {
-    echo "Installing SignalFx Instrumentation for .NET ..."
-    $api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
-    $module_name = "Splunk.SignalFx.DotNet.psm1"
-    echo "Downloading .NET Instrumentation installer ..."
-    $download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $module_name } | Select-Object -Property browser_download_url,name
-    $dotnet_auto_path = Join-Path $tempdir $download.name
-    Invoke-WebRequest -Uri $download.browser_download_url -OutFile $dotnet_auto_path
-    Import-Module $dotnet_auto_path
     echo "Installing SignalFx Dotnet Auto Instrumentation..."
     Install-SignalFxDotnet
 
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -576,16 +576,11 @@ if ($with_dotnet_instrumentation) {
     } else {
         # signalfx-dotnet-tracing github repository API
         $api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
-
         # File pattern to search for
         $pattern = "signalfx-dotnet-tracing-*-x64.msi"
-
-        echo "Finding latest MSI to download ..."
+        echo "Finding latest .NET Instrumentation MSI to download ..."
         $download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $pattern } | Select-Object -Property browser_download_url,name
-
-        # Download installer MSI to Temp
         $msi = Join-Path $env:temp $download.name
-        echo "Downloading $download.browser_download_url ..."
         Invoke-WebRequest -Uri $download.browser_download_url -OutFile $msi
     }
 
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -68,10 +68,10 @@
     (OPTIONAL) Whether to install and configure .NET tracing to forward .NET application traces to the local collector (default: $false)
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true
-.PARAMETER signalfx_env
+.PARAMETER deployment_env
     (OPTIONAL) A system-wide SignalFx "environment" used by .NET instrumentation. Sets the SIGNALFX_ENV environment variable. Ignored if -with_dotnet_instrumentation is false.
     .EXAMPLE
-    .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true -signalfx_env staging
+    .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true -deployment_env staging
 .PARAMETER bundle_dir
     (OPTIONAL) The location of your Smart Agent bundle for monitor functionality (default: C:\Program Files\Splunk\OpenTelemetry Collector\agent-bundle)
     .EXAMPLE
@@ -123,7 +123,7 @@ param (
     [string]$msi_path = "",
     [string]$collector_msi_url = "",
     [string]$fluentd_msi_url = "",
-    [string]$signalfx_env = "",
+    [string]$deployment_env = "",
     [bool]$UNIT_TEST = $false
 )
 
@@ -565,9 +565,9 @@ if ($with_dotnet_instrumentation) {
     echo "Installing SignalFx Dotnet Auto Instrumentation..."
     Install-SignalFxDotnet
 
-    if ($signalfx_env -ne "") {
-        echo "Setting SIGNALFX_ENV environment variable to $signalfx_env ..."
-        update_registry -path "$regkey" -name "SIGNALFX_ENV" -value "$signalfx_env"
+    if ($deployment_env -ne "") {
+        echo "Setting SIGNALFX_ENV environment variable to $deployment_env ..."
+        update_registry -path "$regkey" -name "SIGNALFX_ENV" -value "$deployment_env"
     } else {
         echo "SIGNALFX_ENV environment variable not set. Unless otherwise defined, will appear as 'unknown' in the UI."
     }
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -126,7 +126,7 @@ param (
     [string]$collector_version = "",
     [bool]$with_fluentd = $true,
     [bool]$with_dotnet_instrumentation = $false,
-    [string]$instrumentation_exclude_processes = "Powershell.exe,dotnet.exe",
+    [string]$instrumentation_exclude_processes = "Powershell.exe;dotnet.exe",
     [string]$bundle_dir = "",
     [ValidateSet('test','beta','release')][string]$stage = "release",
     [string]$msi_path = "",
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -176,14 +176,6 @@ $fluentd_config_dir = "$fluentd_base_dir\etc\td-agent"
 $fluentd_config_path = "$fluentd_config_dir\td-agent.conf"
 $fluentd_service_name = "fluentdwinsvc"
 
-$dotnet_tracing_install_dir = "\Program Files\SignalFx\.NET Tracing"
-try {
-    Resolve-Path $env:SYSTEMDRIVE
-    $dotnet_tracing_base_dir = "${env:SYSTEMDRIVE}$dotnet_tracing_install_dir"
-} catch {
-    $dotnet_tracing_base_dir = $dotnet_tracing_install_dir
-}
-
 # check that we're not running with a restricted execution policy
 function check_policy() {
     $executionPolicy  = (Get-ExecutionPolicy)
@@ -400,10 +392,6 @@ if ($with_fluentd -And (service_installed -name "$fluentd_service_name")) {
 
 if ($with_fluentd -And (Test-Path -Path "$fluentd_base_dir\bin\fluentd")) {
     throw "$fluentd_base_dir\bin\fluentd is already installed. Remove/Uninstall fluentd and re-run this script."
-}
-
-if ($with_dotnet_instrumentation -And (Test-Path -Path "$dotnet_tracing_base_dir\SignalFx.Tracing.ClrProfiler.Native.dll")) {
-    throw "SignalFx .NET tracing is already installed. Remove/Uninstall SignalFx .NET tracing and re-run this script."
 }
 
 if ($ingest_url -eq "") {
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -101,10 +101,6 @@
     (OPTIONAL) Specify the URL to the Fluentd MSI package to install (default: "https://packages.treasuredata.com/4/windows/td-agent-4.1.0-x64.msi")
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -fluentd_msi_url https://my.host/td-agent-4.1.0-x64.msi
-.PARAMETER dotnet_instrumentation_msi_url
-    (OPTIONAL) Specify the URL to the SignalFx .NET Instrumentation MSI package to install (default: "https://github.com/signalfx/signalfx-dotnet-tracing/releases/download/v0.2.9/signalfx-dotnet-tracing-0.2.9-x64.msi")
-    .EXAMPLE
-    .\install.ps1 -access_token "ACCESSTOKEN" -dotnet_instrumentation_msi_url https://my.host/signalfx-dotnet-tracing-0.2.9-x64.msi
 .PARAMETER msi_path
     (OPTIONAL) Specify a local path to a Splunk OpenTelemetry Collector MSI package to install instead of downloading the package.
     If specified, the -collector_version and -stage parameters will be ignored.
@@ -569,29 +565,18 @@ if ($with_fluentd) {
 
 if ($with_dotnet_instrumentation) {
     echo "Installing SignalFx Instrumentation for .NET ..."
-    if ($dotnet_instrumentation_msi_url) {
-        $dotnet_tracing_msi_name = "signalfx-dotnet-tracing.msi"
-        echo "Downloading $dotnet_instrumentation_msi_url..."
-        download_file -url "$dotnet_instrumentation_msi_url" -outputDir "$tempdir" -fileName "$dotnet_tracing_msi_name"
-        $msi = (Join-Path "$tempdir" "$fluentd_msi_name")
-    } else {
-        # signalfx-dotnet-tracing github repository API
-        $api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
-        # File pattern to search for
-        $pattern = "signalfx-dotnet-tracing-*-x64.msi"
-        echo "Finding latest .NET Instrumentation MSI to download ..."
-        $download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $pattern } | Select-Object -Property browser_download_url,name
-        $msi = Join-Path $env:temp $download.name
-        Invoke-WebRequest -Uri $download.browser_download_url -OutFile $msi
-    }
+    # signalfx-dotnet-tracing github repository API
+    $api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
+    # File pattern to search for
+    $module_name = "install.psm1"
+    echo "Finding latest .NET Instrumentation installer to download ..."
+    $download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $module_name } | Select-Object -Property browser_download_url,name
+    $tracing_module_installer_path = Join-Path $Env:ProgramFiles\WindowsPowerShell\Modules $download.name
+    Invoke-WebRequest -Uri $download.browser_download_url -OutFile $tracing_module_installer_path
 
     # Install downloaded MSI
     echo "Starting install process ..."
-    Start-Process msiexec.exe -Wait -ArgumentList "/I $msi /quiet"
-
-    # Cleanup
-    echo "Cleaning up ..."
-    Remove-Item $msi
+    Install-SignalFxDotnet()
 
     echo "Setting environment variables for instrumentation ..."
     update_registry -path "$regkey" -name "COR_ENABLE_PROFILING" -value "1"
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -126,6 +126,7 @@ param (
     [string]$collector_version = "",
     [bool]$with_fluentd = $true,
     [bool]$with_dotnet_instrumentation = $false,
+    [string]$instrumentation_exclude_processes = "Powershell.exe,dotnet.exe",
     [string]$bundle_dir = "",
     [ValidateSet('test','beta','release')][string]$stage = "release",
     [string]$msi_path = "",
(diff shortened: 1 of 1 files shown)


atoulme:
I'm not sure if there are tests around this, is this ps1 file loaded up in the installer? I know there are some tests with Ansible. I also know we have an open issue to test Windows. I'd take your word for it that it works well for this PR.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -64,6 +64,18 @@
     (OPTIONAL) Whether to install and configure fluentd to forward log events to the collector (default: $true)
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_fluentd $false
+.PARAMETER with_dotnet_tracing
+    (OPTIONAL) Whether to install and configure .NET tracing to forward .NET application traces to the local collector (default: $false)
```
pjanotti, line 68: @theletterf do we have something better than `observability data`?

```suggestion
    (OPTIONAL) Whether to install and configure .NET instrumentation to forward .NET application observability data to the local collector (default: $false)
```

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -64,6 +64,18 @@
...
+    .EXAMPLE
+    .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_tracing $true
+.PARAMETER signalfx_service_name
```
pjanotti, line 71: Q.: what are the plans when there is more than on app on the box? The typical case would be multiple apps on Windows IIS.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -89,6 +101,10 @@
     (OPTIONAL) Specify the URL to the Fluentd MSI package to install (default: "https://packages.treasuredata.com/4/windows/td-agent-4.1.0-x64.msi")
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -fluentd_msi_url https://my.host/td-agent-4.1.0-x64.msi
+.PARAMETER dotnet_tracing_msi_url
+    (OPTIONAL) Specify the URL to the SignalFx .NET Tracing MSI package to install (default: "https://github.com/signalfx/signalfx-dotnet-tracing/releases/download/v0.2.9/signalfx-dotnet-tracing-0.2.9-x64.msi")
```
pjanotti, line 105: Don't know if there is a syntax for that, but, ideally the version should be a parameter or constant.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -109,11 +125,15 @@ param (
     [bool]$insecure = $false,
     [string]$collector_version = "",
     [bool]$with_fluentd = $true,
+    [bool]$with_dotnet_tracing = $false,
```
pjanotti, line 128: Isn't this going to be manageable when more runtimes are added? Anyway, an alternative can be searched when we get to that point.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -157,6 +177,16 @@ $fluentd_config_dir = "$fluentd_base_dir\etc\td-agent"
 $fluentd_config_path = "$fluentd_config_dir\td-agent.conf"
 $fluentd_service_name = "fluentdwinsvc"
 
+$dotnet_tracing_msi_name = "signalfx-dotnet-tracing-0.2.9-x64.msi"
```
pjanotti, line 180: Ideally the version number is a parameter/const.

pjanotti: Same for bitness, there could be someone using Windows x86

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -375,6 +405,10 @@ if ($with_fluentd -And (Test-Path -Path "$fluentd_base_dir\bin\fluentd")) {
     throw "$fluentd_base_dir\bin\fluentd is already installed. Remove/Uninstall fluentd and re-run this script."
 }
 
+if ($with_dotnet_tracing -And (Test-Path -Path "$dotnet_tracing_base_dir\SignalFx.Tracing.ClrProfiler.Native.dll")) {
+    throw "SignalFx .NET tracing is already installed. Remove/Uninstall SignalFx .NET tracing and re-run this script."
```
pjanotti, line 409: It seems worth to automatically run the uninstall or at least give an option to do it.

pjanotti (commented):
@pmcollins going in the right direction 👍 I will share with other folks that have been working with download/install scripts on the dotnet team.

internal/buildscripts/packaging/installer/install.ps1
pmcollins: Yeah I was kind of on the fence about that -- the reason that I didn't is because we're not doing that with fluentd in this script, and thought maybe it should be a separate effort for the entire script (possibly even handled by Jeff when he gets back). But not a big deal if we want to it now.

internal/buildscripts/packaging/installer/install.ps1
pmcollins: Interesting. It looks like this whole script is for "amd64" though, isn't it?

internal/buildscripts/packaging/installer/install.ps1
pjanotti: Yes, SFx.NET does that. I'm thinking about the wizard in this case: how we communicate that to the user?

internal/buildscripts/packaging/installer/install.ps1
pjanotti: ah, if we are limited to amd64 then... but we did had some users with 32bit versions of Windows in the past. On x64 for box the installer already installs both the x64 and x86 binaries.

internal/buildscripts/packaging/installer/install.ps1
RassK: It can also default to latest public release. Otherwise this script needs PR in every release.
https://github.com/signalfx/signalfx-dotnet-tracing#automated-download-and-installation-for-windows

internal/buildscripts/packaging/installer/install.ps1
RassK: I guess we could just have a switch map. When switch is enabled it downloads the sub-script from the repo artifacts. This way the actual team managing specific language can always auto update the script when needed. Plus version parity stays on if specific release needs something special.

internal/buildscripts/packaging/installer/install.ps1
RassK: I think Windows Server 2008 was the last OS with 32bit. I really don't want to support it 😆

pjanotti: I didn't know that @RassK !!!! 😮

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -535,6 +574,33 @@ if ($with_fluentd) {
...
+if ($with_dotnet_tracing) {
+    echo "Downloading $dotnet_tracing_dl_url..."
+    download_file -url "$dotnet_tracing_dl_url" -outputDir "$tempdir" -fileName "$dotnet_tracing_msi_name"
+    $dotnet_tracing_msi_path = (Join-Path "$tempdir" "$dotnet_tracing_msi_name")
+
+    echo "Installing $dotnet_tracing_msi_path ..."
+    Start-Process msiexec.exe -Wait -ArgumentList "/qn /norestart /i `"$dotnet_tracing_msi_path`""
+    echo "- Done"
+
+    update_registry -path "$regkey" -name "COR_ENABLE_PROFILING" -value "1"
```
RassK, line 586: Isn't it going to enable profiling globally? (machine level variable)
That could cause crashes and even disable remote login, due there are many local applications built on .NET. Specially it doesn't work well with PowerShell (which is also .NET based).

pmcollins: Yes, definitely -- the idea is to have the customer just install one thing and auto-instrumentation just work ("zero config"). Can there be, or is there, a filter in the .NET library to not instrument some applications?

RassK: https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/advanced-config.md#global-management-settings
`SIGNALFX_PROFILER_EXCLUDE_PROCESSES`

Unfortunately we didn't work out whole list, what should be disabled. PowerShell, Server Manager, IIS Manager are the first ones that come to my mind.

internal/buildscripts/packaging/installer/install.ps1
RassK: If someone figures it out, then I guess the same solution could be applied to this issue >
https://github.com/signalfx/splunk-otel-collector/pull/2068#discussion_r986215896

aunshc: For the wizard, we can provide an option to customers to set the service name as they would like upfront. It would definitely help to allow explicit naming per app instrumented on the host/VM - based on the above, is this not possible today with both the SFx and OTel .NET libraries?

internal/buildscripts/packaging/installer/install.ps1
aunshc: I'm aware of an effort that the Java folks worked on to automatically update the release version in the Collector with every release of the agent - https://signalfuse.atlassian.net/browse/APMI-3304, would be helpful to check in to see if a similar approach is possible for .NET as well.

internal/buildscripts/packaging/installer/install.ps1
aunshc: +1 the ability to exclude processes from profiling is important. Similarly, is there a way to filter processes for trace instrumentation?

RassK: Yes, we didn't separate the logic, 1 env works for all.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -64,6 +64,18 @@
     (OPTIONAL) Whether to install and configure fluentd to forward log events to the collector (default: $true)
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_fluentd $false
+.PARAMETER with_dotnet_tracing
```
Kielek, line 67: Consider, with_dotnet_instrumentation.
The change should be propagated to other places like `dotnet_tracing_msi_url`, etc.

internal/buildscripts/packaging/installer/install.ps1
RassK: Yes, it's definitely possible in .NET also, although for me it's seems too minimal. The collector project should have too much knowledge about it's satellite projects. Instead of version only, you can download whole script.

internal/buildscripts/packaging/installer/install.ps1
RassK: Explicitly setting is no problem at all. I guess @pjanotti just meant that it's not basically 0 config (you just setup the collector and instrumentation and you are good to go). While in Windows it's very common to have multiple services on one instance, the default behaviour just doesn't work in UX wise (if the wizard is asking to setup global service name).

internal/buildscripts/packaging/installer/install.ps1
pmcollins: I like this idea. The thing is that the msi is called signalfx-dotnet-tracing-xyz.msi, so I ended up using this terminology everywhere. Will fix.

pmcollins: Having second thoughts, since mixing dotnet_instrumentation and dotnet_tracing (e.g. https://github.com/signalfx/signalfx-dotnet-tracing) might be confusing.

internal/buildscripts/packaging/installer/install.ps1
pjanotti: @aunshc @pmcollins @RassK we have to be careful with the term "profiling" in the .NET context. When we say "profiling" we mean "a .NET profiler" or "CLR profiler" that is a term to identify a piece of code that attaches to a .NET process that has capabilities of doing bytecode instrumentation. We also use the same technology to do the actual "Always on Profiling", ie.: thread sampling/CPU profiling/Memory profiling, features The `COR_ENABLE_PROFILING` above is about attaching a ".NET Profiler" to the process if it will generate traces, metrics, or thread sampling data is controlled by other settings.

To @aunshc point: yes, it is possible to enable/disable "always profiling" separately from distributed tracing settings, see https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/always-on-profiling.md#configuration-settings

internal/buildscripts/packaging/installer/install.ps1
pjanotti: In principle, I like @RassK's idea: the .NET side controls the script that downloads and installs the .NET instrumentation the collector just need to know about the script. The point that seems a problem to me is regarding the Wizard: if we need a new option there or vice-versa, how we track this type of change?

internal/buildscripts/packaging/installer/install.ps1
pjanotti: We are keeping the name `signalfx-dotnet-tracing` because of legacy but it will deliver other features like metrics and Always On Profiling. In light of that I think it is better to use the more general `instrumentation` instead.

internal/buildscripts/packaging/installer/install.ps1
pellared: Setting `COR_ENABLE_PROFILING` globally can bring catastrophic side-effects.

Lots of processes on Windows are .NET. It leads to instrumenting too much. 

At some point, we changed our docs to make sure that one does NOT enable it globally and also added a troubleshooting topic: https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/troubleshooting.md#high-cpu-usage

We are not setting these env vars in the MSI on purpose

pellared: We can also consider setting these if the user provides process names that should be instrumented via `SIGNALFX_PROFILER_PROCESSES` (allowlist). 

Getting a proper value for `SIGNALFX_PROFILER_EXCLUDE_PROCESSES` (blocklist) may be hard. But maybe this is the way to go... We could achieve "zero-config" this way.

internal/buildscripts/packaging/installer/install.ps1
aunshc: The goal with 0 config here is to get started with APM with the first deploy of the Collector (as OOTB as possible) with no explicit step or configuration for the instrumentation agents at that point. Customers should be able to customize (exclude processes, set names) if needed thereafter with recommended documentation for steps on the machine. Ideally, we would like such customization to move to the UI in the future but that is not possible today. Circling back to our discussion - makes sense, I see this as an opportunity for us to remove the global service name from the wizard and letting the .NET infer it to avoid confusion. 

@pjanotti @RassK is there any documentation on how can customers explicitly configure names per app vs the env var for the whole machine?

internal/buildscripts/packaging/installer/install.ps1
aunshc: Since we are majorly targeting net new customers for this effort, the way we went about it for java was to expect the latest agent version being pulled down as part of the Collector deploy (latest version as well). As part of a future improvements, we can definitely provide an option to allow a manual intervention by customers to set the version as they would like from the wizard, reflecting on the installer instructions thereafter.

internal/buildscripts/packaging/installer/install.ps1
aunshc: Got it, thanks for the explanation! Good to know that there is a separate way to enable/disable always on profiling. Is there a way to associate the value of this setting to a specific .NET process vs the host globally?

pellared: @aunshc This is not a way to disable "always on profiling" but the whole auto-instrumentation 😬 

[Here](https://github.com/signalfx/signalfx-dotnet-tracing/blob/v0.2.9/docs/advanced-config.md#alwayson-profiling-settings) are the settings for AlwaysOnProfiling if you are interested.

internal/buildscripts/packaging/installer/install.ps1
RassK: We currently instruct users to set process scoped environment variables. In the advanced section we also mention how to do it differently (through a config file).

Refs >
* https://github.com/signalfx/signalfx-dotnet-tracing#instrument-a-net-application-on-windows
* https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/advanced-config.md#advanced-configuration

RassK: Which brings a rough idea to me. If UI would allow to select root paths of the interested applications, it can plant a config file there (unfortunately for .NET only SignalFx / Splunk variables work, as these can be picked up alternatively). 
The second issue is devops pipelines, there could be a config option that removes all items in the development path, before deploying a new version. That means the most effective UX / UI would be in the tool itself (for eg a Octopus plugin).

internal/buildscripts/packaging/installer/install.ps1
RassK: Yep, with environment variables you can choose the right scope.... but the smaller the scope gets, the harder is to inject these variables. It's easy to setup global variables, but then these values can leak to unwanted processes. 

Why is it hard to have a small scope? - because mostly you just don't manually execute server's service. The simple .exe / .dll might be there but it's usually started by 3rd party: eg a webserver (IIS), scheduler, service watch etc. So you have to locate the master service and inject a config into startup.

internal/buildscripts/packaging/installer/install.ps1
RassK: I think that choosing the version is smaller issue than moving the script control back to the original repo.
The question here is not about the functionality but separation concerns between repos. The collector can still do what it's doing, but it first needs to download a install script which then downloads and installs a right version.
So the logic (control) how the instrumentation is installed, is maintained by the most knowledgeable part (in this specific case it would be the .net team). 
This also assures that you are downloading a right setup script for the right version. Handling this in the collector's setup script is not a good idea.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -109,11 +125,15 @@ param (
     [bool]$insecure = $false,
     [string]$collector_version = "",
```
RassK, line 126: nit: parameter naming convention error
https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/strongly-encouraged-development-guidelines?view=powershell-7.2#use-pascal-case-for-parameter-names

internal/buildscripts/packaging/installer/install.ps1
aunshc: Yes @pellared ,makes sense - was referring to Paulo's comment above with https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/docs/always-on-profiling.md#configuration-settings.

internal/buildscripts/packaging/installer/install.ps1
pmcollins: @RassK I like this idea. There's a snippet in the markdown file that installs the latest version, but I'm not seeing a script in your repo. If you think it would be appropriate, would someone on your team be able to create it and put it in the right place?

RassK: PR here https://github.com/signalfx/signalfx-dotnet-tracing/pull/635

RassK: note: we may still need to standardize module naming if we adapt this pattern. Otherwise, there can be conflicts with other languages.

RassK: [Splunk.SignalFx.DotNet.psm1](https://github.com/signalfx/signalfx-dotnet-tracing/releases/download/v0.2.9/Splunk.SignalFx.DotNet.psm1)

Module name is now updated to `Splunk.SignalFx.DotNet.psm1` for better compatibility.

internal/buildscripts/packaging/installer/install.ps1
pmcollins: I think this is a fine idea, but we're not doing that for the other installers in this script. What do you think about making that a separate effort?

internal/buildscripts/packaging/installer/install.ps1
pmcollins: Nice idea. I think we can punt on this until we bring in the other languages though. Since we're generating the script args, I don't think backwards compatibility is going to be a problem.

internal/buildscripts/packaging/installer/install.ps1
pmcollins: Good to know. Perhaps it would be best to tackle this, along with some of the other items you all have mentioned when @jeffreyc-splunk gets back.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -535,6 +566,42 @@ if ($with_fluentd) {
     echo "- Started"
 }
 
+if ($with_dotnet_instrumentation) {
+    echo "Installing SignalFx Instrumentation for .NET ..."
+    $api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
+    $module_name = "Splunk.SignalFx.DotNet.psm1"
+    echo "Downloading .NET Instrumentation installer ..."
+    $download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $module_name } | Select-Object -Property browser_download_url,name
+    $dotnet_auto_path = Join-Path $env:temp $download.name
```
RassK, line 575: There might be a chance that the temp path is somehow deleted. I ran into this issue with one of my VMs. 
Here is the fix I did for SignalFx and similar to OpenTelemetry.
https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/Splunk.SignalFx.DotNet.psm1#L7

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -535,6 +566,42 @@ if ($with_fluentd) {
...
+    update_registry -path "$regkey" -name "CORECLR_PROFILER" -value "{B4C89B0F-9908-4F73-9F59-0D77C5A06874}"
+
+    if ($instrumentation_exclude_processes -ne "") {
+      echo "Setting SIGNALFX_PROFILER_EXCLUDE_PROCESSES environment variable to $instrumentation_exclude_processes ..."
+      update_registry -path "$regkey" -name "SIGNALFX_PROFILER_EXCLUDE_PROCESSES" -value "$instrumentation_exclude_processes"
+    }
+
+    if ($signalfx_service_name -ne "") {
+        echo "Setting SIGNALFX_SERVICE_NAME environment variable to $signalfx_service_name ..."
+        update_registry -path "$regkey" -name "SIGNALFX_SERVICE_NAME" -value "$signalfx_service_name"
```
RassK, line 594: This works only with single service. If more than one services are installed to that machine, all of them will report the same service name.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -535,6 +566,42 @@ if ($with_fluentd) {
...
+
+    echo "Setting environment variables for instrumentation ..."
+    update_registry -path "$regkey" -name "COR_ENABLE_PROFILING" -value "1"
+    update_registry -path "$regkey" -name "COR_PROFILER" -value "{B4C89B0F-9908-4F73-9F59-0D77C5A06874}"
+    update_registry -path "$regkey" -name "CORECLR_ENABLE_PROFILING" -value "1"
+    update_registry -path "$regkey" -name "CORECLR_PROFILER" -value "{B4C89B0F-9908-4F73-9F59-0D77C5A06874}"
+
+    if ($instrumentation_exclude_processes -ne "") {
+      echo "Setting SIGNALFX_PROFILER_EXCLUDE_PROCESSES environment variable to $instrumentation_exclude_processes ..."
+      update_registry -path "$regkey" -name "SIGNALFX_PROFILER_EXCLUDE_PROCESSES" -value "$instrumentation_exclude_processes"
```
RassK, line 589: When SignalFx is enabled globally, at least PowerShell must be definitely disabled. I would also add IIS manager and Server manager to the exclusion list. (these 3 are known not to be instrumented ever, but PowerShell is especially bad. If remote server is managed over remote PowerShell session, it will crash the connection and server is never accessible. At least one SWAT connected to that issue.)

internal/buildscripts/packaging/installer/install.ps1
pmcollins: Looks like this scenario is handled on line 161. I've updated this code to use `$tempdir`.

internal/buildscripts/packaging/installer/install.ps1
RassK: I think this kind of logic should be in the PS module actually. 
Currently `$dotnet_tracing_install_dir` `$dotnet_tracing_base_dir` `SignalFx.Tracing.ClrProfiler.Native.dll` values are hardcoded into the collector's script, but these can change / be specific to the version. 

Should do something like > 
```powershell
if ($with_dotnet_instrumentation -And (Test-Path -Path Get-SignalFxInstallDirectory))
```

pmcollins: I have made this change. However I just realized that the tracing/profiling installer runs after other things have been installed, whereas before, the test for whether anything has already been installed happened before anything got installed, so we had the ability to abort installing anything unless we knew that everything was going to succeed. Perhaps the tracing module should export another function (an is_installed test) that we can call early on in the process, before we've installed anything.

RassK: Linking this chance here also > 
https://github.com/signalfx/signalfx-dotnet-tracing/pull/653

RassK: @pmcollins I updated the module, it's now merged.

This should work now > 

```csharp
if ($with_dotnet_instrumentation -and Get-IsSignalFxInstalled)
```

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -535,6 +554,42 @@ if ($with_fluentd) {
...
+    Invoke-WebRequest -Uri $download.browser_download_url -OutFile $dotnet_auto_path
+    Import-Module $dotnet_auto_path
+    echo "Installing SignalFx Dotnet Auto Instrumentation..."
+    Install-SignalFxDotnet
+
+    echo "Setting environment variables for instrumentation ..."
+    update_registry -path "$regkey" -name "COR_ENABLE_PROFILING" -value "1"
+    update_registry -path "$regkey" -name "COR_PROFILER" -value "{B4C89B0F-9908-4F73-9F59-0D77C5A06874}"
+    update_registry -path "$regkey" -name "CORECLR_ENABLE_PROFILING" -value "1"
+    update_registry -path "$regkey" -name "CORECLR_PROFILER" -value "{B4C89B0F-9908-4F73-9F59-0D77C5A06874}"
```
RassK, line 573: @pjanotti @pellared @pmcollins since we already have PS module. I'm thinking of moving these 4 variables also to the newly created module. Something like `Register-SignalFxForGlobal`. This bit seems to be the last specifics to the dotnet, so collector's script will be very clean and without any specific's knowledge. 

+ There is a chance that cor profiler's guid will change. Eg when implementing multiplexer etc. So, any possibility to change should be in version specific setup script.

pellared: But what are we trying to achieve here? Instrument `csrss.exe`?

pmcollins: No, I haven't tested this extensively and was hoping to get more info from your team about performance implications and a testing strategy. I hope you can make it to tomorrow's meeting about this at 12:30 eastern so we can discuss.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -64,6 +64,21 @@
...
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_fluentd $false
+.PARAMETER with_dotnet_instrumentation
+    (OPTIONAL) Whether to install and configure .NET tracing to forward .NET application traces to the local collector (default: $false)
+    .EXAMPLE
+    .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true
+.PARAMETER instrumentation_exclude_processes
+    (OPTIONAL) A semicolon-delimited list of process names to be excluded from auto-instrumentation (default: Powershell.exe;dotnet.exe)
+    .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true -instrumentation_exclude_processes "Powershell.exe;dotnet.exe;myprogram.exe"
+.PARAMETER signalfx_service_name
```
aunshc, line 74: Keeping the transition to OTel .NET auto instrumentation in mind once it is stable in the future (alpha today) and we want to move away from signalfx references, I suggest renaming the service name and env parameters to avoid these changes in the future @pmcollins  - wdyt?

pmcollins: Should they be renamed as follows?

`signalfx_service_name` -> `service_name`
`signalfx_env` -> `profiling_env`

I'm not sure we need to do this though because we generate the arguments and download the script that processes those arguments on demand. When we move over to OTel, we can just start using OTel naming without worrying about compatibility.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -64,6 +64,21 @@
     (OPTIONAL) Whether to install and configure fluentd to forward log events to the collector (default: $true)
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_fluentd $false
+.PARAMETER with_dotnet_instrumentation
+    (OPTIONAL) Whether to install and configure .NET tracing to forward .NET application traces to the local collector (default: $false)
+    .EXAMPLE
+    .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true
+.PARAMETER instrumentation_exclude_processes
+    (OPTIONAL) A semicolon-delimited list of process names to be excluded from auto-instrumentation (default: Powershell.exe;dotnet.exe)
```
pjanotti, line 72: Please include also Visual Studio: devenv.exe

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -109,11 +124,15 @@ param (
     [bool]$insecure = $false,
     [string]$collector_version = "",
     [bool]$with_fluentd = $true,
+    [bool]$with_dotnet_instrumentation = $false,
+    [string]$instrumentation_exclude_processes = "Powershell.exe;dotnet.exe",
```
pjanotti, line 128: Same as above: 
```suggestion
    [string]$instrumentation_exclude_processes = "Powershell.exe;dotnet.exe;devenv.exe",
```

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -535,6 +554,42 @@ if ($with_fluentd) {
...
+    echo "Downloading .NET Instrumentation installer ..."
+    $download = (Invoke-WebRequest $api | ConvertFrom-Json).assets | Where-Object { $_.name -like $module_name } | Select-Object -Property browser_download_url,name
+    $dotnet_auto_path = Join-Path $tempdir $download.name
+    Invoke-WebRequest -Uri $download.browser_download_url -OutFile $dotnet_auto_path
+    Import-Module $dotnet_auto_path
+    echo "Installing SignalFx Dotnet Auto Instrumentation..."
+    Install-SignalFxDotnet
+
+    echo "Setting environment variables for instrumentation ..."
+    update_registry -path "$regkey" -name "COR_ENABLE_PROFILING" -value "1"
```
pjanotti, line 570: Per meeting today let's exclude the COR* and CORECLR* env vars from here.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -535,6 +554,42 @@ if ($with_fluentd) {
...
+    update_registry -path "$regkey" -name "CORECLR_ENABLE_PROFILING" -value "1"
+    update_registry -path "$regkey" -name "CORECLR_PROFILER" -value "{B4C89B0F-9908-4F73-9F59-0D77C5A06874}"
+
+    if ($instrumentation_exclude_processes -ne "") {
+      echo "Setting SIGNALFX_PROFILER_EXCLUDE_PROCESSES environment variable to $instrumentation_exclude_processes ..."
+      update_registry -path "$regkey" -name "SIGNALFX_PROFILER_EXCLUDE_PROCESSES" -value "$instrumentation_exclude_processes"
+    }
+
+    if ($signalfx_service_name -ne "") {
+        echo "Setting SIGNALFX_SERVICE_NAME environment variable to $signalfx_service_name ..."
```
pjanotti, line 581: I still have misgivings in offering service name when we are targeting IIS apps. If there is more than one on the box it doesn't work well. I would rather remove service name from the wizard.

pjanotti (commented):
Just minor stuff per our latest conversation.

internal/buildscripts/packaging/installer/install.ps1
pjanotti: although here is the script not the wizard 🤦‍♂️

internal/buildscripts/packaging/installer/install.ps1
pellared:
```suggestion
    [string]$instrumentation_exclude_processes = "Powershell.exe;dotnet.exe",devenv.exe,com.docker.service,com.docker.service.exe"
```
It looks like it can be without `.exe` suffix (can you double-check it @pjanotti on your machine?)

However https://docs.docker.com/desktop/windows/permission-requirements/ mentions the `.exe` suffix 🤷

RassK: I'm not sure what we try to achieve here. If we don't go with globally enabled instrumentation, this list here is non important.

If we want to store these just in case, then I would add:
Server Manager - ServerManager.exe
IIS Manager - InetMgr.exe

pellared: > If we don't go with globally enabled instrumentation, this list here is non important.

You are right 😆

internal/buildscripts/packaging/installer/install.ps1
aunshc: Got it, thanks Pablo. Since these are parameters exposed as options for the installer, it would be best to rename the environment option upfront (because of the feedback we have received about naming in the past) while the SIGNALFX_ENV env variable is set underneath. I would suggest the following:
```
signalfx_env -> deployment_env
```

internal/buildscripts/packaging/installer/install.ps1
pmcollins: Makes sense. I've removed this override.

internal/buildscripts/packaging/installer/install.ps1
pmcollins: I removed all of the env vars that were set here. Do some need to stay? If so, perhaps they could be set by the instrumentation install script?

internal/buildscripts/packaging/installer/install.ps1
pmcollins: This has been removed 🙂

aunshc: +1 we are not moving forward with other .NET processes and the ability to exclude processes for the first version, only IIS apps. @pmcollins has removed that option for now based on yesterday's discussion, we will keep the above processes in mind for future iterations.

internal/buildscripts/packaging/installer/install.ps1
pmcollins: The service name override option is now gone. The wizard should set the sfx env.

pmcollins: Sorry, Aunsh, I didn't see your message when I responded. I'll change sfx env to deployment env shortly.

aunshc: No worries, I added a comment around the same time the changes we pushed I believe. Thanks!

pmcollins: Ok, this is done.

pjanotti (approved):

jvoravong (approved):

PR #2077, opened 2022-10-05
remove some exports from the smartagent receiver

atoulme:
Unexport all types that can be made package private for the smart agent receiver.
This is a change that should not affect any logic in the code.

### DIFF
--- a/internal/receiver/smartagentreceiver/output.go
+++ b/internal/receiver/smartagentreceiver/output.go
@@ -35,11 +35,11 @@ import (
 
 const internalTransport = "internal"
 
-// Output is an implementation of a Smart Agent FilteringOutput that receives datapoints, events, and dimension updates
+// output is an implementation of a Smart Agent FilteringOutput that receives datapoints, events, and dimension updates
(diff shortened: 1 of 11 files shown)


dmitryax (approved):

internal/receiver/smartagentreceiver/output.go
```diff
@@ -162,37 +162,37 @@ func getLoneSFxExporter(host component.Host, exporterType collectorConfig.DataTy
 
 }
 
-func (output *Output) AddDatapointExclusionFilter(filter dpfilters.DatapointFilter) {
+func (output *output) AddDatapointExclusionFilter(filter dpfilters.DatapointFilter) {
```
dmitryax, line 165:
```suggestion
func (out *output) AddDatapointExclusionFilter(filter dpfilters.DatapointFilter) {
```
just to avoid overriding the type

atoulme:
Not sure why checks don't run.

rmfitzpatrick (approved):

PR #2083, opened 2022-10-05
Add timestamp processor

atoulme:
This PR adds a new processor that is able to manipulate the timestamps of records processed by the OpenTelemetry collector.

The processor has a single configuration entry, offset, which accepts negative or positive offsets with time.Duration string values such as "+6h" or "-3h".


pkg/processor/timestamp/config.go
```diff
@@ -0,0 +1,65 @@
...
+// Validate checks if the processor configuration is valid
+func (cfg *Config) Validate() error {
+	if cfg.Offset[0] != '-' && cfg.Offset[0] != '+' {
+		return errorInvalidOffsetFormat
+	}
+	_, err := time.ParseDuration(cfg.Offset[1:])
+	return err
+}
+
+func (cfg *Config) offsetFn() func(pcommon.Timestamp) pcommon.Timestamp {
```
dmitryax, line 49:
```suggestion
func offsetFn(offset string) func(pcommon.Timestamp) pcommon.Timestamp {
```

Just doesn't seem like a part of config

pkg/processor/timestamp/config.go
```diff
@@ -0,0 +1,65 @@
...
+	backwards := cfg.Offset[0] == '-'
```
dmitryax, line 50: I don't think you need to do this. negative duration can be parsed as well and applied with `time.Add`

pkg/processor/timestamp/metric_processor.go
```diff
@@ -0,0 +1,88 @@
...
+						for l := 0; l < metric.Histogram().DataPoints().Len(); l++ {
+							dp := metric.Histogram().DataPoints().At(l)
+							dp.SetStartTimestamp(offsetFn(dp.StartTimestamp()))
+							dp.SetTimestamp(offsetFn(dp.Timestamp()))
+							for m := 0; m < dp.Exemplars().Len(); m++ {
+								e := dp.Exemplars().At(m)
+								e.SetTimestamp(offsetFn(e.Timestamp()))
+							}
+						}
+					case pmetric.MetricTypeNone:
```
dmitryax, line 55: looks like a redundant line

pkg/processor/timestamp/config.go
```diff
@@ -0,0 +1,65 @@
...
+
+var _ config.Processor = (*Config)(nil)
+
+// Validate checks if the processor configuration is valid
+func (cfg *Config) Validate() error {
+	if cfg.Offset[0] != '-' && cfg.Offset[0] != '+' {
+		return errorInvalidOffsetFormat
+	}
+	_, err := time.ParseDuration(cfg.Offset[1:])
+	return err
```
dmitryax, line 46:
```suggestion
	return fmt.Errorf("invalid offset format %s: %w", cfg.Offset, err)
```

pkg/processor/timestamp/config.go
```diff
@@ -0,0 +1,65 @@
...
+	Offset                   string                   `mapstructure:"offset"`
+}
+
+var _ config.Processor = (*Config)(nil)
+
+// Validate checks if the processor configuration is valid
+func (cfg *Config) Validate() error {
+	if cfg.Offset[0] != '-' && cfg.Offset[0] != '+' {
+		return errorInvalidOffsetFormat
+	}
```
dmitryax, line 44: not needed

pkg/processor/timestamp/config.go
atoulme: No, I tried and it didn't work. I expected it would but "-3h" was parsed a 3*time.Hour.

pkg/processor/timestamp/metric_processor.go
atoulme: just to make sure I list them all. I can remove.

pkg/processor/timestamp/config.go
dmitryax: https://go.dev/play/p/rFFirsgMX53 seems to work...

atoulme: Ah, I'm not reproducing. I'll check some more.

dmitryax (approved):

pkg/processor/timestamp/metric_processor.go
atoulme: I added a default clause to error out if the metric type is not supported to future proof the impl.

docs/components.md
```diff
@@ -58,6 +58,7 @@ new release. In addition, any of these components may be removed prior to the
...
 
-| Receivers | Processors | Exporters                                     | Extensions |
-|-----------|------------|-----------------------------------------------|------------|
-|           |   [span_metrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanmetricsprocessor)         | [pulsar](../internal/exporter/pulsarexporter) |            |
+| Receivers | Processors                                                                                                                 | Exporters                                     | Extensions |
+|-----------|----------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------|------------|
+|           | [span_metrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanmetricsprocessor) | [pulsar](../internal/exporter/pulsarexporter) |            |
+|           | [timestamp](https://github.com/signalfx/splunk-otel-collector/tree/main/pkg/processor/timestamp)                           |                                               |            |
```
rmfitzpatrick, line 64: nit, this can be a relative link

go.mod
```diff
@@ -498,3 +499,5 @@ replace github.com/googleapis/gnostic v0.5.6 => github.com/googleapis/gnostic v0
 
 // required to drop dependency on deprecated git.apache.org/thrift.git
 exclude go.opencensus.io v0.19.1
+
+replace github.com/signalfx/splunk-otel-collector/processor/timestampprocessor => ./pkg/processor/timestamp
```
rmfitzpatrick, line 503: nit, do you mind sharing with the current tests replace directive to start a central location for the project components?

internal/components/components.go
```diff
@@ -85,6 +85,7 @@ import (
 	"github.com/signalfx/splunk-otel-collector/internal/extension/smartagentextension"
 	"github.com/signalfx/splunk-otel-collector/internal/receiver/databricksreceiver"
 	"github.com/signalfx/splunk-otel-collector/internal/receiver/smartagentreceiver"
+	timestampprocessor "github.com/signalfx/splunk-otel-collector/processor/timestampprocessor"
```
rmfitzpatrick, line 88: nit, unnecessary named import

pkg/processor/timestamp/config.go
```diff
@@ -0,0 +1,54 @@
...
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+const (
+	zeroTs = pcommon.Timestamp(0)
+)
+
+type Config struct {
+	config.ProcessorSettings `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct
+	Offset                   string                   `mapstructure:"offset"`
```
rmfitzpatrick, line 31: Field comments can be useful for doc generation if we want to use https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/configschema

internal/components/components.go
atoulme: OK, this hid a bigger problem. Good call. Fixing.

pkg/processor/timestamp/go.mod
```diff
@@ -0,0 +1,42 @@
+module github.com/signalfx/splunk-otel-collector/processor/timestampprocessor
+
+go 1.18
```
rmfitzpatrick, line 3: the project is on 1.19 so it's probably best to match that. Being in step could be problematic for custom builds depending on the use case though...

pkg/processor/timestamp/config.go
```diff
@@ -0,0 +1,55 @@
...
+// Validate checks if the processor configuration is valid
+func (cfg *Config) Validate() error {
+	_, err := time.ParseDuration(cfg.Offset)
+	if err != nil {
+		return fmt.Errorf("invalid offset format %s: %w", cfg.Offset, err)
+	}
+	return nil
+}
+
+func offsetFn(offset time.Duration) func(pcommon.Timestamp) pcommon.Timestamp {
```
rmfitzpatrick, line 46: maybe `factory.go` is a more logical place for this since it's used there?

pkg/processor/timestamp/metric_processor.go
```diff
@@ -0,0 +1,91 @@
...
+							}
+						}
+					case pmetric.MetricTypeSummary:
+						for l := 0; l < metric.Summary().DataPoints().Len(); l++ {
+							dp := metric.Summary().DataPoints().At(l)
+							dp.SetStartTimestamp(offsetFn(dp.StartTimestamp()))
+							dp.SetTimestamp(offsetFn(dp.Timestamp()))
+						}
+					default:
+						return pmetric.Metrics{}, fmt.Errorf("unsupported metric type: %v", metric.Type())
```
rmfitzpatrick, line 84: Is it better to log and continue/multierr instead of erroring out in total?

pkg/processor/timestamp/config.go
atoulme: I forgot to move it. I will.

pkg/processor/timestamp/go.mod
dmitryax: Actually why do we need this to be added as a module? I don't think we have other components as modules in our distro

atoulme: We have other asks to move to modules. We are preparing for this.

dmitryax: Makes sense

pkg/processor/timestamp/metric_processor.go
atoulme: To me, if this error is thrown, we're unable to process and it's better to call it out.

PR #2089, opened 2022-10-07
Add vale linter github action

atoulme:


.github/workflows/vale.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
+        with:
+          repository: splunk/vale-splunk-style-guide
+          path: vale-splunk-style-guide
+
+      - run: |
+          cat <<EOF > .vale.ini
```
theletterf, line 29: Is this Vale ini in conflict with the StylesPath below?

.github/workflows/vale.yml
```diff
@@ -0,0 +1,45 @@
...
+          StylesPath = vale-splunk-style-guide/styles
+          MinAlertLevel = suggestion
+          
+          [*.{ md,rst }]
+          BasedOnStyles = Splunk
+          
+          Vocab = splunk
```
theletterf, line 36: This one is not needed — we still don't have a vocabulary

theletterf (changes requested):

.github/workflows/vale.yml
atoulme: Sorry, I don't follow. What are you referring to?

.github/workflows/vale.yml
theletterf: Guess it does no harm!

.github/workflows/vale.yml
theletterf: Not sure how the setup works for a pipeline/action, but, essentially, the `.vale.ini` file already contains the path to styles. The path in the file must match whatever path in the local filesystem. Wondering if that's the cause for the pipeline not to work.

atoulme: I can try entering a nonsensical path and see if it gets picked up.

PR #2100, opened 2022-10-11
Run lint for modules

atoulme:
Built on top of #2099


.github/workflows/build-and-test.yml
```diff
@@ -112,7 +112,7 @@ jobs:
       - name: Lint
         run: |
           make install-tools
-          make -j4 checklicense impi lint misspell
+          make -j4 for-all checklicense impi lint misspell
```
rmfitzpatrick, line 115: I think for this to work `CMD` needs to be set but that would require addressing missing targets in the examples or adding a go mod subgroup

```suggestion
          make -j4 for-all CMD="make checklicense impi lint misspell"
```

atoulme:
This PR just got a bit more involved :/ all the module issues I found by running the command properly are resolved.

rmfitzpatrick (commented):
thanks!

rmfitzpatrick (approved):

jvoravong (approved):

PR #2109, opened 2022-10-13
define discovery confmap.Provider config

rmfitzpatrick:
These changes introduce the basic Config form for the upcoming discovery confmap.Provider to allow adding to and modifying the Collector config with individual component yaml files and runtime/observer endpoint obtained properties (not included). This will also provide the basis for a pre-service discovery receiver configured with `<receiver-name>.discovery.yaml` files with per-observer configuration items.


internal/confmapprovider/discovery/config.go
```diff
@@ -0,0 +1,356 @@
...
+			return loadEntry(typeExtension, path, c.Extensions)
+		case isProcessorEntryPath(path):
+			return loadEntry(typeProcessor, path, c.Processors)
+		case isReceiverEntryPath(path):
+			if isReceiverToDiscoverEntryPath(path) {
+				return loadEntry(typeReceiverToDiscover, path, c.ReceiversToDiscover)
+			}
+			return loadEntry(typeReceiver, path, c.Receivers)
+		default:
+			c.logger.Debug(fmt.Sprintf("Disregarding %s", path))
```
atoulme, line 214: Can we use zap fields?

internal/confmapprovider/discovery/config.go
```diff
@@ -0,0 +1,356 @@
...
+func (ReceiverToDiscoverEntry) ErrorF(path string, err error) error {
+	return errorF(typeReceiverToDiscover, path, err)
+}
+
+func (c *Config) Load(discoveryDPath string) error {
+	if c == nil {
+		return fmt.Errorf("config must not be nil to be loaded (use NewConfig())")
+	}
+	return filepath.WalkDir(discoveryDPath, func(path string, d fs.DirEntry, err error) error {
+		c.logger.Debug(fmt.Sprintf("loading component: WalkDir(%s, %#v, %v)", path, d, err))
```
atoulme, line 192: Same if we can use zap fields

internal/confmapprovider/discovery/config.go
```diff
@@ -0,0 +1,356 @@
...
+}
+
+func unmarshalYaml(path string, out any) error {
+	contents, err := os.ReadFile(filepath.Clean(path))
+	if err != nil {
+		return fmt.Errorf("failed reading file %q: %w", path, err)
+	}
+
+	if err = yaml.Unmarshal(contents, out); err != nil {
+		return fmt.Errorf("failed parsing %q as yaml: %w", path, err)
```
atoulme, line 323: Do we stop walking the directory if one file is bad? Can we test for that and make sure we log and continue?

internal/confmapprovider/discovery/config.go
rmfitzpatrick: I'd prefer not to for a single line since the format is cumbersome imo, unless you think these should all be machine readable.

atoulme: in the past I've been told to avoid concatenations when logging and prefer to use late interpolation.

internal/confmapprovider/discovery/config.go
rmfitzpatrick: On this first pass I'm not under the impression it's a good idea to try to use a config that has invalid sections (the collector also tends to error out earlier on invalid component config). `WalkDir()` will end on the first non-skipable error. Updated to clear the config maps on error to be clear in this regard and added an invalid item test, but I'm happy to update to be more intuitive.

internal/confmapprovider/discovery/config.go
rmfitzpatrick: 👍 updated

internal/confmapprovider/discovery/config.go
atoulme: that said this was in java, go might not be all that bad at it

internal/confmapprovider/discovery/config.go
atoulme: The problem I see often in customer envs is that when we fail, we crash, throw an error such "cannot read the yaml", and the customer is left picking up the pieces.
Given that we now have a few of those files in these directories, it might be good to help customers get by.

Note that in the root config case, the service definition contains the logger severity level - meaning they are cues we log, but might not be seen by the customer.

This is out of scope for this PR, but error resilience - bad inputs, horrible configs - we will need to address that to make our lives easier.

atoulme (approved):
I see the addition of support for Windows, LGTM

internal/confmapprovider/discovery/config.go
rmfitzpatrick: I'd hope in the future that we offer something like a full service otto the collector hosts locally that can be used to created a validated config that the collector service uses on a reload. If we fully adopt opamp in product hopefully a local web service would be provided that uses it as well.

atoulme: That's cool. I believe that might be a future epic - make our lives easier when it's debug time.

PR #2185, opened 2022-10-28
Fix example config for Pulsar exporter, units are nanoseconds

lhotari:
- setting `partitions_auto_discovery_interval` to `1` causes a flood of `CommandPartitionedTopicMetadata` requests to brokers. The broker doesn't log these requests with info level, and it's hard to detect the issue. 
  - [The default value in pulsar-client-go is 1 minute](https://github.com/apache/pulsar-client-go/blob/v0.8.1/pulsar/producer_impl.go#L47) which is 60*10^9. Set the example value to the default.

- setting `batching_max_publish_delay` to `10` will make batching ineffective since the unit is nanoseconds. 
  - [The default value in pulsar-client-go is 10 milliseconds](https://github.com/apache/pulsar-client-go/blob/v0.8.1/pulsar/producer_impl.go#L38) which is 10^7. Set the example value to the default.

### DIFF
--- a/internal/exporter/pulsarexporter/testdata/config.yaml
+++ b/internal/exporter/pulsarexporter/testdata/config.yaml
@@ -23,10 +23,12 @@ exporters:
       compression_level: default
       batch_builder_type: 1
       disable_batching: false
-      batching_max_publish_delay: 10
+      # unit is nanoseconds (10^-9), set to 10 milliseconds in nanoseconds
+      batching_max_publish_delay: 10000000
       batching_max_messages: 1000
       batching_max_size: 128000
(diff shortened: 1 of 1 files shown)


eolivelli (approved):
great catch !

zzzming:
That's an excellent finding! Thanks @lhotari 

time.Duration is in nano seconds
https://pkg.go.dev/time#Duration

pmcollins (approved):
For the record, the Pulsar exporter in this repo was developed for a very narrow use case and is not really intended for general use. As a result, we really have no way to test this, but since this change is for an example config, I think it's low risk. Thanks for fixing.

PR #2195, opened 2022-10-31
add initial discovery mode

rmfitzpatrick:
These changes add new `--discovery` and `--dry-run` cmdline options to run a limited preflight collector service using any `.discovery.yaml` receivers and observers from the config directory and embed them into an internal discovery receiver instance whose successful status logs lead to injecting the discovered receivers into a receiver creator for the final config. If `--dry-run` is set, then the final config will be written to stdout before exiting instead of running the service.

Along with this new functionality, a new common internal discovery package is being added to avoid repeated logic in addition to new config server testutils collector container helpers to load the initial and effective configs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### 💡 Enhancements 💡
 
+- Experimental --discovery and --dry-run functionality [#2195](https://github.com/signalfx/splunk-otel-collector/pull/2195)
+
 ## v0.62.0
 
 This Splunk OpenTelemetry Collector release includes changes from the [opentelemetry-collector v0.62.0](https://github.com/open-telemetry/opentelemetry-collector/releases/tag/v0.62.0) and the [opentelemetry-collector-contrib v0.62.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.62.0) releases.
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -59,7 +59,8 @@ func main() {
 
 	confMapConverters := collectorSettings.ConfMapConverters()
 	configServer := configconverter.NewConfigServer()
-	confMapConverters = append(confMapConverters, configServer)
+	dryRun := configconverter.NewDryRun(collectorSettings.IsDryRun())
+	confMapConverters = append(confMapConverters, dryRun, configServer)
 
 	discovery, err := discovery.New()
 	if err != nil {
--- a/internal/configconverter/dry_run.go
+++ b/internal/configconverter/dry_run.go
@@ -1,0 +1,49 @@
+// Copyright Splunk, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configconverter
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"go.opentelemetry.io/collector/confmap"
+	"gopkg.in/yaml.v2"
+)
+
+var _ confmap.Converter = (*DryRun)(nil)
+
+type DryRun struct {
+	enabled bool
+}
+
+func NewDryRun(enabled bool) DryRun {
+	return DryRun{enabled: enabled}
+}
+
+// Convert is intended to be called as the final service confmap.Converter so
+// that it has access to the final config before exiting, if enabled.
+func (dr DryRun) Convert(_ context.Context, conf *confmap.Conf) error {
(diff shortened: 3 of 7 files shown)


internal/receiver/discoveryreceiver/receiver.go
```diff
@@ -166,7 +164,7 @@ func (d *discoveryReceiver) consumerLoop(loopStarted *sync.WaitGroup) {
 func (d *discoveryReceiver) createAndSetReceiverCreator() error {
 	receiverCreatorFactory, receiverCreatorConfig, err := d.config.receiverCreatorFactoryAndConfig(d.endpointTracker.correlations)
 	if err != nil {
-		return nil
+		return err
```
rmfitzpatrick, line 167: ☹️

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -0,0 +1,525 @@
...
+// duration, and tear them down before returning the discovery config.
+func (d *discoverer) discover(cfg *Config) (map[string]any, error) {
+	discoveryReceivers, discoveryObservers, err := d.createDiscoveryReceiversAndObservers(cfg)
+	if err != nil {
+		d.logger.Error("failed preparing discovery components", zap.Error(err))
+		return nil, err
+	}
+
+	for observerID, observer := range discoveryObservers {
+		d.logger.Debug(fmt.Sprintf("starting %s", observerID.String()))
```
atoulme, line 113:
```suggestion
		d.logger.Debug(fmt.Sprintf("starting observer %s", observerID.String()))
```

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -0,0 +1,525 @@
...
+			d.logger.Info(
+				fmt.Sprintf("%s startup failed. Won't proceed with %s-based discovery", observerID.String(), observerID.Type()),
+				zap.Error(e),
+			)
+			return nil, nil
+		}
+	}
+
+	for receiverID, receiver := range discoveryReceivers {
+		d.logger.Debug(fmt.Sprintf("starting %s", receiverID.String()))
```
atoulme, line 124:
```suggestion
		d.logger.Debug(fmt.Sprintf("starting receiver %s", receiverID.String()))
```

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -0,0 +1,525 @@
...
+func (d *discoverer) discover(cfg *Config) (map[string]any, error) {
+	discoveryReceivers, discoveryObservers, err := d.createDiscoveryReceiversAndObservers(cfg)
+	if err != nil {
+		d.logger.Error("failed preparing discovery components", zap.Error(err))
+		return nil, err
+	}
+
+	for observerID, observer := range discoveryObservers {
+		d.logger.Debug(fmt.Sprintf("starting %s", observerID.String()))
+		if e := observer.Start(context.Background(), d); e != nil {
```
atoulme, line 114: Could we take a context with a timeout there?

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -0,0 +1,525 @@
...
+
+	for receiverID, receiver := range discoveryReceivers {
+		d.logger.Debug(fmt.Sprintf("starting %s", receiverID.String()))
+		if err = receiver.Start(context.Background(), d); err != nil {
+			return nil, err
+		}
+	}
+
+	fmt.Fprintf(os.Stderr, "Discovering for next %s...\n", d.duration.String())
+	time.Sleep(d.duration)
```
atoulme, line 131: Can we use a timeout ticker allowing us to exit early in case the collector needs to stop, using a context.Done goroutine?

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -0,0 +1,525 @@
...
+	}
+
+	for receiverID, receiver := range discoveryReceivers {
+		d.logger.Debug(fmt.Sprintf("starting %s", receiverID.String()))
+		if err = receiver.Start(context.Background(), d); err != nil {
+			return nil, err
+		}
+	}
+
+	fmt.Fprintf(os.Stderr, "Discovering for next %s...\n", d.duration.String())
```
atoulme, line 130: Is this ok? Using stderr this way? I want to understand if we have a policy about stderr.

atoulme (approved):

internal/confmapprovider/discovery/discoverer.go
rmfitzpatrick: I'm not aware of a policy but all the non-service loggers write to stderr. The intention is that `--dry-run` output is a usable config so that something like `SPLUNK_CONFIG_YAML=$(otelcol --discovery) otelcol` is valid but this isn't the case atm until there's a final dry run converter stage that would support all config methods.

internal/confmapprovider/discovery/discoverer.go
rmfitzpatrick: good call, updated for all other the component methods

internal/confmapprovider/discovery/discoverer.go
rmfitzpatrick: tbh I'm not sure I'm following the request since I think sleep calls can be interrupted but updated to listen for the background context closing.

atoulme: my java shows. You got the gist of the idea.

PR #2229, opened 2022-11-04
add integration test for envvar config source

ccordi:


tests/general/envvar_expansion_test.go
```diff
@@ -99,3 +108,94 @@ func TestExpandedYamlViaEnvConfigSource(t *testing.T) {
 	expectedResourceMetrics := tc.ResourceMetrics("yaml_from_env.yaml")
 	require.NoError(t, tc.OTLPReceiverSink.AssertAllMetricsReceived(t, *expectedResourceMetrics, 30*time.Second))
 }
+
+func TestCollectorProcessWithEnvVarConfig(t *testing.T) {
```
rmfitzpatrick, line 112: I think this test can be easily updated to use the collector container and new initial and effective config methods I just added: https://github.com/signalfx/splunk-otel-collector/blob/main/tests/testutils/collector_container.go#L253.

The existing env var tests could be used as a basis: https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/envvar_expansion_test.go#L46

PR #2231, opened 2022-11-04
Add docker observer discovery mode integration test

rmfitzpatrick:
These changes also:

1. move discovery observer config validation to initial component loading so that invalid configs don't stay hidden until service time.
2. pin an underlying docker connection dep* to a more working version that uses the default net.Dialer
3. adds a few testcontainers builder methods that were helpful for writing the test including being able to set the container's user ids.
4. prevents repeated integration test runs by just calling the test cmd in `tests` w/ `./...`


internal/confmapprovider/discovery/discoverer.go
```diff
@@ -236,10 +238,19 @@ func (d *discoverer) createObserver(observerID config.ComponentID, cfg *Config)
...
 	}
-	if err = observerCfgMap.Unmarshal(observerConfig); err != nil {
+
+	if err = config.UnmarshalExtension(observerCfgMap, observerConfig); err != nil {
 		return nil, fmt.Errorf("failed unmarshaling %s config: %w", observerID.String(), err)
 	}
 
+	if ce := d.logger.Check(zap.DebugLevel, "unmarshalled observer config"); ce != nil {
+		if c, e := yaml.Marshal(observerConfig); e != nil {
+			d.logger.Debug("failed unmarshalling observer config for logging", zap.Error(e))
```
atoulme, line 248: why not use `ce` here? also it failed marshaling to yaml, not unmarshaling, correct?

rmfitzpatrick: The `ce` message is already `"unmarshalled observer config"` and in this failure path it's disregarded since the content isn't renderable. Corrected the wording, thanks

atoulme (approved):

PR #2187, opened 2022-10-28
Add windowsperfcounters receiver from contrib

chrislbs:

PR #2246, opened 2022-11-08
Add logstransformprocessor from contrib

chrislbs:
We are leveraging the splunk hec receiver, but it does not support the operators that are present in the logs transform processor. I was considering updating the splunk hec receiver in contrib to add the operator functionality as many core receivers do, but I figured I'd see if we could get this in to this repo as an additional measure.

We're needing to manipulate data in our log events that are not able to be handled by the existing transform processor.

This PR is built on top of https://github.com/signalfx/splunk-otel-collector/pull/2187 .

I ended up using the auto-reformatter available in GoLand to update the tables in the components.md file.


aunshc:
Thanks @chrislbs for the context and adding the relevant components to the distribution with this PR. I'm on the Product Management team at Splunk and would love to understand your use cases for logs collection and manipulation before this is merged and supported moving forward.

Can you provide more context about your organization and objectives you are looking to achieve with these components? Happy to chat over a call, scheduled with the help of your organization's Splunk account team as well. Thanks!

atoulme:
I have rebased and fixed the test failure - let's see if CI agrees.

atoulme (approved):

PR #2249, opened 2022-11-10
Add splunk_skip_repo option to ansible

savec80:
Hello team,
Can you please review PR for skipping use of Splunk repo for installing otel package.
We would like to utilize your role, though we can't rely on external repo and would like to keep rpm locally.
Since we are using RH satellite, we can't just provide splunk_repo_base_url var as url is different.
As a reference there is a parity of this in Smart Agent role: https://github.com/signalfx/signalfx-agent/issues/1628
Thanks.


atoulme:
Looks good but need more docs on the boolean flag - just like what smart agent provides. We can help there.

rmfitzpatrick (approved):

savec80:
@atoulme that's fine, thanks for your help.

PR #2250, opened 2022-11-11
java zc: don't generate service name by default

pmcollins:
This change adds support to the Java Linux auto instrumentation configuration file for a new, optional parameter called `generate_svc_name`. If left unset, it defaults to false, which means that the .so will not generate a service name and add it to the environment of the soon-to-be running Java process. This differs from prior behavior where the .so did generate the service name, but generating the service name in the .so is no longer optimal because the Java instrumentation library itself will generate the service name if it hasn't been set.


atoulme (approved):
LGTM

PR #2271, opened 2022-11-17
Change ConfigSource interface to be more like confmap.Provider

bogdandrutu:
Main differences still:
* No Scheme() func and no scheme in the "selector";
* Extra params passed in Retrieve;


internal/configprovider/manager_test.go
```diff
@@ -197,15 +196,6 @@ map:
   k1: v1
 `},
 				"invalid_yaml_str": {Value: ":"},
-				"valid_yaml_byte_slice": {Value: []byte(`
```
rmfitzpatrick, line 200: Why is this being removed?

bogdandrutu: We dropped the support for the configsources to retrieve bytes, only strings are supported. That is fine since the bytes were converted to strings anyway in the `confmap.Conf` while unmarshaling.

rmfitzpatrick (approved):

dmitryax (approved):

PR #2284, opened 2022-11-21
skip puppet focal for github.com/puppetlabs/puppetlabs-release/issues/271

rmfitzpatrick:
Temporarily skipping these tests to unblock builds since we aren't pushing puppet changes in immediate future.


atoulme (approved):
Approved. This will fix the build for now.

PR #2309, opened 2022-12-01
Make it possible to use a different docker repository when building the image

atoulme:


rmfitzpatrick (approved):

cmd/otelcol/Dockerfile
```diff
@@ -1,7 +1,8 @@
-FROM alpine:3.17.0 as certs
+ARG DOCKER_REPO
```
rmfitzpatrick, line 1:
```suggestion
ARG DOCKER_REPO=docker.io
```
otherwise I think the default is invalid. Or may be better to require the registry to end w/ `/` so the default is from the system.

PR #2329, opened 2022-12-07
Add kubeutils for integration tests

rmfitzpatrick:
These changes update the recent kind cluster helper to provide more expanded resource helpers capable of generating manifests for config maps, daemonsets, and cluster role (bindings).


tests/testutils/endpoint.go
```diff
@@ -36,6 +36,7 @@ type portpair struct {
 // provided that there is no race by some other code to grab the same port
 // immediately.
 func getAvailableLocalAddress(t testing.TB) string {
+	// ln, err := net.Listen("tcp", "0.0.0.0:0")
```
atoulme, line 39: Is this comment meaningful?

tests/testutils/kubeutils/kind_cluster.go
```diff
@@ -0,0 +1,180 @@
...
+	Testcase     *testutils.Testcase
+	Clientset    *kubernetes.Clientset
+	ExposedPorts map[uint16]uint16
+	Name         string
+	Kubeconfig   string
+	Config       string
+}
+
+func NewKindCluster(t *testutils.Testcase) *KindCluster {
+	t.TB.TempDir()
```
atoulme, line 69: I think this is not necessary, please advise?

tests/testutils/endpoint.go
rmfitzpatrick: cruft, removed.

tests/testutils/kubeutils/kind_cluster.go
rmfitzpatrick: Was testing this helper but opted not to require deletion on cleanup.

atoulme (approved):

rmfitzpatrick:
I think rolling back the ubuntu runners invalidated the cached deps so the lint initialization issue should be transient. Will address in a separate PR if not the case.

PR #2331, opened 2022-12-08
Use a fixed version of Ubuntu to run tests

atoulme:


rmfitzpatrick (approved):

rmfitzpatrick (changes requested):
Please ensure no agent dep regressions and that `make for-all CMD='make tidy'` passes.

rmfitzpatrick (approved):

PR #2339, opened 2022-12-09
Remove handling of unsupported --mem-ballast-size-mib command line argument

dmitryax:
The argument was removed from the core collector in v0.36.0. We kept supporting it in our distro but the support is broken. We never remove that argument as we do for others and it's being passed to the core collector in https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/main.go#L105 where it fails with the following error:

```
2022/12/08 16:25:34 main.go:106: application run finished with error: unknown flag: --mem-ballast-size-mib
```

So I don't think it makes sense to bring back support of it.

### DIFF
--- a/cmd/otelcol/config/collector/upstream_agent_config.yaml
+++ b/cmd/otelcol/config/collector/upstream_agent_config.yaml
@@ -25,8 +25,7 @@ extensions:
   zpages:
     #endpoint: 0.0.0.0:55679
   memory_ballast:
-    # In general, the ballast should be set to 1/3 of the collector's memory, the limit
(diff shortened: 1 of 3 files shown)


rmfitzpatrick:
Considering this hasn't worked for a bit it makes sense to remove as you have, but it could also be restored but no longer passed to the service w/ `removeFlag()`, which would be a simpler change (fwiw).

atoulme (approved):

dmitryax:
I don't think there is a reason to keep and maintain an additional configuration option which is likely not being used given that there were no complaints since 0.36.0

Issue #2213, opened 2022-11-02
Outdated Kubernetes instructions


juandspy:
### Summary

I tried deploying the Kubernetes version of the collector ([this one](https://github.com/signalfx/splunk-otel-collector/blob/8578955e71e77decc56720d0e1b71d699b82948b/examples/kubernetes-yaml/splunk-otel-collector-gateway.yaml)) and it gives me the following error:
```
2022/11/02 11:31:48 settings.go:269: Set config to [/conf/otel-collector-config.yaml]
2022/11/02 11:31:48 settings.go:333: Set ballast to 168 MiB
2022/11/02 11:31:48 settings.go:349: Set memory limit to 460 MiB
Error: unknown flag: --mem-ballast-size-mib
2022/11/02 11:31:48 main.go:99: application run finished with error: unknown flag: --mem-ballast-size-mib
```

What do we have to do with [that flag](https://github.com/signalfx/splunk-otel-collector/blob/8578955e71e77decc56720d0e1b71d699b82948b/examples/kubernetes-yaml/splunk-otel-collector-gateway.yaml#L205)?

Related PRs: https://github.com/signalfx/splunk-otel-collector/pull/567

### How to reproduce

```
❯ docker run --rm -e SPLUNK_REALM=us1 -e SPLUNK_ACCESS_TOKEN=test quay.io/signalfx/splunk-otel-collector:0.62.0 /otelcol --mem-ballast-size-mib=168
2022/11/02 11:33:20 settings.go:286: Set config to /etc/otel/collector/gateway_config.yaml
2022/11/02 11:33:20 settings.go:333: Set ballast to 168 MiB
2022/11/02 11:33:20 settings.go:349: Set memory limit to 460 MiB
Error: unknown flag: --mem-ballast-size-mib
2022/11/02 11:33:20 main.go:99: application run finished with error: unknown flag: --mem-ballast-size-mib
```

atoulme:
Thank you for your report. We have since created a complete helm chart you might prefer to use to deploy the collector. Please see https://github.com/signalfx/splunk-otel-collector-chart.

We will look at updating or deleting this example.

juandspy:
@atoulme thanks. For me it was easier to use this template with a config map, service and deployment. It worked like a charm removing the `--mem-ballast-size-mib` flag.

Issue #660, opened 2021-08-24
Using  --version switch does not print version


dgcom:
Running the following command should print version information:
```
$ ./otelcol_linux_amd64 --version
2021/08/24 13:37:44 main.go:228: Unable to find the default configuration file (cmd/otelcol/config/collector/gateway_config.yaml) or (/etc/otel/collector/gateway_config.yaml)
```
Compare to properly working upstream:
```
$ ./otelcontribcol_linux_amd64 --version
otelcontribcol version v0.33.0
```

PR #2343, opened 2022-12-11
Rework command line arguments parsing

dmitryax:
This PR solves two problems:

1. In the current implementation, command flags specific to the Splunk distro are dropped from the arguments list passed to the collector core service by their names. It means that all the flag values are unintentionally passed downstream. Also, flags with `=` are not removed from the list and cause failures downstream. For example, `./bin/otelcol --config-dir=/tmp/conf ...` fails with:

```
2022/12/10 19:15:52 main.go:106: application run finished with error: unknown flag: --config-dir
```

And `removeFlag` function is called from methods that are not expected to be mutating by their definition like `ResolverURIs`.

2. Currently, we rely on the downstream collector core service to display `--help` flag output and get the following:

```
Usage:
  otelcol [flags]

Flags:
      --config --config=file:/path/to/first --config=file:path/to/second   Locations to the config file(s), note that only a single location can be set per flag entry e.g. --config=file:/path/to/first --config=file:path/to/second. (default [])
      --feature-gates Flag                                                 Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature.
  -h, --help                                                               help for otelcol
      --set func                                                           Set arbitrary component config property. The component has to be defined in the config file and the flag has a higher precedence. Array config properties are overridden and maps are joined. Example --set=processors.batch.timeout=2s
  -v, --version                                                            version for otelcol
```

It's not just incomplete but also provides a misleading description for `--config` flag: we don't support config schemas, so using any of the provided examples will fail.  This change defines all the flags in our distro and makes the program use that definition in `--help` output instead of relying on the collector core service. Experimental flags like `--configd`, `--config-dir` and `--discovery` are hidden for now, but can be easily enabled. It provides the following output for `--help` flag:

```
Usage of otelcol:
      --config string          Locations to the config file(s), note that only a single location can be set per flag entry e.g. --config=/path/to/first --config=path/to/second. (default "[]")
      --dry-run                Don't run the service, just show the configuration
      --feature-gates string   Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature. (default "[]")
      --no-convert-config      Do not translate old configurations to the new format automatically. By default, old configurations are translated to the new format for backward compatibility.
      --set string             Set arbitrary component config property. The component has to be defined in the config file and the flag has a higher precedence. Array config properties are overridden and maps are joined. Example --set=processors.batch.timeout=2s (default "[]")
  -v, --version                Version of the collector.
```

Now we also explicitly pass all the flags that are handled by the collector core service instead of passing what we don't use. The reason for that is that there are no plans in the collector core to add any new flags. Most of the old flags were migrated to the configuration. And, even if something will be added, it potentially can be incompatible with our distro and introduce some broken flags to the `--help` output without us noticing.

And, as an opportunistic simplification, this change replaces an unnecessary `Settings` interface with a struct.

### DIFF
--- a/internal/settings/settings.go
+++ b/internal/settings/settings.go
@@ -16,7 +16,6 @@ package settings
 
 import (
 	"fmt"
-	"io"
 	"log"
 	"os"
 	"strconv"
@@ -30,8 +29,6 @@ import (
 )
 
 const (
-	DefaultUndeclaredFlag = -1
-
 	APIURLEnvVar              = "SPLUNK_API_URL"
 	BallastEnvVar             = "SPLUNK_BALLAST_SIZE_MIB"
 	ConfigEnvVar              = "SPLUNK_CONFIG"
@@ -60,47 +57,11 @@ const (
 	ConfigDScheme       = "splunk.configd"
 )
 
-type Settings interface {
-	// ResolverURIs returns the collector config provider resolver uris for the collector service
-	ResolverURIs() []string
-	// ConfMapConverters returns the collector config provider resolver confmap.Converters for the collector service
-	ConfMapConverters() []confmap.Converter
-	// ServiceArgs are the sanitized, adjusted args to be used in updating os.Args[1:] for the collector service
-	ServiceArgs() []string
-	// IsDryRun returns whether --dry-run mode was requested
-	IsDryRun() bool
-}
-
-func New(args []string) (Settings, error) {
-	f, err := newFlags(args)
-	if err != nil {
-		return nil, err
-	}
-
-	// immediate exit paths, no further setup required
-	if f.helpFlag || f.versionFlag {
-		return f, nil
-	}
-
-	if err = checkRuntimeParams(f); err != nil {
-		return nil, err
-	}
-
-	if err = setDefaultEnvVars(); err != nil {
-		return nil, err
-	}
-
-	return f, nil
-}
-
-var _ Settings = (*flags)(nil)
-
-type flags struct {
+type Settings struct {
 	configPaths     *stringArrayFlagValue
 	setProperties   *stringArrayFlagValue
 	configDir       *stringPointerFlagValue
-	serviceArgs     []string
-	helpFlag        bool
+	colCoreArgs     []string
 	versionFlag     bool
 	noConvertConfig bool
 	configD         bool
@@ -108,29 +69,46 @@ type flags struct {
 	dryRun          bool
 }
 
-func (f *flags) ResolverURIs() []string {
+func New(args []string) (*Settings, error) {
+	s, err := parseArgs(args)
+	if err != nil {
+		return nil, err
+	}
+
+	// immediate exit paths, no further setup required
+	if s.versionFlag {
+		return s, nil
+	}
+
+	if err = checkRuntimeParams(s); err != nil {
+		return nil, err
+	}
+
+	if err = setDefaultEnvVars(); err != nil {
+		return nil, err
+	}
+
+	return s, nil
+}
+
+// ResolverURIs returns config provider resolver URIs for the core collector service.
+func (s *Settings) ResolverURIs() []string {
 	var configPaths []string
-	if configPaths = f.configPaths.value; len(configPaths) == 0 {
+	if configPaths = s.configPaths.value; len(configPaths) == 0 {
 		if configEnvVal := os.Getenv(ConfigEnvVar); len(configEnvVal) != 0 {
 			configPaths = []string{"file:" + configEnvVal}
 		}
 	}
 
-	configDir := getConfigDir(f)
+	configDir := getConfigDir(s)
 
-	if f.dryRun {
-		removeFlag(&f.serviceArgs, "--dry-run")
-	}
-
-	if f.configD {
-		removeFlag(&f.serviceArgs, "--configd")
+	if s.configD {
 		configPaths = append(configPaths, fmt.Sprintf("%s:%s", ConfigDScheme, configDir))
 	}
 
-	if f.discoveryMode {
-		removeFlag(&f.serviceArgs, "--discovery")
+	if s.discoveryMode {
 		// discovery uri must come last to successfully merge w/ other config content
-		configPaths = append(configPaths, fmt.Sprintf("%s:%s", DiscoveryModeScheme, f.configDir))
+		configPaths = append(configPaths, fmt.Sprintf("%s:%s", DiscoveryModeScheme, s.configDir))
 	}
 
 	configYaml := os.Getenv(ConfigYamlEnvVar)
@@ -145,32 +123,27 @@ func (f *flags) ResolverURIs() []string {
 	}
 }
 
-func getConfigDir(f *flags) string {
+func getConfigDir(f *Settings) string {
 	configDir := DefaultConfigDir
 	if envConfigDir, ok := os.LookupEnv(ConfigDirEnvVar); ok {
 		configDir = envConfigDir
 	}
 
 	if f.configDir.value != nil {
-		removeFlag(&f.serviceArgs, "--config-dir")
 		configDir = f.configDir.String()
 	}
 
 	return configDir
 }
 
-func (f *flags) ConfMapConverters() []confmap.Converter {
+// ConfMapConverters returns confmap.Converters for the collector core service.
+func (s *Settings) ConfMapConverters() []confmap.Converter {
 	confMapConverters := []confmap.Converter{
 		// nolint: staticcheck
-		overwritepropertiesconverter.New(f.setProperties.value), // support until there's an actual replacement
+		overwritepropertiesconverter.New(s.setProperties.value), // support until there's an actual replacement
 	}
 
-	if f.noConvertConfig {
-		// the collector complains about this flag if we don't remove it. Unfortunately,
-		// this must be done manually since the flag library has no functionality to remove
-		// args
-		removeFlag(&f.serviceArgs, "--no-convert-config")
-	} else {
+	if !s.noConvertConfig {
 		confMapConverters = append(
 			confMapConverters,
 			configconverter.RemoveBallastKey{},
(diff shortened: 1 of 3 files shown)


internal/settings/settings.go
```diff
@@ -60,77 +57,58 @@ const (
 	ConfigDScheme       = "splunk.configd"
 )
 
-type Settings interface {
```
rmfitzpatrick, line 63: > And, as an opportunistic simplification, this change replaces an unnecessary Settings interface with a struct.

The aim of this interface was to help clarify the intended consumer usage or available settings while hiding all the underlying flag/state management that was scattered throughout `main.go`. What are we gaining by using the exported internal struct directly?

dmitryax: Any IDE or editor can show all available methods and point you right to the implementation, which is much better than pointing to an interface declaration that force you to look around for the actual implementation after that. Also, all the methods will be listed the same way as they are in the go doc https://pkg.go.dev/github.com/signalfx/splunk-otel-collector@v0.66.0/internal/settings#Settings. So would disagree that using interfaces for one struct makes developer live any better.

internal/settings/settings.go
```diff
@@ -182,64 +155,83 @@ func (f *flags) ConfMapConverters() []confmap.Converter {
...
-	flagSet := flag.FlagSet{}
-	// we don't want to be responsible for tracking all supported collector service
-	// flags, so allow any we don't use and defer parsing to the actual service
-	flagSet.ParseErrorsWhitelist.UnknownFlags = true
-
-	var cpArgs []string
-	cpArgs = append(cpArgs, args...)
+// parseArgs returns new Settings instance from command line arguments.
+func parseArgs(args []string) (*Settings, error) {
+	flagSet := flag.NewFlagSet("otelcol", flag.ContinueOnError)
```
rmfitzpatrick, line 170: As commented the aim here is to not have to track or manage upstream options in our init logic unless we are using them directly. Why is this no longer desired?

rmfitzpatrick: >The reason for that is that there are no plans in the collector core to add any new flags.

Do you have an issue for this?

>it potentially can be incompatible with our distro and introduce some broken flags to the --help output without us noticing.

Do you mind clarifying why this expectation is something to avoid and why it's better to not adopt features as upstream provides them?

dmitryax: > Do you have an issue for this?

No

> Do you mind clarifying why this expectation is something to avoid and why it's better to not inherit* features as upstream provides them?

Because additional flags introduce some optional functionality. And that functionality potentially can be incompatible with our distro, which will lead to an additional broken flag in `--help` output without us noticing. If that flag, for whatever reason, is required, we will catch it during the upgrade of the core dependency

internal/settings/settings.go
dmitryax: > It's more for maintainability 

IMHO adding unnecessary interfaces reduces maintainability

> establish a clear convention for how future settings should be added

I think it's rather an unnecessarily complicated convention for such a simple interface. Not sure if there is even a reason to expose them as methods... If they are all just getters using data from the struct that was set during initialization, we can even expose them as precalculated struct fields. That will put them all together in one place as it was with the interface. And it'll help to avoid extra work if they would be used more than once in the future. What do you think?

internal/settings/settings.go
dmitryax: If we still want to define all known flags (including core's) in our distro but pass everything that we don't know to the core just to potentially catch a new flag added to the core without documenting it, we can do that. But it will require manual argument parsing. I don't think it is worth it. I'd rather keep an eye on new flags added to the core (which is very unlikely to happen any time soon) and add them properly along with documentation in our distro, the same way as we do for components.

internal/settings/settings.go
rmfitzpatrick: Not to get too caught up in a dispute of personal preference, but they were interfaces because interfaces are collections of methods and not fields. The accessors do the computing and return the setting value, but could definitely be improved.  They were exposed via the settings interface to encapsulate the ~legacy flag helpers while migrating to a central setting store to prevent breaking changes.

Processing them w/ initial parsing stages in `New()` and using exported fields is acceptable to me and would address their idempotence issues.

internal/settings/settings.go
rmfitzpatrick: I'm still under the impression we probably don't want to define any flags we aren't using and the current problems we are running into are longstanding implementation defects. Are we not already manually parsing args atm and in your changes? I can see an inverse `flagSetToArgs()` that removes our distro-only flags and values as a major improvement to the current `removeFlag()`.

internal/settings/settings.go
dmitryax: > Processing them w/ initial parsing stages in New() and using exported fields is acceptable to me and would address their idempotence issues.

Sounds good, I was thinking about doing that but didn't want to go too far. I can do it in this or separate PR

internal/settings/settings.go
dmitryax: > I'm still under the impression we probably don't want to define any flags we aren't using and the current problems we are running into are longstanding implementation defects. 

Speaking about `--help` documentation, we either define flags it in our distro or use the core's definition. We cannot do both. Given that we have a lot of additional flags in our distro and core `--config` flag is different, I don't think we should use core. So I believe we need to define it in our distro. If we are doing so, we should do it for all the flags. Why would we skip supported flags that are implemented in a different place like `--version` and `--feature-gates`?

> Are we not already manually parsing args atm and in your changes? I can see an inverse flagSetToArgs() that removes our distro-only flags and values as a major improvement to the current removeFlag().

We are not, `flagSetToArgs` is a reverse operation. It just generated specific arguments from parsed flags. All the parsing is done by the `pflag` library

rmfitzpatrick: > We are not, flagSetToArgs is a reverse operation. It just generated specific arguments from parsed flags. All the parsing is done by the pflag library

We are using it to determine filter the args to pass to the otelcol service, which I would call manual parsing regardless of the library doing it. My suggestion (unvetted) is to use what you're doing but inverting it to remove our distro specific args (maybe w/ a shlex-like dep?) to generate the collector service args.

dmitryax: Can you elaborate on reasoning behind this suggestion? Why do we want to pass all unknown arguments?

rmfitzpatrick: I think on the whole I'm under the impression we should thinly wrap the actual otelcol service/cobra command using as reasonably minimal flag logic as possible. Unfortunately this is kind of opposed to how we are currently doing things and and have done them in the past, for various reasons. These gh issues suggest to me we've previously been overzealous w/ flag parsing and not great supporting otelcol service args or expected startup logic: https://github.com/signalfx/splunk-otel-collector/issues/97, https://github.com/signalfx/splunk-otel-collector/issues/660, and for https://github.com/signalfx/splunk-otel-collector/issues/2213 your recent PR gives another example of how tracking upstream flags can become easily stale: https://github.com/signalfx/splunk-otel-collector/pull/2339.

rmfitzpatrick: As another anecdote, I think the removal of the overwrite properties converter and the new "set" cobra function will require us to either drop our "set" flag option or port the converter to our distro before stripping the "set" options. I can see advantages with either approach (no loss in existing functionality but more maintenance requirements vs. potential breaking changes but more reliance on upstream functionality).

dmitryax: We cannot tinily wrap  the actual otelcol service/cobra command until we at least support the same set of flags which is not the case for `--config` and make sure that out custom flags are not needed to be displayed. I'm all in with that approach but not sure we want to keep broken `--help` output until that.

As for `--set` I believe we should mark it as deprecated and ask users to use `--config yaml:...` instead. But for that, we need support of config schemas first.

rmfitzpatrick: why would `--set` be deprecated if it's now in the core flags? https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/flags.go#L51

What are you referring to by not supporting "config schemas"? We do have config providers that are selected by the config schemes if that's what you mean https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/main.go#L76 and we should fix our incorrect config flag logic if it doesn't reflect what is supported by core.

dmitryax: > why would --set be deprecated if it's now in the core flags? https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/flags.go#L51

To avoid having workarounds. It's still unclear if we will keep `--set` in the core. But we can leave it's support in out distro also, we just need to copy the code from the core for now.

> What are you referring to by not supporting "config schemas"? We do have config providers that are selected by the config schemes if that's what you mean https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/main.go#L76 and we should fix our incorrect config flag logic if it doesn't reflect what is supported by core.

Yes, I mean config schemas in `--config` flag. Currently, our distro shows the following help output for that flag:

```
      --config --config=file:/path/to/first --config=file:path/to/second   Locations to the config file(s), note that only a single location can be set per flag entry e.g. --config=file:/path/to/first --config=file:path/to/second. (default [])
```

But if I run ` splunk-otel-collector git:(main) ./bin/otelcol --config=file:./conf.yaml`, I get 
```
2022/12/14 11:16:11 main.go:47: invalid settings detected: unable to find the configuration file file:./conf.yaml
```

Do you want to fix `--config` flag sooner than merging this PR? Also, we will never be able to show any help for our own flag, do we want that?

rmfitzpatrick: Yes, I think not treating scheme-prefixed config values as valid for their latest use in the resolver uris is a bug and should be addressed. We can access the otelcol command flags so would be able to provide their output in help when we're ready to advertise them (assuming there isn't some indirection/internalization planned in the future):

```diff
 func runInteractive(settings service.CollectorSettings) error {
        cmd := service.NewCommand(settings)
+       cmd.Flags().Bool("some-made-up-flag", true, "this is in the collector service help content.")
        if err := cmd.Execute(); err != nil {
                return fmt.Errorf("application run finished with error: %w", err)
        }
```

```
bin/otelcol -h
Usage:
  otelcol [flags]

Flags:
      --config --config=file:/path/to/first --config=file:path/to/second   Locations to the config file(s), note that only a single location can be set per flag entry e.g. --config=file:/path/to/first --config=file:path/to/second. (default [])
      --feature-gates Flag                                                 Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature.
  -h, --help                                                               help for otelcol
      --set func                                                           Set arbitrary component config property. The component has to be defined in the config file and the flag has a higher precedence. Array config properties are overridden and maps are joined. Example --set=processors.batch.timeout=2s
      --some-made-up-flag                                                  this is in the collector service help content. (default true)
  -v, --version                                                            version for otelcol
```

dmitryax: We need the flags in advance, accessing `cmd.Flags()` will only add the `--help` output, so we need to duplicate all the definitions before that. It's a hacky but good workaround which we can add later on after we fix `--config` flags. We can still merge this one and make those changes as the following steps:
1. Fix out --config flag
2. Migrate to this proposed workaround
WDYT?

rmfitzpatrick: I think it'd be best to have a planning issue where we plan to separate pre and in-service flags, auditing why we do what we do in general (some of which likely has an expired rationale), and investigate whether its necessary (or wise) to wrap any collector flag.

Would you be ok with this pr fixing the config scheme check, migrating the parsing logic to the struct ~constructor so we can use its exposed fields, and not adding any new flags?

dmitryax: Not sure what to plan here. We have a few issues with parsing, I'm suggesting fixing them in 3 steps:
1. Merge this PR to fix the`--config` doc, add missing docs, and fix parsing arguments with "=".
2. Add support to schemas in `--config`
3. Update flags parsing to use `--help` output from core

I still disagree that we should pass all unknown flags, we can discuss that separately and apply another PR if needed.

dmitryax: And want the passing all unknown flags to be a separate PR to assess how much complexity it adds

dmitryax: Actually, if we don't pass unknown attributes in step 3, we can end up with some "unsupported" arguments that potentially can be added to docs in the core but removed on our side. So I can do that along with the PR for step 3, and we can discuss there if worth doing or not

rmfitzpatrick: If we are fronting the collector service properly it will support its own args as it does for the dep version we use. Our distro shouldn't take responsibility of it imo.

rmfitzpatrick: maybe it makes sense to fix the config parsing and other issues alone since we aren't in alignment on how to move forward otherwise.

dmitryax: Fixing each issue alone requires adding more hacks on top of what we have. I tried to come up with one set of related changes that doesn't require doing that.

atoulme (approved):
Approved - will use this discussion for tech debt around this module.

PR #2360, opened 2022-12-13
Add explicit vault creds-helper calls

atoulme:


.gitlab-ci.yml
```diff
@@ -212,6 +212,9 @@ build-linux-image:
       - ARCH: [amd64, arm64, ppc64le]
   retry: 2
   script:
+    - |
+      DOCKER_PATH="artifactory:v2/cloud/role/docker-reader-role"
```
rmfitzpatrick, line 216: Would it be better to define an anchor once and use aliases where needed?

```yaml
.docker-reader-role: &docker-reader-role
    'creds-helper init && eval $(creds-helper docker --eval "artifactory:v2/cloud/role/docker-reader-role")'

script:
  - *docker-reader-role
  -  make docker-otelcol ...
```

atoulme: It would be better! I didn't think of that. I will add to the PR.

atoulme: Trying this out on a separate gitlab build

.gitlab-ci.yml
```diff
@@ -212,6 +216,7 @@ build-linux-image:
       - ARCH: [amd64, arm64, ppc64le]
   retry: 2
   script:
+    - *docker-reader-role
```
dmitryax, line 219: Is `DOCKER_PATH` env needed anywhere downstream? 

can it be just ?

```
- creds-helper init && eval $(creds-helper docker --eval "artifactory:v2/cloud/role/docker-reader-role")

atoulme: right, I simplified to use that instead.

PR #2367, opened 2022-12-15
Remove github api calls, use github's latest link

atoulme:
We are being throttled by Github because we call in various places its API from public workers. Instead, use the `latest` release link.

### DIFF
--- a/instrumentation/packaging/fpm/common.sh
+++ b/instrumentation/packaging/fpm/common.sh
@@ -29,7 +29,7 @@ JAVA_AGENT_INSTALL_PATH="/usr/lib/splunk-instrumentation/splunk-otel-javaagent.j
 CONFIG_INSTALL_PATH="/usr/lib/splunk-instrumentation/instrumentation.conf"
 
 JAVA_AGENT_RELEASE_PATH="${FPM_DIR}/../java-agent-release.txt"
-JAVA_AGENT_RELEASE_URL="https://github.com/signalfx/splunk-otel-java/releases/download/"
+JAVA_AGENT_RELEASE_URL="https://github.com/signalfx/splunk-otel-java/releases/"
 
 POSTINSTALL_PATH="$FPM_DIR/postinstall.sh"
 PREUNINSTALL_PATH="$FPM_DIR/preuninstall.sh"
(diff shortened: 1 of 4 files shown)


instrumentation/packaging/fpm/common.sh
```diff
@@ -57,20 +57,7 @@ download_java_agent() {
...
-        fi
-    fi
-
-    api_url="$JAVA_AGENT_RELEASE_URL/tags/$tag"
-    dl_url="$( curl -sL "$api_url" | jq -r '.assets[] .browser_download_url' | grep "splunk-otel-javaagent.jar" | grep -v ".asc$" )"
-    if [ -z "$dl_url" ]; then
-        echo "Failed to get the agent download url from $api_url" >&2
-        exit 1
-    fi
+    dl_url="$JAVA_AGENT_RELEASE_URL/$tag/splunk-otel-javaagent.jar"
```
rmfitzpatrick, line 60: This may prevent `latest` tags from working. Do you know if that's a problem for our pipeline and maybe we should keep the tag fetching api call?

```
curl https://github.com/signalfx/splunk-otel-java/releases/download/latest/splunk-otel-javaagent.jar
Not Found%
```

atoulme: No, latest with github resolves to the latest release. See https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases 
```
To link directly to a download of your latest release asset that was manually uploaded, link to /owner/name/releases/latest/download/asset-name.zip.
```

rmfitzpatrick: So the `releases/download/$tag/<artifact>` release url should be updated to `releases/latest/download/<artifact>` if the tag is "latest"?

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -392,12 +392,11 @@ $tempdir = create_temp_dir -tempdir $tempdir
 
 if ($with_dotnet_instrumentation) {
     echo "Installing SignalFx Instrumentation for .NET ..."
-    $api = "https://api.github.com/repos/signalfx/signalfx-dotnet-tracing/releases/latest"
     $module_name = "Splunk.SignalFx.DotNet.psm1"
+    $download = "https://github.com/signalfx/signalfx-dotnet-tracing/releases/download/latest/Splunk.SignalFx.DotNet.psm1"
```
rmfitzpatrick, line 396:
```suggestion
    $download = "https://github.com/signalfx/signalfx-dotnet-tracing/releases/latest/download/Splunk.SignalFx.DotNet.psm1"
```

instrumentation/packaging/fpm/common.sh
atoulme: Hmm, I'll review. I thought the build would pick this up.

rmfitzpatrick (approved):
🥳

PR #2380, opened 2022-12-16
Move OTLP and expose hec to send logs to splunk HEC logs endpoint

atoulme:
This is a breaking change of the default gateway configuration to allow to use the log gateway, specifically for profiling logs.

### DIFF
--- a/cmd/otelcol/config/collector/gateway_config.yaml
+++ b/cmd/otelcol/config/collector/gateway_config.yaml
@@ -57,6 +57,8 @@ receivers:
     # Whether to preserve incoming access token and use instead of exporter token
     # default = false
     #access_token_passthrough: true
+  splunk_hec/logs:
+    endpoint: 0.0.0.0:8088
   zipkin:
     endpoint: 0.0.0.0:9411
 
@@ -107,6 +109,9 @@ exporters:
     access_token: "${SPLUNK_ACCESS_TOKEN}"
     realm: "${SPLUNK_REALM}"
     sync_host_metadata: true
+  splunk_hec/logs:
+    token: "${SPLUNK_ACCESS_TOKEN}"
+    endpoint: "https://ingest.$SPLUNK_REALM.signalfx.com/v1/log"
   # Debug
   #logging:
     #loglevel: debug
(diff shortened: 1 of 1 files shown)


rmfitzpatrick (approved):

dmitryax:
Should we remove [these lines](https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/agent_config.yaml#L184-L185) then?

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -130,6 +135,10 @@ service:
       processors: [memory_limiter, batch, resourcedetection/internal]
       exporters: [signalfx/internal]
     logs:
-      receivers: [otlp, signalfx]
+      receivers: [signalfx]
       processors: [memory_limiter, batch]
       exporters: [signalfx]
+    logs/hec:
```
dmitryax, line 141: This can be just `logs` if we rename the one above

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -130,6 +135,10 @@ service:
...
+      receivers: [otlp, splunk_hec/logs]
+      processors: [ memory_limiter, batch ]
+      exporters: [ splunk_hec/logs ]
```
dmitryax, line 144: Just for consistency

```suggestion
      processors: [memory_limiter, batch]
      exporters: [splunk_hec/logs]
```

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -57,6 +57,8 @@ receivers:
     # Whether to preserve incoming access token and use instead of exporter token
     # default = false
     #access_token_passthrough: true
+  splunk_hec/logs:
+    endpoint: 0.0.0.0:8088
```
dmitryax, line 61: Do we need to expose it if we recommend using otlp between the agent and gateway?

atoulme: We might need to expose it for profiling, which uses HEC.

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -130,6 +135,10 @@ service:
       processors: [memory_limiter, batch, resourcedetection/internal]
       exporters: [signalfx/internal]
     logs:
```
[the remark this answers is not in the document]
atoulme: That would help, indeed. Thanks!

atoulme:
Sorry, I'm not feeling like this is ready to merge as I don't have tests to go with this, and it's a breaking change.

cmd/otelcol/config/collector/gateway_config.yaml
dmitryax: > We might need to expose it for profiling, which uses HEC.

Really? Do profiling libraries send HEC format? I was under impression it's OTLP

PR #2381, opened 2022-12-16
Move to use the upstream oracledbreceiver component

atoulme:


dmitryax (approved):
LGTM assuming the configuration interface and emitted metrics are the same

atoulme:
Integration tests test that we get the same metric data.

PR #2382, opened 2022-12-17
make lint build and unit tests execute in parallel

atoulme:


Makefile
```diff
@@ -71,6 +71,15 @@ for-all:
 	 	$${CMD} ); \
 	done
 
+.PHONY: $(ALL_GO_MODULES)
```
rmfitzpatrick, line 74: Are these two needed or can a single definition work without making the module directories phony targets themselves?

```
.PHONY: for-all-target $(ALL_GO_MODULES)
for-all-target: $(ALL_GO_MODULES)
$(ALL_GO_MODULES):
	@echo "Running target '$(TARGET)' in module '$@'"
	$(MAKE) -C $@ $(TARGET)
```

.github/workflows/build-and-test.yml
```diff
@@ -115,8 +115,8 @@ jobs:
       - name: Lint
         run: |
           make install-tools
-          make -j4 for-all CMD="make checklicense impi lint misspell"
-  
+          make -j4 for-all-target TARGET="checklicense impi lint misspell"
```
rmfitzpatrick, line 118: Testing this out I think `ALL_GO_MODULES` doesn't include the root directory because of the final `^./` filter so we may be unintentionally losing coverage for the project?

atoulme: ah that might explain why it's faster ;) checking.

Makefile
atoulme: `.PHONY: $(ALL_GO_MODULES)` is needed otherwise the Makefile will tell you there's no need to run for all the modules as they are unchanged. By making the targets phony, you tell Makefile to not look at the folder timestamps and run the tasks no matter what.

.github/workflows/build-and-test.yml
atoulme: good catch! Fixed.

atoulme:
It looks like this is just pushing the runner past its resource limits, and we get slowness loading packages, which triggers timeouts. I think we can abandon this approach.

PR #2407, opened 2022-12-21
Fix latest git tag detection

atoulme:
Detect the latest tag using `git ls-remote`, as checkouts no longer contain all tags.


dmitryax:
I can help to review this, but there is not enough context. Can we add a description to the PR, at least?

atoulme:
It doesn't look like my fix took, fwiw. Going to review a bit more.

PR #2408, opened 2022-12-22
fix version in ci/cd

atoulme:


atoulme:
I broke off some of this PR into smaller PRs so we could see better what this PR achieves specifically related to version management.

PR #2410, opened 2022-12-22
zero config: add disable svcname generation capability

pmcollins:
Service name generation can now be disabled by setting `generate_service_name=false` in
the config file. In a later release, service name generation will be disabled by
default.


instrumentation/src/config.c
```diff
@@ -72,10 +77,14 @@ void split_on_eq(char *string, struct kv *pair) {
     pair->v = string;
 }
 
-bool eq_true(char *v) {
+bool str_eq_true(char *v) {
     return v != NULL && !streq("false", v) && !streq("FALSE", v) && !streq("0", v);
 }
 
+bool str_eq_false(char *v) {
```
pmcollins, line 84: I'm going to consolidate these two functions into `str_to_bool(char *v, bool default)` in a separate PR.

atoulme (approved):
LGTM, can merge next year :)

PR #2411, opened 2022-12-23
Prepare upgrade to 0.68.0

atoulme:


internal/exporter/httpsinkexporter/config_test.go
```diff
@@ -21,29 +21,30 @@ import (
...
 )
 
 func TestLoadConfig(t *testing.T) {
-	factories, err := componenttest.NopFactories()
-	assert.NoError(t, err)
 
-	factory := NewFactory()
-	factories.Exporters[typeStr] = factory
-	cfg, err := otelcoltest.LoadConfigAndValidate(path.Join(".", "testdata", "config.yaml"), factories)
+	configs, err := confmaptest.LoadConf(path.Join(".", "testdata", "config.yaml"))
```
rmfitzpatrick, line 29: Is removing `LoadConfigAndValidate` dependency update related or are you wanting to minimize the config validation to the components under test?

atoulme: it's a double whammy. Some of the calls in this code are now deprecated, and therefore make lint fail. The way this test works is also antiquated per otel contrib guidelines, where they use this new approach instead.

rmfitzpatrick (approved):

crobert-1 (approved):

PR #2439, opened 2023-01-03
ensure --dry-run doesn't report expanded config values

rmfitzpatrick:
Currently `--dry-run` will provide the final config values with all config source and environment variables having been resolved/expanded. This is not desirable for dynamic* content. These changes add a new `configprovider.Hook` mechanism to provide additional components access of pre-resolving config content like* the config server currently does. The config server converter is updated to conform to the new interface.

These also pick up existing missing imports and add an integration test with an additional testutil container helper.

### DIFF
--- a/internal/configprovider/config_source_provider_test.go
+++ b/internal/configprovider/config_source_provider_test.go
@@ -22,13 +22,12 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/collector/component"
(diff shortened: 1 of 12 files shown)


internal/configprovider/config_source_provider_test.go
```diff
@@ -126,13 +145,26 @@ func TestConfigSourceConfigMapProvider(t *testing.T) {
 				} else {
 					assert.ErrorContains(t, err, tt.wantErr)
 					assert.Nil(t, r)
-					return
+					break
```
rmfitzpatrick, line 148: I think this was an existing bug in the test that missed shutdown coverage

atoulme (approved):

PR #2445, opened 2023-01-04
Provide unexpanded configs in discovery mode

rmfitzpatrick:
In the same vein as https://github.com/signalfx/splunk-otel-collector/pull/2439 these changes ensure that successfully discovered target entities don't report their expanded config values from the discovery confmap provider* (and by extension `--dry-run`), which would make any configs using env vars only valid for running on a matching environment.

They also update all discovery mode integration tests to vet this w/ a new docker exec helper and add a new k8s integration test (the bulk of the changes).

### DIFF
--- a/internal/confmapprovider/discovery/discoverer.go
+++ b/internal/confmapprovider/discovery/discoverer.go
@@ -406,7 +406,7 @@ func (c *Config) observersForDiscoveryMode() []component.ID {
 }
 
 func (d *discoverer) addUnexpandedReceiverConfig(receiverID, observerID component.ID, cfg map[string]any) {
-	d.logger.Info(fmt.Sprintf("adding unexpanded config[%s][%s]: %v\n", receiverID.String(), observerID.String(), cfg))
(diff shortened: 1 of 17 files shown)


internal/confmapprovider/discovery/discoverer.go
```diff
@@ -246,11 +250,7 @@ func (d *discoverer) createObserver(observerID component.ID, cfg *Config) (otelc
 	}
 
 	if ce := d.logger.Check(zap.DebugLevel, "unmarshalled observer config"); ce != nil {
-		if c, e := yaml.Marshal(observerConfig); e != nil {
-			d.logger.Debug("failed marshaling observer config for logging", zap.Error(e))
-		} else {
-			ce.Write(zap.String("config", string(c)))
-		}
+		ce.Write(zap.String("config", fmt.Sprintf("%#v\n", observerConfig)))
```
rmfitzpatrick, line 253: This fixes a bug where I was trying to marshal unexported structs so we're using their value representation instead.

atoulme (approved):

PR #2450, opened 2023-01-06
updated docker compose dependencies

jrich523:
Updated the docker-compose.yaml files to wait for splunk to become healthy by using the depends_on condition that works off the splunk containers healthcheck


atoulme (approved):
Thank you for your contribution!

PR #2451, opened 2023-01-07
Introduce a new example showing how to recombine log entries

atoulme:


crobert-1 (approved):
Looks great, thanks for adding an example here! Examples are always so helpful with trying to set things up.

Not required at all for this PR, but once it's merged it would be good to add some links to this example elsewhere, so it's easier to find. Would it make sense to add a link in contrib repo's `recombine` documentation? Or is this too Splunk specific?

atoulme:
not sure. Might do a blog post.

PR #2460, opened 2023-01-10
Add -f to curl downloads

atoulme:
Add the -f flag to curl downloads in our packaging scripts, so failure to download a component fails the script.

### DIFF
--- a/internal/buildscripts/packaging/fpm/common.sh
+++ b/internal/buildscripts/packaging/fpm/common.sh
@@ -77,7 +77,7 @@ download_smart_agent() {
     local dl_url=""
 
     if [ "$tag" = "latest" ]; then
-        tag=$( curl -sL "$SMART_AGENT_RELEASE_URL/latest" | jq -r '.tag_name' )
+        tag=$( curl -sfL "$SMART_AGENT_RELEASE_URL/latest" | jq -r '.tag_name' )
         if [ -z "$tag" ]; then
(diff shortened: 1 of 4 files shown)


internal/buildscripts/packaging/fpm/common.sh
```diff
@@ -77,7 +77,7 @@ download_smart_agent() {
     local dl_url=""
 
     if [ "$tag" = "latest" ]; then
-        tag=$( curl -sL "$SMART_AGENT_RELEASE_URL/latest" | jq -r '.tag_name' )
+        tag=$( curl -sfL "$SMART_AGENT_RELEASE_URL/latest" | jq -r '.tag_name' )
         if [ -z "$tag" ]; then
             echo "Failed to get tag_name for latest release from $SMART_AGENT_RELEASE_URL/latest" >&2
             exit 1
```
crobert-1, line 83: For a few of these updated curl commands, there's a check directly after to see if the result of the curl command has any contents (`if [ -z "$tag" ]; then`). Is that a valid check for success? If it is, it may be a clearer error message than the script failure. If it's an equivalent check, we can probably remove `-f` when we check the result in the script.

rmfitzpatrick (approved):

internal/buildscripts/packaging/fpm/common.sh
crobert-1: Perfect, thanks for clarifying! Definitely stick with the `-f` then.

crobert-1 (approved):

PR #2365, opened 2022-12-15
upgrade containerd to 1.6.12 and unpeg from 1.5.13

hughesjj:
[details](https://github.com/signalfx/splunk-otel-collector/network/updates/544389848)

PR #2466, opened 2023-01-12
Remove override to address CVE

hughesjj:
See [advisories/GHSA-2qjp-42……5j-52j9](https://github.com/advisories/GHSA-2qjp-425j-52j9)](https://github.com/advisories/GHSA-2qjp-425j-52j9)


go.mod
```diff
@@ -515,7 +515,6 @@ replace github.com/docker/go-connections => github.com/docker/go-connections v0.
 replace (
 	github.com/Masterminds/goutils => github.com/Masterminds/goutils v1.1.1
 	github.com/apache/thrift => github.com/apache/thrift v0.16.0
-	github.com/containerd/containerd => github.com/containerd/containerd v1.5.13
```
crobert-1, line 518: If we remove this, will the dependency upgrade take it to the `1.6.X` release? If so, does everything still work fine on `1.6.X`? If we're worried about it we could just replace `1.5.13` with `1.5.16` instead of deleting the line.

jeffreyc-splunk: I think this was intended to be upgraded as part of https://github.com/signalfx/splunk-otel-collector/pull/2365.

jeffreyc-splunk (approved):

go.mod
crobert-1: Awe, missed that, thanks!

crobert-1 (approved):

PR #2473, opened 2023-01-13
Update core/contrib deps to latest

jeffreyc-splunk:


rmfitzpatrick:
Do you mind updating the pkg component deps as well? https://github.com/signalfx/splunk-otel-collector/tree/main/pkg

PR #2483, opened 2023-01-17
Use GITHUB_OUTPUT instead of set-output

jeffreyc-splunk:
`set-output` is being deprecated: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


atoulme:
changes seem to look good, build failure is unrelated.

crobert-1 (approved):

PR #2488, opened 2023-01-18
Switch to gcp from gce 

hughesjj:
as specified in breaking changes, need to convert these
* https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/10348

I'm sure there's some cleaner ways to organize the code and am receptive to any suggestions if you have some.

### DIFF
--- a/cmd/otelcol/config/collector/full_config_linux.yaml
+++ b/cmd/otelcol/config/collector/full_config_linux.yaml
@@ -204,7 +204,7 @@ processors:
   # Enables the resource detection processor with default settings
   # Full configuration here: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor
   # NOTE: This processor settings need to be change when using this processor
-  #resouredection:
+  #resourcedetection:
     # # a list of resource detectors to run, valid options are: "env", "system",  "gcp", "ec2", "ecs", "elastic_beanstalk"
     #detectors: [ <string> ]
     # # determines if existing resource attributes should be overridden or preserved, defaults to true


internal/configconverter/normalize_gcp.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"go.opentelemetry.io/collector/confmap"
+	"log"
+)
+
+type NormalizeGcp struct{}
+
+func (NormalizeGcp) Convert(_ context.Context, in *confmap.Conf) error {
+	if in == nil {
+		return fmt.Errorf("cannot NormalizeGcp on nil *confmap.Conf")
```
rmfitzpatrick, line 28: I'm not sure the precedent for this is correct and simply returning nil would probably be better.

internal/configconverter/normalize_gcp.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	const resourceDetector = "processors::resourcedetection::detectors"
```
rmfitzpatrick, line 31: components can be named so you'll want to match against an optional `/<name>`. Otherwise we could just get the sub-config and not need to iterate.

internal/configconverter/normalize_gcp.go
```diff
@@ -0,0 +1,78 @@
...
+	out := map[string]any{}
```
rmfitzpatrick, line 32: updating in place makes more sense to me than recreating a new confmap

internal/configconverter/normalize_gcp.go
```diff
@@ -0,0 +1,78 @@
...
+	nonNormalizedGcpDetectorFound := false
+	for _, k := range in.AllKeys() {
+		v := in.Get(k)
+		match := resourceDetector == k
+		if !match {
+			out[k] = v
+		} else {
+			switch v.(type) {
+			case string:
```
rmfitzpatrick, line 41: detectors have always been a yaml list so would this ever be a string? https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/dfcd49a8cbf68f10a4d77aff453c2e1e6a5256b3#diff-6c201bcf6478be7ec71d2ed94132853dfa188e033c9c432048d19744ffdf2553R28

internal/configconverter/normalize_gcp_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/confmap/confmaptest"
+)
+
+func TestNormalizeGcp(t *testing.T) {
+	expectedCfgMap, err := confmaptest.LoadConf("testdata/normalize_gcp/upstream_agent_config_post_migration.yaml")
```
rmfitzpatrick, line 41: imo it greatly improves readability and lessens maintenance requirements to use `confmap.NewFromStringMap()` directly instead of having separate files.

internal/configconverter/normalize_gcp.go
rmfitzpatrick: who are we warning and to what end? Converters are used by the core service: https://github.com/open-telemetry/opentelemetry-collector/blob/776c6b38443f431915493febc3a283961a97d0db/confmap/resolver.go#L178

rmfitzpatrick: thanks, missed update

internal/configconverter/testdata/gcp-deprecation.yaml
```diff
@@ -0,0 +1,35 @@
+receivers:
```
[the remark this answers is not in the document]
rmfitzpatrick: fwiw I think its more expansive checks could be helpful in testing

cmd/otelcol/config/collector/full_config_linux.yaml
```diff
@@ -205,7 +205,7 @@ processors:
   # Full configuration here: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor
   # NOTE: This processor settings need to be change when using this processor
   #resouredection:
```
atoulme, line 207:
```suggestion
  #resourcedetection:
```

atoulme (approved):

PR #2538, opened 2023-02-07
add support for always on profiling to zero config

pmcollins:
This change adds three new configuration parameters that individually turn on the following flags:

-Dsplunk.profiler.enabled=true
-Dsplunk.profiler.memory.enabled=true
-Dsplunk.metrics.enabled=true

These are turned off by default, so existing users will not be affected at this time.


instrumentation/README.md
```diff
@@ -58,6 +58,31 @@ Java instrumentation jar. Typically, it will be set to something like:
...
 
+### disable_telemetry (optional)
+
+Set this value to `true` to disable the preloader from sending the `splunk.linux-autoinstr.executions` metric to the
+local collector. Default: `false`.
+
+### generate_service_name (optional)
+
+Set this value to `false` to prevent the preloader from setting the `OTEL_SERVICE_NAME` environment variable.
+Default: `true`.
```
rmfitzpatrick, line 69: Is there a doc or brief description you could provide of what the `false` behavior is?

instrumentation/src/splunk.c
```diff
@@ -8,14 +8,17 @@
 #include <unistd.h>
 #include <stdio.h>
 
-#define ENV_VAR_LEN 512
+#define MAX_ENV_VAR_LEN 512
```
rmfitzpatrick, line 11: Is this a known limit for `JAVA_TOOL_OPTIONS`?

instrumentation/src/splunk.c
```diff
@@ -148,15 +155,26 @@ void set_env_var(logger log, const char *var_name, const char *value) {
...
+    char java_tool_options[MAX_ENV_VAR_LEN] = JAVA_TOOL_OPTIONS_PREFIX;
     char log_line[MAX_LOG_LINE_LEN] = "";
     size_t jar_path_len = strlen(cfg->java_agent_jar);
     if (jar_path_len > MAX_CONFIG_ATTR_LEN) {
         sprintf(log_line, "jar_path too long: got %zu chars, max %d chars", jar_path_len, MAX_CONFIG_ATTR_LEN);
         log_warning(log, log_line);
         return;
     }
-    strcat(java_tool_options, (*cfg).java_agent_jar);
+    strncat(java_tool_options, cfg->java_agent_jar, MAX_ENV_VAR_LEN - strlen(java_tool_options) - 1);
```
rmfitzpatrick, line 166: does it make sense to track remaining length for each of these updates instead of what may be silent noops with partial properties?

instrumentation/README.md
jeffreyc-splunk: Which takes precedence if both `generate_service_name` is `true` and `service_name` is set?

pmcollins: Turns out there really isn't a doc yet that describes in detail what the instrumentation library does, although there is [this](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/resource-providers). But I've added a couple of sentences to explain the implications of the false behavior.

Regarding precedence, `generate_service_name=true` will cause the .so generate the service name either dynamically or via the `service_name` override. If it's false, no service name will be set, regardless of the value of the `service_name` override. I've added an explanation to the readme as well as a test case.

instrumentation/src/splunk.c
pmcollins: No, it's just an arbitrary value that's roughly an order of magnitude larger than what it will need.

instrumentation/src/splunk.c
pmcollins: I wasn't able to cause truncation using the maximum string lengths that the .so allows, but I've made some changes to check for truncation anyway to future proof this function (or in case I missed something), and I've added some unit tests to test the limits.

instrumentation/src/splunk.c
rmfitzpatrick: SO suggests it's hardcoded to 1024 in the jdk: https://stackoverflow.com/a/41792192

rmfitzpatrick (approved):

PR #2569, opened 2023-02-16
Split build and integration test into separate jobs

jeffreyc-splunk:
Avoid unnecessary rebuilds when re-running failed tests.


rmfitzpatrick (approved):
Thanks!

PR #2574, opened 2023-02-16
Remove smartagent/signalfx-forwarder from metrics pipeline, signalfxreceiver should be used

bogdandrutu:


bogdandrutu:
@flands maybe you know the reason https://github.com/signalfx/splunk-otel-collector/pull/218#discussion_r604336005. But here are the problems:
* Same formats are fully supported in the signalfx receiver, so confusing users between which to use.
* signalfx-forwarder does not passthrough the token.

dmitryax (commented):
I'm good with this change. The only problem is that now users cannot migrate from smart-agent to collector without migrating their old instrumentations. Maybe we add a deprecation warning first stating the limitation of the receiver?

bogdandrutu:
> I'm good with this change. The only problem is that now users cannot migrate from smart-agent to collector without migrating their old instrumentations. Maybe we add a deprecation warning first stating the limitation of the receiver?

Why? Can you clarify?

dmitryax:
I mean, currently, users can replace signalfx-agent with the collector on the default config, and all the instrumentation will continue working. After this change, they need to either change the instrumentations to send to the new port before the switching from smart-agent (data loss expected) or replace the agent with the collector with a manually added smartagent/signalfx-forwarder receiver. I'm good with the second option, but we need to document this step in the migration doc. 

Another option would be to add a deprecation warning log message to the `smartagent/signalfx-forwarder` mentioning its limitations until we fully deprecate the smart agent

atoulme:
I would be in favor that we close this PR and instead start with a warning message and a deprecation schedule for the signalfx-forwarder. We will work the kinks. If no one objects to this course of action in the next 30 days, I will close this PR.

dmitryax:
> I would be in favor that we close this PR and instead start with a warning message and a deprecation schedule for the signalfx-forwarder. We will work the kinks. If no one objects to this course of action in the next 30 days, I will close this PR.

Sounds good to me

PR #2601, opened 2023-02-22
Add discoverybundler, initial embedded bundle.d, and enabled properties

rmfitzpatrick:
These changes add a new `bundle.d` embedded filesystem for discovery config bundling. Also provide a new `discoverybundler` command to help with discovery config creation (limited to property helpers initially). Also including `enabled` properties to allow disabling bundled receiver instances out of the gate and some minor discovery logging enhancements.

~I will be adding a more detailed readme to this branch but opening as draft for now.~ Readmes have been updated.


internal/confmapprovider/discovery/README.md
```diff
@@ -35,17 +35,68 @@ graph LR
...
   subgraph 5a[receivers]
     5 --> 5a1>otlp.yaml]
-    5a1 --> 5b1[[otlp:<br>protocols:<br>grpc]]
+    5a1 --> 5b1[[otlp:<br>protocols:<br>grpc:]]
   end
 ```
 
-This component is currently exposed in the Collector via the `--configd` option with corresponding
-`--config-dir <config.d path>` and `SPLUNK_CONFIG_DIR` option and environment variable to load
+This component is currently supported in the Collector settings via the `--configd` option with corresponding
```
pmcollins, line 42: I'm not a writer of any kind, but I believe that some writing can be made easier to understand by using the active voice and by introducing new concepts at the end of a phrase or sentence.

What do you think about going in this direction instead? (I'm not sure if what I've written is correct btw)

> To enable the discovery confmap.Provider, use the `--configd` option when starting the Collector. This will load configurations from the default directory, `/etc/otel/collector/config.d`. To load configurations from a different directory, either add a `--config-dir <config.d path>` option or set the `SPLUNK_CONFIG_DIR` environment variable to a directory containing your configuration files. Once started, the confmap.Provider has two major modes of operation: it can either generate a configuration and then start a Collector with that configuration, or it can generate a configuration and just print it to stdout. To have the Collector just print its generated configuration, use the `--dry-run` flag. (meta: not sure what is meant by `terminating` as currently written; does that mean it has to be the last flag?)

rmfitzpatrick: >I'm not a writer of any kind

No need to sell yourself short ;)

I'm not a member of the "use active voice" community but want to avoid a style-driven impasse in favor of clarity. I'd prefer to have the component as the main subject as this readme isn't intended to be end user documentation at this time (cc @theletterf). Will take another pass to separate its settings w/ clearer descriptions (maybe a table?) and appreciate your suggestion.

"terminating" meant it will terminate the process and the option itself isn't order dependent.

pmcollins: Ok, may be just the way my brain works (I'm also striving for clarity). Will defer to docs folks on this.

pmcollins (approved):

PR #2630, opened 2023-02-28
Update OTLP TLS insecure config option formatting

crobert-1:
Configuration for the OTLP exporter "insecure" option changed a long time ago to move under the "tls" map item. We have a lot of tests using the old format, so this updates them to match the new.

### DIFF
--- a/tests/general/container_test.go
+++ b/tests/general/container_test.go
@@ -128,7 +128,8 @@ func TestConfigYamlEnvVar(t *testing.T) {
 exporters:
   otlp:
     endpoint: "${OTLP_ENDPOINT}"
-    insecure: true
+    tls:
+      insecure: true
 
 service:
   pipelines:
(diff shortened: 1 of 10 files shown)


atoulme (approved):

tests/general/container_test.go
```diff
@@ -128,7 +128,8 @@ func TestConfigYamlEnvVar(t *testing.T) {
 exporters:
   otlp:
     endpoint: "${OTLP_ENDPOINT}"
-    insecure: true
+    tls:
```
rmfitzpatrick, line 131: Do you mind keeping this and adding a comment saying it's intentional to ensure that converters apply to this config mechanism?

crobert-1: Sure, I can open another PR undoing this change. I saw a few tests with the incorrect format on purpose that I left, but I didn't know this was also intentionally out-dated.

PR #2642, opened 2023-03-02
Upgrade internal/tools versions

evantorrie:
This was triggered by failure to build with golang-1.20. Versions of golangci-lint prior to 1.50 would often crash/hang with golang-1.20 due to that version's addition of slice-to-array. See https://github.com/golangci/golangci-lint/issues/3420

Since I was also here, I upgraded other versions of tools to their most recent releases where appropriate.


dmitryax (approved):

dmitryax:
Thanks, @evantorrie!

PR #2650, opened 2023-03-03
Changes to the pulsarexporter configuration to prepare with using upstream

atoulme:
These configuration changes sync the options with what is offered by the upstream pulsar exporter offered by the contrib project. Along with the changes from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/19184, these changes will allow down the road to move to the upstream exporter.


pmcollins (approved):
Thanks for taking this on. LGTM.

PR #2668, opened 2023-03-07
adds mdatagen as a tool/adopts into splunk otel distribution

hughesjj-splunk:
If this looks good I can add the `//go:generate mdatagen metadata.yaml` to more receivers.

Open to changes on the naming of the make target, feel free to ping me on slack or dm me here.


jeffreyc-splunk:
I think we'll need to add a `make install-tools` step before this step: https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/integration-test.yml#L48

Makefile
```diff
@@ -115,6 +115,11 @@ gendependabot:
 		  echo "  - package-ecosystem: \"docker\"\n    directory: \"$${dir:1}\"\n    schedule:\n      interval: \"daily\"" >> ${DEPENDABOT_PATH} ); \
 	done
 
+.PHONY: generate
```
jeffreyc-splunk, line 118: Does this target need to be explicitly called during CI?

Makefile
```diff
@@ -115,6 +115,11 @@ gendependabot:
...
+generate:
```
jeffreyc-splunk, line 119: Does this make sense?
```suggestion
generate: install-tools
```

hughesjj-splunk: oh duh that's how to do it, thank you!

Makefile
hughesjj-splunk: Actually, no.  I'd wager you might have some good thoughts on another Makefile etc to place this in.


that said I could see it advantageous as a "wait you actually reran generate and there's no delta from the commit" check if you think it's wise to add such

jeffreyc-splunk:
Need to a `make install-tools` step before https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/integration-test.yml#L48

jeffreyc-splunk:
Do we need to add a `make generate` step before building the collector binaries (https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/build-and-test.yml#L222)?

rmfitzpatrick:
> Do we need to add a `make generate` step before building the collector binaries (https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/build-and-test.yml#L222)?

At least at this initial stage I think generation should be done manually (read: intentionally like the rest of source changes). Immediate implications to auto-generating would be fieldalignment and lint issues that need to modify the commit contents.

Makefile
```diff
@@ -115,6 +115,11 @@ gendependabot:
 		  echo "  - package-ecosystem: \"docker\"\n    directory: \"$${dir:1}\"\n    schedule:\n      interval: \"daily\"" >> ${DEPENDABOT_PATH} ); \
 	done
 
+.PHONY: generate
+generate: install-tools
```
rmfitzpatrick, line 119: something like
```suggestion
generated-metrics: install-tools
```
?

.github/workflows/integration-test.yml
```diff
@@ -45,6 +45,7 @@ jobs:
             /home/runner/go/pkg/mod
             /home/runner/.cache/go-build
           key: v1-go-pkg-mod-${{ runner.os }}-${{ hashFiles('**/go.mod', '**/go.sum') }}
+      - run: make install-tools
```
rmfitzpatrick, line 48: Why would this be necessary for integration tests?

rmfitzpatrick:
The internal/tools system works w/ side-effect imports so the mdatagen cmd would need to be added: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/tools/tools.go#L21

atoulme (approved):

jeffreyc-splunk:
Should we add a new github job to run `make generate-metrics`, and fail if there are uncommitted changes, similar to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/build-and-test.yml#L184?

dmitryax (approved):

PR #2671, opened 2023-03-08
Build linux arm64 bundle

jeffreyc-splunk:
- Update github workflows and gitlab pipeline to build the arm64 bundle and include it in the arm64 deb/rpm/tar packages and docker image
- Update sanity tests to ensure the arm64 packages and image include the bundle

Note: We currently cache the image layers for previous bundle builds in github workflows, but the arm64 bundle build may take over an hour due to qemu slowness.  This is not an issue in gitlab since we can build natively on the splunk-provided arm runners.


hughesjj-splunk:
looks good so far, will approve upon passing tests/moving out of draft

atoulme (approved):
LGTM

.gitlab-ci.yml
```diff
@@ -790,6 +789,8 @@ github-release:
       if [[ "${CI_COMMIT_TAG:-}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
         mv dist/assets/agent-bundle_linux_amd64.tar.gz dist/assets/agent-bundle_${CI_COMMIT_TAG#v}_linux_amd64.tar.gz
         mv dist/assets/agent-bundle_linux_amd64.tar.gz.asc dist/assets/agent-bundle_${CI_COMMIT_TAG#v}_linux_amd64.tar.gz.asc
+        mv dist/assets/agent-bundle_linux_arm64.tar.gz dist/assets/agent-bundle_${CI_COMMIT_TAG#v}_linux_arm64.tar.gz
```
rmfitzpatrick, line 792: I think it would be helpful to discuss w/ @aunshc before adding this to the release artifacts. We gain a lot from these changes without this addition imo.

jeffreyc-splunk: My initial reasoning for releasing the bundle separately was for users of the standalone `otelcol` binary that also want to use the smart agent extension/receiver.  If this is something that we don't want to support, then I can revert these changes.

aunshc: Got it, missed that. Been chatting with Paulo about a similar scenario for OTel .NET auto instrumentation too - there is a possible use case for this where developers want to try the smart agent receiver/extension on their arm systems. This is not a high priority use case at the moment to be supported, we can revert these changes for now.

jeffreyc-splunk: Thanks @aunshc.  I'll exclude it from the release artifacts.

PR #2672, opened 2023-03-08
Deserialize dependency test workflow

jeffreyc-splunk:
- Refactor dependency test workflow into separate jobs
- Only run downstream builds/tests if there are upstream core/contrib updates
- Use test splitting from https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/integration-test.yml for integration tests
- Helps prevent failures due to lack of disk space: https://github.com/signalfx/splunk-otel-collector/actions/runs/4364071868

### DIFF
--- a/.github/workflows/deps-test.yml
+++ b/.github/workflows/deps-test.yml
@@ -9,36 +9,27 @@ on:
 
 env:
   OTEL_VERSION: latest
+  GO_VERSION: 1.19.6
+  GOTESPLIT_TOTAL: "10"
 
 concurrency:
   group: deps-test-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
-  deps-test:
+  # Update the core/contrib dependencies to ${{ env.OTEL_VERSION }}.
+  # If there are any changes to the go.mod and go.sum files, cache them and the downloaded dependencies.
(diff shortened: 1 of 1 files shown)


.github/workflows/deps-test.yml
```diff
@@ -95,3 +63,181 @@ jobs:
...
+            **/go.mod
+            **/go.sum
+            ~/go/pkg/mod
+          key: update-deps-${{ github.run_id }}
+      - uses: actions/setup-go@v3
+        if: steps.restore-cache.outputs.cache-hit == 'true'
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - run: make test
+        if: steps.restore-cache.outputs.cache-hit == 'true'
```
hughesjj-splunk, line 117: Wait `cache-hit=='true'` means we should build things?  Wouldn't it be the opposite (`cache-hit != 'true'` )?

jeffreyc-splunk: For this workflow, we only need to run the tests when there are new upstream core/contrib releases.  After running the `update-deps` script, it checks if the go.mod/go.sum files are changed.  These files are then cached only if there are changes, and the downstream build/test jobs only run when there is a cache hit.

jeffreyc-splunk:
Closing in favor of #2692

PR #2675, opened 2023-03-09
support .env file to load env

### LINKED ISSUE #2643: Feature Request: command line option for `NO_WINDOWS_SERVICE` env var
I am working on otel TA for Splunk Universal Forwarder. 
We are having some trouble for windows service account that runs Splunk but doesn't have permission to change env var.
It would be great if `NO_WINDOWS_SERVICE` feature can be controlled some other way than env var such as command line option.

atoulme:
Fixes #2643

This adds an autoload of any .env file located in the working directory when invoking the collector.

A .env file may contain environment variables that then get loaded as part of the env vars seen by the collector.

This is a lightweight approach to fixing the issue.

### DIFF
--- a/cmd/otelcol/main.go
+++ b/cmd/otelcol/main.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"os"
 
+	_ "github.com/joho/godotenv/autoload"
 	flag "github.com/spf13/pflag"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/confmap"
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/hashicorp/vault-plugin-auth-gcp v0.15.0
 	github.com/hashicorp/vault/api v1.9.0
 	github.com/jaegertracing/jaeger v1.41.0
+	github.com/joho/godotenv v1.3.0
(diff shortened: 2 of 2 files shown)


rmfitzpatrick:
Can you update the advanced configuration section of the readme and add an integration test for the functionality?

cmd/otelcol/main.go
```diff
@@ -22,6 +22,7 @@ import (
 	"log"
 	"os"
 
+	_ "github.com/joho/godotenv/autoload"
```
rmfitzpatrick, line 25: What is the behavior of invalid .env content and would it be better to use the library directly to marshal the error to a proper handler, like in main or `internal/settings`?

hughesjj-splunk:
> Can you update the advanced configuration section of the readme and add a unit or integration test* for the functionality?

+1 to this, and specifically it'd be nice to have an invocation example in the `README`

cmd/otelcol/main.go
rmfitzpatrick: imo I'd expect there to be a logged warning about the validity of the file w/ the parsing error and for none of the env vars in the file to be sourced but can see other options being desirable. Terminating the collector seems overkill.

atoulme: thanks, I'll factor that in.

PR #2720, opened 2023-03-13
Add test for custom collectd plugin

atoulme:
This adds a new integration test that checks explicitly that the smart agent receiver can work with custom collectd plugins.

### DIFF
--- a/tests/receivers/custom_upstat/custom_upstat_test.go
+++ b/tests/receivers/custom_upstat/custom_upstat_test.go
@@ -1,0 +1,45 @@
+// Copyright Splunk, Inc.
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build integration
+
+package tests
+
+import (
+	"path"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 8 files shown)


dloucasfx (approved):

tests/testutils/collector_process.go
```diff
@@ -49,6 +49,10 @@ type CollectorProcess struct {
 	Fail             bool
 }
 
+func (collector CollectorProcess) WithBoundDirectory(path string, mountPoint string) (Collector, error) {
+	return &collector, ErrUnsupportedFeature
```
atoulme, line 53: @rmfitzpatrick I expect there is a better way that will be obvious to you - please let me know.

tests/receivers/custom_upstat/custom_upstat_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/signalfx/splunk-otel-collector/tests/testutils"
+)
+
+func TestCustomUpstatIntegration(t *testing.T) {
+	path, err := filepath.Abs(path.Join(".", "testdata", "custom"))
+	require.NoError(t, err)
+	testutils.AssertAllMetricsReceived(t, "all.yaml", "custom_upstat.yaml",
+		nil, []testutils.CollectorBuilder{func(collector testutils.Collector) testutils.Collector {
+			collector, err := collector.WithBoundDirectory(path, "/var/collectd-python/upstat")
```
rmfitzpatrick, line 36: Since this is a container requirement only I don't think we need to modify the helper types just yet. After more cases require this a better system may emerge.

```go
		func(c testutils.Collector) testutils.Collector {
			custom, err := filepath.Abs(filepath.Join(".", "testdata", "custom"))
			require.NoError(t, err)
			if cc, ok := c.(*testutils.CollectorContainer); ok {
			    cc.Container = cc.Container.WithMount(testcontainers.BindMount(custom, "/var/collectd-python/upstat"))
			    return cc.WithEnv(map[string]string{"PLUGIN_FOLDER": "/var/collectd-python/upstat"})
            }
            return collector.WithEnv(map[string]string{"PLUGIN_FOLDER": custom})
		},
```

atoulme: ok sure!

PR #2728, opened 2023-03-15
Update agent_config.yaml

theletterf:
Replaced deprecated setting of `logging` exporter, as per upstream documentation.


jeffreyc-splunk:
Should we make the same change for all of our other [configs](https://github.com/signalfx/splunk-otel-collector/tree/main/cmd/otelcol/config/collector)?

rmfitzpatrick:
Thanks for this. Would you mind including updates to full_config_linux.yaml, upstream_agent_config.yaml, gateway_config.yaml, and otlp_config_linux.yaml as well?

deployments/ansible/molecule/custom_vars/custom_collector_config.yml
```diff
@@ -24,9 +24,9 @@ processors:
 
 exporters:
   logging/debug:
-    loglevel: debug
+    verbosity: detailed
   logging/info:
-    loglevel: info
+    verbosity: normal
```
jeffreyc-splunk, line 29: @atoulme We should revert this since this ansible test is specifically testing with an older version of the collector.

dmitryax (approved):

PR #2729, opened 2023-03-15
Add java zc options to ansible

jeffreyc-splunk:
For https://signalfuse.atlassian.net/browse/OTL-1886

### DIFF
--- a/deployments/ansible/molecule/with_instrumentation/converge.yml
+++ b/deployments/ansible/molecule/with_instrumentation/converge.yml
@@ -9,6 +9,11 @@
     splunk_otel_auto_instrumentation_version: 0.50.0
     splunk_otel_auto_instrumentation_resource_attributes: deployment.environment=test
     splunk_otel_auto_instrumentation_service_name: test
+    splunk_otel_auto_instrumentation_generate_service_name: false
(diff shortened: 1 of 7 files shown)


deployments/ansible/molecule/with_instrumentation/converge.yml
```diff
@@ -9,6 +9,10 @@
     splunk_otel_auto_instrumentation_version: 0.50.0
     splunk_otel_auto_instrumentation_resource_attributes: deployment.environment=test
     splunk_otel_auto_instrumentation_service_name: test
+    splunk_otel_auto_instrumentation_disable_telemetry: true
```
pmcollins, line 12: Excuse my ignorance about Ansible, but what do these lines do?

jeffreyc-splunk: These are the config options when testing the playbook.  We then verify this option was set correctly in the config file: https://github.com/signalfx/splunk-otel-collector/pull/2729/files#diff-2f799af543408e9ef525d0c2dbe4fb5bd5e166333c31a47aa486a1ade1081371R55

pmcollins (approved):

PR #2737, opened 2023-03-16
Dynamically create test matrix for chef tests

jeffreyc-splunk:
Ensures all platforms/suites defined in [kitchen.yml](https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/chef/kitchen.yml) and [kitchen.windows.yml](https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/chef/kitchen.windows.yml) are tested without manually updating the workflow file.

### DIFF
--- a/.github/workflows/chef-test.yml
+++ b/.github/workflows/chef-test.yml
@@ -47,14 +47,30 @@ jobs:
         id: get-linux-matrix
         run: |
           distros=$(yq '[.platforms[] | "\"" + (.name | sub("_", "-") | sub("\.", "")) + "\""] | join(", ")' kitchen.yml)
+          if [ -z "$distros" ]; then
+            echo "Failed to get platforms from kitchen.yml!" >&2
+            exit 1
+          fi
           suites=$(yq '[.suites[] | "\"" + (.name | sub("_", "-") | sub("\.", "")) + "\""] | join(", ")' kitchen.yml)
+          if [ -z "$suites" ]; then
+            echo "Failed to get suites from kitchen.yml!" >&2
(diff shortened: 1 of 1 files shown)


.github/workflows/chef-test.yml
```diff
@@ -37,13 +37,35 @@ jobs:
...
+  chef-kitchen-matrix:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Get linux distros and suites
+        id: get-linux-matrix
+        run: |
+          distros=$(yq '[.platforms[] | "\"" + (.name | sub("_", "-") | sub("\.", "")) + "\""] | join(", ")' kitchen.yml)
```
rmfitzpatrick, line 49: Does it make sense to add a helper script to generate these vars that can be sourced instead of requiring an action run? It could contain the same invocations or maybe generates a static file that changes on platform/suite additions.

rmfitzpatrick: on second thought since the matrix is the only use case it probably doesn't make sense unless the content would be consumed by an additional process, but there's the small risk of uncaught yq issues by not having the intermediary format.

rmfitzpatrick (approved):

.github/workflows/chef-test.yml
jeffreyc-splunk: Yeah, we just want to create the `matrix` json objects for the downstream matrix configs at run time.  I'll add some checks to make sure that `distros` and `suites` are not empty.

PR #2740, opened 2023-03-17
add properties.discovery.yaml support

rmfitzpatrick:
These changes add the ability to specify discovery properties in a single `config.d/properties.discovery.yaml` file that are merged w/ env/cli set properties.

edit: these changes also fix a [bug I introduced](https://github.com/signalfx/splunk-otel-collector/blob/cb4e27e2dc4010d93ed3e8abbe1d5e9e790ace7e/internal/confmapprovider/discovery/config.go#L497) where `configDirRootRegex` was using a group in a range exclusion.

### DIFF
--- a/internal/confmapprovider/discovery/README.md
+++ b/internal/confmapprovider/discovery/README.md
@@ -102,3 +102,49 @@ successfully started observers.
 
 
 By default, the Discovery mode is provided with pre-made discovery config components in [`bundle.d`](./bundle/README.md).
+
+
+### Discovery properties
+
+Configuring discovery components is performed by merging discovery properties with the config.d receivers
+and extensions `*.discovery.yaml` files. Discovery properties are of the form:
+
+```yaml
+splunk.discovery.receivers.<receiver-type(/name)>.config.<field>(<::subfield>)*: <value>
+splunk.discovery.extensions.<observer-type(/name)>.config.<field>(<::subfield>)*: <value>
+splunk.discovery.receivers.<receiver-type(/name)>.enabled: <true or false>
+splunk.discovery.extensions.<observer-type(/name)>.enabled: <true or false>
+
+# Examples
+splunk.discovery.receivers.prometheus_simple.config.labels::my_label: my_label_value
+splunk.discovery.receivers.prometheus_simple.enabled: true
+
+splunk.discovery.extensions.docker_observer.config.endpoint: tcp://localhost:8080
+splunk.discovery.extensions.k8s_observer.enabled: false
+```
+
+These properties can be in `config.d/properties.discovery.yaml` or specified at run time with `--set` command line options.
+
+Each discovery property also has an equivalent environment variable form using `_x<hex pair>_` encoded delimiters for
+non-word characters `[^a-zA-Z0-9_]`:
+
+```bash
+SPLUNK_DISCOVERY_RECEIVERS_receiver_x2d_type_x2f_receiver_x2d_name_CONFIG_field_x3a__x3a_subfield=value
+SPLUNK_DISCOVERY_EXTENSIONS_observer_x2d_type_x2f_observer_x2d_name_CONFIG_field_x3a__x3a_subfield=value
+SPLUNK_DISCOVERY_RECEIVERS_receiver_x2d_type_x2f_receiver_x2d_name_ENABLED=<true or false>
+SPLUNK_DISCOVERY_EXTENSIONS_observer_x2d_type_x2f_observer_x2d_name_ENABLED=<true or false>
+
+# Examples
+SPLUNK_DISCOVERY_RECEIVERS_prometheus_simple_CONFIG_labels_x3a__x3a_my_label="my_username"
+SPLUNK_DISCOVERY_RECEIVERS_prometheus_simple_ENABLED=true
+
+SPLUNK_DISCOVERY_EXTENSIONS_docker_observer_CONFIG_endpoint="tcp://localhost:8080"
+SPLUNK_DISCOVERY_EXTENSIONS_k8s_observer_ENABLED=false
+```
+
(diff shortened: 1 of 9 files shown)


internal/confmapprovider/discovery/README.md
```diff
@@ -101,4 +101,50 @@ successfully started observers.
...
+splunk.discovery.receivers.prometheus_simple.config.labels::my_label: my_label_value
+splunk.discovery.receivers.prometheus_simple.enabled: true
+
+splunk.discovery.extensions.docker_observer.config.endpoint: tcp://localhost:8080
+splunk.discovery.extensions.k8s_observer.enabled: false
+```
+
+These properties can be in `config.d/properties.discovery.yaml` or specified at run time with `--set` command line options.
+
+Each discovery property also has an equivalent environment variable form using `_x<hex pair>_` encoded delimiters for
```
rmfitzpatrick, line 128: I plan on adding a cli helper of some kind in the future to make producing these env vars possible for the user similar to what we offer in the bundle.d template helpers.

internal/confmapprovider/discovery/testdata/invalid-properties.d/properties.discovery.yaml
```diff
@@ -0,0 +1 @@
+invalid
```
crobert-1, line 1: Nit: should there be an empty newline at the end of this file?

tests/general/discoverymode/docker_observer_discovery_test.go
```diff
@@ -127,8 +130,11 @@ func TestDockerObserver(t *testing.T) {
 							"config": map[string]any{
 								"collection_interval": "1s",
 								"labels": map[string]any{
-									"label_one": "${LABEL_ONE_VALUE}",
-									"label_two": "${LABEL_TWO_VALUE}",
+									"label_four":  "actual.label.four.value",
```
crobert-1, line 133: I realize this has likely no functional impact at all, but is there a reason these labels aren't in order? Is the comparison somehow order sensitive and we're just trying to match what we get back from the collector container's initial config?

internal/confmapprovider/discovery/testdata/config.d/properties.discovery.yaml
```diff
@@ -0,0 +1,2 @@
+splunk.discovery.receivers.smartagent/postgresql.config.params.unused_param: param_value
+splunk.discovery.extensions.docker_observer.config.timeout: 1s
```
crobert-1, line 2: I'm not following, what's the purpose of this file and the `invalid-properties.d/properties.discovery.yaml` file?

internal/confmapprovider/discovery/README.md
crobert-1, line 128: Is this format `_x<hex pair for the given delimeter>_` some kind of encoding standard? I'm not sure I've seen this.

internal/confmapprovider/discovery/README.md
```diff
@@ -101,4 +101,50 @@ successfully started observers.
...
+non-word characters `[^a-zA-Z0-9_]`:
```
crobert-1, line 129: Could you maybe provide an specific example of the mapping from the yaml format to the equivalent environment variable form? It took me quite a while to understand how exactly this mapping worked.

crobert-1 (commented):
Thanks for all of the discovery property examples and clear priority list for different ways to supply properties, super helpful!

tests/general/discoverymode/docker_observer_discovery_test.go
rmfitzpatrick: [Maps are unordered](https://go.dev/ref/spec#Map_types) in go and [in yaml](https://yaml.org/spec/1.2.2/#3221-mapping-key-order) so no. The reason was they were declared alphabetically but I added `label_five` at the top w/o resorting.

internal/confmapprovider/discovery/README.md
rmfitzpatrick: https://github.com/signalfx/splunk-otel-collector/pull/2494#discussion_r1083085874

I'm not familiar with a word character charset that I could use w/ idn or stringprep so defined the "wordify" encoding.

internal/confmapprovider/discovery/README.md
rmfitzpatrick: The guidance for individual receivers will include their form for c/p and I'm still thinking about how to expose the helper w/ a cli: https://github.com/signalfx/splunk-otel-collector/blob/cb4e27e2dc4010d93ed3e8abbe1d5e9e790ace7e/internal/confmapprovider/discovery/bundle/templatefunctions.go#L33

internal/confmapprovider/discovery/testdata/config.d/properties.discovery.yaml
rmfitzpatrick: They are used in tests `TestConfig` and `TestInvalidConfigDirContents`.

internal/confmapprovider/discovery/testdata/invalid-properties.d/properties.discovery.yaml
rmfitzpatrick: added one for good measure.

crobert-1: I don't notice either way except for GitHub reviews, since it explicitly shows if a file doesn't have an empty line at the end. Since most do I just assume that's the standard, so this comment was just a consistency check.

internal/confmapprovider/discovery/README.md
crobert-1: That's fair, thanks for the reference!

crobert-1 (approved):

atoulme (approved):

PR #2762, opened 2023-03-20
Fix app name for prometheus-federation example

jeffreyc-splunk:


atoulme (approved):
I can reproduce. Approved.

PR #2764, opened 2023-03-21
fix #2753 add support for RHEL binary compatible distros to the install script

### LINKED ISSUE #2753: Add support for Rockylinux and Almalinux in the linux installer script splunk-otel-collector.sh
Would it be possible to add support for rockylinux and almalinux in the linux installer script: splunk-otel-collector.sh ?

Currently we get an error message when trying to install splunk-otel-collector on rockylinux

williake:
Added check for RHEL binary compatible distros to the installation script such that the splunk-otel-collector can be installed with the script on distros such as Almalinux, Rockylinux and Scientific Linux

### DIFF
--- a/internal/buildscripts/packaging/installer/install.sh
+++ b/internal/buildscripts/packaging/installer/install.sh
@@ -565,35 +565,35 @@ install() {
       ;;
     *)
       case "$distro_like" in
-      *rhel*)
-        if [ -z "$distro_version" ]; then
-          echo "The distribution version could not be determined" >&2
-          exit 1
-        fi
-        install_yum_package "libcap"
-        if [ "$skip_collector_repo" = "false" ]; then
-          install_collector_yum_repo "$stage"
-        fi
-        install_yum_package "splunk-otel-collector" "$collector_version"
-        if [ -n "$td_agent_version" ]; then
-          if [ "$skip_fluentd_repo" = "false" ]; then
-            install_td_agent_yum_repo "$td_agent_version"
+        *rhel*)
+          if [ -z "$distro_version" ]; then
+            echo "The distribution version could not be determined" >&2
+            exit 1
           fi
-          install_yum_package "td-agent" "$td_agent_version"
-          if command -v yum >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


internal/buildscripts/packaging/installer/install.sh
```diff
@@ -28,6 +28,13 @@ get_distro() {
   echo "$distro"
 }
 
+get_distro_like() {
```
jeffreyc-splunk, line 31: To avoid redundancy with the installation/uninstallation steps below, maybe we can just update the `get_distro` function for `ID_LIKE`?  Something like:
```
 get_distro() {
   local distro="$(. /etc/os-release 2>/dev/null && echo ${ID:-} || true)"
+  local distro_like="$(. /etc/os-release 2>/dev/null && echo ${ID_LIKE:-} || true)"
+  case "$distro_like" in
+    *rhel*)
+      distro="rhel"
+      ;;
+  esac
 
   echo "$distro"
 }
```

williake: Good idea! I agree with your proposed changes. There is less code that has to be changed and less duplicate code

aunshc:
@williake @jeffreyc-splunk my understanding is ScientificLinux is on maintenance support (ending in June 2024). To clarify the versions for RockyLinux and AlmaLinux, would 8.0+ be supported?

jeffreyc-splunk:
> @williake @jeffreyc-splunk my understanding is ScientificLinux is on maintenance support (ending in June 2024). To clarify the versions for RockyLinux and AlmaLinux, would 8.0+ be supported?

@aunshc The changes would support all RHEL-like distros 7 and newer.

jeffreyc-splunk:
> I'm absolutely leaning towards merging this but want to be careful.

I think we should wait for input from @aunshc about whether or not we want to "officially" support all RHEL-like distros.

aunshc:
We will be not moving forward with support with this PR at the moment based on existing business context and requirements. @williake please can you create an IDEA at [ideas.splunk.com](http://ideas.splunk.com/) for this request? Thanks!

Issue #2536, opened 2023-02-06
Support file inclusion in the config in the same way that community Collector does


tigrannajaryan:
The Otel Collector [allows including files](https://opentelemetry.io/docs/collector/configuration/) from the config, e.g.:

```yaml
receivers:
  otlp:
    protocols:
      grpc:

exporters: ${file:receivers.yaml}

service:
  extensions: [ ]
  pipelines:
    traces:
      receivers:  [ otlp ]
      processors: [  ]
      exporters:  [ otlp ]
```

This does not work with Splunk distro of the Collector.

rmfitzpatrick:
This is currently available with the include config source: https://github.com/signalfx/splunk-otel-collector/tree/main/internal/configsource/includeconfigsource

```
config_sources:
  include:

receivers:
  otlp:
    protocols:
      grpc:

exporters: ${include:exporters.yaml}
```

However, we should support this by reconciling the different config managers so that known schemes are forwarded to the registered file provider: https://github.com/signalfx/splunk-otel-collector/blob/f9bb21e8d82303fd42906e9553ca4b45c58d526f/cmd/otelcol/main.go#L97

tigrannajaryan:
It is good that we have an equivalent feature. I think it is also important to avoid diverging / to converge on both source code and how it is used in the config (e.g. `file` vs `include`).

rmfitzpatrick:
This should be resolved by https://github.com/signalfx/splunk-otel-collector/pull/2893

rmfitzpatrick:
Resolved by https://github.com/signalfx/splunk-otel-collector/commit/70b75c75a2d99f9c968fc1f2fec01ce5f0a6fa84

PR #2893, opened 2023-03-29
Refactor config sources into provider wrapper and support all service providers

rmfitzpatrick:
These changes get us closer to reconciling the original config sources and manager with the core resolver and providers that were based on them. In the process they remove unused/unnecessary fields and method requirements for the config source interfaces along with some general renaming to help with accuracy/maintainability. They also add support for existing service confmap.Provider's that would otherwise require settings declarations in our config without upstream equivalents, resolving https://github.com/signalfx/splunk-otel-collector/issues/2536.


tests/testutils/testcase.go
```diff
@@ -231,27 +231,34 @@ func (t *Testcase) ShutdownOTLPReceiverSink() {
 }
 
 func GetCollectorImageOrSkipTest(t testing.TB) string {
-	image := os.Getenv("SPLUNK_OTEL_COLLECTOR_IMAGE")
-	if strings.TrimSpace(image) == "" {
+	image := getCollectorImage()
```
rmfitzpatrick, line 234: I'm going to open another PR w/ the helper changes this branch uses.

rmfitzpatrick: https://github.com/signalfx/splunk-otel-collector/pull/2895

tests/general/testdata/resource_metrics/incompat_env_config_source_labels.yaml
```diff
@@ -10,21 +10,21 @@ resource_metrics:
               state: used
               single-dollar: an-envvar-value-suffix
               single-dollar-no-curly-braces: prefix-an-envvar-value
-              double-dollar: ${env:AN_ENVVAR}-suffix
```
rmfitzpatrick, line 13: Because of https://github.com/open-telemetry/opentelemetry-collector/pull/7055 I think this is now guaranteed to be expanded by the upstream provider. Existing compatibility persists in the non-`${env:..}` case.

internal/confmapprovider/configsource/provider.go
```diff
@@ -0,0 +1,165 @@
...
+	return factories
+}()
+
+type Hook interface {
+	OnNew()
+	OnRetrieve(scheme string, retrieved map[string]any)
+	OnShutdown()
+}
+
+type Provider interface {
```
pmcollins, line 57: Can we make this easier to understand for reviewers and maintainers? Looks like this interface is called Provider, its method is called Provider, and the method accepts a Provider and returns a Provider :). Do we need the interface -- it seems like the only implementer is ProviderWrapper (which is exported). And can we rename the `Provider` function to `WrapProvider` (or something) if that's what it does?

rmfitzpatrick: It's a `configsource.Provider` that takes a `confmap.Provider` and returns a `confmap.Provider`. I initially named the method `Wrap` but thought there were too many "wrap" prefixes so relied on the package name to help differentiate. The interface is helpful imo to help keep the surface area minimal while allowing changes to the implementation. Went ahead and unexported the implementation since the `confmap.Provider` is what's important. Also updated to use the clearer method name and added some doc comments.

pmcollins (approved):
Thanks for doing this.

PR #2894, opened 2023-03-30
New example: substring logs down to a specific length

atoulme:
This example showcases how to use the `Substring` ottl function in the transform processor to limit the length of a particular log record.


examples/otel-logs-truncate-splunk/README.md
```diff
@@ -0,0 +1,16 @@
+# Splunk HEC Example
```
rmfitzpatrick, line 1: What's the difference between the existing `splunk-hec-<...>` and `otel-logs-<...>` examples and can this field not be added to an existing w/ documentation so that we don't need an additional one?

atoulme: not really thought through this. Maybe those examples also need to give live elsewhere. @theletterf wdyt?

examples/otel-logs-truncate-splunk/otel-collector-config.yml
```diff
@@ -0,0 +1,58 @@
+receivers:
```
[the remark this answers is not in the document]
atoulme: sure thing done.

rmfitzpatrick (approved):

PR #2901, opened 2023-04-03
Add E2E test for zero-config for Windows IIS .NET apps

pjanotti:
Creates an E2E test to validate that the deployment of the SFx.NET via the install script is working for .NET Framework apps hosted on Windows IIS.


atoulme (approved):

pmcollins (approved):
Nice!

.github/workflows/build-and-test.yml
```diff
@@ -487,6 +487,62 @@ jobs:
...
+    runs-on: windows-2022
+    needs: [ windows-msi ]
+    steps:
+      - name: Check out the codebase.
+        uses: actions/checkout@v3
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
```
jeffreyc-splunk, line 500:
```suggestion
        uses: actions/setup-go@v4
        with:
          go-version: ${{ env.GO_VERSION }}
          cache-dependency-path: '**/go.sum'
```

.github/workflows/build-and-test.yml
```diff
@@ -487,6 +487,62 @@ jobs:
             throw "Failed to uninstall fluentdwinsvc."
           }
 
+  windows-zeroconfig-e2e-test:
+    runs-on: windows-2022
```
jeffreyc-splunk, line 491: Is there any value in also testing on `windows-2019`?

pjanotti: In principle no: there shouldn't be any difference affecting the versions shipped with the .NET instrumentation, nor the app itself. There is always a possibility of some difference, but, it is extremely unlikely in this case.

tests/zeroconfig/windows/README.md
```diff
@@ -0,0 +1,19 @@
+# Tests for Splunk Zero Configuration of .NET applications hosted on Windows IIS
+
+Tests in this directory validate that automatic instrumentation of
+.NET and .NET Framework applications hosted on Windows IIS are working under
+a deployment following the Zero Configuration procedures.
+
+## Requirements to run the tests locally
+
+The GitHub job `windows-zeroconfig-e2e-test` in the [`build-and-test`](../../../.github/workflows/build-and-test.yml)
```
rmfitzpatrick, line 9: Having all the setup in ci logic can make maintainability and debugging difficult. Would it be reasonable to migrate to a ps1 script and invoke there?

tests/zeroconfig/windows/testdata/expected_resource_metrics.yaml
```diff
@@ -0,0 +1,6 @@
+resource_metrics:
+  - attributes:
+    scope_metrics:
+      - metrics:
+          - name: otelcol_exporter_sent_spans
```
rmfitzpatrick, line 5: Do you mind adding all the expected metrics including labels? `<ANY>` can be used as a value for dynamic ones where we mainly care about the attr presence/key.

.github/workflows/build-and-test.yml
```diff
@@ -487,6 +487,63 @@ jobs:
             throw "Failed to uninstall fluentdwinsvc."
           }
 
+  windows-zeroconfig-e2e-test:
```
rmfitzpatrick, line 490: I'm not sure this needs to run w/ `build-and-test` and think a new zero config workflow could be added that is limited to `internal/buildscripts/packaging/msi` and `instrumentation` content changes.

.github/workflows/build-and-test.yml
```diff
@@ -487,6 +487,63 @@ jobs:
...
+
+      - name: Build Docker compose
+        working-directory: tests/zeroconfig/windows/testdata/
+        run: |
+          docker compose build
+
+      - name: Run the test
+        working-directory: tests/zeroconfig/windows/
+        run: |
+          go test -timeout 5m -run ^TestWindowsIISInstrumentation$ -tags zeroconfig -v
```
rmfitzpatrick, line 541: `-tags zeroconfig` could be used if added to the test file so we don't need to filter w/ `-run`

.github/workflows/build-and-test.yml
jeffreyc-splunk: Only downside to moving to a new workflow is that it would require duplicating the build for all upstream deps (otelcol, agent bundle, and msi).

Alternatively, maybe we can try something like https://github.com/marketplace/actions/paths-changes-filter to only run the job when there are relevant changes?

tests/zeroconfig/windows/testdata/expected_resource_metrics.yaml
pjanotti: > Do you mind adding all the expected metrics including attributes*?

Most of the metrics are independent of the .NET instrumentation, or do you mean only the ones expected after the request to the instrumented application is done?

.github/workflows/build-and-test.yml
pjanotti: If adding the 3rd party action is acceptable it sounds good to me - this action seems reputable. Related Q.: is this enough to guarantee that it runs at least once for each release?

tests/zeroconfig/windows/README.md
pjanotti: I personally prefer the separate action steps: typically you run a few of them and then repeat a specific one to debug/investigate. My guess is that the script will be run only in CI, for debug/investigation devs will copy and paste particular lines of the script. That said I will make the change so we can better judge it.

.github/workflows/build-and-test.yml
rmfitzpatrick: We could also run this for merges to main for general tracking coverage?

jeffreyc-splunk: If we want to use that action, I think we can add the action to the upstream `windows-msi` job and configure this job to always run for `main` or for relevant changes.  Maybe something like:
```
windows-msi:
  outputs:
    zeroconfig: ${{ steps.filter.outputs.zeroconfig }}
  steps:
      uses: actions/checkout@v3
    - uses: dorny/paths-filter@v2
      id: filter
      with:
        filters: |
          zeroconfig:
            - path1
            - path2
            - ...
```
and then
```
windows-zeroconfig-e2e-test:
  needs: [ windows-msi ]
  if: ${{ github.ref_name == 'main' || needs.windows-msi.outputs.zeroconfig == 'true' }}
  steps: ...
```

tests/zeroconfig/windows/testdata/expected_resource_metrics.yaml
pjanotti: Yes, this is an internal collector metric. The SFx.NET metrics are disabled by default so I didn't want to rely on them.

tests/zeroconfig/windows/README.md
pjanotti: After writing it, I liked it :) - let's see if it passes CI ...

tests/zeroconfig/windows/testdata/expected_resource_metrics.yaml
pjanotti: I have a story to add the resource spans to testutils, the plan is to switch this test to actually check for the spans.

tests/zeroconfig/windows/README.md
rmfitzpatrick: Thanks!

tests/zeroconfig/windows/testdata/expected_resource_metrics.yaml
rmfitzpatrick: Adding resource spans checks sounds great though it took me a bit to understand the evaluated flow here. You're installing two collectors, one for zero config w/ ingest url configured to be the signalfx receiver of the second collector, which is configured to report to the test's otlp sink?

tests/zeroconfig/windows/testdata/Dockerfile.pipeline.collector
```diff
@@ -0,0 +1,27 @@
...
+
+# WARNING: Passing access token as a build argument is to facilitate manual validation of this Dockerfile. It shouldn't be used in production.
+ARG verify_access_token=false
+ARG access_token=some_token
+ENV VERIFY_ACCESS_TOKEN=$verify_access_token
+ENV ACCESS_TOKEN_TMP=$access_token
+RUN `
+    $token = $Env:ACCESS_TOKEN_TMP; `
+    $collector_msi_path = (dir "c:\setup\splunk-otel-collector-*.msi").FullName; `
+    c:\setup\install.ps1 -access_token $token -msi_path $collector_msi_path -with_fluentd $false -ingest_url "http://10.1.1.1:4318"
```
rmfitzpatrick, line 21: mind configuring it to use a reserved/unused address in the off chance something leads it to report to the otlp sink directly?

```suggestion
    c:\setup\install.ps1 -access_token $token -msi_path $collector_msi_path -with_fluentd $false -ingest_url "http://192.0.2.1:12345"
```

tests/zeroconfig/windows/testdata/expected_resource_metrics.yaml
pjanotti: yes, I tried to document that on the test code, may require some re-wording to make that clear... see https://github.com/signalfx/splunk-otel-collector/pull/2901/files#diff-164dad5cd795a3f716eea69cefc15b2062e4d9d766cfa9ee103ae943716a40f1R41

.github/workflows/build-and-test.yml
pjanotti: I guess this is the last pending thing to merge this PR. I will take a look in the afternoon PT.

pjanotti: I opted to have a dedicated job instead of piggybacking on `windows-msi` since the job is very light weight, but, no strong opinion about that, see https://github.com/signalfx/splunk-otel-collector/pull/2901/commits/2ce7fb89dca76f1569d7968a2857d99b9877eeda

tests/zeroconfig/windows/windows_iis_test.go
```diff
@@ -0,0 +1,111 @@
...
+	// New-NetFirewallRule -DisplayName 'zc-iis-test' -Direction Inbound -LocalAddress 10.1.1.1 -LocalPort 4318 -Protocol TCP -Action Allow -Profile Any
+	//
+	// The command to remove the rule is:
+	//
+	// Remove-NetFirewallRule -DisplayName 'zc-iis-test'
+	//
+	// The OTLP also requires the Docker compose network to be created before it is initialized, since
+	// the address is part of that network.
+	//
+	otlp, err := testutils.NewOTLPReceiverSink().WithEndpoint("10.1.1.1:4318").Build()
```
rmfitzpatrick, line 68: Thanks for revealing a pain point. Ideally this would be possible w/ the testcase helpers but hardcoding the endpoint complicates it. Will be sure to add an option for this and update in the future.

.github/workflows/build-and-test.yml
```diff
@@ -487,6 +487,55 @@ jobs:
...
+      changed: ${{ steps.filter.outputs.changed }}
+    steps:
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
+          changed:
+            - 'internal/buildscripts/packaging/installer/install.ps1'
+            - 'internal/buildscripts/packaging/msi/**'
+            - 'pkg/signalfx-agent/bundle/**'
```
jeffreyc-splunk, line 506: Should the tests also be included?
```suggestion
            - 'pkg/signalfx-agent/bundle/**'
            - 'tests/zeroconfig/**'
```

pjanotti: Good point @jeffreyc-splunk pending...

rmfitzpatrick (approved):
Thanks!

jeffreyc-splunk (approved):

PR #2921, opened 2023-04-05
[lightprometheusreceiver] initial working draft

dloucasfx:
This is the initial working draft for `lightprometheusreceiver`


rmfitzpatrick:
Would it be better to update the simple prometheus receiver that we already contributed instead of adding a new one? With prometheus, simple prometheus and the prometheus-exporter monitor this would be the fourth prometheus receiver for scraping metric views.

rmfitzpatrick:
Please add an integration test. There's an existing one w/ internal metrics that could be used as a basis: https://github.com/signalfx/splunk-otel-collector/blob/main/tests/receivers/prometheus/internal_prometheus_test.go

internal/receiver/lightprometheusreceiver/scraper.go
```diff
@@ -0,0 +1,223 @@
...
+				dp := gauge.DataPoints().AppendEmpty()
+				dp.SetDoubleValue(fm.GetGauge().GetValue())
+				dp.SetTimestamp(now)
+				for _, l := range fm.GetLabel() {
+					if l.GetValue() != "" {
+						dp.Attributes().PutStr(l.GetName(), l.GetValue())
+					}
+				}
+			}
+		case dto.MetricType_HISTOGRAM, dto.MetricType_GAUGE_HISTOGRAM:
```
jinja2, line 172: Would we consider adding an option to this receiver to drop certain type of prom metrics?
I have a use case where I want to collect prom metrics from some k8s components but am not really interested in the histogram metrics because the buckets are generated dynamically and w/o native UI support for histogram in O11y, visualizing these is not a good experience. So I would just like to not collect them instead of adding exclusion rules for specific metrics.

dloucasfx: good point, yes we can and I think we should. @atoulme wdyt?

internal/receiver/lightprometheusreceiver/factory.go
```diff
@@ -0,0 +1,57 @@
...
+	"go.opentelemetry.io/collector/receiver/scraperhelper"
+)
+
+const typeStr = "lightprometheus"
+
+func NewFactory() receiver.Factory {
+	return receiver.NewFactory(
+		typeStr,
+		createDefaultConfig,
+		receiver.WithMetrics(createMetricsReceiver, component.StabilityLevelDevelopment),
```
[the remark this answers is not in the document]
atoulme: Yes, we could move it to alpha: https://github.com/open-telemetry/opentelemetry-collector#stability-levels

internal/receiver/lightprometheusreceiver/scraper.go
```diff
@@ -0,0 +1,223 @@
...
+				}
+				dp.SetSum(fm.GetSummary().GetSampleSum())
+				dp.SetCount(fm.GetSummary().GetSampleCount())
+				for _, l := range fm.GetLabel() {
+					if l.GetValue() != "" {
+						dp.Attributes().PutStr(l.GetName(), l.GetValue())
+					}
+				}
+			}
+		case dto.MetricType_UNTYPED:
```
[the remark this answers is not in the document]
atoulme: Not sure. Gauge is the most generic metric type and we should be good with it for now.

internal/receiver/lightprometheusreceiver/scraper.go
dloucasfx: right, filtering at the source will save us a few bytes... anything we can do save one bit of memory at a time :-)

atoulme:
> Would it be better to update the simple prometheus receiver that we already contributed instead of adding a new one? 
True, the aim is to test it here and move it upstream, and see if we can use simpleprometheusreceiver. Tough though as it changes the config.

internal/receiver/lightprometheusreceiver/scraper.go
atoulme: let's do it in a separate PR. Less is more.

atoulme: https://signalfuse.atlassian.net/browse/OTL-2009 is filed to follow up.

PR #2924, opened 2023-04-06
remove set library

atoulme:
This PR removes the gopkg.in/fatih/set.v0 library in favor of a switch statement.


dmitryax (approved):
Nice 👍

PR #2927, opened 2023-04-06
Improve error handling for windows installers

jeffreyc-splunk:
For the windows installer script and chocolatey installer:
- Add retries (max of 3) when starting/stopping the collector and fluentd services
- Check the exit code when installing MSI's
- Move service starts to the end, after all components have been installed
- Update failure messages
- Add new workflow to test windows installer script
- Update installation tests

Known issue:
- CI tests may occasionally fail when the installers try to start the collector service with our default configs, e.g. https://github.com/signalfx/splunk-otel-collector/actions/runs/4872776707/jobs/8691911641
  ```
  failed to start service: failed to start extensions: listen tcp 0.0.0.0:55679: bind: An attempt was made to 
  access a socket in a way forbidden by its access permissions.; failed to shutdown pipelines: no existing 
  monitoring routine is running; no existing monitoring routine is running; no existing monitoring routine is 
  running; no existing monitoring routine is running
  ```
  Still unsure if it's an environmental issue with github runners, or with the collector service itself.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -451,17 +451,29 @@ jobs:
           path: ./dist
 
       - name: Installation test
+        env:
+          token: fake-token
+          realm: fake-realm
+          memory: "256"
         run: |
           $ErrorActionPreference = 'Stop'
           Set-PSDebug -Trace 1
           $msi_path = Resolve-Path .\dist\splunk-otel-collector*.msi
           $env:VERIFY_ACCESS_TOKEN = "false"
-          .\internal\buildscripts\packaging\installer\install.ps1 -access_token "testing123" -realm "test" -msi_path "$msi_path" -mode "${{ matrix.MODE }}" -with_fluentd $${{ matrix.WITH_FLUENTD }}
+          .\internal\buildscripts\packaging\installer\install.ps1 -access_token "${{ env.token }}" -realm "${{ env.realm }}" -msi_path "$msi_path" -mode "${{ matrix.MODE }}" -memory "${{ env.memory }}" -with_fluentd $${{ matrix.WITH_FLUENTD }}
           Start-Sleep -s 30
-          powershell.exe -File .github\workflows\scripts\win-test-services.ps1 -mode "${{ matrix.MODE }}" -with_fluentd "${{ matrix.WITH_FLUENTD }}"
-          powershell.exe -File .github\workflows\scripts\win-test-support-bundle.ps1 -mode "${{ matrix.MODE }}" -with_fluentd "${{ matrix.WITH_FLUENTD }}"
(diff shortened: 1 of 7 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -399,13 +399,13 @@ jobs:
       - name: Installation test
         run: |
           $ErrorActionPreference = 'Stop'
-          Set-PSDebug -Trace 1
+          # Set-PSDebug -Trace 1
```
pjanotti, line 402: I don't know from the top of my head how to do it, but, instead of commenting out this line it may be worth making it use an input for the workflow, so it has a desired default for the automatic runs but you can provide a different value when running it manually.

.github/workflows/scripts/win-test-services.ps1
```diff
@@ -1,29 +1,46 @@
...
     [string]$mode = "agent",
+    [string]$access_token = "testing123",
+    [string]$realm = "test",
+    [string]$memory = "512",
     [string]$with_fluentd = "true"
 )
 
 $ErrorActionPreference = 'Stop'
-Set-PSDebug -Trace 1
+# Set-PSDebug -Trace 1
```
pjanotti, line 10: Similar suggestion to the CI workflow: use a parameter with the desired default.

internal/buildscripts/packaging/choco/splunk-otel-collector/tools/common.ps1
```diff
@@ -108,4 +132,24 @@ For more information execute:
...
 }
 
+function install_msi([string]$path) {
+    Write-Host "Installing $path ..."
+    $startTime = Get-Date
+    $proc = (Start-Process msiexec.exe -Wait -PassThru -ArgumentList "/qn /norestart /i `"$path`"")
+    if ($proc.ExitCode -ne 0 -and $proc.ExitCode -ne 3010) {
+        $err = "The installer failed with error code ${proc.ExitCode}."
+        try {
+            $events = (Get-WinEvent -ProviderName "MsiInstaller" | Where-Object { $_.TimeCreated -ge $startTime })
```
pjanotti, line 142: 🥇

pjanotti (approved):
Took a quick look since it was mentioned in the other PR that I was reviewing: LGTM. Feel free to ping me if you want another review when is not a draft anymore.

.github/workflows/build-and-test.yml
jeffreyc-splunk: Thanks.  I just commented it out as a temporary change to decrease the noise during my testing.  We can use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs to change this for manual runs, if necessary.

rmfitzpatrick (approved):

PR #2929, opened 2023-04-07
Fixing systemd metadata so sendLoadState, sendSubState and sendActiveState options work again

dloucasfx:


crobert-1 (approved):
I'm still unsure why `gauge.substate.running` doesn't have a `group`, but I left a comment on the [other PR](https://github.com/signalfx/signalfx-agent/pull/2778) so I won't duplicate it here.

Looks good otherwise!

dloucasfx:
> I'm still unsure why `gauge.substate.running` doesn't have a `group`, but I left a comment on the [other PR](https://github.com/signalfx/signalfx-agent/pull/2778) so I won't duplicate it here.
> 
> Looks good otherwise!

this is a default metric, the only metric that gets sent without any option enabled.

PR #2931, opened 2023-04-08
[SPIKE] Remove unused code

atoulme:
Remove code that is not actively used by the smart agent receiver from the vendored Smart Agent code, removing the dependency to the AWS SDK Go v1 and the signalfx-go libraries in particular.


rmfitzpatrick (changes requested):
monitorcodegen is used to generate monitor metric files and shouldn't be deleted

PR #2949, opened 2023-04-11
Update exception messages in install.ps1

theletterf:
Edited exception messages for style and clarity.

### DIFF
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -399,7 +399,7 @@ if ($with_dotnet_instrumentation) {
     Invoke-WebRequest -Uri $download -OutFile $dotnet_autoinstr_path -UseBasicParsing
     Import-Module $dotnet_autoinstr_path
     if (Get-IsSignalFxInstalled) {
-        throw "SignalFx Instrumentation for .NET is already installed. Remove or uninstall the SignalFx Instrumentation for .NET and rerun this script."
+        throw "SignalFx Instrumentation for .NET is already installed. Remove or uninstall SignalFx Instrumentation for .NET and rerun this script."
     }
 }
 
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -495,7 +495,7 @@ if ($config_path -Eq "") {
     if (Test-Path -Path "$old_config_path") {
         $config_path = $old_config_path
     } else {
-        throw "Valid Collector configuration file not found. Rerun the script with Administrator rights."
+        throw "Valid Collector configuration file not found."
     }
 }
 
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -176,7 +176,7 @@ function check_policy() {
     if ($executionRestricted) {
         throw @"
 You can't import or run scripts with execution policy $executionPolicy.
-Change you execution policy to RemoteSigned or similar:
+Change your execution policy to RemoteSigned or similar:
         PS> Set-ExecutionPolicy RemoteSigned
 For more information, run the following command:
         PS> Get-Help about_execution_policies
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -175,10 +175,10 @@ function check_policy() {
     $executionRestricted = ($executionPolicy -eq "Restricted")
     if ($executionRestricted) {
         throw @"
-Your execution policy is $executionPolicy, this means you will not be able import or use any scripts including modules.
-To fix this change you execution policy to something like RemoteSigned.
+You can't import or run scripts with execution policy $executionPolicy.
+Change you execution policy to RemoteSigned or similar:
         PS> Set-ExecutionPolicy RemoteSigned
-For more information execute:
+For more information, run the following command:
         PS> Get-Help about_execution_policies
 "@
     }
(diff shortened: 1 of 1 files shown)


internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -495,7 +495,7 @@ if ($config_path -Eq "") {
     if (Test-Path -Path "$old_config_path") {
         $config_path = $old_config_path
     } else {
-        throw "The installed splunk-otel-collector package does not include a supported config file!"
+        throw "Valid Collector configuration file not found. Rerun the script with Administrator rights."
```
theletterf, line 498: @pjanotti Would this be accurate as a message/solution?

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -175,10 +175,10 @@ function check_policy() {
     $executionRestricted = ($executionPolicy -eq "Restricted")
     if ($executionRestricted) {
         throw @"
-Your execution policy is $executionPolicy, this means you will not be able import or use any scripts including modules.
-To fix this change you execution policy to something like RemoteSigned.
+You can't import or run scripts with execution policy $executionPolicy.
+Change you execution policy to RemoteSigned or similar:
```
pmcollins, line 179:
```suggestion
Change your execution policy to RemoteSigned or similar:
```

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -399,7 +399,7 @@ if ($with_dotnet_instrumentation) {
     Invoke-WebRequest -Uri $download -OutFile $dotnet_autoinstr_path -UseBasicParsing
     Import-Module $dotnet_autoinstr_path
     if (Get-IsSignalFxInstalled) {
-        throw "SignalFx Instrumentation for .NET is already installed. Remove/Uninstall SignalFx Instrumentation for .NET and re-run this script."
+        throw "SignalFx Instrumentation for .NET is already installed. Remove or uninstall the SignalFx Instrumentation for .NET and rerun this script."
```
pmcollins, line 402:
```suggestion
        throw "SignalFx Instrumentation for .NET is already installed. Remove or uninstall SignalFx Instrumentation for .NET and rerun this script."
```

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -362,7 +362,7 @@ $ErrorActionPreference = 'Stop'; # stop on all errors
 # check administrator status
 echo 'Checking if running as Administrator...'
 if (!(check_if_admin)) {
-    throw 'You are not currently running this installation under an Administrator account.  Installation aborted!'
+    throw 'Installer is running without Administrator rights. Installation failed.'
```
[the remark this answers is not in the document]
pjanotti: Thanks @jeffreyc-splunk, it seems that we can switch to require later this year https://learn.microsoft.com/en-us/lifecycle/announcements/windows-server-2012-r2-end-of-support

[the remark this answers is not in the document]
pjanotti: I wonder what version is present on an up-to-date Windows Server 2012, but, not curious enough to check it...

internal/buildscripts/packaging/installer/install.ps1
pjanotti: Correct, in the case, at hand, the MSI fails to install but the error only bubbles up when checking for the config. I was guessing that MSI failure was due to the lack of rights, but, given the `check_if_admin` that doesn't seem to be the case.

internal/buildscripts/packaging/installer/install.ps1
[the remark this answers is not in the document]
jeffreyc-splunk: Not sure either.  According to https://learn.microsoft.com/en-us/powershell/scripting/install/powershell-support-lifecycle?view=powershell-7.3#powershell-end-of-support-dates, looks like Microsoft technically only supports 7.2 and newer right now.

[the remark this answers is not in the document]
pjanotti: 7.2 is for `pwsh` the portable Powershell on top of .NET - `Windows Powershell` life cycle is covered by the respective Windows that it was shipped with.

internal/buildscripts/packaging/installer/install.ps1
theletterf: Can we leave the suggested change as is then?

theletterf:
@pjanotti @jeffreyc-splunk Made some edits. PTAL!

internal/buildscripts/packaging/installer/install.ps1
jeffreyc-splunk: > While this was my initial suspicion for the case at hand, there is a check above if the script is running as admin or not, unless the check is incorrect it seems that this case was already accounted for. It will be better to validate the result of running the MSI, i.e.: instead of the current code
> 
> ```powershell
> echo "Installing $msi_path ..."
> Start-Process msiexec.exe -Wait -ArgumentList "/qn /norestart /i `"$msi_path`""
> echo "- Done"
> ```
> 
> the script should have something like: https://github.com/signalfx/signalfx-dotnet-tracing/blob/main/Splunk.SignalFx.DotNet.psm1#L52-L57

@pjanotti is correct.  Currently, the script does not fail even if the msi installation is unsuccessful (instead, the script fails at the config file check since it didn't get installed).  The script needs to be updated to capture and check the exit code from `Start-Process` as @pjanotti suggested, such that the config file check only fails if the msi was installed successfully but did not contain a valid config file (which I can do in a separate PR).

pjanotti: > Can we leave the suggested change as is then?

@theletterf I would rather leave the text without mentioning the admin rights and let @jeffreyc-splunk follow-up with a more general fix.
```suggestion
        throw "Valid Collector configuration file not found."
```

pjanotti (approved):
Just one minor suggestion from my side and then LGTM.

jeffreyc-splunk (approved):
LGTM with @pjanotti's suggestion.

Also, I'm testing some script changes in https://github.com/signalfx/splunk-otel-collector/pull/2927 to better handle errors.

theletterf:
@jeffreyc-splunk Can we merge?

PR #2951, opened 2023-04-11
Fix debian upgrade command in linux-installer.md

jeffreyc-splunk:


docs/getting-started/linux-installer.md
```diff
@@ -157,7 +157,7 @@ To upgrade the Collector, run the following commands on your system (requires
 - Debian:
   ```sh
   sudo apt-get update
-  sudo apt-get --only-upgrade splunk-otel-collector
+  sudo apt-get install --only-upgrade splunk-otel-collector
```
hughesjj-splunk, line 160: nit: Is it ubuntu or debian going for `apt` instead of `apt-get` now?

hughesjj-splunk: (Or are these flags only found in `apt-get`?)

PR #2954, opened 2023-04-12
install goimports w/ tools

rmfitzpatrick:
`make fmt` uses goimports but we don't currently provide it as a dependency.


internal/tools/go.mod
```diff
@@ -2,9 +2,16 @@ module github.com/signalfx/splunk-otel-collector/internal/tools
...
+	github.com/dgrijalva/jwt-go => github.com/form3tech-oss/jwt-go v3.2.5+incompatible
+	github.com/form3tech-oss/jwt-go => github.com/golang-jwt/jwt/v4 v4.4.3
+	github.com/gorilla/websocket => github.com/gorilla/websocket v1.4.1
+	golang.org/x/crypto => golang.org/x/crypto v0.5.0
+)
+
 require (
 	github.com/client9/misspell v0.3.4
-	github.com/golangci/golangci-lint v1.51.2
+	github.com/golangci/golangci-lint v1.52.2
```
rmfitzpatrick, line 14: I think this bump brought in linter updates so I've updated the required changes instead of disabling the check.

PR #2959, opened 2023-04-14
httpsinkexporter: add support for metrics and filtering

owais:
- now supports metrics datatype.
- also supports filtering by name and attributes (metrics only)
- query URL has changed from `/` to `/spans` and `/metrics`
- fixed some other subtle bugs/issues

### DIFF
--- a/internal/exporter/httpsinkexporter/factory.go
+++ b/internal/exporter/httpsinkexporter/factory.go
@@ -34,6 +34,7 @@ func NewFactory() exporter.Factory {
 	return exporter.NewFactory(
 		typeStr,
 		createDefaultConfig,
+		exporter.WithMetrics(createMetricsExporter, component.StabilityLevelDevelopment),
 		exporter.WithTraces(createTracesExporter, component.StabilityLevelDevelopment),
 	)
 }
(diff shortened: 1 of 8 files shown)


atoulme (approved):
Approved, I have reached out to the team for a second look however.

internal/exporter/httpsinkexporter/factory.go
```diff
@@ -31,10 +31,12 @@ const (
 
 // NewFactory creates a factory for httpsink exporter.
 func NewFactory() exporter.Factory {
+	sinkFactory := newSinkFactory()
```
rmfitzpatrick, line 34: Does this ensure there's only one sink regardless of desired endpoint? What if multiples are wanted?

```
httpsink/one:
  endpoint: localhost:12345
httpsink/two:
  endpoint: localhost:23456
```

owais: This ensures there is one sink (http server) per exporter instance. Your example will start two sink servers. This was introduced so both metrics and spans start only one server per exporter instance. In the previous implementation they'd have tried to start the server twice for a single exporter instance.

owais: You're right. There is a bug in this logic. Will fix shortly.

rmfitzpatrick (approved):

owais:
Thanks @rmfitzpatrick. All passing now.

PR #2964, opened 2023-04-14
Log endpoint id w/ discovery status

rmfitzpatrick:
These changes add the observer provided endpoint ids to discovery status logs. Also including a minor discovery mode observer creation setting change* to make sure the ID is set where the host observer expects* it.


pmcollins (approved):

rmfitzpatrick:
>         	Messages:   	Error, do this: mount -t proc proc /proc
I'm adding a skip check for arm for now. I've not tracked down what's up but for some reason ps through qemu doesn't work w/ our bundled tools, and I have a feeling it's related to our patchelf use.

PR #3004, opened 2023-04-21
discovery receiver: provide zap fields to regexp and expr

rmfitzpatrick:
In some cases the most helpful information from a logged statement is in its zap fields, which are currently unavailable to the statement evaluator. These changes add all fields not including `name`, `caller`, and `stacktrace` to the matched pattern for regexp and expr match types. Also includes adding an `ExprEnv` object for accessing fields that make invalid identifiers.

These changes are breaking for both regexp and expr. It's my opinion that these are ok to land without transitional guidance as the component is still in development.

regexp: Since the pattern is no longer just the message, `^$` start and end identifiers will likely no longer lead to matches. `message: ` prefixes can reasonably* replace `^`, and `$` will persist where no fields are logged.

expr: The env has been updated to use `message` instead of `msg`.

### DIFF
--- a/internal/receiver/discoveryreceiver/README.md
+++ b/internal/receiver/discoveryreceiver/README.md
@@ -1,7 +1,7 @@
 # Discovery Receiver
 
 | Status                   |                  |
-| ------------------------ | ---------------- |
+|--------------------------|------------------|
 | Stability                | [in-development] |
 | Supported pipeline types | logs             |
 | Distributions            | [Splunk]         |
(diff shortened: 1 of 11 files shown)


internal/receiver/discoveryreceiver/statement_evaluator.go
```diff
@@ -62,14 +63,22 @@ func newStatementEvaluator(logger *zap.Logger, id component.ID, config *Config,
...
-			exprEnv: func(pattern string) map[string]any {
-				return map[string]any{"msg": pattern}
-			},
-			id: id,
+			id:            id,
 		},
 		encoder: encoder,
 	}
+
+	se.evaluator.exprEnv = func(pattern string) map[string]any {
```
pmcollins, line 71: Looks like this is defining evaluator behavior in a foreign constructor using an anonymous function. Can we encapsulate the construction and behavior of `evaluator` in evaluator.go?

internal/receiver/discoveryreceiver/statement_evaluator.go
```diff
@@ -83,6 +76,16 @@ func newStatementEvaluator(logger *zap.Logger, id component.ID, config *Config,
 	return se, nil
 }
 
+// exprEnv will unpack logged statement message and field content for expr program use
+func (se *statementEvaluator) exprEnv(pattern string) map[string]any {
```
pmcollins, line 80: I guess I was thinking that since this method only uses the logger field of statementEvaluator it's not cohesive to this type. Instead, it's used to extend behavior of an evaluator, so perhaps it should live there? (And it could just be a function that accepts a logger.)

pmcollins (approved):

internal/receiver/discoveryreceiver/statement_evaluator.go
rmfitzpatrick: I may not be following your suggestion since the metric evaluator defines its own env function as is expected w/ derived types: https://github.com/signalfx/splunk-otel-collector/blob/e7064dc00394f18368c6bdcdf86083831be4761c/internal/receiver/discoveryreceiver/metric_evaluator.go#L62

internal/receiver/discoveryreceiver/README.md
```diff
@@ -290,45 +290,75 @@ Flags: 0
...
+| `watch_observers` (required) | []string                  | <no value> | The array of Observer extensions to receive Endpoint events from                                                                                                                                     |
+| `log_endpoints`              | bool                      | false      | Whether to emit log records for Observer Endpoint events                                                                                                                                             |
+| `embed_receiver_config`      | bool                      | false      | Whether to embed a base64-encoded, minimal Receiver Creator config for the generated receiver as a reported metrics `discovery.receiver.rule` resource attribute value for status log record matches |
+| `receivers`                  | map[string]ReceiverConfig | <no value> | The mapping of receiver names to their Receiver sub-config                                                                                                                                           |
 
 ### ReceiverConfig
```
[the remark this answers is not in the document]
rmfitzpatrick: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/configschema

Issue #3005, opened 2023-04-21
Fail the build if `make generate finds change`


atoulme:
Should we add a new github job/step to run `make generate-metrics` and fail if there are uncommitted changes, similar to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/build-and-test.yml#L184?

_Originally posted by @jeffreyc-splunk in https://github.com/signalfx/splunk-otel-collector/issues/2668#issuecomment-1518060538_

jeffreyc-splunk:
Added `generate-metrics` job to github workflow in #3007.

PR #3007, opened 2023-04-21
Add generate-metrics job

jeffreyc-splunk:
For #3005:
- Add job to run `make generate-metrics` and fail if there are changes
- Update Makefile to run `make fmt` after mdatagen to ensure changes are formatted
- Change the `gofmt` job to run `make fmt` instead of only `gofmt -s -w .`

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -68,9 +68,32 @@ jobs:
       - name: gofmt
         run: |
           make install-tools
-          gofmt -s -w .
+          make fmt
           if ! git diff --exit-code; then
-            echo "One or more Go files are not formatted correctly. Run 'gofmt -s -w .' and push the changes."
+            echo "One or more Go files are not formatted correctly. Run 'make fmt' and push the changes."
+            exit 1
+          fi
+
+  generate-metrics:
+    name: generate-metrics
+    # Use 20.04.5 until https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16450 is resolved
+    runs-on: ubuntu-20.04
+    needs: [setup-environment]
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -68,9 +68,32 @@ jobs:
...
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache-dependency-path: '**/go.sum'
+
+      - name: generate-metrics
+        run: |
+          make generate-metrics
+          if ! git diff --exit-code; then
```
rmfitzpatrick, line 95: fieldalignment updated the databricks receiver MetricsBuilder and this didn't fail.

jeffreyc-splunk: It originally did (see https://github.com/signalfx/splunk-otel-collector/actions/runs/4767265273/jobs/8475311980) before adding `make fmt` to the `generate-metrics` target.

rmfitzpatrick: So make fmt would then cause a git diff though right? If we aren't committing the fieldalignment changes I think I missing something.

jeffreyc-splunk: I think that `internal/receiver/databricksreceiver/internal/metadata/generated_metrics.go` currently in main is correctly formatted.  But if we run `make generate-metrics` without `make fmt`, it is generated unformatted and causes a diff.

rmfitzpatrick: sorry, I'm of course assuming our source wasn't fmt'ed.

rmfitzpatrick (approved):

PR #3008, opened 2023-04-21
[receiver/lightprometheus] Fix histogram translation

dmitryax:
Histograms are incorrectly translated to OTel schema and histogram buckets are dropped on signalfx exporter, so they don’t appear on the backend.

1. In the prometheus schema, each bucket is a cumulative value over lower buckets. In otel, it’s not the case

2. Otel schema expects number of bounds to be always equal to number of counts + 1

3. Light prometheus doesn’t set AggregationTemporality for histograms

This change fixes the these issues. See https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a6c7576fed18ad7287fc85504b9a33c665d20a04/receiver/prometheusreceiver/internal/metricfamily.go#L100 as reference for the translation in the prometheus receiver.


internal/receiver/lightprometheusreceiver/scraper_test.go
```diff
@@ -120,12 +120,7 @@ istio_requests_total{response_code="foo-bar-value",reporter="foo-bar-value",sour
...
 # TYPE istio_request_bytes histogram
-istio_request_bytes_bucket{response_code="200",reporter="foo-bar-value",source_workload="foo-bar-value",source_workload_namespace="foo-bar-value",source_principal="foo-bar-value",source_app="foo-bar-value",source_version="foo-bar-value",source_cluster="foo-bar-value",destination_workload="foo-bar-value",destination_workload_namespace="foo-bar-value",destination_principal="foo-bar-value",destination_app="foo-bar-value",destination_version="foo-bar-value",destination_service="foo-bar-value",destination_service_name="foo-bar-value",destination_service_namespace="foo-bar-value",destination_cluster="foo-bar-value",request_protocol="foo-bar-value",response_flags="foo-bar-value",grpc_response_status="foo-bar-value",connection_security_policy="unknown",source_canonical_service="foo-bar-value",destination_canonical_service="foo-bar-value",source_canonical_revision="foo-bar-value",destination_canonical_revision="foo-bar-value",le="0.5"} 0
```
rmfitzpatrick, line 123: Would additive changes be helpful rather than losing this coverage?

dmitryax: This change actually rather increases the coverage. This is a broken/cut histogram output missing the required `+Inf` bucket. So it has to be fixed. I reduced it to one bucket to cover that particular path additionally. There is another valid histogram in this file already.

atoulme (approved):

internal/receiver/lightprometheusreceiver/scraper_test.go
atoulme: Yes, that's a good call. Thanks!

PR #3040, opened 2023-05-01
update openshift deps and ignore undesired version in dependabot

rmfitzpatrick:
Need to ignore [v3.9.0](https://pkg.go.dev/github.com/openshift/api?tab=versions). Using a replace rule in the main go.mod since contrib has similarly accepted dependabot udpate.


atoulme (approved):

jvoravong (approved):

.github/dependabot.yml
```diff
@@ -60,13 +62,17 @@ updates:
...
+        versions: ["v3.9.0+incompatible"]
     schedule:
       interval: "weekly"
   - package-ecosystem: "gomod"
     directory: "/pkg/signalfx-agent"
     ignore:
       - dependency-name: "go.opentelemetry.io/*"
       - dependency-name: "github.com/open-telemetry/*"
+      - dependency-name: "github.com/openshift/*"
+        versions: ["v3.9.0+incompatible"]
```
jeffreyc-splunk, line 75: Should these changes be added to the 'gendependabot' target in the Makefile instead?

rmfitzpatrick: oh right, forgot about that...

jeffreyc-splunk: Shall I add a simple github job to fail if this file is manually changed?

rmfitzpatrick: Updated but I'm not sure how platform agnostic the changes are. imo we should adopt a templated approach in future changes.

PR #3047, opened 2023-05-01
fix flaky test by retrying and waiting if server does not respond in time

hughesjj:
Could consider moving this to an end-to-end test or similar as well

### DIFF
--- a/internal/receiver/prometheusremotewritereceiver/client_test.go
+++ b/internal/receiver/prometheusremotewritereceiver/client_test.go
@@ -74,7 +74,7 @@ func (prwc *MockPrwClient) SendWriteRequest(wr *prompb.WriteRequest) error {
 			break
 		}
 		if errors.Is(err, syscall.ECONNREFUSED) {
-			retry -= 1
+			retry--
 			time.Sleep(2 * time.Second)
 		}
 	}


internal/receiver/prometheusremotewritereceiver/client.go
```diff
@@ -65,8 +66,17 @@ func (prwc *MockPrwClient) SendWriteRequest(wr *prompb.WriteRequest) error {
 
 	ctx, cancel := context.WithTimeout(context.Background(), prwc.Timeout)
 	defer cancel()
-
-	err = prwc.Client.Store(ctx, compressed)
+	retry := 3
```
rmfitzpatrick, line 69: Can this be moved to a `_test.go` file since it's not intended to be used by the actual component?

internal/receiver/prometheusremotewritereceiver/client.go
```diff
@@ -65,8 +66,17 @@ func (prwc *MockPrwClient) SendWriteRequest(wr *prompb.WriteRequest) error {
...
+	for retry > 0 {
+		err = prwc.Client.Store(ctx, compressed)
+		if nil == err {
+			break
+		}
+		if strings.Contains(err.Error(), "connection refused") {
```
rmfitzpatrick, line 75: maybe something like
```suggestion
		if errors.Is(err, syscall.ECONNREFUSED) {
```

also should we silently and immediately retry if not?

rmfitzpatrick (approved):

atoulme (approved):

PR #3051, opened 2023-05-02
Switch Windows IIS zero-config test to check for traces instead of metrics

pjanotti:
Stop using indirect metrics to validate the Windows IIS zero-config deployment for .NET Framework applications. Now the test checks for the expected spans instead, see [expected spans file](https://github.com/signalfx/splunk-otel-collector/pull/3051/files#diff-721729486953b4987e7865221feabbbcb1ba055349169c999929bc238bc4283b).

This PR adds a `ResourceTraces` to `testutils\telemetry`. This type has a few differences from the other o11y signals (`metrics` and `logs`):

- It doesn't have a "relaxed match" mode: the use of `<ANY>` on the attributes suffices. This can become an issue if spans expected by the tests change frequently; if that happens, adding a relaxed matching for the span attributes can be useful.
- Spans have IDs that should be ignored by integration tests; as such, they were not added to the respective types. Validating parent-children relations hasn't been used frequently for this type of test.

Due to these differences, the types and tests for `ResourceTraces` omitted a few functions and methods that are present on the types for metrics and logs:

Not implemented methods/functions:

1. `FillDefaultValues`: no need to add version build to the traces.
2. `Validate()`: no need to validate the traces. Metrics has metric types to parse; logs always return `nil`.
3. No need at this time to implement the relaxed equals for spans, so there is no need for `Hash`, `Equals`, etc.

Not implemented the related tests, e.g.: `Equivalence`, `Consistency`, and `AllWithMissingAndEmptyAttributes`.

1. TestLogEquivalence
2. TestLogRelaxedEquivalence
3. TestLogAttributeRelaxedEquivalence
4. TestLogHashFunctionConsistency
5. TestFlattenResourceLogsConsistency
6. TestLogContainsAllWithMissingAndEmptyAttributes

### DIFF
--- a/tests/testutils/otlp_receiver_sink.go
+++ b/tests/testutils/otlp_receiver_sink.go
@@ -138,16 +138,7 @@ func (otlp *OTLPReceiverSink) Start() error {
 		return err
 	}
 
-	if err := (*otlp.metricsReceiver).Start(context.Background(), otlp.Host); err != nil {
-		return err
-	}
-
-	if err := (*otlp.tracesReceiver).Start(context.Background(), otlp.Host); err != nil {
-		_ = (*otlp.metricsReceiver).Shutdown(context.Background())
-		return err
-	}
-
-	return nil
+	return (*otlp.metricsReceiver).Start(context.Background(), otlp.Host)
 }
 
 func (otlp *OTLPReceiverSink) Shutdown() error {
@@ -155,15 +146,7 @@ func (otlp *OTLPReceiverSink) Shutdown() error {
 		return err
 	}
 
-	metricsShutdownErr := (*otlp.metricsReceiver).Shutdown(context.Background())
-	tracesShutdownErr := (*otlp.tracesReceiver).Shutdown(context.Background())
-	if metricsShutdownErr != nil {
-		return metricsShutdownErr
-	}
-	if tracesShutdownErr != nil {
-		return tracesShutdownErr
-	}
-	return nil
+	return (*otlp.metricsReceiver).Shutdown(context.Background())
 }
 
 func (otlp *OTLPReceiverSink) AllLogs() []plog.Logs {
--- a/tests/testutils/otlp_receiver_sink.go
+++ b/tests/testutils/otlp_receiver_sink.go
@@ -296,12 +296,12 @@ func (otlp *OTLPReceiverSink) AssertAllMetricsReceived(t testing.TB, expectedRes
 	return err
 }
 
-func (otlp *OTLPReceiverSink) AssertAllTracesReceived(t testing.TB, exptectedResourceTraces telemetry.ResourceTraces, waitTime time.Duration) error {
+func (otlp *OTLPReceiverSink) AssertAllTracesReceived(t testing.TB, expectedResourceTraces telemetry.ResourceTraces, waitTime time.Duration) error {
 	if err := otlp.assertBuilt("AssertAllTracesReceived"); err != nil {
 		return err
 	}
 
-	if len(exptectedResourceTraces.ResourceSpans) == 0 {
+	if len(expectedResourceTraces.ResourceSpans) == 0 {
 		return fmt.Errorf("empty ResourceTraces provided")
 	}
 
@@ -324,7 +324,7 @@ func (otlp *OTLPReceiverSink) AssertAllTracesReceived(t testing.TB, exptectedRes
 		receivedTraces = telemetry.FlattenResourceTraces(receivedResourceTraces)
 
 		var containsAll bool
-		containsAll, err = receivedTraces.ContainsAll(exptectedResourceTraces)
+		containsAll, err = receivedTraces.ContainsAll(expectedResourceTraces)
 		return containsAll
 	}, waitTime, 10*time.Millisecond, "Failed to receive expected traces")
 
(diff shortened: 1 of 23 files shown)


pjanotti:
One of the CI failures comes from the usage of `yaml.Marshal` as part of the string that is verified by the tests. The same pattern is used by the other resource telemetry signals in tests, initially I'm planning to do a quick fix just on the affected tests. It is desirable to revisit this pattern to all 3 signals and update them to avoid random failures. See https://github.com/signalfx/splunk-otel-collector/actions/runs/4856638013/jobs/8656548917?pr=3051#step:8:360

The second CI failure seems unrelated https://github.com/signalfx/splunk-otel-collector/actions/runs/4856638020/jobs/8656387708?pr=3051#step:4:324

tests/testutils/otlp_receiver_sink.go
```diff
@@ -249,3 +295,43 @@ func (otlp *OTLPReceiverSink) AssertAllMetricsReceived(t testing.TB, expectedRes
 
 	return err
 }
+
+func (otlp *OTLPReceiverSink) AssertAllTracesReceived(t testing.TB, exptectedResourceTraces telemetry.ResourceTraces, waitTime time.Duration) error {
```
rmfitzpatrick, line 299:
```suggestion
func (otlp *OTLPReceiverSink) AssertAllTracesReceived(t testing.TB, expectedResourceTraces telemetry.ResourceTraces, waitTime time.Duration) error {
```

tests/testutils/otlp_receiver_sink.go
```diff
@@ -125,14 +137,33 @@ func (otlp *OTLPReceiverSink) Start() error {
...
 		return err
 	}
-	return (*otlp.metricsReceiver).Start(context.Background(), otlp.Host)
+
+	if err := (*otlp.metricsReceiver).Start(context.Background(), otlp.Host); err != nil {
+		return err
+	}
+
+	if err := (*otlp.tracesReceiver).Start(context.Background(), otlp.Host); err != nil {
+		_ = (*otlp.metricsReceiver).Shutdown(context.Background())
```
rmfitzpatrick, line 146: I think the otlp receiver uses singletons so as long as the factory methods are called everything is hooked up w/ starting an individual, which is why there isn't an existing logs receiver start call.

tests/testutils/otlp_receiver_sink.go
```diff
@@ -125,14 +137,33 @@ func (otlp *OTLPReceiverSink) Start() error {
...
 }
 
 func (otlp *OTLPReceiverSink) Shutdown() error {
 	if err := otlp.assertBuilt("Shutdown"); err != nil {
 		return err
 	}
-	return (*otlp.metricsReceiver).Shutdown(context.Background())
+
+	metricsShutdownErr := (*otlp.metricsReceiver).Shutdown(context.Background())
+	tracesShutdownErr := (*otlp.tracesReceiver).Shutdown(context.Background())
```
rmfitzpatrick, line 159: similar note as above. If they change the receiver implementation we'd need to address it but may be worth avoiding for now.

tests/testutils/telemetry/traces.go
```diff
@@ -0,0 +1,257 @@
...
+import (
+	"bytes"
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v2"
+)
+
+// ResourceTraces is a convenience type for test helpers and assertions.
+type ResourceTraces struct {
```
rmfitzpatrick, line 26: It's a bit difficult for me not to bikeshed that this should be `ResourceSpans` given the yaml key name and existing convention (which would translate to `AssertAllSpansReceived` in the sink) but I'm ok w/ mirroring the traces envelope given the proto definitions.

tests/testutils/telemetry/traces_test.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Since expected-traces.yaml specify a single resource spans, they will never find resource spans from received.
+	containsAll, err = expected.ContainsAll(received)
+	require.False(t, containsAll)
+	require.Error(t, err)
+	// The function used to generate the string representation of the span doesn't guarantee the order of the fields
+	// test both possibilities to ensure the test passes in all platforms.
+	errMsg := err.Error()
+	containsExpectedMissingSpanMsg :=
+		strings.Contains(errMsg, "Missing Spans: [name: span_one\nattributes:\n  one_attr: one_value\n  two_attr: two_value\n]") ||
```
rmfitzpatrick, line 231: I'm under the impression yaml marshaling is sorted/deterministic?

pjanotti: At least not being Windows and macOS https://github.com/signalfx/splunk-otel-collector/actions/runs/4856638013/jobs/8656548917?pr=3051#step:8:360

tests/testutils/telemetry/traces.go
pjanotti: I had the same feeling but thought in following the proto definitions...

tests/testutils/otlp_receiver_sink.go
pjanotti: I wasn't aware of that, let me remove it and test it locally.

tests/testutils/otlp_receiver_sink.go
pjanotti: Fixed via 4fc2a2715278b83f01b0e268ba7682e61f376f61

tests/testutils/telemetry/traces_test.go
pjanotti: I mean the test was passing locally on my Windows machine but failed on the mac when I tried a local repro.

pjanotti: Fixed via 31f0a1c0e0122ada5a7ec4dabbdca9e21444ebbd thanks for insisting on this issue.

rmfitzpatrick: nice, wasn't sure if I was misreading or conflating w/ something else

PR #3052, opened 2023-05-02
relocate signalfx-agent to internal

rmfitzpatrick:


Makefile.Common
```diff
@@ -26,12 +26,13 @@ ALL_PKG_DIRS := $(shell $(GOCMD) list -f '{{ .Dir }}' ./... | sort)
...
                                 -not -path '*/third_party/*' \
                                 -not -path '*/local/*' \
-                                -not -path '*/pkg/signalfx-agent/*' \
+                                -not -path '*/internal/signalfx-agent/*' \
                                 -type f | sort)
 
 # All source code and documents. Used in spell check.
 ALL_SRC_AND_DOC := $(shell find $(ALL_PKG_DIRS) -name "*.md" -o -name "*.go" -o -name "*.yaml" \
                                 -not -path '*/third_party/*' \
+                                -not -path '*/internal/signalfx-agent/*' \
```
crobert-1, line 35: Why are we removing this from spell check? Was it already excluded?

atoulme: I think it was meant to be removed before, as per the ALL_SRC variable before. Given that this is internal, we won't publish this API anywhere, so it probably makes sense to stop trying to spellcheck it.

PR #1045, opened 2021-12-15
Add jmx receiver to components

rrupapara:
Adding the jmx receiver as component.

Jmx [jar](https://repo1.maven.org/maven2/io/opentelemetry/contrib/opentelemetry-jmx-metrics/) bundling has also been taken place as part of adopting jmx receiver in all the installation of splunk-otel-collector

Incorporated new methods in `tests/testutils/container.go` as part of auto dependency updates of `github.com/testcontainers/testcontainers-go` from `v0.11.1` to `v0.12.0`

All the test cases have been passed

PR #3054, opened 2023-05-02
Add JMX receiver support

crobert-1:
This is essentially using all of the logic from [an earlier PR to add the JMX receiver](https://github.com/signalfx/splunk-otel-collector/pull/1045), updated for current main branch. The purpose of this PR is to add the JMX receiver from the contrib repository to the Splunk distribution of the OpenTelemetry Collector.

Integration test is currently failing, I'm investigating. Posting now so it can get some earlier feedback.


internal/buildscripts/packaging/jmx-metric-gatherer-release.txt
```diff
@@ -0,0 +1 @@
+1.7.0-alpha
```
rmfitzpatrick, line 1: This is quite old: https://central.sonatype.com/artifact/io.opentelemetry.contrib/opentelemetry-jmx-metrics/1.25.1-alpha/versions

PR #3065, opened 2023-05-04
Discovery receiver and smartagent/postgres rule improvements

rmfitzpatrick:
These changes provide a new `append_pattern` option for discovery receiver statement evaluation to not require debug logging to see the originating statements and source them in the smartagent/postgres rules, with a couple of corrections that the feature led to.

They also update the default placeholder value to "splunk.discovery.default" to provide a bit more clarity.

### DIFF
--- a/internal/confmapprovider/discovery/bundle/bundle.d/receivers/smartagent-postgresql.discovery.yaml
+++ b/internal/confmapprovider/discovery/bundle/bundle.d/receivers/smartagent-postgresql.discovery.yaml
@@ -11,9 +11,9 @@ smartagent/postgresql:
       type: postgresql
       connectionString: 'sslmode=disable user={{.username}} password={{.password}}'
       params:
-        username: splunk.discovery
-        password: splunk.discovery
-      masterDBName: splunk.discovery
+        username: splunk.discovery.default
+        password: splunk.discovery.default
+      masterDBName: splunk.discovery.default
   status:
     metrics:
       successful:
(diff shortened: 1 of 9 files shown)


internal/confmapprovider/discovery/bundle/bundle.d/receivers/smartagent-postgresql.discovery.yaml
```diff
@@ -52,17 +61,12 @@ smartagent/postgresql:
...
           log_record:
             severity_text: info
+            append_pattern: true
             body: >-
               Make sure your PostgreSQL database has
               `shared_preload_libraries = 'pg_stat_statements'`
               in the postgresql.conf file and that
               `CREATE EXTENSION IF NOT EXISTS pg_stat_statements;`
               has been run for each database you would like to monitor.
-        - regexp: 'could not monitor postgresql server: failed to determine total_time column name'
```
rmfitzpatrick, line 61: This prefix occurs more frequently than I thought, including w/ the above unreachable network issue (definitely not warranting a partial status).

internal/confmapprovider/discovery/bundle/bundle.d/receivers/smartagent-postgresql.discovery.yaml.tmpl
```diff
@@ -21,26 +21,35 @@
             body: PostgreSQL receiver is working!
     statements:
       failed:
+        - regexp: 'connect: network is unreachable'
+          first_only: true
+          log_record:
+            severity_text: info
+            append_pattern: true
+            body: The container cannot be reached by the collector. Is it in a shared network?
```
theletterf, line 29:
```suggestion
            body: The container cannot be reached by the Collector. Make sure they're in the same network.
```

atoulme (approved):

PR #3068, opened 2023-05-05
Add JMX receiver support

crobert-1:
This change adds support for the JMX receiver, which is implemented in the contrib repo.

Note: This receiver is not currently tested in a containerized environment.


atoulme (approved):

rmfitzpatrick:
>Note: This receiver is not currently tested when the collector is running inside a container.

For clarity, I believe this is referring to the metric gatherer jar not being bundled in the collector image (or any of the installation paths) and we inherit the general usage requirement of the upstream collector: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/jmxreceiver#jar_path-default-optopentelemetry-java-contrib-jmx-metricsjar

rmfitzpatrick (approved):

PR #3105, opened 2023-05-10
Docs feedback

hughesjj:
Implement feedback from docs team

1. Using `metadata.Type` from `generated_status.go` (generated via `metadata.yaml` and `mdatagen`)
2. Rewording and clarification


crobert-1 (commented):
Don't want to be a bother here, but could we split this into two PRs? I'm thinking it may be good to split out the mdatagen related changes into its own PR.

theletterf (approved):

PR #3149, opened 2023-05-19
Add workflow to lint examples

jeffreyc-splunk:
Run `make lint` for PRs with go-related changes in `examples`.

### DIFF
--- a/.github/workflows/lint-examples.yml
+++ b/.github/workflows/lint-examples.yml
@@ -1,0 +1,42 @@
+name: lint-examples
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/lint-examples.yml'
+      - 'examples/**/*.go'
+      - 'examples/**/go.mod'
+      - 'examples/**/go.sum'
+
+concurrency:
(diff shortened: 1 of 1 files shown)


jeffreyc-splunk:
Failure is the same issue with the `signalfxgatewayprometheusremotewritereceiver` example that exists in main:
```
Error: fake-metrics-generator.go:42:11: ineffectual assignment to err (ineffassign)
	timeout, err := model.ParseDuration("2s")
	         ^
make: *** [../../Makefile.Common:96: lint] Error 1
```

dmitryax:
Why do we need a separate workflow for the examples? Can we reuse the existing `build-and-test`?

jeffreyc-splunk:
> Why do we need a separate workflow for the examples? Can we reuse the existing `build-and-test`?

My thought was that if a PR only has changes in `examples`, we won't need to run the entire build-and-test workflow, but can still catch issues before it is merged (like #3144 which is currently causing the lint job to fail for the main branch).

rmfitzpatrick:
lint error should be resolved by https://github.com/signalfx/splunk-otel-collector/pull/3167

Issue #3146, opened 2023-05-19
The customer is facing issues implementing the ability to include YAML file inside a another YAML file


salapatt:
https://github.com/signalfx/splunk-otel-collector/tree/main/internal/configsource/includeconfigsource#:~:text=The%20config%20source%20can%20be%20used%20to%20insert%20complete%20sections%20of%20the%20configuration%3A

When you include another yaml file, such as:
`${include:/etc/configs/exporters.yaml}`

Any env vars within that file are interpreted literally. So, we can't have:
`exporters: ${include:/etc/configs/exporters.yaml}`

where exporters.yaml is:
```
signalfx:
  access_token: "${SPLUNK_ACCESS_TOKEN}"
  api_url: "${SPLUNK_API_URL}"
  ingest_url: "${SPLUNK_INGEST_URL}"
  sync_host_metadata: true
```
Here, the collector will try hitting the api_url: "${SPLUNK_API_URL}" as opposed to what is actually stored in that env var.

dil-mlucas:
Thanks for the assist here Shibu. This was created from a splunk support case I had opened up, but I am afraid this is only part of the problem. I will outline the rest below.

There is no ability to include multiple yaml injections, or the steps to do so are not outlined in documentation. It would be very useful if you could do something like:

receivers: ${include:/etc/configs/receiversOne.yaml}, ${include:/etc/configs/receiversTwo.yaml}

But there is nothing in the docs that alludes to this capability.

Our final problem lies with the smartagent/telegraf/exec plugin. When attempting to have config such as:

smartagent/exec:
type: telegraf/exec
command: 'go run /opt/splunk/scripts/myGoScript.go'
intervalSeconds: 600
telegrafParser:
dataFormat: value
dataType: string

The collector fails to find golang in the $PATH of the system, even though it is clearly outlined. I suspect this is because telegraf/exec was written in golang using the os.exec package, which by default does not have awareness of $PATH or other env vars. We need the ability for telegraf/exec to understand env/path vars, whether that be a flag we can set in the plugin, or if it is rewritten with different use of the os.exec package.

rmfitzpatrick:
Thanks for the report @salapatt and @dil-mlucas. Unfortunately glob style imports in the include config source aren't currently supported and we're investigating a desired approach to be able to land the functionality. It's possible that improving the file config provider per https://github.com/open-telemetry/opentelemetry-collector/issues/7177 would be preferable.

[...]

PR #3168, opened 2023-05-22
Add missing expand config converter

rmfitzpatrick:
The core default settings use the expand converter, which we should support to avoid the first part of https://github.com/signalfx/splunk-otel-collector/issues/3146 on recursive value loading.

These changes also update the file confmap provider and include config source tests to exercise both forms (and adopt current helper practices for the latter).


atoulme (approved):

atoulme (approved):

dil-mlucas:
Hello @rmfitzpatrick and thanks for the work here. If I am understanding correctly, you were not able to reproduce the issues I faced with the include config source and environment variables?

rmfitzpatrick:
@dil-mlucas I closed this issue because it has unintended consequences and requires a more exacting change in the config resolution process. Using `"${env:ENV_VAR}"` is a workaround in the meantime.

PR #3170, opened 2023-05-22
[configsource/include] add support to pattern match and multiple include files

dloucasfx:
I am not sure if this breaks other "include" use cases, opening this PR for initial review and if this is something acceptable I will add a bunch of test cases.

### DIFF
--- a/internal/configsource/includeconfigsource/source.go
+++ b/internal/configsource/includeconfigsource/source.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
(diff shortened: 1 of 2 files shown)


rmfitzpatrick:
Please add an integration test to https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/configsources/include_test.go that exercises these changes. You could use the changes from https://github.com/signalfx/splunk-otel-collector/pull/3168 as a basis.

dloucasfx:
actually, this does not work as we use template and with multiple files, we only end up using the first parsed file (root).
Closing PR

PR #3175, opened 2023-05-23
evaluate --set property values as yaml

rmfitzpatrick:
These changes will unmarshal `--set` property values as yaml. This is to match the existing core service functionality of converting `--set` flags to yaml, though without the yaml confmap provider directives we don't currently include.

### DIFF
--- a/internal/configconverter/overwrite_properties.go
+++ b/internal/configconverter/overwrite_properties.go
@@ -21,11 +21,13 @@ package configconverter
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/knadh/koanf/maps"
 	"github.com/magiconair/properties"
 	"go.opentelemetry.io/collector/confmap"
+	"gopkg.in/yaml.v3"
 )
 
 type converter struct {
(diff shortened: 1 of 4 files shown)


jvoravong (approved):

internal/configconverter/overwrite_properties.go
```diff
@@ -57,7 +59,12 @@ func (c *converter) Convert(_ context.Context, conf *confmap.Conf) error {
 	// Create a map manually instead of using properties.Map() to not expand the env vars.
 	parsed := make(map[string]interface{}, props.Len())
 	for _, key := range props.Keys() {
-		value, _ := props.Get(key)
+		var value any
+		if raw, ok := props.Get(key); ok {
```
crobert-1, line 63: Will `ok` ever be false here? I assume the answer is no, since we're iterating over the keys then getting the value for each. If it's always going to be true we can remove the check.

I know it may be "safer" to keep it, but I was a bit concerned about setting the `parsed[key] = value` when value hasn't been updated and is just `any` now, instead of any empty string like it would have been before. This is an irrelevant concern since there will always be a value.

crobert-1 (approved):

internal/configconverter/overwrite_properties.go
rmfitzpatrick: What do we gain from ignoring the contains bool and treating the default empty string as the value for the property if it wasn't actually in the properties map (in the off chance that the properties parser/Keys method led to this flow)? Since we don't only want string representations of the values I'd think a nil default makes more sense.

crobert-1: I don't think we actually gain anything from using an empty string vs. the nil default. I was more just pointing out that the new `if` makes the change a little less clear than it is, and isn't really necessary. 

It's fine to keep it with me, it just made me overthink the change.

rmfitzpatrick: We'll likely want to adopt the yaml confmap provider and incorporate an equivalent translation from https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/flags.go#L49 so it's temporary (and will drop the overkill dep).

PR #3176, opened 2023-05-23
Apply config converters to `--dry-run` content.

rmfitzpatrick:
These changes fix a bug where config converter updates aren't propagated to* `--dry-run` content such that the reported config doesn't match that provided to the actual collector service.

~requires https://github.com/signalfx/splunk-otel-collector/pull/3175~

### DIFF
--- a/internal/configconverter/discovery.go
+++ b/internal/configconverter/discovery.go
@@ -77,7 +77,10 @@ func getServiceExtensions(out map[string]any) (map[string]any, []any, error) {
 	service := map[string]any{}
 	var serviceExtensions []any
 	if s, hasService := out["service"]; hasService && s != nil {
-		service = s.(map[string]any)
+		var ok bool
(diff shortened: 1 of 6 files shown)


theletterf (approved):

internal/configconverter/discovery.go
```diff
@@ -77,7 +77,10 @@ func getServiceExtensions(out map[string]any) (map[string]any, []any, error) {
 	service := map[string]any{}
 	var serviceExtensions []any
 	if s, hasService := out["service"]; hasService && s != nil {
-		service = s.(map[string]any)
+		var ok bool
+		if service, ok = s.(map[string]any); !ok {
```
rmfitzpatrick, line 81: Though not directly related to the fix, this issue was revealed during testing and is on the same path.

jinja2 (approved):

PR #3179, opened 2023-05-24
Migrate the prometheus monitor to use pmetric data directly

atoulme:
This changes the prometheus-exporter monitor to directly convert data to pmetric data instead of datapoints, which are then converted to pmetric data. As result, the monitor uses less memory.


rmfitzpatrick:
I think bypassing the sfx datapoint conversion removes support for the existing filtering logic and would be considered a breaking change. Can we do something like adding and documenting a new config option that defaults to false so this is opt in w/ known caveats?

internal/signalfx-agent/pkg/monitors/prometheusexporter/conversion.go
```diff
@@ -24,66 +21,108 @@ func counterExtractor(m *dto.Metric) float64 {
...
+		dp := sum.DataPoints().AppendEmpty()
+		dp.SetDoubleValue(e(m))
+		for i := range m.Label {
+			dp.Attributes().PutStr(m.Label[i].GetName(), m.Label[i].GetValue())
+		}
+		dp.Attributes().PutStr("system.type", "prometheus-exporter")
+	}
+}
+
+func makeSummaryDatapoints(sm pmetric.ScopeMetrics, name string, ms []*dto.Metric) {
```
rmfitzpatrick, line 75: I'm not sure these or histograms have existing test coverage. Updating the [prometheus-exporter integration test](https://github.com/signalfx/splunk-otel-collector/tree/main/tests/receivers/smartagent/prometheus-exporter) to include them (and exercise both suggested paths) would be helpful in vetting the changes.

pkg/receiver/smartagentreceiver/output.go
```diff
@@ -195,6 +196,15 @@ func (out *output) Copy() types.Output {
 	return &cp
 }
 
+func (out *output) SendMetrics(metrics pmetric.Metrics) {
+	if out.nextMetricsConsumer == nil {
+		return
+	}
+	ctx := out.reporter.StartMetricsOp(context.Background())
+	err := out.nextMetricsConsumer.ConsumeMetrics(context.Background(), metrics)
```
rmfitzpatrick, line 204: the monitor configs have an `extra_dimensions` field that should be applied to these

internal/signalfx-agent/pkg/monitors/prometheusexporter/conversion.go
atoulme: Yes, I'll create a separate PR to add an integration test with histograms, and then rebase this PR to use the same integration test and ensure backwards compatibility.

pkg/receiver/smartagentreceiver/output.go
atoulme: I'll create an integration test that uses `extra_dimensions`, and use it to see if we can drive backwards compatibility, similar to what we have for histograms.

atoulme:
We still can go a bit further on this one, and use the pprof extension to compare memory usage in tests. That would help test memory usage over time.

atoulme:
Some manual testing:
Before:
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O     PIDS
c592fe864131   otelcollector   182.82%   676.8MiB / 9.729GiB   6.79%     531MB / 1.31MB   65.5kB / 0B   14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O     PIDS
c592fe864131   otelcollector   137.58%   839.9MiB / 9.729GiB   8.43%     549MB / 1.37MB   65.5kB / 0B   14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O     PIDS
c592fe864131   otelcollector   137.58%   839.9MiB / 9.729GiB   8.43%     549MB / 1.37MB   65.5kB / 0B   14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O     PIDS
c592fe864131   otelcollector   152.88%   787.9MiB / 9.729GiB   7.91%     568MB / 1.44MB   65.5kB / 0B   14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O     PIDS
c592fe864131   otelcollector   152.88%   787.9MiB / 9.729GiB   7.91%     568MB / 1.44MB   65.5kB / 0B   14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O         BLOCK I/O     PIDS


After:
b80982607e00   otelcollector   147.53%   108.6MiB / 9.729GiB   1.09%     585MB / 1.62MB   0B / 0B     14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O   PIDS
b80982607e00   otelcollector   149.49%   119.2MiB / 9.729GiB   1.20%     605MB / 1.68MB   0B / 0B     14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O   PIDS
b80982607e00   otelcollector   149.49%   119.2MiB / 9.729GiB   1.20%     605MB / 1.68MB   0B / 0B     14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O   PIDS
b80982607e00   otelcollector   146.83%   128.2MiB / 9.729GiB   1.29%     625MB / 1.73MB   0B / 0B     14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O   PIDS
b80982607e00   otelcollector   146.83%   128.2MiB / 9.729GiB   1.29%     625MB / 1.73MB   0B / 0B     14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O   PIDS
b80982607e00   otelcollector   167.24%   308.5MiB / 9.729GiB   3.10%     636MB / 1.77MB   0B / 0B     14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O   PIDS
b80982607e00   otelcollector   167.24%   308.5MiB / 9.729GiB   3.10%     636MB / 1.77MB   0B / 0B     14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O   PIDS
b80982607e00   otelcollector   149.49%   223.2MiB / 9.729GiB   2.24%     655MB / 1.82MB   0B / 0B     14
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT     MEM %     NET I/O          BLOCK I/O   PIDS
b80982607e00   otelcollector   149.49%   223.2MiB / 9.729GiB   2.24%     655MB / 1.82MB   0B / 0B     14

internal/signalfx-agent/pkg/monitors/prometheusexporter/conversion.go
```diff
@@ -117,3 +186,10 @@ func labelsToDims(labels []*dto.LabelPair) map[string]string {
 	}
 	return dims
 }
+
+func applyDimensions(attrs pcommon.Map, extraDimensions map[string]string) {
```
rmfitzpatrick, line 190: Can this be moved to the output method like is done for `SendDatapoints` to avoid the requirement of adding to all monitors? I think adding the missing filtering pass there is also still required unless I'm missing it.

atoulme: Yes, I'll do that, that should be easier to repurpose indeed.

atoulme: Done.

rmfitzpatrick: To clarify I'm requesting that the existing functionality for the sfx data be added to your new `SendMetrics` method for compatibility and to avoid having to add the logic in monitors: https://github.com/signalfx/splunk-otel-collector/blob/main/pkg/receiver/smartagentreceiver/output.go#L205

internal/signalfx-agent/pkg/monitors/prometheusexporter/prometheus.go
```diff
@@ -133,30 +134,37 @@ func (m *Monitor) Configure(conf *Config) error {
 
 	var ctx context.Context
 	ctx, m.cancel = context.WithCancel(context.Background())
+	applyDimensions := func(attrs pcommon.Map) {
+		attrs.PutStr("system.type", m.monitorName)
+		for k, v := range conf.ExtraDimensions {
+			attrs.PutStr(k, v)
+		}
+	}
```
rmfitzpatrick, line 142: Mind shifting this to smartagentreceiver/output.go so all monitors pick it up for free?

internal/signalfx-agent/pkg/monitors/prometheusexporter/conversion.go
```diff
@@ -24,66 +22,110 @@ func counterExtractor(m *dto.Metric) float64 {
...
 	// upper bound value
 	case dto.MetricType_HISTOGRAM:
-		return makeHistogramDatapoints(*mf.Name, mf.Metric)
+		makeHistogramDatapoints(sm, *mf.Name, mf.Metric, applyDimensions)
 	default:
-		return nil
 	}
 }
 
-func makeSimpleDatapoints(name string, ms []*dto.Metric, dpf dpFactory, e extractor) []*datapoint.Datapoint {
```
rmfitzpatrick, line 49: Does it make more sense to have flagged, additive changes so that rolling back can occur in the same version (or this is opt-in). Moving everything in one go can set users up for poor experience.

atoulme: Yes, it does. A feature gate would make the most sense. I can do this, I didn't think this changeset would get so involved to be honest.

internal/signalfx-agent/pkg/monitors/prometheusexporter/conversion.go
```diff
@@ -24,66 +22,110 @@ func counterExtractor(m *dto.Metric) float64 {
...
+	metric := sm.Metrics().AppendEmpty()
+	metric.SetName(name)
+	sum := metric.SetEmptySum()
+	sum.SetAggregationTemporality(pmetric.AggregationTemporalityCumulative)
+	sum.SetIsMonotonic(true)
+	for _, m := range ms {
+		dp := sum.DataPoints().AppendEmpty()
+		dp.SetDoubleValue(e(m))
+		for i := range m.Label {
+			dp.Attributes().PutStr(m.Label[i].GetName(), m.Label[i].GetValue())
```
rmfitzpatrick, line 70: You can get the attributes reference once and mutate instead for forming intermediate types w/ each call.

internal/signalfx-agent/pkg/utils/filter/filter.go
```diff
@@ -21,6 +21,7 @@ type StringFilter interface {
 // StringMapFilter matches against the values of a map[string]string.
 type StringMapFilter interface {
 	Matches(map[string]string) bool
+	MatchesAny(map[string]any) bool
```
rmfitzpatrick, line 24: Rather than using raw maps, what about accepting `pcommon.Map` so we can the `Range` and value `AsString()` helpers?

internal/signalfx-agent/pkg/monitors/prometheusexporter/conversion.go
atoulme: Is the PR now in line with your request above? Just making sure I am understanding correctly.

atoulme:
OK, this is still very much WIP work, and I think I can peel off more things:
* Make a separate PR for the integration test with filters (might test multiple filters)
* Make a separate PR for filters with pmetric, to make the changes here a bit saner to review, and might need to introduce more test coverage to make sure we don't break anything
* Add the output feature gate with pmetric vs datapoint for the prometheusexporter specifically.

internal/signalfx-agent/go.mod
```diff
@@ -124,6 +124,7 @@ require (
 	github.com/kr/pretty v0.3.1
 	github.com/signalfx/signalfx-agent/pkg/apm v0.0.0-00010101000000-000000000000
 	github.com/smartystreets/goconvey v1.8.0
+	go.opentelemetry.io/collector/pdata v1.0.0-rc2
```
dmitryax, line 127: Let's use the latest version `v1.0.0-rcv0012`?

internal/signalfx-agent/pkg/monitors/prometheusexporter/prometheus.go
```diff
@@ -133,30 +134,37 @@ func (m *Monitor) Configure(conf *Config) error {
...
 
-func fetchPrometheusMetrics(fetch fetcher) ([]*datapoint.Datapoint, error) {
+func fetchPrometheusMetrics(fetch fetcher, applyDimensions func(attrs pcommon.Map)) (pmetric.Metrics, error) {
+	metrics := pmetric.NewMetrics()
 	metricFamilies, err := doFetch(fetch)
 	if err != nil {
-		return nil, err
+		return metrics, err
 	}
+	sm := metrics.ResourceMetrics().AppendEmpty().ScopeMetrics().AppendEmpty()
```
dmitryax, line 162: We probably shouldn't do this for empty `metricFamilies` Let's return early if that's the case?

internal/signalfx-agent/pkg/core/dpfilters/overridable.go
```diff
@@ -47,3 +49,23 @@ func (f *overridableDatapointFilter) Matches(dp *datapoint.Datapoint) bool {
...
+	case pmetric.MetricTypeGauge:
+		attributes = m.Gauge().DataPoints().At(0).Attributes()
+	case pmetric.MetricTypeSummary:
+		attributes = m.Summary().DataPoints().At(0).Attributes()
+	case pmetric.MetricTypeSum:
+		attributes = m.Sum().DataPoints().At(0).Attributes()
+	case pmetric.MetricTypeHistogram:
+		attributes = m.Histogram().DataPoints().At(0).Attributes()
+	case pmetric.MetricTypeExponentialHistogram:
+		attributes = m.ExponentialHistogram().DataPoints().At(0).Attributes()
```
dmitryax, line 65: Is this expected that we drop the whole metric only if the first data point matches? It doesn't look like the case for the existing behavior and probably should not be changed this way. 

The filtering logic is probably a bit more complicated. We cannot just pass `pmetric.Metric` and return a `bool`. We need to be able to mutate the `pmetric.Metric` and drop some datapoints.

I'd suggest we merge this change without filtering logic under a disabled feature gate first. Then we handle the filtering and enable the feature gate. WDYT?

cc @rmfitzpatrick

rmfitzpatrick: Good call that the implementations can't be 1:1. I'm ok w/ adding the deeper filtering here or tabling until a later pass w/ clear disclaimer in release notes. Either way the feature should be gated, though my preference is for a monitor config update that's false by default to avoid a global config mechanism but can see it either way.

dmitryax: If we table it, my preference would be not to introduce any support for filtering config under the feature gate (we can fail to start if this config is used). As the next step, we introduce the correct filtering, then enable the feature gate. This filtering solution doesn't seem right to give it to users even under a feature gate.

rmfitzpatrick: To clarify I was referring to adding the config option and removing the filtering in this PR (maybe failing out if filter fields are used like you suggest) and then adding filtering support in a later one. I wouldn't suggest using the current implementation w/ its known issues.

internal/signalfx-agent/go.mod
atoulme: Done.

internal/signalfx-agent/pkg/monitors/prometheusexporter/prometheus.go
atoulme: Done.

atoulme:
Closing as old. We may want to delete more smart agent code to simplify this work.

PR #3184, opened 2023-05-25
Add a test for histograms with the smart agent receiver prometheus-exporter monitor

atoulme:
This is just a test change, no functionality of the collector is changed. It adds an integration test to ensure we maintain backwards compatibility in how we manage histogram conversion from prometheus metrics.


tests/receivers/smartagent/prometheus-exporter/prometheus_exporter_test.go
```diff
@@ -18,13 +18,24 @@
 package tests
 
 import (
+	"path"
 	"testing"
 
 	"github.com/signalfx/splunk-otel-collector/tests/testutils"
 )
 
+var httpd = []testutils.Container{testutils.NewContainer().WithContext(
```
rmfitzpatrick, line 27: nit, mind adding to the test directly since not shared?

tests/receivers/smartagent/prometheus-exporter/testdata/httpd/metrics
```diff
@@ -0,0 +1,46 @@
+# HELP istio_agent_cert_expiry_seconds The time remaining, in seconds, before the certificate chain will expire. A negative    value indicates the cert is expired.
+# TYPE istio_agent_cert_expiry_seconds gauge
+istio_agent_cert_expiry_seconds{resource_name="default"} 7449418.02275500029
```
rmfitzpatrick, line 3: I can look into fixing values equivalence checks to detect deviations (unless you'd like to in your translation progress?). Since these are static values it would hopefully be trivial to add them to the resource metrics.

rmfitzpatrick (approved):

tests/receivers/smartagent/prometheus-exporter/testdata/httpd/metrics
atoulme: That’d be useful, sure! I don’t mean to impose on you though.

tests/receivers/smartagent/prometheus-exporter/prometheus_exporter_test.go
atoulme: Sure thing.

atoulme: Fixed.

PR #3189, opened 2023-05-26
Replace deprecated logging exporter parameter

theletterf:
The config files in the repo are still using the deprecated `loglevel: debug` for the `logging` exporter. This setting should now be `verbosity: detailed` as per https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/loggingexporter/README.md

### DIFF
--- a/deployments/ansible/molecule/custom_vars/custom_collector_config.yml
+++ b/deployments/ansible/molecule/custom_vars/custom_collector_config.yml
@@ -24,7 +24,7 @@ processors:
 
 exporters:
   logging/debug:
-    verbosity: detailed
+    loglevel: debug
   logging/info:
     loglevel: info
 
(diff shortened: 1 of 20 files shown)


atoulme (approved):

deployments/ansible/molecule/custom_vars/custom_collector_config.yml
```diff
@@ -24,7 +24,7 @@ processors:
 
 exporters:
   logging/debug:
-    loglevel: debug
+    verbosity: detailed
```
jeffreyc-splunk, line 27: Let's revert this since this file is only used for testing installation of an older version of the collector that doesn't support this change.

jeffreyc-splunk (changes requested):

atoulme (approved):

jeffreyc-splunk (approved):

atoulme:
Merging as I don't think @theletterf has merge rights. Thanks @theletterf!

PR #3232, opened 2023-06-05
pin corrected opentelemetry-collector-contrib/pkg/stanza

rmfitzpatrick:
Incorporates the filelog receiver bug fix


dmitryax (approved):
Thanks @rmfitzpatrick!

PR #3235, opened 2023-06-06
tests: adopt signalfx-go 1.32.0 

rmfitzpatrick:
Adopts signalfx-go 1.32.0 and signalfx-go/signalflow/v2 w/ https://github.com/signalfx/signalfx-go/pull/202. Includes minor e2e test changes as the only consumer.


atoulme (approved):

atoulme:
Maybe 20s could be wrapped with a t.Eventually, but that's ok, it's better than before.

rmfitzpatrick:
> Maybe 20s could be wrapped with a t.Eventually, but that's ok, it's better than before.

Good call. I'd have to update the assertion helpers to be able to do that and can do so with the next end to end update that breaks out a central secrets/fixture helper for another case.

PR #3238, opened 2023-06-09
support splunk.discovery mapping in properties.discovery.yaml

rmfitzpatrick:
These changes introduce a new mapped form for discovery.properties.yaml content:

```yaml
splunk.discovery:
  extensions:
    observer_type:
      enabled: false
  receivers:
    receiver_type/receiver_name:
      enabled: false
      config:
        field: value
```

The properties in this form are compatible with the existing `--set` style ones in the same file* and have a lower sourcing precedence to account for collisions. Invalid properties loaded from the file will be ignored and logged as warnings so as not to interfere with collector startup.

In order to reuse an existing `mergeMaps` util it was migrated to a new internal package.


theletterf (approved):
I like this. Seems simpler!

internal/confmapprovider/discovery/README.md
```diff
@@ -145,6 +164,7 @@ SPLUNK_DISCOVERY_EXTENSIONS_k8s_observer_ENABLED=false
 The priority order for discovery config content from lowest to highest is:
```
theletterf, line 164: This doesn't include bundle.d on purpose?

rmfitzpatrick: Good call, added.

internal/confmapprovider/discovery/properties/conf_test.go
```diff
@@ -0,0 +1,115 @@
...
+			path:          filepath.Join(".", "testdata", "invalid-mapping-unknown-component.yaml"),
+		},
+		{
+			expectedError: `invalid property "splunk.discovery.123" (parsing error): splunk.discovery:1:18: unexpected token "123" (expected ("receivers" | "extensions") <dot> ComponentID <dot> (("config" <dot>) | "enabled") (<string> | <dot> | <forwardslash>)*); invalid property "splunk.discovery.processors.a_processor.config.some_field" (parsing error): splunk.discovery:1:18: unexpected token "processors" (expected ("receivers" | "extensions") <dot> ComponentID <dot> (("config" <dot>) | "enabled") (<string> | <dot> | <forwardslash>)*); invalid property "splunk.discovery.receivers.a_receiver.unknown.some_field" (parsing error): splunk.discovery:1:57: unexpected token "<EOF>" (expected <dot> (("config" <dot>) | "enabled") (<string> | <dot> | <forwardslash>)*); invalid property "splunk.discovery.receivers.another_receiver/with-name.another_unknown.some_other_field" (parsing error): splunk.discovery:1:87: unexpected token "<EOF>" (expected <dot> (("config" <dot>) | "enabled") (<string> | <dot> | <forwardslash>)*)`,
```
jvoravong, line 85: Maybe add a comment about how this test expects to work.

rmfitzpatrick: renamed the test to be more descriptive

PR #3258, opened 2023-06-12
render commented bundle.d to default linux config.d

rmfitzpatrick:
These changes add `--commented` and `--dir` options to the discoverybundler to render commented-out discovery config templates to another directory and do so for a new default config.d directory at cmd/otelcol/config/collector/config.d.linux to be sourced in installers in subsequent PRs.

The aim of these changes is for discovery mode users to more easily alter bundle.d rules by being able to modify their content that have higher priority than the bundled resources if using this directory as the default config-dir (will be the case once adopted in the packages and installers).


jvoravong (approved):

cmd/otelcol/config/collector/config.d.linux/extensions/k8s-observer.discovery.yaml
```diff
@@ -0,0 +1,6 @@
+#####################################################################################
+# This file is generated by the Splunk Distribution of the OpenTelemetry Collector. #
+#####################################################################################
+# k8s_observer:
+#   enabled: true
+#   auth_type: serviceAccount
```
rmfitzpatrick, line 6: Just realized that observer properties aren't nested in `config` mappings, which I'll update in a follow-up PR for consistency.

PR #3259, opened 2023-06-12
Update signalfx-go in used in agent generate call

rmfitzpatrick:
Also adds pipefail to avoid false negatives: https://github.com/signalfx/splunk-otel-collector/actions/runs/5247407672/jobs/9477595540#step:4:2289


jeffreyc-splunk (approved):
Thanks!

PR #3267, opened 2023-06-14
kubernetes-cluster: sync node labels as properties on node name dimension

jinja2:
Adds a configuration to the smartagent/kubernetes-cluster receiver to optional sync node labels as properties on the kubernetes_node dimension.

### DIFF
--- a/internal/signalfx-agent/pkg/monitors/kubernetes/cluster/monitor.go
+++ b/internal/signalfx-agent/pkg/monitors/kubernetes/cluster/monitor.go
@@ -60,6 +60,9 @@ type Config struct {
 	// with a value of `0` corresponding to "False", `1` to "True", and `-1`
 	// to "Unknown".
 	NodeConditionTypesToReport []string `yaml:"nodeConditionTypesToReport" default:"[\"Ready\"]"`
+	// If set to true, the kubernetes_node dimension, in addition to the kubernetes_node_uid dimension, will get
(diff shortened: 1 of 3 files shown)


atoulme (approved):

internal/signalfx-agent/pkg/monitors/kubernetes/cluster/monitor.go
```diff
@@ -60,6 +60,9 @@ type Config struct {
 	// with a value of `0` corresponding to "False", `1` to "True", and `-1`
 	// to "Unknown".
 	NodeConditionTypesToReport []string `yaml:"nodeConditionTypesToReport" default:"[\"Ready\"]"`
+	// If set to true, the kubernetes_node dimension, in addition to the kubernetes_node_uid dimension, will get
+	// properties about each respective node synced to it. Do not enable this, if node names in the cluster are reused.
+	AddPropertiesNodeName bool `yaml:"addPropertiesNodeName" default:"false"`
```
rmfitzpatrick, line 65: readability nits or something to this affect? I think "logDimensionUpdates" may be the only precedent.

```suggestion
	// If set to true the `kubernetes_node` dimension, in addition to the `kubernetes_node_uid` dimension, will get
	// properties about each respective node synced to it. Do not enable this if node names in the cluster are reused (can lead to colliding or stale properties).
	AddPropertiesNodeName bool `yaml:"updatesForNodeDimension" default:"false"`
```

rmfitzpatrick (approved):

PR #3268, opened 2023-06-14
Use pointers to atomic ints instead of atomic ints

hughesjj:
Thanks @rmfitzpatrick for the catch, going to see if this fixes it.

Still would love to know why this doesn't reproduce for me locally....


atoulme (approved):

rmfitzpatrick:
I think this fixes a portion of the flake, but the main issue from what I can see is in the WaitGroup and test-only ready method that could be addressed in another PR.

rmfitzpatrick (approved):

PR #3271, opened 2023-06-15
feat: filelog receiver config equivalent to Splunk Addon for Unix and…

weliasz:
… Linux File and Directory Inputs

### DIFF
--- a/cmd/otelcol/config/collector/full_config_linux.yaml
+++ b/cmd/otelcol/config/collector/full_config_linux.yaml
@@ -622,6 +622,6 @@ service:
       processors: [memory_limiter, batch]
       exporters: [signalfx]
     logs:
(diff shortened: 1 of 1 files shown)


cmd/otelcol/config/collector/full_config_linux.yaml
```diff
@@ -580,6 +622,6 @@ service:
       processors: [memory_limiter, batch]
       exporters: [signalfx]
     logs:
-      receivers: [fluentforward, otlp]
+      receivers: [fluentforward, otlp, filelog]
```
atoulme, line 625: I would recommend that we revert this line change for now and land the rest of the contribution.
```suggestion
      receivers: [fluentforward, otlp]
```

atoulme (approved):

PR #3276, opened 2023-06-16
check for nil before coercing to string

pmcollins:
A customer referenced an empty file in their SA config and translatesfx panicked because that resulted in a nil being coerced to string.  This change tests the unmarshaled variable for nil before attempting to coerce it to a string.


pmcollins:
Failed check is unrelated to this change.

rmfitzpatrick (approved):

pmcollins:
Thanks for the review, Ryan.

PR #3303, opened 2023-06-22
Correct prometheus normalization package bump

rmfitzpatrick:
Also adding integration tests that use metrics from a quarkus demo app.


dmitryax (approved):
Thank you, @rmfitzpatrick!

PR #3304, opened 2023-06-22
Add GHA workflow to poll for new agent

breedx-splk:
Rather than the agent having to push PRs to every repo that might depend on it, let's turn that around and have repos that need to look for new agents do that themselves.

This checks for a new java agent version every 4 hours and, if found, updates the `java-agent-release.txt` file and creates a pull request.

A similar (and slightly more complicated) version exists in the helm repo now as well.

### DIFF
--- a/.github/workflows/update_java_agent.yaml
+++ b/.github/workflows/update_java_agent.yaml
@@ -1,0 +1,40 @@
+name: Check for new java agent
+
+on:
+  schedule:
+    - cron: "45 */4 * * *"
+  workflow_dispatch:
+
+env:
+  RELEASE_FILE: instrumentation/packaging/java-agent-release.txt
+  LATEST_API: https://api.github.com/repos/signalfx/splunk-otel-java/releases/latest
+
+jobs:
+  maybe_update_java:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.5.3
+      - name: swizzle version
+        id: swizzle_version
+        run: |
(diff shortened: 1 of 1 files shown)


jeffreyc-splunk:
We actually have something similar in our gitlab pipeline to update our openjdk dependency that runs daily, see:
https://github.com/signalfx/splunk-otel-collector/blob/main/.gitlab-ci.yml#L137
https://github.com/signalfx/splunk-otel-collector/blob/main/.gitlab/update-openjdk.sh

We can use this to handle new java agent releases.

breedx-splk:
> We actually have something similar in our gitlab pipeline to update our openjdk dependency that runs daily, see: https://github.com/signalfx/splunk-otel-collector/blob/main/.gitlab-ci.yml#L137 https://github.com/signalfx/splunk-otel-collector/blob/main/.gitlab/update-openjdk.sh
> 
> We can use this to handle new java agent releases.

Yeah, [seems familiar](https://github.com/signalfx/splunk-otel-java/blob/main/scripts/propagate-version.sh#L64). 😉 

So are you saying that you don't want this change because it is inconsistent with the existing update scriptery? Do you want to take on that work then? I'd really like to remove the push from the java agent release code....

jeffreyc-splunk:
I can add it our gitlab pipeline.  You can remove the push from the java agent repo.

breedx-splk:
> I can add it our gitlab pipeline. You can remove the push from the java agent repo.

Great, thanks!

PR #3324, opened 2023-06-29
Remove fluentbit extension from config yaml

### LINKED ISSUE #2472: Cannot use deprecrated fluentbit extension
[full_config_linux.yaml](https://github.com/signalfx/splunk-otel-collector/blob/6893428573f475a3f0b3cef282f39ac0a1a15022/cmd/otelcol/config/collector/full_config_linux.yaml#L491) needs to be updated. It is still showing how to use fluentbit but when trying to use it, it doesn't work. Per [6721](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/6721), it has been deprecated.

```
Jan 13 10:35:06 techsummit otelcol[70887]: 2023/01/13 10:35:06 main.go:111: application run finished with error: failed to get config: cannot unmarshal the configuration: 1 error(s) decoding:
Jan 13 10:35:06 techsummit otelcol[70887]: * error decoding 'extensions': unknown extensions type: "fluentbit" for id: "fluentbit" (valid values: [memory_ballast ecs_task_observer docker_observer http_forwarder k8s_observer smartagent zpages ecs_observer health_check file_storage host_observer pprof])
```

atoulme:
Fixes #2472

### DIFF
--- a/cmd/otelcol/config/collector/full_config_linux.yaml
+++ b/cmd/otelcol/config/collector/full_config_linux.yaml
@@ -485,20 +485,6 @@ exporters:
 ###############################################################################
 extensions:
 
-  # Enables the fluentbit extension
-  # Full configuration here: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/fluentbitextension
-  # NOTE: These settings should be updated prior to use
-  fluentbit:
-    executable_path: /usr/src/fluent-bit/build/bin/fluent-bit
-    tcp_endpoint: 127.0.0.1:8006
-    config: |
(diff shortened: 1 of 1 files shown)


dmitryax:
I wonder why we need this config file at all. I don't see a lot of value in it. If it's already broken, we can just remove it from the distro instead of keep fixing it going forward while no one is using it.

dmitryax:
I see this config is referenced in the docs once. So we'd need to update the docs as well. We can probably handle that later and merge this fix for now

dmitryax (approved):

PR #3329, opened 2023-06-29
expand postgres discovery rules

rmfitzpatrick:
These changes improve the smartagent/postgresql discovery rules to not rely on just the default port and also include a partial metric match that will occur if the receiver can reach a pg instance without having the required extension enabled.


atoulme:
Those changes also impact nginx discovery rules to make the regexp a bit more stringent. Changes look good to me.

atoulme (approved):

PR #3330, opened 2023-06-29
Split the logs pipeline in the default configuration

dmitryax:
Split the logs pipeline in the default configuration in a way that profiling data is always sent to Splunk Observability endpoint while other logs can be sent to another hec endpoint configured with `SPLUNK_HEC_URL` and `SPLUNK_HEC_TOKEN` environment variables.

### DIFF
--- a/cmd/otelcol/config/collector/ecs_ec2_config.yaml
+++ b/cmd/otelcol/config/collector/ecs_ec2_config.yaml
@@ -129,12 +129,18 @@ exporters:
     access_token: "${SPLUNK_ACCESS_TOKEN}"
     realm: "${SPLUNK_REALM}"
     correlation:
-  # Logs + Profiling
+  # Logs
   splunk_hec:
     token: "${SPLUNK_HEC_TOKEN}"
     endpoint: "${SPLUNK_HEC_URL}"
     source: "otel"
     sourcetype: "otel"
+    profiling_data_enabled: false
+  # Profiling
+  splunk_hec/profiling:
+    token: "${SPLUNK_ACCESS_TOKEN}"
+    endpoint: "${SPLUNK_INGEST_URL}/v1/log"
+    log_data_enabled: false
 
 service:
   extensions: [health_check, http_forwarder, zpages, memory_ballast]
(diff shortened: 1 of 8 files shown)


atoulme (approved):

cmd/otelcol/config/collector/ecs_ec2_config.yaml
```diff
@@ -129,12 +129,20 @@ exporters:
...
     endpoint: "${SPLUNK_HEC_URL}"
     source: "otel"
     sourcetype: "otel"
+    profiling_data_enabled: false
+  # Profiling
+  splunk_hec/profiling:
+    token: "${SPLUNK_ACCESS_TOKEN}"
+    endpoint: "${SPLUNK_INGEST_URL}/v1/log"
+    source: "otel"
+    sourcetype: "otel"
```
rmfitzpatrick, line 144: are these used by the backend?

rmfitzpatrick (approved):

cmd/otelcol/config/collector/ecs_ec2_config.yaml
dmitryax: Good question. I asked the profiling team

theletterf (changes requested):
All example files should be updated as well.

cmd/otelcol/config/collector/ecs_ec2_config.yaml
gsmirnov-splk: While @vovencij is on vacation, I can confirm that we are not using these fields in any way on the server side.

aunshc: If these are optional fields and not used for profiling, I suggest we remove them from the default configuration of this dedicated profiling hec exporter.

dmitryax: Removed these fields from the profiling exporters

dmitryax:
> All example files should be updated as well.

Updated examples where it makes sense

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -129,7 +138,11 @@ service:
       receivers: [prometheus/internal]
       processors: [memory_limiter, batch, resourcedetection/internal]
       exporters: [signalfx/internal]
-    logs:
-      receivers: [otlp, signalfx]
+    logs/signalfx:
+      receivers: [signalfx]
       processors: [memory_limiter, batch]
       exporters: [signalfx]
+    logs:
```
rmfitzpatrick, line 145: great fix

rmfitzpatrick (approved):

jvoravong (approved):

PR #3332, opened 2023-06-29
expand collectd/mysql discovery rules

rmfitzpatrick:
Similar to https://github.com/signalfx/splunk-otel-collector/pull/3329 these changes don't rely on mysql's default port and expand discovery rules to support more options.


atoulme (approved):
LGTM

PR #3350, opened 2023-07-06
Update databricks receiver metrics to add units, and move to mdatagen 0.81.0

atoulme:


internal/tools/go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/golangci/golangci-lint v1.53.3
 	github.com/google/addlicense v1.1.1
 	github.com/jstemmer/go-junit-report v1.0.0
-	github.com/open-telemetry/opentelemetry-collector-contrib/cmd/mdatagen v0.80.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/cmd/mdatagen v0.81.0
```
rmfitzpatrick, line 17: I think the resulting go.sum changes need to be included for the installation step to work

atoulme: sorry, fixed.

jinja2 (approved):

internal/receiver/databricksreceiver/metadata.yaml
```diff
@@ -215,181 +221,212 @@ metrics:
...
+    unit: By
     gauge:
       value_type: double
     attributes:
       [cluster.id, spark.app.id, pipeline.id, pipeline.name]
   databricks.spark.executor_metrics.major_gc.count:
     enabled: true
-    description: n/a
+    description: Total major GC count. For example, the garbage collector is one of MarkSweepCompact, PS MarkSweep, ConcurrentMarkSweep, G1 Old Generation and so on.
+    unit: "{gc}"
```
rmfitzpatrick, line 291: fwiw I see in the spark receiver they use `{gc_operation}`. I'd think the plural form of that could preferable but think `{gc}` is ok too since it's shorter.

internal/receiver/databricksreceiver/metadata.yaml
```diff
@@ -215,181 +221,212 @@ metrics:
...
+    unit: By
     gauge:
       value_type: double
     attributes:
       [cluster.id, spark.app.id, pipeline.id, pipeline.name]
   databricks.spark.executor_metrics.process_tree.jvm_rss.memory:
     enabled: true
-    description: n/a
+    description: "Resident Set Size: number of pages the process has in real memory. This is just the pages which count toward text, data, or stack space. This does not include pages which have not been demand-loaded in, or which are swapped out. Enabled if spark.executor.processTreeMetrics.enabled is true."
+    unit: "{pages}"
```
rmfitzpatrick, line 381: interesting they don't get the bytes and would require you to know the page size.

rmfitzpatrick (approved):

internal/receiver/databricksreceiver/metadata.yaml
atoulme: maybe I interpreted the docs wrong, fwiw.

internal/receiver/databricksreceiver/metadata.yaml
atoulme:
```suggestion
    unit: "{gc_operations}"
```

atoulme:
Going to merge this and I will need to follow up on the hadoop tests as they have become too flaky.

PR #3351, opened 2023-07-06
Reorder start of server to call cancel after closing server

atoulme:
Small change to ensure we follow to some extent the logic of the Shutdown method.

### DIFF
--- a/internal/receiver/signalfxgatewayprometheusremotewritereceiver/receiver.go
+++ b/internal/receiver/signalfxgatewayprometheusremotewritereceiver/receiver.go
@@ -74,16 +74,19 @@ func (receiver *prometheusRemoteWriteReceiver) Start(ctx context.Context, host c
 		Host:               host,
 		Parser:             newPrometheusRemoteOtelParser(),
 	}
-	ctx, receiver.cancel = context.WithCancel(ctx)
-	server, err := newPrometheusRemoteWriteServer(cfg)
-	if err != nil {
-		return err
-	}
-	if nil != receiver.server {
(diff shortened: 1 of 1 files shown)


internal/receiver/signalfxgatewayprometheusremotewritereceiver/receiver.go
```diff
@@ -74,17 +74,20 @@ func (receiver *prometheusRemoteWriteReceiver) Start(ctx context.Context, host c
...
+	if receiver.server != nil {
 		err := receiver.server.close()
 		if err != nil {
 			return err
 		}
 	}
+	if receiver.cancel != nil {
+		receiver.cancel()
+	}
+	ctx, receiver.cancel = context.WithCancel(ctx)
```
rmfitzpatrick, line 86: I believe this should be based on the background context (but in actuality it is going to be that looking at the service flow): https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/component.go#L41

>	// In that case make sure that the long-running operation does not use the context passed
	// to Start() function since that context will be cancelled soon and can abort the long-running
	// operation. Create a new context from the context.Background() for long-running operations.

rmfitzpatrick (approved):

internal/receiver/signalfxgatewayprometheusremotewritereceiver/receiver.go
atoulme: I know this was copied from another receiver, most receivers take the approach of tying to the original context. Probably worth enforcing through tests in contrib.

PR #3352, opened 2023-07-06
Unexport more from signalfxprometheusremotewritereceiver

atoulme:
Limit the Go API exposed by the component to as little as possible.

### DIFF
--- a/internal/receiver/signalfxgatewayprometheusremotewritereceiver/client_test.go
+++ b/internal/receiver/signalfxgatewayprometheusremotewritereceiver/client_test.go
@@ -30,12 +30,12 @@ import (
 	"github.com/prometheus/prometheus/storage/remote"
 )
 
-type MockPrwClient struct {
+type mockPrwClient struct {
 	Client  remote.WriteClient
 	Timeout time.Duration
 }
 
-func NewMockPrwClient(addr string, path string, timeout time.Duration) (MockPrwClient, error) {
+func newMockPrwClient(addr string, path string, timeout time.Duration) (mockPrwClient, error) {
 	URL := &config.URL{
 		URL: &url.URL{
 			Scheme: "http",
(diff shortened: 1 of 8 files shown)


internal/receiver/signalfxgatewayprometheusremotewritereceiver/prometheus_to_otel.go
```diff
@@ -129,19 +120,6 @@ func (prwParser *prometheusRemoteOtelParser) scaffoldNewMetric(ilm pmetric.Scope
 	return nm
 }
 
-// addBadRequests is used to report write requests with invalid data
-func (prwParser *prometheusRemoteOtelParser) addBadRequests(ilm pmetric.ScopeMetrics, start time.Time, end time.Time) {
```
rmfitzpatrick, line 133: was this deemed unnecessary for bw compatibility?

atoulme: Amazingly once I unexported more structs, lint started to complain that this particular code wasn't executed.

rmfitzpatrick: Maybe because line 66 was removed accidentally during the adjustment?

atoulme: No it was removed first and this method went next.

Here is the problem as explained through lint: lint once all methods were unexported was able to ascertain that the method `AddSfxCompatibilityMetrics` with the constant parameters 0, 0 and 0.

Those parameters govern how many missing data points, invalid data points and nan data points were reported as part of the export.

I set to work and made sure to create tests that added at least one missing data point (no name on the metric) and nan data point (NaN value). I however could not find a way to report an invalid data point.

That's because the invalid counter is never incremented. It is defined here: https://github.com/signalfx/splunk-otel-collector/pull/3352/files#diff-21fa25669c776099992651cf01d4f1de4164e767ccd0e4ea448994bde9a81f83L50 and yet never used, so it will always report zero. I could not find a way to change that state, so I instead removed the counter, the metric, and the testing methods around it.

atoulme: I'm going to nix my last commit and we can discuss next week what best action to take.

rmfitzpatrick (approved):

PR #3369, opened 2023-07-11
OTL-2195: Disable fluentd by default in installer scripts

jeffreyc-splunk:


rmfitzpatrick:
I think we'll need an equivalent filelog receiver addition in our agent config for this to not be so breaking. That would then be problematic for windows and when running as a container so can see that three agent configs would be necessary.

jeffreyc-splunk:
> I think we'll need an equivalent filelog receiver addition in our agent config for this to not be so breaking. That would then be problematic for windows and when running as a container so can see that three agent configs would be necessary.

Based on discussions, disabling fluentd installation by default does not require new/updated configs for the filelog receiver.

rmfitzpatrick (approved):

docs/getting-started/linux-installer.md
```diff
@@ -183,14 +185,21 @@ To upgrade the Collector, run the following commands on your system (requires
 
 ### Fluentd Configuration
 
-> If log collection is not required, run the installer script with the
-> `--without-fluentd` option to skip installation of Fluentd and the
-> plugins/dependencies listed below.
+If log collection with Fluentd is required and your platform is [supported](
```
theletterf, line 188: Is required by whom?

docs/getting-started/linux-installer.md
```diff
@@ -183,14 +185,21 @@ To upgrade the Collector, run the following commands on your system (requires
...
 
-By default, the Fluentd service will be installed and configured to forward log
-events with the `@SPLUNK` label to the Collector (see below for how to add
-custom Fluentd log sources), and the Collector will send these events to the
-HEC ingest endpoint determined by the `--realm SPLUNK_REALM` option, e.g.
+The Fluentd deb/rpm package (`td-agent`) will be installed, and the `td-agent`
+service will be configured to forward log events with the `@SPLUNK` label to
+the Collector (see below for how to add custom Fluentd log sources). The
+Collector will then send these events to the HEC ingest endpoint determined by
+the `--realm SPLUNK_REALM` option, i.e.
```
theletterf, line 202: Does it make sense to still default the HEC endpoint to the realm?

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -677,7 +677,7 @@ Options:
                                         (default: https://ingest.REALM.signalfx.com/v2/trace)
   --uninstall                           Removes the Splunk OpenTelemetry Collector for Linux.
   --with[out]-fluentd                   Whether to install and configure fluentd to forward log events to the collector.
-                                        (default: --with-fluentd)
+                                        (default: --without-fluentd)
```
theletterf, line 680: So --without-fluentd is preserved for backward compatibility? What happens if both are used? Is evaluation order based? Just thinking about conflicts.

docs/getting-started/linux-installer.md
jeffreyc-splunk: I believe this was handled in https://github.com/signalfx/splunk-otel-collector/pull/3330.

internal/buildscripts/packaging/installer/install.sh
jeffreyc-splunk: - Correct, similar to how we currently support `--with-fluentd` even though it was the default option.
- The last option specified in the installation command will be used since we parse the options serially.

PR #3373, opened 2023-07-13
Set explicit stemcell requirement for Tanzu Tile

crobert-1:
To allow the Tanzu Tile to work in TAS v3+ environments, the stemcell version must be explicitly set, or it won't run. [Reference for context.](https://docs.vmware.com/en/Tile-Developer-Guide/3.0/tile-dev-guide/tile-generator.html#stemcells-13)

Tested on TAS v2.13, v3.0, and v4.0

### DIFF
--- a/deployments/cloudfoundry/tile/tile.yml
+++ b/deployments/cloudfoundry/tile/tile.yml
@@ -16,6 +16,10 @@ description: |
   This also includes a BOSH release of the Splunk OpenTelemetry Collector, called "splunk-otel-collector"
   that can be used in deployments to put the collector on individual VMs.
 
+stemcell_criteria:
+  os: 'ubuntu-jammy'
+  version: '1.148'
+
 # Specify the packages to be included in the tile.
 packages:
   - name: splunk_collector_firehose_nozzle


deployments/cloudfoundry/tile/tile.yml
```diff
@@ -16,6 +16,10 @@ description: |
   This also includes a BOSH release of the Splunk OpenTelemetry Collector, called "splunk-otel-collector"
   that can be used in deployments to put the collector on individual VMs.
 
+stemcell_criteria:
+  os: 'ubuntu-jammy'
+  version: '1.148'
```
rmfitzpatrick, line 21: What's this version reference?

rmfitzpatrick: I see now, but I guess we're moving from https://network.tanzu.vmware.com/products/stemcells-ubuntu-xenial#/releases/1303059?

crobert-1: Right, previously the default was `>= ubuntu-xenial 621.584`, but for some reason this rendered all stemcells invalid with the tile on TAS v3 and v4 (I tried xenial, bionic, and jammy based stemcells). I contacted the Pivotal team but have not heard back for the reasoning here.

rmfitzpatrick (approved):

PR #3374, opened 2023-07-13
[receiver/databricks] add retry/backoff on http 429s

pmcollins:
A customer with a presumably large Databricks deployment encountered rate limit errors (HTTP 429) while using the Databricks receiver. This change adds retry/backoff logic for when a 429 is encountered.


atoulme (approved):
LGTM - I'm ok with the default exponential backoff settings, I think: https://pkg.go.dev/github.com/cenkalti/backoff/v4#pkg-constants

rmfitzpatrick (approved):

PR #3413, opened 2023-07-26
Add basicauth extension

atoulme:
This adds the basicauth extension to our distribution of the collector. This addition is supported by an integration test checking that we can scrape a Prometheus endpoint protected by basic authentication.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - (Splunk) Discovery mode: Ensure all successful observers are used in resulting receiver creator instance ([#3391](https://github.com/signalfx/splunk-otel-collector/pull/3391))
 
+### 💡 Enhancements 💡
+
+- Add support for basicauth extension ([]())
 ## v0.81.0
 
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 - (Splunk) Discovery mode: Ensure all successful observers are used in resulting receiver creator instance ([#3391](https://github.com/signalfx/splunk-otel-collector/pull/3391))
 
+### 💡 Enhancements 💡
+
+- Add support for basicauth extension ([]())
```
crobert-1, line 11:
```suggestion
- Add support for basicauth extension ([#3413](https://github.com/signalfx/splunk-otel-collector/pull/3413))
```

internal/components/components.go
```diff
@@ -124,6 +125,7 @@ func Get() (otelcol.Factories, error) {
 		smartagentextension.NewFactory(),
 		zpagesextension.NewFactory(),
 		ballastextension.NewFactory(),
+		basicauthextension.NewFactory(),
```
crobert-1, line 128: In general, these factories are in alphabetical order. I know we got off track with the extensions before your change, so you don't have to do it here, but it'd be nice to arrange all of these in alphabetical order.

crobert-1 (approved):

internal/components/components.go
atoulme: good call, reordered

PR #3421, opened 2023-07-28
Network interface configuration

atoulme:
Add support for configuring the network interface on which the collector receivers will listen


cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -9,14 +9,19 @@
 # - SPLUNK_HEC_TOKEN: The Splunk HEC authentication token
 # - SPLUNK_HEC_URL: The Splunk HEC endpoint URL, e.g. https://ingest.us0.signalfx.com/v1/log
 # - SPLUNK_INGEST_URL: The Splunk ingest URL, e.g. https://ingest.us0.signalfx.com
+# - SPLUNK_NETWORK_INTERFACE: The network interface the agent receivers listen on. Defaults to 0.0.0.0.
```
theletterf, line 12: @atoulme What's the difference here between "network interface" and "IP address"? What other values could this setting have other than the classic IPv4 address?

atoulme: A network interface is designated via an IP address. A network interface represents a communication hardware device offered by the computer to connect to a network. The loopback network interface is typically associated with 127.0.0.1, while an Ethernet network interface will use a fixed IP, typically bound to a local network such as 10.0.0.1 or 192.168.2.1. The 0.0.0.0 network interface is special as it indicates to bind to all existing network interfaces available on the computer.

theletterf: Right, 0.0.0.0 == localhost, etc. I was just wondering if the label is clear enough—sometimes correct definitions are not the clearest labels. I trust your judgement here though. Thanks for the explanation!

atoulme: No 0.0.0.0 is not localhost, these are different network interfaces. I agree with you that the name is confusing, though I feel is more precise.

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -9,14 +9,15 @@
 # - SPLUNK_HEC_TOKEN: The Splunk HEC authentication token
 # - SPLUNK_HEC_URL: The Splunk HEC endpoint URL, e.g. https://ingest.us0.signalfx.com/v1/log
 # - SPLUNK_INGEST_URL: The Splunk ingest URL, e.g. https://ingest.us0.signalfx.com
+# - SPLUNK_NETWORK_INTERFACE: The network interface the agent receivers listen on.
```
rmfitzpatrick, line 12: I think denoting the directionality is helpful since `SPLUNK_NETWORK_INTERFACE` could be ambiguous and may be misinterpreted as an address to reach splunk

```suggestion
# - SPLUNK_LISTEN_INTERFACE: The network interface the agent receivers listen on.
```

atoulme: Thanks, this is better. I will adopt it across.

rmfitzpatrick (dismissed):
You've added a new required env var without any enforcement on service startup (breaking). Please add to https://github.com/signalfx/splunk-otel-collector/blob/main/internal/settings/settings.go using the memory env vars as an example.

edit: we should be setting the default value in settings to not make it required imo***

atoulme:
Thanks! I was also looking for good ways to test this. I caught a few failing tests. I'll take any hint on what else we can add to test.

rmfitzpatrick:
> Thanks! I was also looking for good ways to test this. I caught a few failing tests. I'll take any hint on what else we can add to test.

The settings test suite has env var helpers to detect what's set by the collector: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/settings/settings_test.go#L204.

We also have a default config test that should always reflect any changes: https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/default_config_test.go. I think it's good that it didn't require any changes though I'm not 100% on what the ":<port>" addr parsing semantics are atm.

jeffreyc-splunk:
We should also add the new var to the choco package as a new parameter:
- description: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/buildscripts/packaging/choco/splunk-otel-collector/splunk-otel-collector.nuspec
- add and handle the new parameter: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/buildscripts/packaging/choco/splunk-otel-collector/tools/chocolateyinstall.ps1

jeffreyc-splunk:
We'll also need to add support to our [deployments](https://github.com/signalfx/splunk-otel-collector/tree/main/deployments) for the new env var when these changes are released.

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -76,23 +77,23 @@ receivers:
       - job_name: 'otel-collector'
         scrape_interval: 10s
         static_configs:
-        - targets: ['0.0.0.0:8888']
+        - targets: ["${SPLUNK_LISTEN_INTERFACE}:8888"]
```
rmfitzpatrick, line 80: This is the address to query, so if we want to influence the internal prom server address we should specify it in `service::telemetry::metrics::address`

tests/general/default_config_test.go
```diff
@@ -35,8 +35,9 @@ func TestDefaultGatewayConfig(t *testing.T) {
 		"",
 		func(collector testutils.Collector) testutils.Collector {
 			return collector.WithArgs().WithEnv(map[string]string{
-				"SPLUNK_ACCESS_TOKEN": "not.real",
-				"SPLUNK_REALM":        "not.real",
+				"SPLUNK_LISTEN_INTERFACE": "0.0.0.0",
```
rmfitzpatrick, line 38: To vet the default behavior it's probably best to remove this and if you'd like add new tests to confirm the env var is picked up where expected.

cmd/otelcol/config/collector/agent_config.yaml
atoulme: and I wonder if we should hardcode to 127.0.0.1, instead of tying this up to this env var.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Offer an installer option to configure the network interface used by the collector ()[]
```
rmfitzpatrick, line 4:
```suggestion
- Use `SPLUNK_LISTEN_INTERFACE` and associated installer option to configure the network interface used by the collector for default configurations ([#3421](https://github.com/signalfx/splunk-otel-collector/pull/3421))
```

cmd/otelcol/config/collector/agent_config.yaml
atoulme: That might be a separate PR that I could land first.

rmfitzpatrick: I'm wrong, appears the default behavior w/ `:8888` resolves to `:::8888` so this seems ok as is though an additional env var may be warranted in the future.

tests/general/default_config_test.go
```diff
@@ -35,8 +35,9 @@ func TestDefaultGatewayConfig(t *testing.T) {
 		"",
 		func(collector testutils.Collector) testutils.Collector {
 			return collector.WithArgs().WithEnv(map[string]string{
-				"SPLUNK_ACCESS_TOKEN": "not.real",
-				"SPLUNK_REALM":        "not.real",
+				"SPLUNK_NETWORK_INTERFACE": "0.0.0.0",
```
rmfitzpatrick, line 38: I think this will help confirm the default behavior doesn't now require this env var, but may be missing something.

```suggestion
```

tests/general/default_config_test.go
```diff
@@ -197,8 +199,9 @@ func TestDefaultAgentConfig(t *testing.T) {
 			return collector.WithArgs(
 				"--config", "/etc/otel/collector/agent_config.yaml",
 			).WithEnv(map[string]string{
-				"SPLUNK_ACCESS_TOKEN": "not.real",
-				"SPLUNK_REALM":        "not.real",
+				"SPLUNK_NETWORK_INTERFACE": "0.0.0.0",
```
rmfitzpatrick, line 202:
```suggestion
```

cmd/otelcol/config/collector/agent_config.yaml
atoulme: Right, if you don't mind, this might be later instead of landing it first.

tests/general/default_config_test.go
atoulme: ouch this is a bad merge 🤦

atoulme:
Going to squash to one commit and make sure the diff makes sense.

rmfitzpatrick (approved):

atoulme:
@jeffreyc-splunk please take a look when you have a moment. I am rerunning some of the ansible tests assuming that they're just flaky.

jeffreyc-splunk:
> @jeffreyc-splunk please take a look when you have a moment. I am rerunning some of the ansible tests assuming that they're just flaky.

Yeah, unfortunately virtualbox/vagrant on the github runners is flaky.  May take a couple of retries to get all of those tests to pass.

jeffreyc-splunk:
Not sure if it needs to be called out directly, but the new option in the installers will only be applicable for new installations or when the default unmodified config files are upgraded to this new version.  If the customer uses a custom/modified config file, they will need to manually replace the values in their config with this env var for these changes to take effect.

atoulme:
I can maybe call that out in a subsequent PR if you don't mind, with a changelog entry that calls out to customers how to take advantage of this if they have an existing install.

jeffreyc-splunk (approved):

PR #3482, opened 2023-08-10
Remove sending a metric as part of auto-instrumentation

atoulme:
Since we are standardizing on OTEL_RESOURCE_ATTRIBUTES, we no longer should be emitting metrics when auto-instrumenting applications.


rmfitzpatrick:
What attributes have we standardized on for this to be a noop?

jeffreyc-splunk:
I assume this would be considered a breaking change if any customers are currently relying on this metric?

atoulme:
Moving to a branch for development, merging and closing.

PR #3262, opened 2023-06-13
[receiver/jmx] Add metric gatherer to artifacts

crobert-1:
To properly support the jmx receiver in Docker, the dockerfile has to download the JMX metrics gatherer, and have it available to the collector. This change adds this functionality to the linux/macos dockerfile, as well as the windows dockerfile.

Also, adds an integration test getting metrics from a docker container.

PR #3485, opened 2023-08-10
skip telegraf/proctstat test on arm

rmfitzpatrick:
The limited set of expected metrics appears to vary (seen in testing https://github.com/signalfx/splunk-otel-collector/pull/3262), so skipping for now. When we have actual arm runners expected behavior can be refined.


rmfitzpatrick:
This may actually be an issue w/ https://github.com/signalfx/splunk-otel-collector/pull/3262 though I'm not seeing why atm...

atoulme (approved):

rmfitzpatrick:
addressed in the PR that revealed it: https://github.com/signalfx/splunk-otel-collector/pull/3262

PR #3506, opened 2023-08-14
Add systemd sample file to splunk-otel-auto-instrumention

jeffreyc-splunk:
- Include sample systemd drop-in file in the `splunk-otel-auto-instrumentation` package that can be manually copied to the systemd config directory to enable the java agent for systemd services (this will be handled automatically by the installers if systemd auto instrumentation is enabled)
- Include default `/etc/splunk/zeroconfig_java.conf`
- Remove automatic addition of `libsplunk.so` to `/etc/ld.so.preload` when the package is installed (this will be handled automatically by the installers only if system-wide auto instrumentation is enabled)
- `libsplunk.so` will be automatically removed from `/etc/ld.so.preload` (if it exists) when the package is uninstalled (unchanged from current behavior)
- Add tests that simulate the manual steps above and verify configured attributes for tomcat


instrumentation/packaging/fpm/common.sh
```diff
@@ -24,16 +24,18 @@ PKG_DESCRIPTION="Splunk OpenTelemetry Auto Instrumentation"
...
+JAVA_AGENT_INSTALL_PATH="${INSTALL_DIR}/splunk-otel-javaagent.jar"
+CONFIG_INSTALL_PATH="${INSTALL_DIR}/instrumentation.conf"
+CONFIG_PATH="$REPO_DIR/instrumentation/install/instrumentation.conf"
+EXAMPLES_INSTALL_DIR="${INSTALL_DIR}/examples"
+EXAMPLES_DIR="${FPM_DIR}/examples"
 
 JAVA_AGENT_RELEASE_PATH="${FPM_DIR}/../java-agent-release.txt"
 JAVA_AGENT_RELEASE_URL="https://github.com/signalfx/splunk-otel-java/releases/"
 
-POSTINSTALL_PATH="$FPM_DIR/postinstall.sh"
```
rmfitzpatrick, line 34: What is the ld.so.preload removal process to replace this?

jeffreyc-splunk: The removal happens in preuninstall.sh (see below), which has not been changed. So it should always be removed, even if added externally.

instrumentation/libsplunk.md
```diff
@@ -0,0 +1,181 @@
...
+soon-to-be running Java application. In this way, all Java applications on the host will be automatically instrumented
+by the Splunk OTel Java agent.
+
+Once instrumented, Java executables send traces to a locally running
+[Splunk Open Telemetry Collector](https://github.com/signalfx/splunk-otel-collector)
+(installed separately) and then on to Splunk APM.
+
+## Configuration File
+
+At startup, the shared object reads the config file, `/usr/lib/splunk-instrumentation/instrumentation.conf` which by
```
atoulme, line 23: this will no longer be true, but we can check those docs in the next PR.

instrumentation/packaging/fpm/zeroconfig_java.conf
```diff
@@ -0,0 +1 @@
+JAVA_TOOL_OPTIONS=-javaagent:/usr/lib/splunk-instrumentation/splunk-otel-javaagent.jar
```
atoulme, line 1: need a newline

internal/buildscripts/packaging/tests/helpers/util.py
```diff
@@ -42,6 +42,18 @@ def retry(function, exception, max_attempts=5, interval=5):
...
+    while True:
+        code, output = container.exec_run(cmd)
+        elapsed = time.time() - start_time
+        if code == 0:
+            print(f"'{cmd}' completed in {elapsed}s:\n{output.decode('utf-8')}")
+            return code, output
+        assert elapsed < timeout, f"timed out waiting for '{cmd}':\n{output.decode('utf-8')}"
+        time.sleep(1)
+
+
```
atoulme, line 56: this can be peeled off to a separate PR for main to simplify the diff a bit

instrumentation/packaging/fpm/zeroconfig_java.conf
atoulme: maybe there's a newline, sorry. I can't quite tell from github's web UI.

atoulme (approved):

PR #3509, opened 2023-08-16
Move to gopsutil 3.23.7 and remove the need to set environment variables

### LINKED ISSUE #1340: Unable to configure hostmetrics receiver from container environment

atoulme:
Fixes #1340


jinja2 (approved):

atoulme:
I will merge after the 0.83.0 release is made.

PR #3514, opened 2023-08-18
[DNM] Zero-config C code to inject environment variables for Java and Node.js

atoulme:
** Work in progress **

This PR shows the current work in progress to adopt a zero-config solution that is minimally inserting environment variables into the runtime environment of Java and Node.js processes.

Please feel free to review and help strengthen this approach.


instrumentation/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+#include <string.h>
+#include <unistd.h>
+#include <stdio.h>
+
+#define MAX_LINE_LENGTH 1023
+#define MAX_LINES 50
+
+// TODO change to systemd drop in file paths
+static char *const env_var_file_java = "/etc/splunk/zeroconfig_java.conf";
+static char *const env_var_file_node = "/etc/splunk/zeroconfig_node.conf";
```
atoulme, line 11: @jeffreyc-splunk would you have the path of the systemd drop-in env var files handy? Did you also separate them by language by any chance?

instrumentation/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+
+static char *const allowed_env_vars[] = {
+    "OTEL_SERVICE_NAME",
+    "OTEL_EXPORTER_OTLP_ENDPOINT",
+    "OTEL_RESOURCE_ATTRIBUTES",
+    "SPLUNK_PROFILER_ENABLED",
+    "SPLUNK_PROFILER_MEMORY_ENABLED",
+    "SPLUNK_METRICS_ENABLED",
```
atoulme, line 19: this list needs to be completed. It doesn't represent the whole set yet.

instrumentation/src/main.c
jeffreyc-splunk: If we go with https://github.com/signalfx/splunk-otel-collector/pull/3506, the deb/rpm packages can include sample systemd drop-in files in `/usr/lib/splunk-instrumentation/examples/systemd/` for each language (see https://github.com/signalfx/splunk-otel-collector/pull/3506/files#diff-71d510eec30ef5d3556193068fc6726c24202c439f5bd9f25989cb511ff113f9 for java). We can then add a sample `splunk-otel-js.conf` once we add support for the node.js agent.  These sample files can be used for manual installations.

On the other hand, I'm thinking that the installers will automatically create the actual drop-in file to be loaded by systemd, for example `/usr/lib/systemd/system.conf.d/splunk-otel-auto-instrumentation.conf`, based on the installation options and defaults (see https://github.com/signalfx/splunk-otel-collector/pull/3506/files#diff-4dca014cfa3f56017a31c6a9e698de416a5bcc9cf6450a8ef9d088295f19ca98 for the test config).

instrumentation/src/main.c
jeffreyc-splunk: For java, would we need to support all of the [splunk-specific env vars](https://docs.splunk.com/Observability/gdi/get-data-in/application/java/configuration/advanced-java-otel-configuration.html)?

Any from [otel java](https://opentelemetry.io/docs/instrumentation/java/automatic/agent-config/#configuring-with-environment-variables) and [otel sdk](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/) not included in the above list?

jeffreyc-splunk:
If we go with these changes, should we just use the `LD_PRELOAD` env var for systemd-only instrumentation? So instead of setting the env vars directly in a systemd config like:
```
[Manager]
DefaultEnvironment="JAVA_TOOL_OPTIONS=-javaagent:/usr/lib/splunk-instrumentation/splunk-otel-javaagent.jar"
DefaultEnvironment="OTEL_EXPORTER_OTLP_ENDPOINT=http://1.2.3.4:5678"
DefaultEnvironment="OTEL_RESOURCE_ATTRIBUTES=deployment.environment=my_deployment_environment"
DefaultEnvironment="OTEL_SERVICE_NAME=my_service_name"
DefaultEnvironment="SPLUNK_PROFILER_ENABLED=true"
```
The systemd config would just be something like:
```
[Manager]
DefaultEnvironment="LD_PRELOAD=/usr/lib/splunk-instrumentation/libsplunk.so"
```
And `libsplunk.so` would just read/inject the env vars from a separate config file. This would simplify configuration for the installers since we'd be dealing with a single config file for both systemd and `/etc/ld.so.preload` methods.

atoulme:
That second approach would allow to run everything through the preload function, which we can test for. I'm ok with either way but I tend to the latter option because we can then format the env var file into a simple file to source from.

instrumentation/src/main.c
atoulme: yes we're missing quite a few.

aunshc: Quick question for clarification: Correct me if I am wrong, wouldn't the allowed env variables be injected from the language specific drop-in files and not the constant ones defined here?

atoulme: This allows us to check the file contents, and ensure that only the keys we allow here are injected in the environment of processes.

jeffreyc-splunk: Will this library read `JAVA_TOOL_OPTIONS` and `NODE_OPTIONS` from the conf files, or are you expecting them to be defined somewhere else?  For systemd, we can define them directly in the drop-in file(s), but should they be added to the list for the preloader method?

instrumentation/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+        }
+        // if we have read a string without a newline termination, we have reached the end of file
+        // or the string is past our max buffer size and we have an invalid value and we need to abort.
+
+        // Properly formatted input file contains lines of length less than MAX_LINE_LENGTH and ends with a newline.
+        // If newline character is not read, it means that a line is greater that MAX_LINE_LENGTH or the file does not end with a newline.
+        if (newline == NULL) {
+            break;
+        }
+        if (strlen(buffer) == 0) {
```
pjanotti, line 70: General recommendation: use the "n" versions/alternatives for all functions dealing with `char` buffers in C. Not only `strlen`, but, `memchar` instead of `strchr`, etc.  We can reason that the "n" versions/alternatives aren't needed in some instances, but instead of relying on such reasoning and assumptions, using the "n" versions/alternatives is more straightforward and safer.

instrumentation/src/main.c
```diff
@@ -0,0 +1,99 @@
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <stdio.h>
+
+#define MAX_LINE_LENGTH 1023
+#define MAX_LINES 50
+
+#define ALLOWED_ENV_VARS "OTEL_SERVICE_NAME", "OTEL_EXPORTER_OTLP_ENDPOINT", "OTEL_RESOURCE_ATTRIBUTES", "SPLUNK_PROFILER_ENABLED", "SPLUNK_PROFILER_MEMORY_ENABLED", "SPLUNK_METRICS_ENABLED", "JAVA_TOOL_OPTIONS", "NODE_OPTIONS"
```
rmfitzpatrick, line 9: Are you intending to rely on the conf contents for `JAVA_TOOL_OPTIONS` and `NODE_OPTIONS`?

instrumentation/src/main.c
```diff
@@ -0,0 +1,99 @@
...
+
+static char *const allowed_env_vars[] = {ALLOWED_ENV_VARS};
+static size_t const allowed_env_vars_size = sizeof(allowed_env_vars) / sizeof(*allowed_env_vars);
+
+// TODO change to systemd drop in file paths
```
rmfitzpatrick, line 14: If using the unit files I think we'll want to use a vetted ini? I'm not sure this would be a desirable conf form though since they are ultimately for configuring different things.

instrumentation/src/main.c
rmfitzpatrick: If we are allowing arbitrary input I think instrumentation checksum validation from .rodata should be required. This will add complexity and overhead but the benefit is that we'd only respect known target values.

jeffreyc-splunk:
> That second approach would allow to run everything through the preload function, which we can test for. I'm ok with either way but I tend to the latter option because we can then format the env var file into a simple file to source from.

Instead of the systemd drop-in method, another option could be to include and install a simple [systemd environment generator](https://www.freedesktop.org/software/systemd/man/systemd.environment-generator.html) script to read and inject env vars from a shared config file. This would bypass `libsplunk.so` while still allowing us to manage and source env vars from a single config file with either the preloader or systemd-only methods.

instrumentation/tests/nodejs/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN mkdir -p /etc/splunk/
+
+COPY zeroconfig.conf /etc/splunk/zeroconfig_node.conf
+
+CMD node index.js
+
+ENV LD_PRELOAD /etc/ld.preload.so/libsplunk.so
+
+COPY libsplunk.so /etc/ld.preload.so/libsplunk.so
```
jeffreyc-splunk, line 14: Same comment as the java dockerfile.

instrumentation/src/main.c
jvoravong: The contents of these conf files are required for other auto-instrumentation implementation such as chart/operator implementations. I think these will be required, can supply more details if needed.

instrumentation/src/main.c
jvoravong: Side note: A reoccurring issue we have been noticing with being swappable with upstream auto-instrumentation solutions vs Splunk solutions is Splunk file paths are unique to Splunk vs what upstream uses. I don't think this issue should be addressed in this PR, but it is worth noting for future compatibility.

atoulme: We can make those configurable via macros. I'll do that.

instrumentation/tests/java/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+CMD java -jar Main.jar
+
+ENV OTEL_SERVICE_NAME iknowmyownservicename
+
+ENV ANOTHER_VAR foo
+
+ENV LD_PRELOAD /etc/ld.preload.so/libsplunk.so
+
+COPY libsplunk.so /etc/ld.preload.so/libsplunk.so
```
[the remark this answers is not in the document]
atoulme: sure thing

instrumentation/src/main.c
atoulme: these are definitely required or you cannot inject java or nodejs agents.

instrumentation/src/main.c
atoulme: I can look but I'm bad at C, and I am afraid I don't completely understand what you mean. I'll take a deeper look.

atoulme: OK, I have applied that approach.

instrumentation/src/main.c
atoulme: Done.

instrumentation/src/main.c
atoulme: Sure, I'll look into that.

PR #3536, opened 2023-08-21
Update installer script for systemd auto instrumentation

jeffreyc-splunk:
***NOTE***: Does not currently support any `libsplunk.so` or `splunk-otel-js` changes from the `zero-config-dev` branch since they have not been released yet.

- Add `--with[out]-systemd-instrumentation` options
  - Defaults to `--without-systemd-instrumentation`
  - Installs the existing `splunk-otel-auto-instrumentation` deb/rpm package
  - Creates/configures the `/usr/lib/systemd/system.conf.d/00-splunk-otel-auto-instrumentation.conf` drop-in file with env vars
  - Include the `splunk.zc.method=splunk-otel-auto-instrumentation-<version>-systemd` resource attribute
- The `--with[out]-instrumentation` options behaves as it does now:
  - Defaults to `--without-instrumentation`
  - Installs the existing `splunk-otel-auto-instrumentation` deb/rpm package
  - Updates `/etc/ld.so.preload` for `libsplunk.so` and creates/configures `/usr/lib/splunk-instrumentation/instrumentation.conf`
  - Include the `splunk.zc.method=splunk-otel-auto-instrumentation-<version>` resource attribute
- Add `--otlp-endpoint` option
  - Defaults to `http://${listen_interface}:4317`
  - Only supported with `--with-systemd-instrumentation` since `libsplunk.so` currently doesn't support setting `OTEL_EXPORTER_OTLP_ENDPOINT`
- Update tests to use the changes and tomcat setup for java instrumentation from #3506


atoulme:
I lost track of the state of this PR. Can we merge it into the long term dev branch? I'll review now.

atoulme (approved):

PR #3598, opened 2023-09-08
Bump go modules to 1.20

jeffreyc-splunk:
- Update `tidy` commands
- Add `tidy` github job

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -23,10 +23,6 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  PYTHON_VERSION: '3.10'
-  PIP_VERSION: '22.0.4'
-  REQUIREMENTS_PATH: "internal/buildscripts/packaging/tests/requirements.txt"
-  RESULT_PATH: "~/testresults"
   GO_VERSION: 1.20.8
 
 jobs:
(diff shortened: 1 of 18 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -48,6 +44,30 @@ jobs:
...
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache-dependency-path: '**/go.sum'
+
+      - name: tidy
+        run: |
+          make install-tools
+          make for-all CMD="make tidy"
+          if ! git diff --exit-code; then
```
rmfitzpatrick, line 66: Do you know if changes are required to show the diff content in the action? It'd be helpful to do so instead of a binary check w/o info.

jeffreyc-splunk: This is a similar method we've been using in other jobs, e.g. https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/build-and-test.yml#L96 (see https://github.com/signalfx/splunk-otel-collector/actions/runs/6102809198/job/16562126153 for an example of failure). Hopefully, if there are any go.mod/go.sum changes after the tidy, the job will fail.

atoulme (approved):
LGTM

rmfitzpatrick (approved):

PR #3627, opened 2023-09-11
feat: enabling scriptedinputs receiver in components

weliasz:


cmd/otelcol/config/collector/full_config_linux.yaml
```diff
@@ -167,6 +167,24 @@ receivers:
   fluentforward:
     endpoint: 0.0.0.0:8006
 
+  # Enables scripted inputs receiver
+  # Full configuration here: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/receiver/scriptedinputsreceiver/README.md
+  scripted_inputs/cpu:
```
rmfitzpatrick, line 172: I think we can add a single example since this is for the receiver as a component option and not for documenting its permutations.

rmfitzpatrick:
Please add an integration test along with these changes. You can use `<RE2()>` directives for dynamic resource logs content now.

internal/receiver/scriptedinputsreceiver/script_test.go
```diff
@@ -0,0 +1,33 @@
+// Copyright Splunk, Inc.
```
rmfitzpatrick, line 1: This should be in `tests/receivers/scriptedinputsreceiver` and still in the `tests` package like all the rest are.

rmfitzpatrick: When I run this in the correct location it appears to not be functional in the container:

```
2023-09-13T13:57:03.219Z	error	scriptedinputsreceiver/commander.go:84	Error in cmd wait: %v	{"kind": "receiver", "name": "scripted_inputs/cpu", "data_type": "logs", "error": "exit status 1"}
github.com/signalfx/splunk-otel-collector/internal/receiver/scriptedinputsreceiver.(*commander).watch
	github.com/signalfx/splunk-otel-collector/internal/receiver/scriptedinputsreceiver/commander.go:84
2023-09-13T13:57:13.215Z	warn	scriptedinputsreceiver/operator.go:108	Script didn't complete within configured interval.{script_name 15 0 cpu <nil>}	{"kind": "receiver", "name": "scripted_inputs/cpu", "data_type": "logs"}
```

I think it'd be worth improving stderr/startup failure behavior and resolving this issue before including as a component. I also wonder if all the other supported scripts should be added first as well.

internal/receiver/scriptedinputsreceiver/testdata/resource_logs/cpu.yaml
```diff
@@ -0,0 +1,4 @@
+resource_logs:
+  - scope_logs:
+      - logs:
+          - body: "<RE2(CPU    pctUser  pctSystem    pctIdle)>"
```
rmfitzpatrick, line 4: Can this be expanded to demonstrate more content? attribute checks would be helpful too imo.

weliasz: The output looks like this:
CPU    pctUser  pctSystem    pctIdle
all       3.89        6.8       90.1

What do you suggest to test more?

internal/receiver/scriptedinputsreceiver/script_test.go
weliasz: When I placed the test into proper directory I got the same result, what would be your suggestion on how to diagnose the failure in the container (locally on my MAC it works perfectly)

tests/receivers/scriptedinputs/testdata/config.yaml
```diff
@@ -0,0 +1,2 @@
+scripted_inputs/cpu:
```
rmfitzpatrick, line 1: believe this file can be deleted

internal/receiver/scriptedinputsreceiver/config.go
```diff
@@ -96,6 +97,10 @@ func (c *Config) Validate() error {
 
 // Build will build a stdoutOperator.
 func (c *Config) Build(logger *zap.SugaredLogger) (operator.Operator, error) {
+	if isContainer() {
+		return nil, fmt.Errorf("cannot run scripted inputs receiver in Docker container")
```
rmfitzpatrick, line 101: Do you mind updating the readme for this limitation as well?

```suggestion
		return nil, fmt.Errorf("scriped inputs receiver must be run directly on host and is not supported in container.")
```

internal/receiver/scriptedinputsreceiver/script_test.go
rmfitzpatrick: initially addressed by container restriction but stderr handling should be added in later PR imo.

internal/receiver/scriptedinputsreceiver/config.go
```diff
@@ -96,6 +97,10 @@ func (c *Config) Validate() error {
 
 // Build will build a stdoutOperator.
 func (c *Config) Build(logger *zap.SugaredLogger) (operator.Operator, error) {
+	if isContainer() {
+		return nil, fmt.Errorf("scriped inputs receiver must be run directly on host and is not supported in container")
```
rmfitzpatrick, line 101: Mind updating the readme?

rmfitzpatrick (approved):

PR #3628, opened 2023-09-11
Fix chef tests

jeffreyc-splunk:
Convert chef tests from serverspec to inspec.

Follow-up to https://github.com/signalfx/splunk-otel-collector/pull/3606#issuecomment-1711945223.

### DIFF
--- a/.github/workflows/chef-test.yml
+++ b/.github/workflows/chef-test.yml
@@ -117,10 +117,10 @@ jobs:
         with:
           version: 21.12.720
 
-      - name: gem install kitchen-docker
+      - name: Install kitchen-docker
         run: |
-          $env:PATH = 'C:\opscode\chef-workstation\embedded\bin;' + $env:PATH
-          gem install kitchen-docker -v 2.12.0
+          chef env --chef-license accept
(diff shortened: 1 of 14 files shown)


rmfitzpatrick (approved):

jeffreyc-splunk:
> Do you know what's declaring busser-serverspec as a dependency?

It is a dependency of `serverspec`, which we were using as the verifier in test-kitchen: https://kitchen.ci/docs/verifiers/serverspec/

rmfitzpatrick:
> It is a dependency of serverspec
I wasn't able to find an explicit declaration but wonder if test-kitchen or busser is dynamically installing it based on the verifier name in the suite path. Mainly curious and not tied to this PR content.

jeffreyc-splunk:
> > It is a dependency of serverspec
> 
> I wasn't able to find an explicit declaration but wonder if test-kitchen or busser is dynamically installing it based on the verifier name in the suite path. Mainly curious and not tied to this PR content.*

I couldn't either... I think kitchen just installs it by default for serverspec: https://github.com/signalfx/splunk-otel-collector/actions/runs/6125267267/job/16626924536?pr=3606#step:4:325

PR #3640, opened 2023-09-14
update filelog to latest snapshot to fix regression with default flush period

atoulme:
**Description:**
update filelog to latest snapshot to fix regression with default flush period


tests/receivers/filelog/testdata/syslog
```diff
@@ -5,4 +5,4 @@
 <4>Oct  6 14:45:33 debian kernel: [17099.344322] device veth4f642f0 left promiscuous mode
 <5>Oct  6 14:45:33 debian kernel: [17099.344328] docker0: port 1(veth4f642f0) entered disabled state
 <6>Oct  6 14:45:33 debian avahi-daemon[564]: Withdrawing address record for aa::bb:cc:dd:ee on veth4f642f0.
-<7>Oct  6 14:45:33 debian NetworkManager[567]: <info>  [1665081933.1506] device (veth4f642f0): released from master device docker0
+<7>Oct  6 14:45:33 debian NetworkManager[567]: <info>  [1665081933.1506] device (veth4f642f0): released from master device docker0
```
jinja2, line 8: is the newline removal intentional?

atoulme: It is. The bug that we're fixing here is https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/26691 which caused a regression where the last line of a file is not sent if it doesn't have an EOL.

PR #3642, opened 2023-09-15
Add warnings regarding the change of default value from 0.0.0.0 to 127.0.0.1 in installers

atoulme:
**Description:**
Add warnings regarding the change of default value from `0.0.0.0` to `127.0.0.1` in installers.

This logs a message when the value of the listen interface is set to `0.0.0.0` during installation, on Linux and Windows, in preparation of incoming changes to installer default values.

### DIFF
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -574,10 +574,6 @@ if ($config_path -Eq "") {
     }
 }
 
-if ($network_interface -Eq "0.0.0.0") {
-    echo "Starting with version 0.86.0, the collector installer will change its default network listening interface from 0.0.0.0 to 127.0.0.1. Please consult the release notes for more information and configuration options."
-}
-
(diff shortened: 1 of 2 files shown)


jeffreyc-splunk:
Maybe these messages should be moved to the end of the scripts, so that it is more noticeable and doesn't get lost in the output from the installation?

aunshc:
The messaging LGTM, thanks @atoulme! Would this apply to other deployment approaches (Ansible, Puppet,..) for the same warnings to be relayed or only the installer script?

crobert-1 (approved):

jeffreyc-splunk:
> The messaging LGTM, thanks @atoulme! Would this apply to other deployment approaches (Ansible, Puppet,..) for the same warnings to be relayed or only the installer script?

The other deployments methods (ansible, chef, etc) do not use these scripts.

atoulme:
For now, I have circumscribed the changes to default networking values to our installers and Chocolatey because they characterize VM installations and are primary target for our zero-config approach. We can expand to other installation methods next.

jinja2 (approved):

PR #3684, opened 2023-09-22
removed more dead code.

samiura:
**Description:** Removed unused code


atoulme:
`monitorcodegen` is used in the code base here: internal/signalfx-agent/pkg/monitors/zcodegen/codegen.go

PR #3720, opened 2023-10-03
Allow passing splunk_hec_token for log ingestion into Splunk Cloud to work.

chepati:
This patch is adding the ability to pass splunk_hec_token so that logs can be ingested into Splunk Cloud.


jeffreyc-splunk (approved):
Thanks!

PR #3731, opened 2023-10-05
Redis discovery

greatestusername:
**Description:** 
Adds discovery mode tests and configs for `redis` receiver

**Link to Splunk idea:** 
No idea link. Internal contribution

**Testing:** 
- Added new integration bundled test and test data
- verified that existing redis integration tests work as intended

**Documentation:**


greatestusername:
Come on tests! (especially the windows tests that likely wont work because they require WSL)

Tell me what's broken in our workflow that works on my local!

internal/confmapprovider/discovery/bundle/bundle.d/receivers/redis.discovery.yaml.tmpl
```diff
@@ -0,0 +1,71 @@
...
+  enabled: true
+  rule:
+    docker_observer: type == "container" and any([name, image, command], {# matches "(?i)redis"}) and not (command matches "splunk.discovery")
+    host_observer: type == "hostport" and command matches "(?i)redis" and not (command matches "splunk.discovery")
+    k8s_observer: type == "port" and pod.name matches "(?i)redis"
+  config:
+    default:
+      endpoint: {{ defaultValue }}
+      password: {{ defaultValue }}
+      username: {{ defaultValue }}
```
rmfitzpatrick, line 11: I'm not too familiar with the redis acl system but I think default installations don't require auth info. Does it make sense not to use defaults for the username and password fields? They are still configurable by discovery config properties so I think the guidance can remain as is but we'd not be set up to fail to connect in all cases.

greatestusername: This is a good call out! Will change that so we don't use defaults but leave the statement guidance

internal/confmapprovider/discovery/bundle/bundle.d/receivers/redis.discovery.yaml
```diff
@@ -0,0 +1,73 @@
...
+#####################################################################################
+redis:
+  enabled: true
+  rule:
+    docker_observer: type == "container" and any([name, image, command], {# matches "(?i)redis"}) and not (command matches "splunk.discovery")
+    host_observer: type == "hostport" and command matches "(?i)redis" and not (command matches "splunk.discovery")
+    k8s_observer: type == "port" and pod.name matches "(?i)redis"
+  config:
+    default:
+      endpoint: splunk.discovery.default
```
rmfitzpatrick, line 13: This will be set from the discovered endpoint so I'm not sure we should be setting a default.

internal/confmapprovider/discovery/bundle/bundle.d/receivers/redis.discovery.yaml.tmpl
```diff
@@ -0,0 +1,69 @@
+{{ receiver "redis" }}:
+  enabled: true
+  rule:
+    docker_observer: type == "container" and any([name, image, command], {# matches "(?i)redis"}) and not (command matches "splunk.discovery")
+    host_observer: type == "hostport" and command matches "(?i)redis" and not (command matches "splunk.discovery")
+    k8s_observer: type == "port" and pod.name matches "(?i)redis"
+  config:
+    default:
+      endpoint: {{ defaultValue }}
```
rmfitzpatrick, line 9: This will be set from the discovered endpoint so I'm not sure we should be setting a default.

greatestusername:
argle new tests to investigate :D

internal/confmapprovider/discovery/bundle/bundle.d/receivers/redis.discovery.yaml.tmpl
```diff
@@ -0,0 +1,69 @@
+{{ receiver "redis" }}:
+  enabled: true
+  rule:
+    docker_observer: type == "container" and any([name, image, command], {# matches "(?i)redis"}) and not (command matches "splunk.discovery")
+    host_observer: type == "hostport" and command matches "(?i)redis" and not (command matches "splunk.discovery")
+    k8s_observer: type == "port" and pod.name matches "(?i)redis"
+  config:
+    default:
+      endpoint: '`endpoint`'
```
rmfitzpatrick, line 9: I think this field should already be inserted by the receiver creator. Do you know if this works (if not I can address where necessary)?

```suggestion
```

greatestusername: I was not able to get it to work without adding that in. 

When I remove the whole config section I get a `no receiver configuration specified in config` error.

Also erroring when just removing endpoint and leaving the `config` and `default` fields.

rmfitzpatrick:
Mind updating the readme https://github.com/signalfx/splunk-otel-collector/blob/main/internal/confmapprovider/discovery/README.md#discovery-mode?

rmfitzpatrick (approved):

greatestusername:
> Mind updating the readme https://github.com/signalfx/splunk-otel-collector/blob/main/internal/confmapprovider/discovery/README.md#discovery-mode?

can do!

PR #3733, opened 2023-10-05
Rename instrumentation deb/rpm package

jeffreyc-splunk:
- Need to build/release a new package that includes the [changes](https://github.com/signalfx/splunk-otel-collector/compare/main...zero-config-dev) from the `zero-config-dev` branch to prevent breaking our installers, which are currently only designed to install/configure the existing `splunk-otel-auto-instrumentation` package.
- Installation of this new package, `splunk-otel-auto-instrumentation-ng`, will automatically replace the old package.
- Any ideas for the new package name are welcome. Once we settle on the new name, I'll update the docs accordingly.


jeffreyc-splunk:
Closing for now since this may not be needed (the breaking changes are minimal since it'd be just a new config file).

PR #3754, opened 2023-10-09
otlp_config_linux.yaml - add otlphttp for metrics

Kielek:
**Description:** Add example how to send metrics over otlphttp

**Link to Splunk idea:** Based on created example for https://github.com/signalfx/tracing-examples/pull/169

**Testing:** Tested in https://github.com/signalfx/tracing-examples/pull/169

**Documentation:** N/A?

### DIFF
--- a/cmd/otelcol/config/collector/otlp_config_linux.yaml
+++ b/cmd/otelcol/config/collector/otlp_config_linux.yaml
@@ -76,6 +76,7 @@ processors:
 exporters:
   # Traces
   otlphttp:
+    metrics_endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com/v2/datapoint/otlp"
     traces_endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace/otlp"
     compression: gzip
     headers:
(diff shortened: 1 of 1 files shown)


cmd/otelcol/config/collector/otlp_config_linux.yaml
```diff
@@ -117,7 +118,7 @@ service:
     metrics:
       receivers: [otlp, signalfx, smartagent/signalfx-forwarder]
       processors: [memory_limiter, batch]
-      exporters: [signalfx]
+      exporters: [otlphttp, signalfx]
```
Kielek, line 121: The question is if here we need both exporters. Same for traces.

rmfitzpatrick: The signalfx exporter has a number of built in translations that work best w/ existing built-in content. You are also now adding double publishes.

rmfitzpatrick (changes requested):
This config is for use and not just example. If the ask is to document how to use the otlp http exporter, I'd suggest updating a readme or product doc.

theletterf:
I think this could be the doc that could host the PR's contents: https://docs.splunk.com/observability/en/gdi/opentelemetry/components/otlphttp-exporter.html

PR #3759, opened 2023-10-11
Expose more options to configure kind

atoulme:
**Description:**
Add `Image` and `KubeAdmConfigPatches` options to the `KindCluster` object.

### DIFF
--- a/tests/testutils/kubeutils/kind_cluster.go
+++ b/tests/testutils/kubeutils/kind_cluster.go
@@ -213,6 +213,5 @@ func (k KindCluster) renderConfig() string {
 	require.NoError(k.Testcase, err)
 	err = tpl.Execute(out, k)
 	require.NoError(k.Testcase, err)
-	fmt.Println(out.String())
 	return out.String()
 }
--- a/tests/testutils/kubeutils/kind_cluster.go
+++ b/tests/testutils/kubeutils/kind_cluster.go
@@ -26,6 +26,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/Masterminds/sprig/v3"
 	"github.com/docker/docker/api/types"
 	docker "github.com/docker/docker/client"
 	"github.com/stretchr/testify/require"
@@ -45,7 +46,13 @@ const kindConfigTemplate = `kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
   - role: control-plane
-    image: kindest/node:v1.26.0
+    {{- if .KubeAdmConfigPatches }}
+    kubeadmConfigPatches:
+    {{- range $configPatch := .KubeAdmConfigPatches }}
+      - |
+{{ $configPatch | indent 8 }}
+    {{- end }}
+    {{- end }}
     {{- if .ExposedPorts }}
     extraPortMappings:
     {{- range $hostPort, $containerPort := .ExposedPorts }}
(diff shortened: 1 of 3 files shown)


tests/testutils/kubeutils/kind_cluster.go
```diff
@@ -45,7 +46,13 @@ const kindConfigTemplate = `kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
   - role: control-plane
-    image: kindest/node:v1.26.0
```
rmfitzpatrick, line 48: I've been planning on expanding this to allow specifying arbitrary nodes (to allow specifying multiple control-plane and workers). I can envision cases where we use different images for different instances so would need a mechanism of specifying this field. Did you find it necessary to remove this or can it be templated instead?

atoulme: yes, instead of having this in the config yaml, I made a change to make this a parameter of the cluster config, so it can be set. The kind command takes an `--image` parameter so it can be set separately from the kind config.

atoulme: my use case right now is to use the same kind config with different images. Maybe there's more subtlety down the road for image support.

rmfitzpatrick: Do you templating here instead if it wasn't a necessary change (as in the value is only respected if using the cli option)? Seems unnecessary to add that approach and then to require shifting back in the future.

tests/testutils/kubeutils/kind_cluster_integration_test.go
```diff
@@ -42,6 +42,12 @@ func TestKindCluster(t *testing.T) {
 	portOne, portTwo := uint16(ports[0]), uint16(ports[1])
 	cluster.ExposedPorts[portOne] = 12345
 	cluster.ExposedPorts[portTwo] = 23456
+	cluster.KubeAdmConfigPatches = []string{
+		`kind: InitConfiguration
```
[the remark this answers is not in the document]
atoulme: I do not have other options that I know of.

tests/testutils/kubeutils/kind_cluster.go
rmfitzpatrick: Lost the thread on this being resolved here but since this is a node setting I think it should be templated instead of moving to the cli option.

tests/testutils/kubeutils/kind_cluster.go
```diff
@@ -198,9 +209,10 @@ func (k KindCluster) runKindCmd(args []string) {
 
 func (k KindCluster) renderConfig() string {
 	out := &bytes.Buffer{}
-	tpl, err := template.New("").Parse(kindConfigTemplate)
+	tpl, err := template.New("").Funcs(sprig.TxtFuncMap()).Parse(kindConfigTemplate)
 	require.NoError(k.Testcase, err)
 	err = tpl.Execute(out, k)
 	require.NoError(k.Testcase, err)
+	fmt.Println(out.String())
```
rmfitzpatrick, line 216: If you want to see this output it should be debug logged w/ the test logger imo

```suggestion
```

tests/testutils/kubeutils/kind_cluster.go
atoulme: Oh sorry I thought I answered I don't mind either way. Let me see to it.

PR #3761, opened 2023-10-11
Update otlp endpoint for linux installer script

jeffreyc-splunk:


atoulme (approved):

rmfitzpatrick (approved):
thanks!

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -1170,7 +1171,13 @@ parse_args_and_install() {
   fi
 
   if [ -z "$otlp_endpoint" ]; then
-    otlp_endpoint="http://${listen_interface}:4317"
+    if [ -n "$listen_interface" ]; then
+      otlp_endpoint="http://${listen_interface}:4317"
+    elif [ "$mode" = "agent" ]; then
+      otlp_endpoint="http://127.0.0.1:4317"
+    else
+      otlp_endpoint="http://0.0.0.0:4317"
```
rmfitzpatrick, line 1179: actually since this is only used for the agent exporter env var, would we ever need to use 0.0.0.0?

jeffreyc-splunk: Not sure... I guess the plan here was to just match what would be configured for the collector's otlp receiver via `SPLUNK_LISTEN_INTERFACE`.  But if the java and node.js agents only need to export to 127.0.0.1 (unless explicitly overridden with the `--otlp-endpoint` or `--listen-interface` options), then I can remove 0.0.0.0.

PR #3766, opened 2023-10-12
Add debian 12 support to installer

rmfitzpatrick:
**Description:**
These changes add debian 12 to the installer allowlist and packaging tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 💡 Enhancements 💡
 
 - (Splunk) Update golang to 1.20.10 ([#3770](https://github.com/signalfx/splunk-otel-collector/pull/3770))
+- (Splunk) Add debian 12 support to installer ([#3766](https://github.com/signalfx/splunk-otel-collector/pull/3766))
 
 ## v0.86.0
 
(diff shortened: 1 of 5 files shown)


jeffreyc-splunk:
https://github.com/signalfx/splunk-otel-collector/blob/main/internal/buildscripts/packaging/installer/install.sh#L96

Should we try upgrading fluentd to [v5](https://td-agent-package-browser.herokuapp.com/5/debian) specifically for bookworm since there are no [v4](https://td-agent-package-browser.herokuapp.com/4/debian) packages?

rmfitzpatrick:
> Should we try upgrading fluentd to [v5](https://td-agent-package-browser.herokuapp.com/5/debian) specifically for bookworm since there are no [v4](https://td-agent-package-browser.herokuapp.com/4/debian) packages?

I was about to ask you :). This would increase the scope of these changes to include vetting the existing fluentd config on v5. Anything else you can thing of?

rmfitzpatrick:
Maybe it's best to disallow fluentd at first and then adopt v5 for https://td-agent-package-browser.herokuapp.com/lts/5/amazon/2023 as well in subsequent changes.

jeffreyc-splunk:
> Maybe it's best to disallow fluentd at first and then adopt v5 for https://td-agent-package-browser.herokuapp.com/lts/5/amazon/2023 as well in subsequent changes.

Probably. You'll need to update https://github.com/signalfx/splunk-otel-collector/blob/main/internal/buildscripts/packaging/installer/install.sh#L797 and https://github.com/signalfx/splunk-otel-collector/blob/main/internal/buildscripts/packaging/tests/installer_test.py#L149.

jeffreyc-splunk (approved):
Thanks! We'll need to update the changelog, product docs, and the o11y integration wizard whenever this is released.

PR #3755, opened 2023-10-10
Replace logging with debug

theletterf:
Following the steps of https://github.com/open-telemetry/opentelemetry.io/pull/3279. Solves https://github.com/signalfx/splunk-otel-collector/issues/3708

PR #3768, opened 2023-10-13
Add comments for debug and logging exporter

theletterf:
Follows https://github.com/signalfx/splunk-otel-collector/pull/3755

### DIFF
--- a/cmd/otelcol/config/collector/agent_config.yaml
+++ b/cmd/otelcol/config/collector/agent_config.yaml
@@ -156,7 +156,9 @@ exporters:
     endpoint: "${SPLUNK_GATEWAY_URL}:4317"
     tls:
       insecure: true
-  # Debug
+  # Debug exporter
+  # If using version 0.85 or lower of the Collector
+  # replace 'debug' with 'logging'
   debug:
     verbosity: detailed
 
--- a/cmd/otelcol/config/collector/agent_config.yaml
+++ b/cmd/otelcol/config/collector/agent_config.yaml
@@ -156,10 +156,9 @@ exporters:
     endpoint: "${SPLUNK_GATEWAY_URL}:4317"
     tls:
       insecure: true
-  # Debug exporter
-  # If using version 0.85 or lower of the Collector
-  # replace 'debug' with 'logging'
-  debug:
+  # This component has been deprecated by the "debug" exporter,
+  # which will be adopted in a future release.
+  logging:
     verbosity: detailed
 
 service:
--- a/cmd/otelcol/config/collector/agent_config.yaml
+++ b/cmd/otelcol/config/collector/agent_config.yaml
@@ -158,7 +158,7 @@ exporters:
       insecure: true
   # This component has been deprecated by the "debug" exporter,
   # which will be adopted in a future release.
-  logging:
+  debug:
     verbosity: detailed
 
 service:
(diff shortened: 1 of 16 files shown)


cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -156,7 +156,9 @@ exporters:
     endpoint: "${SPLUNK_GATEWAY_URL}:4317"
     tls:
       insecure: true
-  # Debug
+  # Debug exporter
+  # If using version 0.85 or lower of the Collector
+  # replace 'debug' with 'logging'
   debug:
```
rmfitzpatrick, line 162: Since this is a bundled, default config file I suggest we postpone adopting the new name.

```suggestion
  # This component has been deprecated by the "debug" exporter,
  # which will be adopted in a future release.
  logging:
```

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -107,7 +107,9 @@ exporters:
     access_token: "${SPLUNK_ACCESS_TOKEN}"
     realm: "${SPLUNK_REALM}"
     sync_host_metadata: true
-  # Debug
+  # Debug exporter
+  # If using version 0.85 or lower of the Collector
+  # replace 'debug' with 'logging'
   #debug:
```
rmfitzpatrick, line 113:
```suggestion
  # This component has been deprecated by the "debug" exporter,
  # which will be adopted in a future release.
  # logging:
```

theletterf:
@rmfitzpatrick I only edited comments and now a couple integration tests are failing? How so?

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -156,8 +156,9 @@ exporters:
     endpoint: "${SPLUNK_GATEWAY_URL}:4317"
     tls:
       insecure: true
-  # Debug
-  debug:
+  # This component has been deprecated by the "debug" exporter,
+  # which will be adopted in a future release.
+  logging:
```
jeffreyc-splunk, line 161: I think this should be reverted:
```suggestion
  debug:
```

jeffreyc-splunk: Since `logging` was replaced by `debug` in #3755, the tests are expecting `debug` now.

rmfitzpatrick:
> @rmfitzpatrick I only edited comments and now a couple integration tests are failing? How so?

The test failures are expected since the agent config has been changed. This is what the test is for since changes to this config are meaningful for users. You can update the expected value to reflect the contents at the path detailed in the failure.

cmd/otelcol/config/collector/agent_config.yaml
rmfitzpatrick: I've requested we hold back on this change: https://github.com/signalfx/splunk-otel-collector/pull/3768#discussion_r1358264637

theletterf:
We haven't reverted the other PR yet. What's the decision here, @atoulme ?

theletterf:
@rmfitzpatrick @jeffreyc-splunk TBH, I'm a bit confused right now as to what I should change for this to go through. Ryan, you seem to suggest a partial revert of some config files: could you apply the relevant code suggestions? I feel like I'm cutting red wire from a detonator right now... :-)

jeffreyc-splunk:
> @rmfitzpatrick @jeffreyc-splunk TBH, I'm a bit confused right now as to what I should change for this to go through. Ryan, you seem to suggest a partial revert of some config files: could you apply the relevant code suggestions? I feel like I'm cutting red wire from a detonator right now... :-)

If we're not reverting the previous changes, I think the test failures would be resolved by rebasing the branch with main.

atoulme:
I would like us to be forward looking and adopt the latest, document that previous versions use the previous approach. We can discuss in team meeting and weigh pros and cons a bit more.

PR #3769, opened 2023-10-13
Attempted benchmarking plan

rmfitzpatrick:
This doc provides a rough description of the current state of* our sizing guidance and a couple targets to improve our assessment capabilities.

1. Splunk testbed benchmarking suite w/ testbed improvements.
2. A proposed utilization extension for a ~single place to look for diagnostic info as a user and developer.

### DIFF
--- a/docs/experimental/utilization.md
+++ b/docs/experimental/utilization.md
@@ -1,0 +1,136 @@
+## Enhanced Collector Sizing
+
+As an application deployed in customer environments, the collector should advertise its resource requirements for its
+capabilities. This would aid in avoiding:
+
+1. Failure scenarios and inefficiencies resulting from under and over provisioning.
+2. Guesswork from the user and developer in assessing requirements and their variation resulting from arbitrary use
+cases and changesets (scaling events, upgrades, functional changes, etc.).
+
+Since the collector is an assortment of components with specialized purposes and features, many offering infinite
+configuration permutations that influence utilization, it is generally impossible to accurately predict the requirements
+for a deployment and engagement. Despite this, if individual components make their requirements known somehow, then
+estimated tolerances are possible (though susceptible to unforeseen environmental and nonlinear scaling effects).
+
+### Current Guidance
+
+The current sizing guidelines for the collector are advertised
+[here](https://docs.splunk.com/observability/en/gdi/opentelemetry/sizing.html). The quantified guidance provided is:
+
+```
+With a single CPU core, the Collector can receive, process, or export the following:
+
+- If handling traces, 15,000 spans per second.
+- If handling metrics, 20,000 data points per second.
+- If handling logs, 10,000 log records per second, including Fluentd td-agent, which forwards logs to the fluentforward
+  receiver in the Collector.
+```
+
+and 
+
+```
+Use a ratio of one CPU to 2 GB of memory.
+
+If the Collector handles both trace and metrics data, consider both types of data when planning your deployment. For
+example, 7.5K spans per second plus 10K data points per second requires 1 CPU core.
+
+The Collector does not persist data to disk so no disk space is required.
+```
+
+Unfortunately these statements aren't based on advisable heuristics and don't account for the characteristics of default
+deployments. The remaining guidance details some use case concerns to consider without providing grounded examples or
+usable suggestions. This leaves the effort as a daunting exercise for the user.
+
+The telemetry per second guidance was generated from modified contrib testbed load tests as described
+[here](https://github.com/signalfx/splunk-otel-collector/pull/226). The
+[contrib testbed](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/testbed/README.md)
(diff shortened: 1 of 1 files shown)


docs/experimental/utilization.md
```diff
@@ -0,0 +1,136 @@
...
+### Proposed Roadmap
+
+There are a few immediate possible adoptions and improvements to the testbed that would be helpful in devising more
+accurate guidance:
+
+1. Create Splunk distro benchmarking suite that uses the contrib testbed helpers directly with Splunk-specific testdata 
+   not suitable for upstream hosting. All improvements to the testbed will target upstream however, if accepted.
+1. Use taskset or similar utility to limit collector process to a configurable number and arrangement of cores.
+1. Model from or use recorded telemetry from actual host and cloud services instead of generic load generator content.
+1. Exercise and incorporate exporter helper sending queue configuration to suite (`num_consumers`, queue size tuning,
```
omrozowicz-splunk, line 82: I think this was done in scope of this PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/27251

docs/experimental/utilization.md
```diff
@@ -0,0 +1,136 @@
...
+arbitrary configuration without requiring intensive, upfront instrumentation. I have been unable to find a robust tool
+in the go ecosystem offering these capabilities. If you know of one please share.
+
+### Proposed Roadmap
+
+There are a few immediate possible adoptions and improvements to the testbed that would be helpful in devising more
+accurate guidance:
+
+1. Create Splunk distro benchmarking suite that uses the contrib testbed helpers directly with Splunk-specific testdata 
+   not suitable for upstream hosting. All improvements to the testbed will target upstream however, if accepted.
```
dmitryax, line 79: How different the Splunk-specific test data is, and why is it not suitable for upstream hosting?

dmitryax:
BTW there is an OTel Benchmarking SIG started recently. A couple of folks from the Collector SIG are participating there. I think it makes sense to join the efforts and get some help from them instead of doing it ourselves in isolation.

dmitryax:
@omrozowicz-splunk there are a few. this one https://github.com/open-telemetry/community/issues/1617 is a good start

omrozowicz-splunk:
> @omrozowicz-splunk there are a few. this one [open-telemetry/community#1617](https://github.com/open-telemetry/community/issues/1617) is a good start

Okay, out of this thread, I see only efforts taken are for a couple of SDKs, but don't see any effort already taken specifically related to the collector, and the thread has been inactive since September. I might have just missed it, but if not - I was thinking about proposing tasks raised by @rmfitzpatrick in this PR on a contrib repo as it might be useful for everyone.

As far as I understand these are the ones that are not related to Splunk specifically:
1. Generating load based on recorded telemetry
2. Providing an option to exclude setup and teardown from benchmarking workflows
3. `Use taskset or similar utility to limit collector process to a configurable number and arrangement of cores.` we already have something to check expected CPU usage https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c26444b0069e86501f551dee5d1fbe488c94b051/testbed/testbed/child_process_collector.go#L380 and we can adjust this constraint in every test. I understand this is not enough?

docs/experimental/utilization.md
rmfitzpatrick: Splunk (default) configurations would be an example. Individual component testing would be done upstream.

docs/experimental/utilization.md
rmfitzpatrick: I'm not sure I see where that's the case, specifically parameterized `num_consumers` and `queue_size` values.

rmfitzpatrick:
>we already have something to check expected CPU usage

@omrozowicz-splunk I'm referring to using utilities to restrict the execution of the process to a specific (number of) core(s) or altering the CFS period for containers to determine impact of throughput, not utilization directly.

docs/experimental/utilization.md
omrozowicz-splunk: ah, okay, I meant [this one](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/27251/files#diff-52151c5e2de30ec441196b8816a35b17595d4172595120f8d5c8a7990a9a45a6R101), but this is otlp specific only

docs/experimental/utilization.md
```diff
@@ -0,0 +1,136 @@
...
+1. Use taskset or similar utility to limit collector process to a configurable number and arrangement of cores.
+1. Model from or use recorded telemetry from actual host and cloud services instead of generic load generator content.
+1. Exercise and incorporate exporter helper sending queue configuration to suite (`num_consumers`, queue size tuning,
+   etc.)
+1. Don't include service startup or teardown in benchmarking in workflow analysis (i.e. sleep a few seconds before
```
rmfitzpatrick, line 84: This is actually the case and the errors I've run into are per-component edge cases that don't need to be detailed here.
```suggestion
```

atoulme:
Closing, we are pursuing this approach but there's no need to commit this to the repository.

PR #3793, opened 2023-10-17
Spike with a deep cut of smart agent code deleted

atoulme:
This is a spike - running it via CI to see what tests pass, and it will get cut out into smaller PRs as needed.


atoulme:
I have removed in separate PRs the elements identified in this PR, except for filters. We will need to remove translatesfx first.

PR #3794, opened 2023-10-17
Java instrumentation example

atoulme:
Add an example showing how to inject the Java SDK into a Docker image, run it with the right env vars to point at a local collector, and show the data collected in JSON format in a text file.

### DIFF
--- a/examples/tomcat-java-instrumentation/.gitignore
+++ b/examples/tomcat-java-instrumentation/.gitignore
@@ -1,0 +1,2 @@
+splunk-otel-javaagent.jar
+output.txt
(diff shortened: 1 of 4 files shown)


crobert-1 (changes requested):
It would be good to include a short README about what this demo does and how to use it.

rmfitzpatrick:
I believe this project is a better place for efforts like this: https://github.com/signalfx/tracing-examples.

As an aside, it's not clear to me what purpose these examples have in the collector project that wouldn't be better served as readme or docs improvements for their respective components.

PR #3796, opened 2023-10-18
Remove old filters

atoulme:
Remove old filters code that was used by the Smart Agent but is no longer in use.


cmd/translatesfx/translatesfx/otel_test.go
```diff
@@ -399,219 +394,6 @@ func TestSAExcludesToExpr_MetricNamesAndDimensions(t *testing.T) {
 	assert.Equal(t, expected, ex[0])
 }
 
-func TestFilterTranslation(t *testing.T) {
```
rmfitzpatrick, line 402: I think these tests confirm that sa `metricsToInclude` and `metricsToInclude` filters are correctly translated via `saExcludesToExpr` to the transform process expr rules. Why are they being removed?

atoulme: Sorry, I'll take another look.

atoulme: OK, that's because they are used by translatesfx. I missed that. I will close this PR.

rmfitzpatrick: I think the required agent method was inlined in translatesfx to avoid exporting it and you may be able to just drop any changes to that cmd in this PR. whatever works though

PR #3800, opened 2023-10-18
feat: add all scripts to pipeline

weliasz:
**Description:** Add all scripts to pipeline

### DIFF
--- a/cmd/otelcol/config/collector/full_config_linux.yaml
+++ b/cmd/otelcol/config/collector/full_config_linux.yaml
@@ -169,11 +169,179 @@ receivers:
 
   # Enables scripted inputs receiver
   # Full configuration here: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/receiver/scriptedinputsreceiver/README.md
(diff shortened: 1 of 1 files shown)


rmfitzpatrick:
@weliasz why is this necessary when a single instance could be adequate for an example? We don't include all the variations of receivers in this or any other config.

weliasz:
Not sure if this is necessary. I just wanted to make sure we deliver full config for target deployment. If this would be done some other way and we don't need to worry about it in this file maybe this PR could be closed.

atoulme:
The intent is a bit fuzzy at this point, that's for sure. I guess if we add all those receivers we drive home the point it's ready for testing, and folks can pick it up without building their own yaml.

atoulme:
After discussing with @rmfitzpatrick I think we need to discuss this yaml file and probably just move it out of the way. My suggestion is that we accept this change and merge it, and then cut off this to an example so we can offer it for folks to try it out without having them figure out the yaml.

atoulme (approved):

PR #3820, opened 2023-10-19
fix: script name should be logged instead of script content

weliasz:
**Description:** script name should be logged instead of script content

### DIFF
--- a/internal/receiver/scriptedinputsreceiver/config.go
+++ b/internal/receiver/scriptedinputsreceiver/config.go
@@ -38,7 +38,7 @@ const (
 
 var availableScripts = func() []string {
 	var s []string
-	for sn, _ := range scripts {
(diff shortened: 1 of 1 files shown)


rmfitzpatrick (approved):

internal/receiver/scriptedinputsreceiver/config.go
```diff
@@ -38,7 +38,7 @@ const (
 
 var availableScripts = func() []string {
 	var s []string
-	for _, sn := range scripts {
+	for sn, _ := range scripts {
```
rmfitzpatrick, line 41: lint failures:

```suggestion
	for sn := range scripts {
```

apologies for this error, `scripts` used to be a slice when I first spiked the organization update.

atoulme (approved):

PR #3823, opened 2023-10-19
pass BundleDir as part of config, not as env var

atoulme:
This env var is used as a global variable in the code, when it should be passed in as parameter from the configuration.


rmfitzpatrick:
> This env var is used as a global variable in the code, when it should be passed in as parameter from the configuration.

What is the motivation for these changes? Are you intending to add support for multiple bundles by exposing the field in monitor config later on?

fwiw the SA extension still gets its default bundle w/ this env var, which we also use in the default configs. There's definitely some redundancy here but I'm not clear on what we're gaining w/ your changes.

rmfitzpatrick (approved):

atoulme:
This is just to address tech debt and simplify maintenance, as I'm trying to remove any ambiguity and make the code more maintainable.

PR #3839, opened 2023-10-24
Update OpenTelemetry Dependencies to v0.88.0

srv-gh-o11y-gdi:


crobert-1:
Investigating failures, Ryan commented [here.](https://github.com/signalfx/splunk-otel-collector/pull/3924#issuecomment-1812580652)

crobert-1:
From [docs](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/monitoring.md#queue-length):
```
The `otelcol_exporter_enqueue_failed_spans`, `otelcol_exporter_enqueue_failed_metric_points` and `otelcol_exporter_enqueue_failed_log_records` indicate the number of span/metric points/log records failed to be added to the sending queue. This may be cause by a queue full of unsettled elements, so you may need to decrease your sending rate or horizontally scale collectors.
```
For a single test under `receivers/lightprometheus:TestInternalPrometheusMetrics`, I attempted to set `queue_size` to 1, but we still didn't get the metric back. Maybe we need to increase rate of sending data in the OTLP sink test?

Once thing to consider, the test expects failed queueing of all telemetry types (metrics, traces, and logs), irrelevant of what kind of telemetry types we're sending. This is now an invalid expectation, so we will have to update the expected metrics.

crobert-1:
Test update: By setting the `collection_interval: 5ms` in the `lightprometheus` receiver's configuration, I was able to start getting the metric back. The test still fails since it's not getting failed enqueue metrics log and trace pipelines, but as described in my last comment, we should no longer expected enqueue failures for telemetry types that aren't being sent.

I think it would be good to test some numbers a bit more to find what collection interval is best. `5ms` works, but logs get pretty bloated for failed tests, so we can try to find a good balance here.

rmfitzpatrick:
@crobert-1 I think it's ok to table triggering the metrics in another pass specifically for the internal metrics tests and remove them from the expected resource_metrics for now. Reducing collection intervals doesn't seem like it would actually trigger reporting and it's a product of flake.

crobert-1:
I filed #3926 (happy to move it elsewhere if it could be tracked more easily) to properly trigger the metrics. For now I'll remove the config change in favor of simply removing missing metrics.

tests/helm/helm_test.go
```diff
@@ -110,6 +110,13 @@ splunkObservability:
   tracesEnabled: false
   logsEnabled: false
 
+clusterReceiver:
+  config:
+    receivers:
+      k8s_cluster:
+        # required until breaking opencensus.resourcetype is released
+        # https://github.com/signalfx/splunk-otel-collector-chart/commit/1c06636ec4f0beff90c54773cb3e645df048261e
+        resource_attributes: null
```
rmfitzpatrick, line 119: In the past we have added config converters to comply with breaking changes in default configuration. Since this was a shortlived item in the helm chart I'm not sure it's required but this should be documented in at least that project's changelog imo.

jvoravong (approved):

atoulme (approved):

PR #3841, opened 2023-10-24
Change installer flag to use `--with[out]-instrumentation-sdk`

atoulme:
**Description:**
Instead of using flags per language, condense the instrumentation SDK selection using a `--with[out]-instrumentation-sdk` flag.

### DIFF
--- a/internal/buildscripts/packaging/installer/install.sh
+++ b/internal/buildscripts/packaging/installer/install.sh
@@ -871,12 +871,11 @@ Auto Instrumentation:
                                         applications running as systemd services.
                                         Cannot be combined with the '--with-instrumentation' option.
                                         (default: --without-systemd-instrumentation)
-  --with[out]-java-instrumentation      Whether to enable Auto Instrumentation for Java.
-                                        (default: --with-java-instrumentation if --with-instrumentation or
+  --with[out]-instrumentation-sdk "<s>" Whether to enable Auto Instrumentation for a specific language. This option
+                                        takes a comma separated set of values representing supported
+                                        auto-instrumentation SDKs. Currently supported: java,node.
+                                        (default: --with-instrumentation-sdk "java,node" if --with-instrumentation or
                                         --with-systemd-instrumentation is also specified)
-  --with[out]-node-instrumentation      Whether to enable Auto Instrumentation for Node.js.
-                                        (default: --with-node-instrumentation if --with-instrumentation or
-                                        --with-systemd-instrumentation' is also specified)
   --npm-command "<command>"             If Auto Instrumentation for Node.js is enabled, npm is required to install the
                                         included Splunk OpenTelemetry Auto Instrumentation for Node.js package with the
                                         following command:
@@ -1196,17 +1195,31 @@ parse_args_and_install() {
       --without-systemd-instrumentation)
         with_systemd_instrumentation="false"
         ;;
-      --with-java-instrumentation)
-        with_java_instrumentation="true"
-        ;;
-      --without-java-instrumentation)
+      --with-instrumentation-sdk)
         with_java_instrumentation="false"
-        ;;
-      --with-node-instrumentation)
-        with_node_instrumentation="true"
-        ;;
-      --without-node-instrumentation)
         with_node_instrumentation="false"
+        echo "$2" | tr ',' '\n' | while read lang; do
+            if [[ "$lang" -eq "java" ]]; then
+                with_java_instrumentation="true"
+            elif [[ "$lang" -eq "node" ]]; then
+                with_node_instrumentation="true"
+            else then
+                echo "Unknown instrumentation SDK $1" >&2
+                usage
+            fi
+        done
+        ;;
+      --without-instrumentation-sdk)
+        echo "$2" | tr ',' '\n' | while read lang; do
+            if [[ "$lang" -eq "java" ]]; then
+                with_java_instrumentation="false"
+            elif [[ "$lang" -eq "node" ]]; then
+                with_node_instrumentation="false"
+            else then
+                echo "Unknown instrumentation SDK $1" >&2
+                usage
+            fi
+        done
         ;;
       --npm-command)
         npm_command="$2"
@@ -1296,13 +1309,13 @@ parse_args_and_install() {
 
   if [ "$with_instrumentation" = "true" ]; then
     if [ "$with_java_instrumentation" = "false" ] && [ "$with_node_instrumentation" = "false" ]; then
-      echo "[ERROR] The --with-instrumentation option was specified, but both --without-java-instrumentation and --without-node-instrumentation options were also specified." >&2
+      echo "[ERROR] The --with-instrumentation option was specified, but --without-instrumentation-sdk java,node options was also specified." >&2
       echo "[ERROR] At least one language must be enabled for auto instrumentation" >&2
       exit 1
     fi
   elif [ "$with_systemd_instrumentation" = "true" ]; then
     if [ "$with_java_instrumentation" = "false" ] && [ "$with_node_instrumentation" = "false" ]; then
-      echo "[ERROR] The --with-systemd-instrumentation option was specified, but both --without-java-instrumentation and --without-node-instrumentation options were also specified." >&2
+      echo "[ERROR] The --with-systemd-instrumentation option was specified, but both --without-instrumentation-sdk java,node was also specified." >&2
       echo "[ERROR] At least one language must be enabled for auto instrumentation" >&2
       exit 1
     fi
--- a/internal/buildscripts/packaging/installer/install.sh
+++ b/internal/buildscripts/packaging/installer/install.sh
@@ -1217,8 +1217,8 @@ parse_args_and_install() {
             elif [[ "$lang" -eq "node" ]]; then
                 with_node_instrumentation="false"
             else
-                echo "Unknown instrumentation SDK $1" >&2
                 usage
+                echo "[ERROR] Unknown instrumentation SDK: $lang" >&2
                 exit 1
             fi
         done
--- a/internal/buildscripts/packaging/installer/install.sh
+++ b/internal/buildscripts/packaging/installer/install.sh
@@ -1315,7 +1315,7 @@ parse_args_and_install() {
     fi
   elif [ "$with_systemd_instrumentation" = "true" ]; then
     if [ "$with_java_instrumentation" = "false" ] && [ "$with_node_instrumentation" = "false" ]; then
-      echo "[ERROR] The --with-systemd-instrumentation option was specified, but both --without-instrumentation-sdk java,node was also specified." >&2
+      echo "[ERROR] The --with-systemd-instrumentation option was specified, but --without-instrumentation-sdk java,node was also specified." >&2
       echo "[ERROR] At least one language must be enabled for auto instrumentation" >&2
       exit 1
     fi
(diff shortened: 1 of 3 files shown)


internal/buildscripts/packaging/installer/install.sh
```diff
@@ -871,12 +871,11 @@ Auto Instrumentation:
                                         applications running as systemd services.
                                         Cannot be combined with the '--with-instrumentation' option.
                                         (default: --without-systemd-instrumentation)
-  --with[out]-java-instrumentation      Whether to enable Auto Instrumentation for Java.
-                                        (default: --with-java-instrumentation if --with-instrumentation or
+  --with[out]-instrumentation-sdk "<s>" Whether to enable Auto Instrumentation for a specific language. This option
+                                        takes a comma separated set of values representing supported
+                                        auto-instrumentation SDKs. Currently supported: java,node.
+                                        (default: --with-instrumentation-sdk "java,node" if --with-instrumentation or
                                         --with-systemd-instrumentation is also specified)
```
pjanotti, line 878: Q not related to this change: if systemd instrumentation the user has to put some info to enable each instrumentation, right? I mean there is data in systemd that triggers each instrumentation, no? Here it sounds like that just adding `--with-systemd-instrumentation` you will get java and node instrumentations, without any other action. Is this correct?

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -1296,13 +1309,13 @@ parse_args_and_install() {
...
     if [ "$with_java_instrumentation" = "false" ] && [ "$with_node_instrumentation" = "false" ]; then
-      echo "[ERROR] The --with-instrumentation option was specified, but both --without-java-instrumentation and --without-node-instrumentation options were also specified." >&2
+      echo "[ERROR] The --with-instrumentation option was specified, but --without-instrumentation-sdk java,node options was also specified." >&2
       echo "[ERROR] At least one language must be enabled for auto instrumentation" >&2
       exit 1
     fi
   elif [ "$with_systemd_instrumentation" = "true" ]; then
     if [ "$with_java_instrumentation" = "false" ] && [ "$with_node_instrumentation" = "false" ]; then
-      echo "[ERROR] The --with-systemd-instrumentation option was specified, but both --without-java-instrumentation and --without-node-instrumentation options were also specified." >&2
+      echo "[ERROR] The --with-systemd-instrumentation option was specified, but both --without-instrumentation-sdk java,node was also specified." >&2
```
pjanotti, line 1318:
```suggestion
      echo "[ERROR] The --with-systemd-instrumentation option was specified, but --without-instrumentation-sdk java,node was also specified." >&2
```

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -1296,13 +1309,13 @@ parse_args_and_install() {
 
   if [ "$with_instrumentation" = "true" ]; then
     if [ "$with_java_instrumentation" = "false" ] && [ "$with_node_instrumentation" = "false" ]; then
-      echo "[ERROR] The --with-instrumentation option was specified, but both --without-java-instrumentation and --without-node-instrumentation options were also specified." >&2
+      echo "[ERROR] The --with-instrumentation option was specified, but --without-instrumentation-sdk java,node options was also specified." >&2
```
pjanotti, line 1312:
```suggestion
      echo "[ERROR] The --with-instrumentation option was specified, but --without-instrumentation-sdk java,node was also specified." >&2
```

pjanotti (approved):
LGTM with 1 Q and 2 minor suggestions.

internal/buildscripts/packaging/installer/install.sh
atoulme: the only difference between the 2 modes is how we inject the env vars. With the /etc/preload.so approach, we have a bit of C code that detects the process and adds the env vars dynamically. With the systemd approach, we add env vars as a config value for systemd services. The downside of /etc/preload.so is the security aspect of intercepting program invocation. The downside of systemd is that it only applies to systemd services. I expect customers can find something that works for them based on these compromises.

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -1196,17 +1195,31 @@ parse_args_and_install() {
...
-      --without-node-instrumentation)
         with_node_instrumentation="false"
+        echo "$2" | tr ',' '\n' | while read lang; do
+            if [[ "$lang" -eq "java" ]]; then
+                with_java_instrumentation="true"
+            elif [[ "$lang" -eq "node" ]]; then
+                with_node_instrumentation="true"
+            else then
+                echo "Unknown instrumentation SDK $1" >&2
+                usage
```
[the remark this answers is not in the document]
atoulme: usage exits 0

internal/buildscripts/packaging/installer/install.sh
pjanotti: ah, thanks - I misunderstood how instrumentation via systemd was configured.

internal/buildscripts/packaging/installer/install.sh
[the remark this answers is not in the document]
jinja2: ah, seems a bit inconsistent though given we are writing to stderr and then exiting with ok status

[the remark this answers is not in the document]
pjanotti: I like Jina's suggestion. The caller of usage should decide the exit code, usage shouldn't have an exit. The code calling usage via `--help` is assuming that https://github.com/signalfx/splunk-otel-collector/blob/b2b8efe795c73c24431e6d21f2d679a6c40c8d1b/internal/buildscripts/packaging/installer/install.sh#L1111-L1115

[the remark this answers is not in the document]
atoulme: OK. I am not sure this should be done in this PR but I can try.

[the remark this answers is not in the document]
pjanotti: Separate PR is indeed better :)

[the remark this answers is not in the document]
atoulme: ok let me land this in a PR before this one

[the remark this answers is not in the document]
atoulme: oh simple PR. Just a one line change. See #3844

pjanotti (approved):

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -871,12 +871,11 @@ Auto Instrumentation:
                                         applications running as systemd services.
                                         Cannot be combined with the '--with-instrumentation' option.
                                         (default: --without-systemd-instrumentation)
-  --with[out]-java-instrumentation      Whether to enable Auto Instrumentation for Java.
-                                        (default: --with-java-instrumentation if --with-instrumentation or
+  --with[out]-instrumentation-sdk "<s>" Whether to enable Auto Instrumentation for a specific language. This option
```
jeffreyc-splunk, line 874: Instead of a comma-separated list of languages, would it be simpler to just allow specifying the options multiple times for each language, e.g. `--with[out]-instrumentation-sdk node --with[out]-instrumentation-sdk java`?

atoulme: No, @aunshc didn't want this

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -1196,17 +1195,33 @@ parse_args_and_install() {
...
-      --without-node-instrumentation)
         with_node_instrumentation="false"
+        echo "$2" | tr ',' '\n' | while read lang; do
+            if [[ "$lang" -eq "java" ]]; then
+                with_java_instrumentation="true"
+            elif [[ "$lang" -eq "node" ]]; then
+                with_node_instrumentation="true"
+            else
+                echo "Unknown instrumentation SDK $1" >&2
+                usage
```
jeffreyc-splunk, line 1208:
```suggestion
                usage
                echo "[ERROR] Unknown instrumentation SDK $1" >&2
```
Since the usage message is pretty long and dense, the error may not be noticeable.

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -1196,17 +1195,33 @@ parse_args_and_install() {
...
+        ;;
+      --without-instrumentation-sdk)
+        echo "$2" | tr ',' '\n' | while read lang; do
+            if [[ "$lang" -eq "java" ]]; then
+                with_java_instrumentation="false"
+            elif [[ "$lang" -eq "node" ]]; then
+                with_node_instrumentation="false"
+            else
+                echo "Unknown instrumentation SDK $1" >&2
+                usage
```
jeffreyc-splunk, line 1221:
```suggestion
                usage
                echo "[ERROR] Unknown instrumentation SDK: $lang" >&2
```

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -1196,17 +1195,33 @@ parse_args_and_install() {
...
+                exit 1
+            fi
+        done
+        ;;
+      --without-instrumentation-sdk)
+        echo "$2" | tr ',' '\n' | while read lang; do
+            if [[ "$lang" -eq "java" ]]; then
+                with_java_instrumentation="false"
+            elif [[ "$lang" -eq "node" ]]; then
```
jeffreyc-splunk, line 1217:
```suggestion
            if [ "$lang" = "java" ]; then
                with_java_instrumentation="false"
            elif [ "$lang" = "node" ]; then
```

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -1196,17 +1195,33 @@ parse_args_and_install() {
...
-      --without-java-instrumentation)
+      --with-instrumentation-sdk)
         with_java_instrumentation="false"
-        ;;
-      --with-node-instrumentation)
-        with_node_instrumentation="true"
-        ;;
-      --without-node-instrumentation)
         with_node_instrumentation="false"
+        echo "$2" | tr ',' '\n' | while read lang; do
```
jeffreyc-splunk, line 1201:
```suggestion
        for lang in $(echo "$2" | tr ',' ' '); do
```
Otherwise the local var assignments below will be lost (see https://www.shellcheck.net/wiki/SC2031).

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -1196,17 +1195,33 @@ parse_args_and_install() {
...
+                with_node_instrumentation="true"
+            else
+                echo "Unknown instrumentation SDK $1" >&2
+                usage
+                exit 1
+            fi
+        done
+        ;;
+      --without-instrumentation-sdk)
+        echo "$2" | tr ',' '\n' | while read lang; do
```
jeffreyc-splunk, line 1214:
```suggestion
        for lang in $(echo "$2" | tr ',' ' '); do
```

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -871,12 +871,11 @@ Auto Instrumentation:
                                         applications running as systemd services.
                                         Cannot be combined with the '--with-instrumentation' option.
                                         (default: --without-systemd-instrumentation)
-  --with[out]-java-instrumentation      Whether to enable Auto Instrumentation for Java.
-                                        (default: --with-java-instrumentation if --with-instrumentation or
+  --with[out]-instrumentation-sdk "<s>" Whether to enable Auto Instrumentation for a specific language. This option
+                                        takes a comma separated set of values representing supported
+                                        auto-instrumentation SDKs. Currently supported: java,node.
```
jeffreyc-splunk, line 876:
```suggestion
                                        auto-instrumentation SDKs. Currently supported: java,node.
                                        Use --with-instrumentation-sdk to enable only the specified language(s),
                                        for example "--with-instrumentation-sdk java".
```

rmfitzpatrick (approved):

PR #3842, opened 2023-10-24
Revert "Bump github.com/golangci/golangci-lint in /internal/tools

jvoravong:
Revert [this](https://github.com/signalfx/splunk-otel-collector/pull/3829) change until all the lint issues can be addressed.

PR #3843, opened 2023-10-24
Simple lint fixes to help support newer version of  github.com/golangci/golangci-lint in /internal/tools

jvoravong:
**Description:** 
- First PR for the easy lint updates to support bumping the golangci-lint version.
- A second PR will come out with less easy updates that will be needed which will involve updating prometheus exporter code.

### DIFF
--- a/internal/signalfx-agent/pkg/core/dpfilters/filter.go
+++ b/internal/signalfx-agent/pkg/core/dpfilters/filter.go
@@ -13,7 +13,7 @@ import (
 type DatapointFilter interface {
 	// Matches takes a datapoint and returns whether it is matched by the
 	// filter
-	Matches(datapointFilter *datapoint.Datapoint) bool
+	Matches(dp *datapoint.Datapoint) bool
 }
 
 // BasicDatapointFilter is designed to filter SignalFx datapoint objects.  It
--- a/internal/signalfx-agent/pkg/core/dpfilters/filter.go
+++ b/internal/signalfx-agent/pkg/core/dpfilters/filter.go
@@ -13,7 +13,7 @@ import (
 type DatapointFilter interface {
 	// Matches takes a datapoint and returns whether it is matched by the
 	// filter
-	Matches(*datapoint.Datapoint) bool
+	Matches(datapointFilter *datapoint.Datapoint) bool
 }
 
 // BasicDatapointFilter is designed to filter SignalFx datapoint objects.  It
--- a/internal/signalfx-agent/pkg/core/propfilters/filter_test.go
+++ b/internal/signalfx-agent/pkg/core/propfilters/filter_test.go
@@ -16,7 +16,7 @@ func TestFilters(t *testing.T) {
 		filteredProperties := f.FilterProperties(properties)
 
 		expectedProperties := map[string]string{"replicaSet": "abc"}
-		assert.Equal(t, filteredProperties, expectedProperties)
+		assert.Equal(t, expectedProperties, filteredProperties)
 	})
 
 	t.Run("Filter a regex property name", func(t *testing.T) {
@@ -27,7 +27,7 @@ func TestFilters(t *testing.T) {
 		filteredProperties := f.FilterProperties(properties)
 
 		expectedProperties := map[string]string{"replicaSet": "abc"}
-		assert.Equal(t, filteredProperties, expectedProperties)
+		assert.Equal(t, expectedProperties, filteredProperties)
 	})
 
 	t.Run("Filter a globbed property name", func(t *testing.T) {
@@ -38,7 +38,7 @@ func TestFilters(t *testing.T) {
 		filteredProperties := f.FilterProperties(properties)
 
 		expectedProperties := map[string]string{"replicaSet": "abc"}
-		assert.Equal(t, filteredProperties, expectedProperties)
+		assert.Equal(t, expectedProperties, filteredProperties)
 	})
 
 	t.Run("Filter a single property value", func(t *testing.T) {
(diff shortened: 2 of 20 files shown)


internal/signalfx-agent/pkg/utils/filter/filter.go
```diff
@@ -15,12 +15,12 @@ import (
...
 // StringFilter matches against simple strings
 type StringFilter interface {
-	Matches(string) bool
+	Matches(stringFilter string) bool
 }
 
 // StringMapFilter matches against the values of a map[string]string.
 type StringMapFilter interface {
-	Matches(map[string]string) bool
+	Matches(stringMapeFilter map[string]string) bool
```
dmitryax, line 23: Curios which linter requires this? Seems unnecessary

atoulme, line 23:
```suggestion
	Matches(stringMapFilter map[string]string) bool
```

atoulme: I can't say if it's useful, but I don't mind if we adopt it.

atoulme (approved):
LGTM

internal/signalfx-agent/pkg/utils/filter/filter.go
pjanotti: Naming the parameters can help understand the expected effect and output of a method. However, the case at hand is not a good example of that since the type name and the method name covers it pretty well.

On a related note: the names seem auto-generated and are not good in this case. They should be either something like `input`, `inputStr`, or simply `str` and so on - the current choices give the impression that one is supposed to pass another filter.

internal/signalfx-agent/pkg/core/dpfilters/filter.go
```diff
@@ -13,7 +13,7 @@ import (
 type DatapointFilter interface {
 	// Matches takes a datapoint and returns whether it is matched by the
 	// filter
-	Matches(*datapoint.Datapoint) bool
+	Matches(datapointFilter *datapoint.Datapoint) bool
```
pjanotti, line 16: Similar to the other comment, perhaps simply `datapoint` or `dp` (since it matches the type name).

internal/signalfx-agent/pkg/core/propfilters/filter_test.go
```diff
@@ -245,8 +245,8 @@ func TestFilters(t *testing.T) {
 		filteredDimension := f.FilterDimension(dim)
 		nodeFilteredDimension := f.FilterDimension(dimNode)
 		expectedProperties := map[string]string{"replicaSet": "abc", "service_uid": "123"}
-		assert.Equal(t, filteredDimension.Properties, expectedProperties)
-		assert.Equal(t, nodeFilteredDimension.Properties, properties)
+		assert.Equal(t, expectedProperties, filteredDimension.Properties)
+		assert.Equal(t, properties, nodeFilteredDimension.Properties)
```
pjanotti, line 249: Shooting from the hip and not from your change but I am questioning if the test is checking the right thing: `properties` and `nodeProperties` are maps defined with the same keys and values. Isn't the case the if you filter both with the same filter they will have the same content? Anyway, not related to this change.

internal/signalfx-agent/pkg/monitors/subproc/core.go
```diff
@@ -192,8 +192,8 @@ func makePipes() (*messageReadWriter, io.ReadCloser, io.WriteCloser, error) {
 // subprocess.  MontiorCore.ConfigureInSubproc handles configuring the
 // subprocess monitor and collecting the result.
 type MessageHandler interface {
-	ProcessMessages(context.Context, MessageReceiver)
-	HandleLogMessage(io.Reader) error
+	ProcessMessages(context context.Context, messageReceiver MessageReceiver)
```
pjanotti, line 195: nit: idiomatic golang
```suggestion
	ProcessMessages(ctx context.Context, messageReceiver MessageReceiver)
```

pjanotti (approved):

internal/signalfx-agent/pkg/monitors/subproc/core.go
jvoravong: Update lines 195 and 196 to use ctx and r.

internal/signalfx-agent/pkg/utils/filter/filter.go
jvoravong: Linter: [inamedparam](https://github.com/signalfx/splunk-otel-collector/actions/runs/6618591762/job/17977433653#step:6:174)
The linter wants more context to be added where possible. Decided to go with short names for this PR. Let's consider using more descriptive names in future code contributions.

Updated the names to be str for StringFilter and StringMapFilter.

internal/signalfx-agent/pkg/core/propfilters/filter_test.go
jvoravong: I think line [228](https://github.com/signalfx/splunk-otel-collector/blob/db32e9b91f8b065fcb4b65f880420295f9d307d7/internal/signalfx-agent/pkg/core/propfilters/filter_test.go#L228) calling [New()](https://github.com/signalfx/splunk-otel-collector/blob/b2b8efe795c73c24431e6d21f2d679a6c40c8d1b/internal/signalfx-agent/pkg/core/propfilters/dimfilter.go#L32) with two different arguments for propertyNames and dimensionValues shows some insight into what is being tested here.

rmfitzpatrick (approved):

rmfitzpatrick:
Mind updating the dep w/ these changes?

jvoravong:
@rmfitzpatrick,
- Reverted "Bump github.com/golangci/golangci-lint in /internal/tools #3842" as part of https://github.com/signalfx/splunk-otel-collector/pull/3842 yesterday.
- Still need to get a second PR out with less easy updates that will be needed for upgrading the lint dependency.

PR #3848, opened 2023-10-27
Use underscore for rpm test packages

jeffreyc-splunk:
Currently, the rpm packages built and released to our test repos in artifactory from the `main` branch have `~post` appended to the version. However, the `~` in the version forces the sorting to be lower than previously released packages with the same base version, i.e. `0.86.0~post` < `0.86.0`. Use `_post` instead to enforce the correct sorting order for test packages.

### DIFF
--- a/instrumentation/packaging/fpm/rpm/build.sh
+++ b/instrumentation/packaging/fpm/rpm/build.sh
@@ -27,8 +27,8 @@ if [[ -z "$VERSION" ]]; then
     VERSION="$( get_version )"
 fi
 
-# rpm doesn't like dashes in the version, replace with tildas
-VERSION="${VERSION/'-'/'~'}"
+# rpm doesn't like dashes in the version, replace with underscore
+VERSION="${VERSION/'-'/'_'}"
 VERSION="${VERSION#v}"
 
 buildroot="$(mktemp -d)"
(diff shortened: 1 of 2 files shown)


jeffreyc-splunk:

jeffreyc-splunk:
recheck

instrumentation/packaging/fpm/rpm/build.sh
```diff
@@ -27,8 +27,8 @@ if [[ -z "$VERSION" ]]; then
     VERSION="$( get_version )"
 fi
 
-# rpm doesn't like dashes in the version, replace with tildas
-VERSION="${VERSION/'-'/'~'}"
+# rpm doesn't like dashes in the version, replace with underscore
+VERSION="${VERSION/'-'/'_'}"
```
crobert-1, line 31: Is there any chance someone's relying on the version to be a `~` here? I'm wondering if we're going to break anyone's automation.

This probably isn't a big deal either way, as long as it's called out in the release notes.

crobert-1 (approved):

instrumentation/packaging/fpm/rpm/build.sh
jeffreyc-splunk: This change only affects packages built from `main` for dev/test. It does not affect any packages for actual releases/tags.

jvoravong (approved):

PR #3854, opened 2023-10-27
use ampere runners for arm64 installer tests

atoulme:
Attempt to use ampere runners for arm64 installer tests.


atoulme:
What a twist :) Python versions available with github actions do not support Linux ARM: https://github.com/actions/python-versions

PR #3873, opened 2023-10-30
testutils: add more manifest types and rendering improvements

rmfitzpatrick:
These changes add deployment, service, and namespace manifest helpers while also moving validation error checks to the `Render()` method. They also add a new `RenderAll()` helper to avoid boilerplate for creating multiple resources.

### DIFF
--- a/tests/testutils/kubeutils/manifests/deployment.go
+++ b/tests/testutils/kubeutils/manifests/deployment.go
@@ -20,19 +20,21 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// DaemonSet is a "k8s.io/api/apps/v1.DaemonSet" convenience type
-type DaemonSet struct {
+// Deployment is a "k8s.io/api/apps/v1.Deployment" convenience type
+type Deployment struct {
+	Labels         map[string]string
+	MatchLabels    map[string]string
 	Namespace      string
 	Name           string
 	ServiceAccount string
-	Labels         map[string]string
 	Containers     []corev1.Container
 	Volumes        []corev1.Volume
+	Replicas       int
 }
 
-const daemonSetTemplate = `---
+const deploymentTemplate = `---
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
 metadata:
 {{- if .Name }}
   name: {{ .Name }}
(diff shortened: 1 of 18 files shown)


tests/testutils/kubeutils/manifests/deployment.go
```diff
@@ -0,0 +1,92 @@
...
+package manifests
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Deployment is a "k8s.io/api/apps/v1.Deployment" convenience type
+type Deployment struct {
```
jinja2, line 24: Why do we prefer object type helpers like this to render the yaml manifest, instead of using yamls with optional go-templating in them when needed? Imo, these structs can become hard to maintain as new fields need to be exposed and readability is worse when compared to yaml.

rmfitzpatrick: This is a good question and I don't have a good answer for you, which probably suggests I'm missing the better way. What you are describing was what I tried to implement with these helpers, but by adopting the api types directly I realize that there is a tradeoff in terseness. I've also realized that these updates are only partially adopted so I'll make another pass to hopefully improve readability in the actual tests.

I would be interested in your suggestions for an interface that allows value accessibility (both in referencing existing and in setting desired) outside of the string representation.

rmfitzpatrick: Updated for more thorough adoption. There's definitely some DRY violations, but I think after https://github.com/signalfx/splunk-otel-collector/pull/3875 lands I can start to add a new collector helper type that takes care of the repeated resources like the one that exists for the container and subprocess.

rmfitzpatrick: A toy example would be something like this that encapsulates all the yaml and templating items for you while retaining access to the `Name` field for all ns objects.

```go
	var namespaces []manifests.Namespace
	for i := 0; i < 100 ; i ++ {
		namespaces = append(namespaces, manifests.Namespace{Name: randomName()})
	}
	cluster.Apply(manifests.RenderAll(t, namespaces...))
```

jinja2: I am gonna think more about how/if we can replace the helpers w/o needing to write manifests from scratch/dealt with runtime objects. Going to resolve this for now since I have a better understanding of the motivation and no good alternative at the moment.

jinja2 (approved):

PR #3874, opened 2023-10-30
Support confmap providers in discovery config

samiura:
Adding option for arbitrary env var to authenticate for dbs on discovery mode

**Link to Splunk idea:** 

### DIFF
--- a/internal/confmapprovider/discovery/discoverer.go
+++ b/internal/confmapprovider/discovery/discoverer.go
@@ -33,6 +33,8 @@ import (
 	"go.opentelemetry.io/collector/config/configtelemetry"
 	"go.opentelemetry.io/collector/confmap"
 	"go.opentelemetry.io/collector/confmap/converter/expandconverter"
+	"go.opentelemetry.io/collector/confmap/provider/envprovider"
+	"go.opentelemetry.io/collector/confmap/provider/yamlprovider"
 	"go.opentelemetry.io/collector/consumer"
 	otelcolextension "go.opentelemetry.io/collector/extension"
 	"go.opentelemetry.io/collector/otelcol"
@@ -54,6 +56,11 @@ import (
 const (
 	durationEnvVar = "SPLUNK_DISCOVERY_DURATION"
 	logLevelEnvVar = "SPLUNK_DISCOVERY_LOG_LEVEL"
+)
+
+var (
+	yamlProvider = yamlprovider.New()
+	envProvider  = envprovider.New()
 )
 
 // discoverer provides the mechanism for a "preflight" collector service
@@ -115,6 +122,35 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
 	}
 	d.propertiesConf = d.propertiesConfFromEnv()
 	return d, nil
+}
+
+func (d *discoverer) resolveConfig(discoveryReceiverRaw map[string]any) (*confmap.Conf, error) {
+	providers := map[string]confmap.Provider{
+		"yaml": yamlProvider,
+		"env":  envProvider,
+	}
+	out, err := yaml.Marshal(discoveryReceiverRaw)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal discovery receiver config for uri: %w", err)
+	}
+	uris := []string{fmt.Sprintf("yaml:%s", out)}
+	resolver, err := confmap.NewResolver(confmap.ResolverSettings{
+		URIs:       uris,
+		Providers:  providers,
+		Converters: []confmap.Converter{d.expandConverter},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create a resolver from the given uris. %w", err)
+	}
+
+	conf, err := resolver.Resolve(context.Background())
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve configuration from the resovler %w", err)
+	}
+	if err = resolver.Shutdown(context.Background()); err != nil {
+		d.logger.Warn("error shutting down resolver", zap.Error(err))
+	}
+	return conf, nil
 }
 
 func (d *discoverer) propertiesConfFromEnv() *confmap.Conf {
@@ -295,9 +331,10 @@ func (d *discoverer) createDiscoveryReceiversAndObservers(cfg *Config) (map[comp
 		}
 
 		discoveryReceiverRaw["receivers"] = receivers
-		discoveryReceiverConfMap := confmap.NewFromStringMap(discoveryReceiverRaw)
-		if err = d.expandConverter.Convert(context.Background(), discoveryReceiverConfMap); err != nil {
-			return nil, nil, fmt.Errorf("error converting environment variables in receiver config: %w", err)
+		discoveryReceiverConfMap, err := d.resolveConfig(discoveryReceiverRaw)
+
+		if err != nil {
+			return nil, nil, fmt.Errorf("error preparing discovery receiver config: %w", err)
 		}
 
 		if err = component.UnmarshalConfig(discoveryReceiverConfMap, discoveryReceiverConfig); err != nil {
--- a/internal/confmapprovider/discovery/README.md
+++ b/internal/confmapprovider/discovery/README.md
@@ -170,12 +170,16 @@ These properties can be in `config.d/properties.discovery.yaml` or specified at 
 
 You can also specify a `--discovery-properties=<filepath.yaml>` argument to disregard `config.d/properties.discovery.yaml` properties and load properties not to be shared with another Collector service, while still benefiting from existing discovery component definitions.
 
-The `config.d/properties.discovery.yaml` file supports specifying the property values directly as well within a mapped form:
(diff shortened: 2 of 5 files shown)


internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +119,38 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
 	return d, nil
 }
 
+func (d *discoverer) discoveryAuthConfFromEnv(discoveryReceiverRaw map[string]any) (*confmap.Conf, error) {
```
rmfitzpatrick, line 122: nit, since this is a general config resolver pass I don't think we need to focus on the auth application. Maybe something like this?

```suggestion
func (d *discoverer) resolveConfig(discoveryReceiverRaw map[string]any) (*confmap.Conf, error) {
```

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +119,38 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
...
+		d.logger.Warn(fmt.Sprintf("Failed to create resolver from yaml provider. %v", err))
+	}
+	uris := []string{fmt.Sprintf("yaml:%s", out[:])}
+	resolver, err := confmap.NewResolver(confmap.ResolverSettings{
+		URIs:       uris,
+		Providers:  providers,
+		Converters: []confmap.Converter{d.expandConverter},
+	})
+	if err != nil {
+		d.logger.Error(fmt.Sprintf("Failed to create a resolver from the given uris:%s and providers:%v. %v", uris, providers, err))
```
rmfitzpatrick, line 141: A lot of the config structs use https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configopaque/opaque.go but I think this still can log secrets. Maybe just include the err to restrict what comes out? (not sure if the err will contain the full input though).

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +119,38 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
 	return d, nil
 }
 
+func (d *discoverer) discoveryAuthConfFromEnv(discoveryReceiverRaw map[string]any) (*confmap.Conf, error) {
+	yamlProvider := yamlprovider.New()
+	envProvider := envprovider.New()
```
rmfitzpatrick, line 124: I believe these can be globals rather than created each call.

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +119,38 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
...
+
+	providers := map[string]confmap.Provider{
+		"yaml": yamlProvider,
+		"env":  envProvider,
+	}
+	out, err := yaml.Marshal(discoveryReceiverRaw)
+	if err != nil {
+		d.logger.Warn(fmt.Sprintf("Failed to create resolver from yaml provider. %v", err))
```
rmfitzpatrick, line 132: this is something we need to bail on:

```suggestion
	if err != nil {
		return nil, fmt.Errorf("failed to marshal discovery receiver config for uri: %w", err)
```

internal/confmapprovider/discovery/discoverer.go
samiura: I have read the CLA Document and I hereby sign the CLA

internal/confmapprovider/discovery/discoverer.go
samiura: err doesn't include the secrets. I will just log err and yes providers do not give us much in this case.

internal/confmapprovider/discovery/discoverer.go
samiura: got it.

samiura:

internal/confmapprovider/discovery/discoverer.go
samiura: :done

rmfitzpatrick:
@samiura do you mind updating this existing test to confirm that unexpanded provider directives are used in the initial config? I think replacing this w/ `${env:LABEL_ONE_VALUE}` could be adequate: https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/discoverymode/testdata/host-observer-config.d/receivers/internal-prometheus.discovery.yaml#L9

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +124,34 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
 	return d, nil
 }
 
+func (d *discoverer) resolveConfig(discoveryReceiverRaw map[string]any) (*confmap.Conf, error) {
+
```
rmfitzpatrick, line 128: nit

```suggestion
```

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +124,34 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
...
+		Providers:  providers,
+		Converters: []confmap.Converter{d.expandConverter},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create a resolver from the given uris. %v", err)
+	}
+
+	conf, err := resolver.Resolve(context.Background())
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve configuration as a conf from the resovler %v", err)
```
rmfitzpatrick, line 149:
```suggestion
		return nil, fmt.Errorf("failed to resolve configuration as a conf from the resolver %w", err)
```

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +124,34 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
...
+	}
+	resolver.Shutdown(context.Background())
```
rmfitzpatrick, line 151: We should at least log this error

samiura: done.

internal/confmapprovider/discovery/discoverer.go
samiura: good catch.

internal/confmapprovider/discovery/discoverer.go
samiura: removed.

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +124,34 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
...
+		Providers:  providers,
+		Converters: []confmap.Converter{d.expandConverter},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create a resolver from the given uris. %w", err)
+	}
+
+	conf, err := resolver.Resolve(context.Background())
+	if err != nil {
+		d.logger.Error(fmt.Sprintf("failed to resolve configuration as a conf from the resovler %v", err))
```
rmfitzpatrick, line 148: we're already returning and it's further raised by the caller
```suggestion
```

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +124,34 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
...
+		return nil, fmt.Errorf("failed to resolve configuration as a conf from the resovler %w", err)
```
rmfitzpatrick, line 149: think a commit was lost

```suggestion
		return nil, fmt.Errorf("failed to resolve configuration from the resolver %w", err)
```

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -117,6 +124,34 @@ func newDiscoverer(logger *zap.Logger) (*discoverer, error) {
...
+	}
+	resolver.Shutdown(context.Background())
```
rmfitzpatrick, line 151:
```suggestion
	if err = resolver.Shutdown(context.Background()); err != nil {
		d.logger.Error(fmt.Errorf("error shutting down resolver: %w", err))
	}
```

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -295,9 +330,10 @@ func (d *discoverer) createDiscoveryReceiversAndObservers(cfg *Config) (map[comp
 		}
 
 		discoveryReceiverRaw["receivers"] = receivers
-		discoveryReceiverConfMap := confmap.NewFromStringMap(discoveryReceiverRaw)
-		if err = d.expandConverter.Convert(context.Background(), discoveryReceiverConfMap); err != nil {
-			return nil, nil, fmt.Errorf("error converting environment variables in receiver config: %w", err)
+		discoveryReceiverConfMap, err := d.resolveConfig(discoveryReceiverRaw)
+
+		if err != nil {
+			return nil, nil, fmt.Errorf("error converting environment variables into a receiver config: %w", err)
```
rmfitzpatrick, line 336:
```suggestion
		if err != nil {
			return nil, nil, fmt.Errorf("error preparing discovery receiver config: %w", err)
```

PR #3875, opened 2023-10-30
testutils: add cluster.WaitForPods()

rmfitzpatrick:
These changes add a new kind cluster method to avoid boilerplate in finding running pods. They also make sure that all the expected control plane pods are running before returning from `cluster.Create()` to avoid race conditions when evaluating telemetry from the control plane.

### DIFF
--- a/tests/testutils/kubeutils/kind_cluster.go
+++ b/tests/testutils/kubeutils/kind_cluster.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"strings"
 	"text/template"
 	"time"
@@ -30,6 +31,8 @@ import (
 	docker "github.com/docker/docker/client"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/multierr"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
@@ -96,6 +99,13 @@ func (k *KindCluster) Create() {
 	require.NoError(k.Testcase, err)
 	k.Clientset, err = kubernetes.NewForConfig(restConfig)
 	require.NoError(k.Testcase, err)
+
+	for _, pod := range []string{
+		"kube-apiserver", "kube-controller-manager", "kube-scheduler", "etcd-cluster", "coredns", "kindnet", "kube-proxy",
+	} {
+		k.WaitForPods(pod, "kube-system", 2*time.Minute)
+	}
+	k.WaitForPods("local-path-provisioner", "local-path-storage", 2*time.Minute)
 }
 
 func (k *KindCluster) Teardown() {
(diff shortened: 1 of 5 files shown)


tests/testutils/kubeutils/kind_cluster.go
```diff
@@ -152,6 +162,54 @@ func (k KindCluster) Apply(manifests string) (stdOut, stdErr bytes.Buffer, err e
...
+				matched = append(matched, pod)
+			}
+		}
+
+		if len(matches) == 0 {
+			running = matched
+			return true
+		}
+		return false
+	}, duration, time.Second)
```
atoulme, line 208: I guess we could check that duration is greater than a second if we fix polling to a second here. Not a big deal, but line 171 also sets 5 seconds for the timeout, should it be changed to match the poll interval?

rmfitzpatrick: Sorry, what's the issue here?

atoulme: I think if duration is less than a second, then polling never runs. The context is set to 5s, so we are running an operation with a timeout of 5s with a polling period of one second.

rmfitzpatrick: I think the implementation of the testify Eventually asserter ensures there's only one invocation at a time and its actual interval is the max of the ticker and call durations.

If someone expects a timeout duration for less than a second I think the resulting assertion failure would be helpful enough to communicate that their expectations don't meet the execution (so the duration would have to be bumped up to give k8s time to get the pod running).

tests/testutils/kubeutils/kind_cluster.go
```diff
@@ -152,6 +162,54 @@ func (k KindCluster) Apply(manifests string) (stdOut, stdErr bytes.Buffer, err e
 	return k.runKubectl(stdin, "apply", "-f", f.Name())
 }
 
+func (k KindCluster) WaitForPods(podNameRegex, namespaceName string, duration time.Duration) []*corev1.Pod {
```
jinja2, line 165: To differentiate from pod being Ready
```suggestion
func (k KindCluster) WaitForPodsRunning(podNameRegex, namespaceName string, duration time.Duration) []*corev1.Pod {
```

 
I see us needing to wait for pods to be ready for traffic vs just in Running phase. I think at least for the collector we should wait for pod Ready if we have healthcheck extension + probes set up. In case you want to add the ready check in this PR, adding the logic (haven't tested it). 

```
import api "k8s.io/kubernetes/pkg/apis/core"

func isPodReady(pod *api.Pod) bool {
	status := pod.Status
	if status == nil {
		return false
	}
	for i := range status.Conditions {
		if status.Conditions[i].Type == api.PodReady {
			return status.Conditions[i].Status == api.ConditionTrue
		}
	}
	return false
}
```

rmfitzpatrick: what about breaking out a `WaitForPodFn(podNameRegex, namespaceName string, duration time.Duration, func(*corev1.Pod) bool)` and updating `WaitForPods` to use the ready status?

rmfitzpatrick: The issue with using ready for collectors in discovery is that their service extensions have yet to be running

jinja2: Ah, that makes sense

jinja2 (approved):

PR #3877, opened 2023-10-31
Respect JAVA_HOME in smartagent receiver

rmfitzpatrick:
**Description:**
These changes make sure that `JAVA_HOME` is only set in the smart agent receiver if currently unset. This breaking change will avoid the scenario where the jmx receiver's target java version isn't configurable otherwise.

**Documentation:**
Since this is a bug fix I haven't documented more than a changelog entry but appreciate suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


atoulme (approved):
Change looks good to me.

PR #3878, opened 2023-11-01
Use "TARGET_ENV: velocity" for msi signing job

jeffreyc-splunk:
Without this var, the "signed" MSI artifact is corrupted/uninstallable.


jeffreyc-splunk:
Signing service has been updated. Not required, for now.

PR #3879, opened 2023-11-02
fix: use proper key formating as per stanza persister

VihasMakwana:
**Description:** Migrating offsets from SCK to SCK-Otel doesn't work. This is because of incorrect keys we use to populate the boltdb cache. 
The proper keys should be
`file_input.knownFiles`, `journald_input.lastReadCursor`
But we prepend `$.` to them and it causes log replication. 
Refer [scoped persister](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.88.0/pkg/stanza/operator/persister.go#L30-L38) for the current format.

~~**Note: I'm currently in the process of testing this on my Mac. I'd appreciate you're review on this by then**~~
Testing: Manually tested it on SCK and SCK-OTeL with locally built image

### DIFF
--- a/cmd/migratecheckpoint/log_library.go
+++ b/cmd/migratecheckpoint/log_library.go
@@ -30,8 +30,8 @@ type Fingerprint struct {
 
 type Reader struct {
 	Fingerprint    *Fingerprint
-	Offset         int64
 	FileAttributes map[string]any
+	Offset         int64
 }
 
 type journaldCursor struct {
--- a/cmd/migratecheckpoint/log_library.go
+++ b/cmd/migratecheckpoint/log_library.go
@@ -29,8 +29,9 @@ type Fingerprint struct {
 }
 
 type Reader struct {
-	Fingerprint *Fingerprint
-	Offset      int64
+	Fingerprint    *Fingerprint
+	Offset         int64
+	FileAttributes map[string]any
 }
 
 type journaldCursor struct {
--- a/cmd/migratecheckpoint/main.go
+++ b/cmd/migratecheckpoint/main.go
@@ -86,7 +86,7 @@ func (m *Migrator) MigrateContainerPos(lines []string) {
 	defer client.Close()
 	_, buf := m.ConvertFilePos(lines)
 
-	err = client.Set("$.file_input.knownFiles", buf.Bytes())
+	err = client.Set("file_input.knownFiles", buf.Bytes())
 	if err != nil {
 		log.Printf("error storing container checkpoints: %v", err)
 	}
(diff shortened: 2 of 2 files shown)


VihasMakwana:

VihasMakwana:
Screenshot 2023-11-03 at 5 22 48 PM

In above image SCK ingested till `Hello 46`.
I continued with SCK-OTeL and it continued from `Hello 47`. No duplicates.

VihasMakwana:
Following is the current behavior (buggy migration). As you can see, each event is read twice. Event though fluentd left at `Hello 46`. This is because of the incorrect key we use in boltdb.

Screenshot 2023-11-03 at 5 27 52 PM

cmd/migratecheckpoint/log_library.go
```diff
@@ -29,8 +29,9 @@ type Fingerprint struct {
 }
 
 type Reader struct {
-	Fingerprint *Fingerprint
-	Offset      int64
+	Fingerprint    *Fingerprint
+	Offset         int64
+	FileAttributes map[string]any
```
atoulme, line 34: what FileAttributes doing?

cmd/migratecheckpoint/main.go
```diff
@@ -110,7 +110,7 @@ func (m *Migrator) MigrateCustomPos(matches []string) {
 				if err != nil {
 					log.Printf("error creating a new DB client for host file checkpoints: %v", err)
 				}
-				err = client.Set("$.file_input.knownFiles", buf.Bytes())
+				err = client.Set("file_input.knownFiles", buf.Bytes())
```
atoulme, line 113: Have we _always_ had this issue? Is there a breaking change in stanza that occurred?

cmd/migratecheckpoint/log_library.go
VihasMakwana: We're trying to convert fluentd's format to the following struct as per OTeL's format. We get `Offset` and `Fingerprint` by reading the fluentd's pos_file. But the field `FileAttributes` is left nil. 
This causes a panic when the filelogreceiver tries to access this field while decoding it (causes nil reference). Setting it to an empty map resolves this.

Screenshot 2023-11-04 at 1 28 06 AM

atoulme: Makes sense, thanks!

cmd/migratecheckpoint/main.go
VihasMakwana: Well, not sure tbh. Need to dig in through commit history.

VihasMakwana: We don't have any tests for migration from SCK. I'll create a PR in the chart repo covering this part.
If I understand correctly, we're moving to Golang-based functional tests, right?

VihasMakwana: @atoulme here's a PR to add functional test cases covering this. https://github.com/signalfx/splunk-otel-collector-chart/pull/1024
PTAL! Thanks

atoulme:
Please fix the CI and we can look at getting this in.

VihasMakwana:
> Please fix the CI and we can look at getting this in.
Done.

atoulme:
Please rebase to get the fix on hadoop tests.

atoulme:
I rebased for you.

atoulme (approved):

PR #3898, opened 2023-11-07
[JMX Path] Resolve msi installation CI test failure

crobert-1:

We've been getting [failures](https://github.com/signalfx/splunk-otel-collector/actions/runs/6789950948/job/18461298543) in the windows packaging test `msi-test`.

Example error:
```
Resolve-Path: D:\a\_temp\d30f3d1d-1fed-4469-b6f0-8903e5936f43.ps1:12
Line |
  12 |  Resolve-Path -Path "$env:SYSTEMDRIVE\opt\opentelemetry-java-contrib-j …
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Cannot find path 'C:\opt\opentelemetry-java-contrib-jmx-metrics.jar' because it does not exist.

Error: Process completed with exit code 1.
```
I installed the MSI in a windows environment and found that the file referenced was in fact installed properly. Update test to check each disk for the installed MSI.


crobert-1:

crobert-1:
Looks like the timeout increase didn't work work at this point. Looking at [past runs](https://github.com/signalfx/splunk-otel-collector/actions/workflows/win-package-test.yml) we can see this was succeeding as of last week, and the download URL for the JMX gatherer hasn't changed. Maybe there's some kind of environment change impacting this. [Here was the first failure](https://github.com/signalfx/splunk-otel-collector/actions/runs/6720318762) I could find, but the change is unrelated to this. I'll continue investigating.

pjanotti:
Alternatively: does it need to be on the `%SystemDrive%\opt`? I guess that it needs a known location because at some time it also needs to set the corresponding env var, but, it shouldn't be a problem if that is under `%ProgramFiles%`. Am I missing something?

Related links:
* https://wixtoolset.org/docs/v3/xsd/wix/directory/
* The built in variable `WindowsVolume` can be helpful https://wixtoolset.org/docs/v3/bundle/bundle_built_in_variables/
* https://learn.microsoft.com/en-us/windows/win32/msi/windowsvolume?redirectedfrom=MSDN

crobert-1:
> Alternatively: does it need to be on the `%SystemDrive%\opt`? I guess that it needs a known location because at some time it also needs to set the corresponding env var, but, it shouldn't be a problem if that is under `%ProgramFiles%`. Am I missing something?

The goal here was to match the [default path](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/jmxreceiver#jar_path-default-optopentelemetry-java-contrib-jmx-metricsjar) of the `jar` as closely as possible.

I assume by `specify a disk in the wix source`, you're referencing when we check the path, not the when building the msi, right?

pjanotti:
> I assume by specify a disk in the wix source, you're referencing when we check the path, not the when building the msi, right?

I mean the MSI source, here: https://github.com/signalfx/splunk-otel-collector/blob/cb8e08e18b79399c1a2cdd3c8390cbf3d3cf0fde/internal/buildscripts/packaging/msi/splunk-otel-collector.wxs#L49-L53

This snippet is not specifying anything for the disk, it could use some of the pre-defined values like in https://github.com/signalfx/splunk-otel-collector/blob/cb8e08e18b79399c1a2cdd3c8390cbf3d3cf0fde/internal/buildscripts/packaging/msi/splunk-otel-collector.wxs#L14-L18

pjanotti:
Regarding the default path. On Windows `/opt/...` will be treated as relative to the current disk, i.e.: the current drive when the process was started.. An environment variable using standard env vars like `ProgramData` won't be subject to this type of issue. It seems to me that the default location of the jar should be different on Windows to avoid this issue.

crobert-1:
> Regarding the default path. On Windows `/opt/...` will be treated as relative to the current disk, i.e.: the current drive when the process was started.. An environment variable using standard env vars like `ProgramData` won't be subject to this type of issue. It seems to me that the default location of the jar should be different on Windows to avoid this issue.

I was able to think about it a bit more, I think that's a good idea. My main concern here was that the existing documentation on the JMX receiver would need to reference this (for the Splunk distribution), but it's pretty out of date. I'll need to make a note of this somewhere obvious so users are aware of this. I'll work on updating to a directory that makes sense here. Thanks!

internal/buildscripts/packaging/msi/splunk-otel-collector.wxs
```diff
@@ -43,14 +43,12 @@
                         <RegistryValue Type="string" Name="SPLUNK_CONFIG" Value="[CommonAppDataFolder]Splunk\OpenTelemetry Collector\agent_config.yaml"/>
                      </RegistryKey>
                   </Component>
+                  <Component Id="JmxMetricsJarComponent" Guid="c2aca15f-813a-4497-9bad-1c97e7bff618">
+                    <File Id="JmxMetricsJar" Name="opentelemetry-java-contrib-jmx-metrics.jar" KeyPath="yes" Source="$(var.JmxMetricsJar)"/>
```
rmfitzpatrick, line 47: Are you introducing breaking changes and doc/usage requirements as the outcome of an attempt to reduce flake?

crobert-1: I think it's also a matter of best practices on Windows. I'm not very familiar with paths on Windows and where files should be installed, so I'm up for correction here. It sounded like the program files directory where the other files installed by the `msi` was a more appropriate place.

This would be a breaking change for users though, you're correct. I'll have to add a note in the changelog if this is the accepted way to go.

The URL isn't actually documented anywhere that I've found. [Splunk's JMX receiver documentation](https://docs.splunk.com/observability/en/gdi/monitors-languages/jmx.html#jmx) is still pointing at the smart agent receiver. It's an ongoing effort to migrate the documentation, so this doesn't have a documentation impact since either path would need to be added to documentation.

rmfitzpatrick: Ok, separating concerns here:

1. I think @pjanotti found the test failure cause, which is likely a real bug in our wix use where we aren't specifying the opt directory's drive and there's a race condition. Doing so would be the fix I'd advocate for, though I may not be tracking what's up here.
2. Windows conventions and the jmx receiver: this is an issue for the contrib component where the default path would be based on the detected goos/runtime. IMO as a distribution (wrapper) for the receiver we have an obligation to fix problems, conform to their implementation transparently, or if not possible/advisable, automatically follow required practices on the user's behalf (maybe w/ a config converter?). I think that's not needed here, or at least would have larger scope than warranted for the observed failures.

crobert-1: 1. @pjanotti would it be acceptable to you to keep the jmx at the same path, and find a way to determinately install and check a specific drive to resolve this? Or is it really bad practice to have the path be `Drive:\opt\`?
2. Yes, and for context, the receiver doesn't do any interesting logic for the path, it doesn't account for OS. So I agree this would be an issue to fix in `contrib`, not here.

pjanotti: This is a "simple problem" which requires thinking carefully about the resolution... At first my preferential fix is to make the change in contrib, defaulting the env var according to `runtime.GOOS` and that should be a path based on `ProgramFiles`. Meanwhile, since that is going to take a bit of time, we can get rid of the test flakiness by updating the check (something using `Test-Path` on all drives on the box) - let me sketch such solution.

pjanotti: Something like the code below should do it

```PowerShell
$found_jmx = $false
(Get-PSDrive -PSProvider FileSystem).Root | ForEach-Object {
    $found_jmx = $found_jmx -Or (Test-Path (Join-Path "$_" "opt\opentelemetry-java-contrib-jmx-metrics.jar"))
}
if (-not $found_jmx) { throw "Could not find expected jar file" }
```

rmfitzpatrick (approved):

PR #3902, opened 2023-11-09
[Tanzu Tile] Fix proxy exclusions bug

crobert-1:

The NO_PROXY environment variable was previously not being set properly by the Tanzu Tile configuration option. This was because the input variable's name was being automatically converted from otel_proxy_no to otel_proxy_false internally, but the variable setting logic was still checking otel_proxy_no. Change name to otel_proxy_exclusions to avoid this.

Manually tested both the BOSH job and Tanzu Tile to ensure changes work properly.

Added some more

### DIFF
--- a/deployments/cloudfoundry/bosh/jobs/splunk-otel-collector/templates/splunk-otel-collector.conf.erb
+++ b/deployments/cloudfoundry/bosh/jobs/splunk-otel-collector/templates/splunk-otel-collector.conf.erb
@@ -1,9 +1,10 @@
 <% if p('otel.proxy.http', '').length > 0 %>
 export HTTP_PROXY=<%= p('otel.proxy.http') %>
+export http_proxy=<%= p('otel.proxy.http') %>
 <% end %>
 <% if p('otel.proxy.https', '').length > 0 %>
 export HTTPS_PROXY=<%= p('otel.proxy.https') %>
 <% end %>
-<% if p('otel.proxy.no', '').length > 0 %>
-export NO_PROXY=<%= p('otel.proxy.no') %>
+<% if p('otel.proxy.exclusions', '').length > 0 %>
+export NO_PROXY=<%= p('otel.proxy.exclusions') %>
 <% end %>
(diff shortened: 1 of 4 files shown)


deployments/cloudfoundry/bosh/jobs/splunk-otel-collector/templates/splunk-otel-collector.conf.erb
```diff
@@ -1,9 +1,10 @@
 <% if p('otel.proxy.http', '').length > 0 %>
 export HTTP_PROXY=<%= p('otel.proxy.http') %>
+export http_proxy=<%= p('otel.proxy.http') %>
```
crobert-1, line 3: Some architectures expect this in lower case (`http_proxy`), so I've added it as a safe-guard here.

deployments/cloudfoundry/tile/tile.yml
```diff
@@ -94,16 +94,16 @@ forms:
   - name: otel_proxy_http
     type: string
     label: HTTP proxy URL
-    description: URL used as the proxy URL for HTTP requests unless overridden by otel_proxy_no
+    description: URL used as the proxy URL for HTTP requests unless overridden by "Proxy exclusions"
```
crobert-1, line 97: This is in the Tanzu Tile UI, so the user wouldn't know what `otel_proxy_no` is referring to. The text box is labelled `Proxy exclusions`, so this makes the text box's description consistent with the UI's label.

deployments/cloudfoundry/tile/tile.yml
```diff
@@ -94,16 +94,16 @@ forms:
...
 
   - name: otel_proxy_https
     type: string
     label: HTTPS proxy URL
-    description: URL used as the proxy for HTTPS requests unless overridden by otel_proxy_no
+    description: URL used as the proxy for HTTPS requests unless overridden by "Proxy exclusions"
     optional: true
 
-  - name: otel_proxy_no
+  - name: otel_proxy_exclusions
```
crobert-1, line 106: Variable name underscores are converted to periods when being used in the BOSH job. (That's why the BOSH job references `otel.proxy.exclusions`, but then the Tanzu Tile references `otel_proxy_exclusions`)

dloucasfx (approved):

atoulme (approved):

PR #3906, opened 2023-11-10
DRAFT - Kafka Discovery configs and testing

greatestusername-splunk:
**Description:** Implements configs and testing for discovery of Kafka instances.
Testing is done in three separate tests due to complications in correctly matching metrics streams when performing discovery on all three different types (broker, producer, consumer) all at once.

Also updates version of Kafka being tested against thanks to Ryan Faircloth


tests/receivers/smartagent/collectd-kafka/bundled_test.go
```diff
@@ -0,0 +1,185 @@
...
+func TestKafkaBrokerMetrics(t *testing.T) {
+	testutils.SkipIfNotContainerTest(t)
+	if runtime.GOOS == "darwin" {
+		t.Skip("unable to share sockets between mac and d4m vm: https://github.com/docker/for-mac/issues/483#issuecomment-758836836")
+	}
+	tc := testutils.NewTestcase(t)
+	defer tc.PrintLogsOnFailure()
+	defer tc.ShutdownOTLPReceiverSink()
+
+	kafkaBroker := kafka.WithName("prefix-kafka-test-broker-suffix").WithEnvVar("START_AS", "broker").WithExposedPorts("7099:7099", "9092:9092").WillWaitForPorts("7099", "9092").WillWaitForLogs("INFO Opening socket connection to server")
```
rmfitzpatrick, line 45: These can be defined as globals, and to keep test overhead down it may be preferable to start shared instances like is done w/ the existing tests.

internal/confmapprovider/discovery/bundle/bundle.d/receivers/smartagent-collectd-kafka-broker.discovery.yaml.tmpl
```diff
@@ -0,0 +1,43 @@
+{{ receiver "smartagent/collectd/kafka_broker" }}:
+  enabled: true
+  rule:
+    docker_observer: type == "container" and any([name, image, command], {# matches "(?i)kafka.*broker"}) and not (command matches "splunk.discovery")
+    host_observer: type == "hostport" and command matches "(?i)kafka.*broker"
+    k8s_observer: type == "port" and pod.name matches "(?i)kafka.*broker"
+  config:
+    default:
+      type: collectd/kafka
+      extraMetrics: ["*"]
```
rmfitzpatrick, line 10: We don't want to enabled all metrics by default since many aren't categorized (custom w/ billing implications).

internal/confmapprovider/discovery/bundle/bundle.d/receivers/smartagent-collectd-kafka-broker.discovery.yaml.tmpl
```diff
@@ -0,0 +1,43 @@
...
+            append_pattern: true
+            body: The container is unable to connnect to the MBean server to retrieve kafka metrics.
+        - regexp: "java.rmi.UnknownHostException: Unknown host"
+          first_only: true
+          log_record:
+            severity_text: info
+            append_pattern: true
+            body: Kafka host was discovered but is unresolvable.
+      partial:
+        - regexp: "clusterName is a required field"
```
rmfitzpatrick, line 35: As a required field I think this will fail regardless of availability so it's a partial status imo.

I think the value is only used as an extra dimension, so could be a good candidate for different observer default values using using the endpoint environments. I'm not seeing a great approach w/ what's available though, so using the standard `default` helper is what I'd suggest.

internal/confmapprovider/discovery/bundle/bundle.d/receivers/smartagent-collectd-kafka-broker.discovery.yaml
```diff
@@ -0,0 +1,47 @@
+#####################################################################################
+#                               Do not edit manually!                               #
+# All changes must be made to associated .tmpl file before running 'make bundle.d'. #
+#####################################################################################
+smartagent/collectd/kafka_broker:
+  enabled: true
+  rule:
+    docker_observer: type == "container" and any([name, image, command], {# matches "(?i)kafka.*broker"}) and not (command matches "splunk.discovery")
```
rmfitzpatrick, line 8: I think the broker/consumer/producer distinction is likely going to be detected by reported metrics based on available mbeans and not running process identifiers. Having the same rules for all kafkas makes more sense to me out of the gate.

internal/confmapprovider/discovery/bundle/bundle.d/receivers/smartagent-collectd-kafka-broker.discovery.yaml
```diff
@@ -0,0 +1,47 @@
...
+    host_observer: type == "hostport" and command matches "(?i)kafka.*broker"
```
rmfitzpatrick, line 9: we'll probably want the `not (command matches "splunk.discovery")` filter here, but for maintainability I think we should be dynamically including to to remove the requirement (will look into this).

PR #3921, opened 2023-11-13
Add an option to pass MSI properties to install.ps1

pjanotti:
**Description:**
Adding an option to pass MSI public properties when installing the collector. This is useful to control various installation properties, see https://learn.microsoft.com/en-us/windows/win32/msi/property-reference#configuration-properties

**Link to Splunk idea:**
Slack thread prompting this change https://splunk.slack.com/archives/CQ17ZQJFP/p1699365452683019

**Testing:**
Validated it locally, waiting on field validation.

**Documentation:**
Script documentation is included, however, I need to follow-up with docs to see if a manual update is needed when this is published.
cc @theletterf

### DIFF
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -106,6 +106,11 @@
     If specified, the -collector_version and -stage parameters will be ignored.
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -msi_path "C:\SOME_FOLDER\splunk-otel-collector-1.2.3-amd64.msi"
+.PARAMETER msi_public_properties
+    (OPTIONAL) Specify public MSI properties to be used when installing the Splunk OpenTelemetry Collector MSI package.
+    For information about the public MSI properties see https://learn.microsoft.com/en-us/windows/win32/msi/property-reference#configuration-properties
+    .EXAMPLE
+    .\install.ps1 -access_token "ACCESSTOKEN" -msi_public_properties "ALLUSERS=1" 
 #>
 
 param (
@@ -126,6 +131,7 @@ param (
     [string]$bundle_dir = "",
     [ValidateSet('test','beta','release')][string]$stage = "release",
     [string]$msi_path = "",
+    [string]$msi_public_properties = "",
     [string]$collector_msi_url = "",
     [string]$fluentd_msi_url = "",
     [string]$deployment_env = "",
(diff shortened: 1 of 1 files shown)


pjanotti:
It may be worth later adding specific options that build on top of this one, e.g.: `-InstallForAllUsers`, but, we can wait to see the actual usage to see if it is worth adding "direct MSI property" options.

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -106,6 +106,11 @@
     If specified, the -collector_version and -stage parameters will be ignored.
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -msi_path "C:\SOME_FOLDER\splunk-otel-collector-1.2.3-amd64.msi"
+.PARAMETER msi_public_properties
+    (OPTIONAL) Specify public MSI properties to be used when installing the Splunk OpenTelemetry Collector MSI package.
+    For information about the public MSI properties see https://learn.microsoft.com/en-us/windows/win32/msi/property-reference#configuration-properties
+    .EXAMPLE
+    .\install.ps1 -access_token "ACCESSTOKEN" -msi_public_properties "ALLUSERS=1"
```
jeffreyc-splunk, line 113: What is the syntax for multiple properties? Space-separated `key=value` pairs?

pjanotti: Here is a good example with 2 properties from https://learn.microsoft.com/en-us/windows/win32/msi/command-line-options

```DOS
msiexec /i testdb.msi INSTALLLEVEL=3 /l* msi.log COMPANYNAME="Acme ""Widgets"" and ""Gizmos."""
```

pjanotti: Just realized that the command-line takes even private properties, will rename accordingly...

crobert-1 (approved):

rmfitzpatrick:
Does it make sense to update https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/win-installer-script-test.yml#L35 to set multiple public properties?

pjanotti:
@rmfitzpatrick 
> Does it make sense to update https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/win-installer-script-test.yml#L35 to set multiple public properties?

In lack of a dedicated test I think passing `-msi_public_properties "ALLUSERS=1"` seems good. I will add it soon.

jinja2 (approved):

PR #3930, opened 2023-11-17
Set env vars for service instead of machine

pjanotti:
**Description:**
On Windows move the environment variables to the service scope. This servers two purposes:

1. Avoid issues with Splunk instrumentations that use some of the same environment variables as the collector and could end up sending data straight to the backend instead of the sending locally to the collector.
2. Reduce the scope of visibility of the access token, which should only be visible to the collector

Thanks @pellared for noticing this issue.

## Review notes

- "Monkey see, monkey does": in retrospect, after learning a bit about some of the tech used here, I wanted to change some other things, e.g.: use an array instead of maps that a later converted to arrays in Chef, but, this is already taking too long and such improvements can be addressed separately.

- Sorting variables when writing to the registry: this facilitates human inspection and tests.

- Choco: if there was a previous version with environment variables on the machine scope, the environment variables are removed from the machine level. Not doing the same for other deployments because they don't seem to support upgrading deployments, even the install script requires removing the previously installed collector.

- Ansible:
  - Removed some redundant restarts of the service on `deployments/ansible/roles/collector/tasks/otel_win_install.yml`, the restart is done only after the environment variables for the service are updated.
  - Tests are failing on CI but passing on mac dev box. 

- Puppet:
  - Unlike the other mass deployments Puppet doesn't support deploying SFx.NET to Windows+IIS.
  - `deployments/puppet/manifests/collector_win_registry.pp`: respective tests are at `internal\buildscripts\packaging\tests\deployments\puppet\puppet_test.py` (`internal/buildscripts/packaging/tests/helpers/win_utils.py` was modified to support new environment variables location).

**Link to Splunk idea:**
N/A

**Testing:**
Local tests and CI.

**Documentation:**
Updating CHANGELOG.md

cc @RassK @theletterf


pjanotti:
Returning to draft until I can look at the test failure.

jeffreyc-splunk:
The msi itself also needs to be updated: https://github.com/signalfx/splunk-otel-collector/blob/main/internal/buildscripts/packaging/msi/splunk-otel-collector.wxs#L42

jeffreyc-splunk:
And we'll have to update our ansible, chef, and puppet modules since they currently only configure `HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment`.

internal/buildscripts/packaging/msi/splunk-otel-collector.wxs
```diff
@@ -39,8 +39,10 @@
                      <RegistryKey Root="HKLM" Key="SYSTEM\CurrentControlSet\Services\EventLog\Application\splunk-otel-collector">
                         <RegistryValue Type="expandable" Name="EventMessageFile" Value="%SystemRoot%\System32\EventCreate.exe"/>
                      </RegistryKey>
-                     <RegistryKey Root="HKLM" Key="SYSTEM\CurrentControlSet\Control\Session Manager\Environment">
-                        <RegistryValue Type="string" Name="SPLUNK_CONFIG" Value="[CommonAppDataFolder]Splunk\OpenTelemetry Collector\agent_config.yaml"/>
+                     <RegistryKey Root="HKLM" Key="SYSTEM\CurrentControlSet\Services\splunk-otel-collector">
```
rmfitzpatrick, line 42: Would adopting this path [here](https://github.com/signalfx/splunk-otel-collector/blob/52ec41e2119ea6db9478bc645ac4a38d9c628359/internal/buildscripts/packaging/installer/install.ps1#L169) be necessary for these changes? not sure if you've already done this at some point in this branch.

pjanotti: Yes, this those paths will need to be update, however, notice that the env vars for the instrumentation itself will stay on that path. In this PR it is done here https://github.com/signalfx/splunk-otel-collector/pull/3930/files#diff-a0762d7dac27b79c910fcb6bf782cadc7428dd4574d065afeabc5ea90eb9538dL583

CHANGELOG.md
```diff
@@ -2,9 +2,15 @@
 
 ## Unreleased
 
+### 🛑 Breaking changes 🛑
+
+- (Splunk) On Windows the `SPLUNK_*` environment variables were moved from the machine scope to the collector service scope.
+  It is possible that some instrumentations are relying on the machine-wide environment variables set by the installation. ([#3930](https://github.com/signalfx/splunk-otel-collector/pull/3930))
```
rmfitzpatrick, line 8: nit, the aim of this statement could be misinterpreted as an ~ambiguous call to action. Maybe something like this throughout the other instances?

```suggestion
- (Splunk) On Windows the `SPLUNK_*` environment variables were moved from the machine scope to the collector service scope. This avoids collisions with other agents and instrumentation. ([#3930](https://github.com/signalfx/splunk-otel-collector/pull/3930))
```

deployments/ansible/molecule/custom_vars/windows-verify.yml
```diff
@@ -36,10 +37,6 @@
       check_mode: yes
       register: service_status
 
-    - name: Assert splunk-otel-collector service status
-      assert:
-        that: not service_status.changed
-
```
rmfitzpatrick, line 42: Is this removal related to the env var changes? Wonder if there should be a new var to dictate if expected.

deployments/ansible/roles/collector/tasks/otel_win_install.yml
```diff
@@ -37,7 +37,6 @@
   ansible.windows.win_package:
     path: "{{otel_msi_package.dest}}"
     state: present
-  notify: "restart windows splunk-otel-collector"
```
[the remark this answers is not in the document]
pjanotti: Yes, the restart should only happen after the environment variables are set for the service.

atoulme (approved):
Approved, feel free to merge if enough CI coverage.

PR #3932, opened 2023-11-17
Add .NET zero-config support

atoulme:
**Description:** 
Adds support for .NET programs as part of zero config.

This is a simple addition of support for dotnet, additional environment variables should be supported as well.


atoulme:
@pjanotti please take a look - this should help get things rolling

instrumentation/tests/dotnet/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0 AS build
+ARG TARGETARCH
+WORKDIR /source
+
+# copy csproj and restore as distinct layers
+COPY *.csproj .
+RUN dotnet restore -a ${TARGETARCH}
```
pjanotti, line 7: Instead of passing the architecture as a parameter you can use `--use-current-runtime` same for `dotnet publish`

instrumentation/tests/dotnet/test.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+set -euox pipefail
+
+arch="${ARCH:-amd64}"
```
pjanotti, line 5: `amd64` wouldn't work for `dotnet --architecture` but per previous comment, you don't need to pass that.

pjanotti: Related Q.: is this `arch` to facilitate `arm64` builds?

pjanotti (commented):
Good start!

instrumentation/tests/dotnet/test.sh
atoulme: yes, the tests run both amd64 and arm, for Java, node, and .NET now.

instrumentation/tests/dotnet/Dockerfile
atoulme: I just tried that and it broke, most likely because we run arm on qemu, and the program gets confused and runs with amd64 when it detects the current runtime.

pjanotti: Hum I'm thinking if this is actually a bug for `dotnet` CLI 🤔. Anyway, if works as it is...

instrumentation/src/main.c
```diff
@@ -11,10 +11,11 @@
 static char *const allowed_env_vars[] = {ALLOWED_ENV_VARS};
 static size_t const allowed_env_vars_size = sizeof(allowed_env_vars) / sizeof(*allowed_env_vars);
 
+#define DOTNET_ENV_VAR_FILE "/etc/splunk/zeroconfig/dotnet.conf"
```
jeffreyc-splunk, line 14: Are there .NET-specific env vars that need be added to `ALLOWED_ENV_VARS` for this file?

atoulme: Yes, I've been chatting about it with Paulo who pointed me to this example:
https://github.com/signalfx/tracing-examples/blob/main/opentelemetry-tracing/opentelemetry-dotnet/automatic/aspnetcore-and-mongodb/InstrumentContainer/Dockerfile

We can use the env vars defined there to be allowed as a starting point.

atoulme: added those @pjanotti please check when you get a sec

pjanotti: You can remove OTEL_DOTNET_AUTO_INTEGRATIONS_FILE, this one is deprecated - I will update the example.

instrumentation/tests/dotnet/test.sh
```diff
@@ -0,0 +1,33 @@
...
+echo "Test presence of OTEL_RESOURCE_ATTRIBUTES"
+echo "$OUTPUT" | grep "OTEL_RESOURCE_ATTRIBUTES=foo=bar,this=that" > /dev/null && echo "Test passes"  || exit 1
+echo "Test presence of OTEL_SERVICE_NAME"
+echo "$OUTPUT" | grep "OTEL_SERVICE_NAME=iknowmyownservicename" > /dev/null && echo "Test passes"  || exit 1
+echo "Test presence of SPLUNK_METRICS_ENABLED"
+echo "$OUTPUT" | grep "SPLUNK_METRICS_ENABLED=abcd" > /dev/null && echo "Test passes"  || exit 1
+echo "Test presence of ANOTHER_VAR"
+echo "$OUTPUT" | grep "ANOTHER_VAR=foo" > /dev/null && echo "Test passes"  || exit 1
+echo "Test absence of FOO"
+echo "$OUTPUT" | grep -v "FOO" > /dev/null && echo "Test passes"  || exit 1
```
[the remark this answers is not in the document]
atoulme: I'll fix this, this is used for java and node.js too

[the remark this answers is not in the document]
atoulme: on mac:
```
echo "$OUTPUT" | grep -v "foo"
$?
```
returns 1

trying on ubuntu!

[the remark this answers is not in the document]
atoulme: I don't reproduce on a ubuntu container:
```
➜ $> docker run -it ubuntu:20.04 bash
root@5ee3cf7464fd:/# OUTPUT="one two three"
root@5ee3cf7464fd:/# set -o pipefail
root@5ee3cf7464fd:/#  echo "$OUTPUT" | grep -v "two" > /dev/null && echo "Test passes"  || echo "Test fails"
Test fails
```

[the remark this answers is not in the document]
atoulme: ok sorry, now I reproduce. I think I have a terrible fix for this, thank you for finding this out!

instrumentation/src/main.c
```diff
@@ -6,15 +6,16 @@
 #define MAX_LINE_LENGTH 1023
 #define MAX_LINES 50
 
-#define ALLOWED_ENV_VARS "OTEL_SERVICE_NAME", "OTEL_EXPORTER_OTLP_ENDPOINT", "OTEL_RESOURCE_ATTRIBUTES", "SPLUNK_PROFILER_ENABLED", "SPLUNK_PROFILER_MEMORY_ENABLED", "SPLUNK_METRICS_ENABLED", "JAVA_TOOL_OPTIONS", "NODE_OPTIONS"
+#define ALLOWED_ENV_VARS "OTEL_SERVICE_NAME", "OTEL_EXPORTER_OTLP_ENDPOINT", "OTEL_RESOURCE_ATTRIBUTES", "SPLUNK_PROFILER_ENABLED", "SPLUNK_PROFILER_MEMORY_ENABLED", "SPLUNK_METRICS_ENABLED", "JAVA_TOOL_OPTIONS", "NODE_OPTIONS", "CORECLR_ENABLE_PROFILING", "CORECLR_PROFILER", "CORECLR_PROFILER_PATH", "DOTNET_ADDITIONAL_DEPS", "DOTNET_SHARED_STORE", "DOTNET_STARTUP_HOOKS", "OTEL_DOTNET_AUTO_INTEGRATIONS_FILE", "OTEL_DOTNET_AUTO_HOME", "OTEL_DOTNET_AUTO_PLUGINS"
```
pjanotti, line 9: `OTEL_DOTNET_AUTO_INTEGRATIONS_FILE` is deprecated and can be removed.

pjanotti (approved):
LGTM, just a non-material nit

instrumentation/Makefile
```diff
@@ -70,7 +70,11 @@ docker-run:
 	docker run --rm -it -v `pwd`:/instr instr-devel
 
 .PHONY: tests
-tests: test-java test-nodejs
+tests: test-java test-nodejs test-dotnet
+
+.PHONY: test-dotnet
+test-dotnet: dist
+	(cd tests/dotnet && ARCH=$(ARCH) ./test.sh)
```
rmfitzpatrick, line 77: A pattern rule can replace the other two targets:

```suggestion
.PHONY: test-dotnet-java-nodejs
test-%: dist
	(cd tests/$* && ARCH=$(ARCH) ./test.sh)
```

I'm not sure of the phony convention mechanics but that's what is used elsewhere in our makefiles.

atoulme: sure, thanks!

rmfitzpatrick: You can delete the following two targets (now dupes)

PR #3954, opened 2023-11-20
wip: restore SA1019 checks

rmfitzpatrick:
in dev: Restoring the checks that were silenced in https://github.com/signalfx/splunk-otel-collector/pull/2873.

PR #3956, opened 2023-11-22
Remove use of deprecated code and defaults in config

atoulme:
This will unblock #3954: it removes the use of deprecated code. As part of this removal, I have removed the defaults set on the field so they are no longer interpreted and tests no longer need to refer to them.


internal/signalfx-agent/pkg/core/config/sources/sources.go
```diff
@@ -36,7 +36,7 @@ type SourceConfig struct {
 	// main disadvantage of watching is slightly greater network and compute
 	// resource usage. This option is not itself watched for changes. If you
 	// change the value of this option, you must restart the agent.
-	Watch *bool `yaml:"watch" default:"true"`
```
pjanotti, line 39: Q.: the previous changes the values were deprecated and were being ignored, but, this one looks like the default is going to change from `true` to `false`. What am I missing in this picture?

pjanotti: Is this deprecated but there is no comment reflecting that?

internal/signalfx-agent/pkg/core/config/config.go
```diff
@@ -218,7 +218,6 @@ type CollectdConfig struct {
 	// If you won't be using any collectd monitors, this can be set to true to
 	// prevent collectd from pre-initializing
 	DisableCollectd bool `yaml:"disableCollectd" default:"false"`
-	// Deprecated: this setting has no effect and will be removed.
```
crobert-1, line 221: Why are we removing the deprecated notices here?

(This might be a duplicate of @pjanotti's question, so feel free to ignore if that's the case)

atoulme: Because I made a mistake. While this code has no usage in go, it is serialized to represent collectd config on disk, and removing it breaks tests. I should not have deprecated it in the first place.

internal/signalfx-agent/pkg/core/config/sources/sources.go
atoulme: this field is part of a struct which is used in only one place, where it is marked deprecated:
https://github.com/signalfx/splunk-otel-collector/blob/main/internal/signalfx-agent/pkg/core/config/config.go#L170

crobert-1 (approved):

PR #3983, opened 2023-11-30
remove curl usage

atoulme:
**Description:**
remove curl from the agent bundle and rely on wget for all operations in the support bundle script

**Testing:**
Manual testing that the change is effective with a local docker build.

### DIFF
--- a/internal/buildscripts/packaging/fpm/etc/otel/collector/splunk-support-bundle.sh
+++ b/internal/buildscripts/packaging/fpm/etc/otel/collector/splunk-support-bundle.sh
@@ -87,7 +87,7 @@ createTempDir() {
 #######################################
 checkCommands() {
     echo "INFO: Checking for commands..."
-    for EXE in systemctl journalctl curl wget pgrep; do
+    for EXE in systemctl journalctl wget pgrep; do
       if ! command -v $EXE &> /dev/null; then
           echo "WARN: $EXE could not be found."
           echo "      Please install to capture full support bundle."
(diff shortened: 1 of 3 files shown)


pjanotti (approved):

internal/buildscripts/packaging/fpm/etc/otel/collector/splunk-support-bundle.sh
```diff
@@ -125,8 +125,8 @@ getConfig() {
     # Also need to get config in memory as dynamic config may modify stored config
     # It's possible user has disabled collecting in memory config
     if timeout 1 bash -c 'cat < /dev/null > /dev/tcp/localhost/55554'; then
-        curl -s http://localhost:55554/debug/configz/initial >"$TMPDIR"/config/initial.yaml 2>&1
-        curl -s http://localhost:55554/debug/configz/effective >"$TMPDIR"/config/effective.yaml 2>&1
+        wget -q http://localhost:55554/debug/configz/initial -O "$TMPDIR"/config/initial.yaml 2>&1
```
rmfitzpatrick, line 128: I think any support changes like this are potentially too impacting for customer envs. imo the client executable should be invoked w/ a variable based on what's available. if we are going to require one, we should include it as a dep in our packages.

internal/signalfx-agent/bundle/Dockerfile
```diff
@@ -352,7 +352,7 @@ ENV useful_bins=" \
   /bin/ss \
   /bin/tar \
   /bin/umount \
-  /usr/bin/curl \
+  /usr/bin/wget \
```
rmfitzpatrick, line 355: If this is regarding a low libcurl cve addressing that at the source would be my suggestion since we don't resolve it w/ this change.

atoulme:
We don't need this PR anymore as we have another approach to update curl with the new agent bundle. Closing.

PR #3990, opened 2023-12-01
Add a deprecation notice to collectd-etcd

atoulme:
This PR has a companion PR on the collectd-etcd repository:
https://github.com/signalfx/collectd-etcd/pull/7

The collectd plugin doesn't support modern versions of etcd and Prometheus should be used instead.

This PR adds a warning notice when starting the monitor to let users know that this monitor is now deprecated and may be removed soon.

### DIFF
--- a/internal/signalfx-agent/pkg/monitors/collectd/etcd/etcd.go
+++ b/internal/signalfx-agent/pkg/monitors/collectd/etcd/etcd.go
@@ -57,6 +57,7 @@ type Monitor struct {
 
 // Configure configures and runs the plugin in collectd
 func (m *Monitor) Configure(conf *Config) error {
+	m.Logger().Warn("The ectd-collectd plugin is deprecated. It will be removed in a future release.")
 	conf.pyConf = &python.Config{
 		MonitorConfig: conf.MonitorConfig,
 		Host:          conf.Host,


internal/signalfx-agent/pkg/monitors/collectd/etcd/etcd.go
```diff
@@ -57,6 +57,7 @@ type Monitor struct {
 
 // Configure configures and runs the plugin in collectd
 func (m *Monitor) Configure(conf *Config) error {
+	m.Logger().Warn("The ectd-collectd plugin is deprecated. It will be removed in a future release.")
```
rmfitzpatrick, line 60: Can you include a note about the expected migration path (including receiver name)?

atoulme:
I opted to point out to the `etcd` plugin which is a thin wrapper around the prometheus endpoint scraper.

internal/signalfx-agent/pkg/monitors/collectd/etcd/etcd.go
```diff
@@ -57,6 +57,7 @@ type Monitor struct {
 
 // Configure configures and runs the plugin in collectd
 func (m *Monitor) Configure(conf *Config) error {
+	m.Logger().Warn("The ectd-collectd plugin is deprecated. Please use the etcd plugin instead. See https://docs.splunk.com/observability/en/gdi/monitors-databases/etcd.html for more information. This plugin will be removed in a future release.")
```
rmfitzpatrick, line 60:
```suggestion
	m.Logger().Warn("The ectd-collectd plugin is deprecated. Please use the etcd monitor instead. See https://docs.splunk.com/observability/en/gdi/monitors-databases/etcd.html for more information. This plugin will be removed in a future release.")
```

rmfitzpatrick (approved):

PR #3991, opened 2023-12-01
remove deprecated netinterface monitor

atoulme:
**Description:**

This removes the deprecated netinterface monitor. This component was [deprecated](https://github.com/signalfx/signalfx-agent/commit/6aafbe1f8b34ecca1d05b12c6f67a652248384ac) 3 years 9 months ago. As noted in the documentation of this monitor, the net-io monitor should be used instead:

```
    **This monitor is deprecated in favor of the `net-io` monitor. Please
    switch to that monitor as this monitor will be removed in a future release
    of the agent.**  Note that the `net-io` monitor uses the `interface`
    dimension to identify the network card instead of the `plugin_instance`
    dimension, but otherwise the metrics are the same.
```


rmfitzpatrick (changes requested):
please replace w/ deprecation notice for at least a release

crobert-1 (approved):

PR #3995, opened 2023-12-01
remove collectd/load monitor

atoulme:
**Description:**

This PR removes the `collectd/load` monitor in favor of the `load` monitor. This component was [deprecated](https://github.com/signalfx/signalfx-agent/commit/f0bd44260f9e80506165be6fce09eb7ee0823bf8) 3 years 11 months ago.

The `collectd/load` monitor is deprecated with the following notice:

```
    **This monitor has been deprecated in favor of the `load` monitor. That
    monitor emits the same metrics and is fully compatible.  This
    `collectd/load` monitor will be removed in a future release of the agent."
```


rmfitzpatrick (changes requested):

PR #3997, opened 2023-12-01
Remove collectd/elasticsearch monitor

atoulme:
**Description:**

This PR removes the `collectd/elasticsearch` monitor in favor of the `elasticsearch` monitor.

As noted in the deprecation notice of the monitor:

```
    Monitors ElasticSearch instances. We strongly recommend using the
    [elasticsearch](./elasticsearch.md) monitor instead, as it will
    scale much better.
```


rmfitzpatrick (dismissed):

crobert-1 (approved):

PR #4000, opened 2023-12-01
Fix use of grep -v

atoulme:
As identified in https://github.com/signalfx/splunk-otel-collector/pull/3932#discussion_r1411312062, tests do not work well to check absence of an environment variable. This is an attempt to ensure the tests work properly.

### DIFF
--- a/instrumentation/tests/java/test.sh
+++ b/instrumentation/tests/java/test.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -euo pipefail
+set -euox pipefail
 
 arch="${ARCH:-amd64}"
 BASE="eclipse-temurin:11"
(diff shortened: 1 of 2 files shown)


instrumentation/tests/java/test.sh
```diff
@@ -25,9 +25,9 @@ echo "$OUTPUT" | grep "SPLUNK_METRICS_ENABLED=abcd" > /dev/null && echo "Test pa
 echo "Test presence of ANOTHER_VAR"
 echo "$OUTPUT" | grep "ANOTHER_VAR=foo" > /dev/null && echo "Test passes"  || exit 1
 echo "Test absence of FOO"
-echo "$OUTPUT" | grep -v "FOO" > /dev/null && echo "Test passes"  || exit 1
+echo "$OUTPUT" | grep "FOO" | grep -v "FOO" > /dev/null && echo "Test passes"  || exit 1
```
jeffreyc-splunk, line 28: Maybe simpler to do something like:
```
if echo "$OUTPUT" | grep -q "FOO"; then
  exit 1
else
  echo "Test passes"
fi

Comment (jinja2) [reply instrumentation/tests/java/test.sh:28]: how about something like
```[[ $OUTPUT != *FOO* ]] && echo "no FOO" || echo "found FOO"```

Comment (atoulme) [reply instrumentation/tests/java/test.sh]: sorry just seeing this. I am not familiar with `-q`, looking.

File: instrumentation/tests/java/test.sh [hunk 2]
```diff
@@ -25,13 +25,14 @@ echo "$OUTPUT" | grep "SPLUNK_METRICS_ENABLED=abcd" > /dev/null && echo "Test pa
 echo "Test presence of ANOTHER_VAR"
 echo "$OUTPUT" | grep "ANOTHER_VAR=foo" > /dev/null && echo "Test passes"  || exit 1
 echo "Test absence of FOO"
-echo "$OUTPUT" | grep -v "FOO" > /dev/null && echo "Test passes"  || exit 1
+echo "$OUTPUT" | (grep FOO || echo "") | grep -v "FOO" > /dev/null && echo "Test passes"  || exit 1
```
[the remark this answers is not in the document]
atoulme: it is, but I thought I moved on to use the regexp Jina offered.

instrumentation/tests/java/test.sh
atoulme: sorry, the regexp approach doesn't work for multiline variables.

PR #4002, opened 2023-12-03
Fix run as service detection

pjanotti:
**Description:**
Fix run as a service detection on Windows. Instead of trying to detect if it is a service or not, for which both `svc.IsAnInteractiveSession()` and `svc.IsWindowsService()` are somehow broken, try to run as a Windows service, if it fails fallback to run as an interactive process. This will add a small cost, which on my machine is under 5 microseconds, to startup in the interactive mode compared to using the `NO_WINDOWS_SERVICE` environment variable. While this value seems fine for startup I'm keeping the `NO_WINDOWS_SERVICE` option instead of deprecating it (it doesn't seem worth the trouble of deprecating it).

**Link to Splunk idea:**
N/A

**Testing:**
Added benchmark to measure the cost of trying to start as a service (notice that when running as a service there is no extra cost).

**Documentation:**
Added entry in changelog.

### DIFF
--- a/cmd/otelcol/main_windows.go
+++ b/cmd/otelcol/main_windows.go
@@ -28,10 +28,30 @@ import (
 )
 
 func run(params otelcol.CollectorSettings) error {
-	// do not need to supply service name when startup is invoked through Service Control Manager directly
+	// There shouldn't be any reason to use NO_WINDOWS_SERVICE anymore, but,
+	// keeping it as a forcing mechanism or if someone is concerned about
+	// the cost of attempting to run as a service before falling back to
+	// interactive mode.
+	//
+	// The BenchmarkSvcRunFail measures the overhead of attempting and failing to
+	// run as a service:
+	//
+	// goos: windows
+	// goarch: amd64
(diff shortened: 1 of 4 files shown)


cmd/otelcol/main_windows.go
```diff
@@ -21,40 +21,41 @@ package main
...
-	} else {
-		return runService(params)
-	}
-}
-
-func checkUseInteractiveMode() (bool, error) {
-	// If environment variable NO_WINDOWS_SERVICE is set with any value other
-	// than 0, use interactive mode instead of running as a service. This should
-	// be set in case running as a service is not possible or desired even
-	// though the current session is not detected to be interactive
```
atoulme, line 43: FWIW this code is copied from https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/builder/internal/builder/templates/main_windows.go.tmpl#L29

atoulme (approved):
Fix LGTM, it should be ported to the collector upstream as well.

PR #4028, opened 2023-12-08
Remove translatesfx

atoulme:
**Description:**
This PR removes translatesfx from the repository. Translatesfx is a tool used to migrate configuration between the smart agent and the OpenTelemetry Collector. As the Smart Agent has now reached end of support, this tool is deprecated and will be removed from the collector next release onwards.


atoulme:
It fails on both PRs recently, need to check main build.

atoulme:
So the build passed on arm because the test skips it there, and it fails on amd64 because it looks for a version of the chart we skipped. I think we can merge this PR as the code doesn't touch this particular execution path, and we need to fix main.

PR #4037, opened 2023-12-08
add start timestamps

atoulme:
**Description:**
Add capture of start timestamp to metrics expressed by the lightprometheusreceiver.
**Testing:**
Can add tests once https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/29718 is implemented.

### DIFF
--- a/internal/receiver/lightprometheusreceiver/scraper.go
+++ b/internal/receiver/lightprometheusreceiver/scraper.go
@@ -160,6 +160,7 @@ func (s *scraper) convertMetricFamilies(metricFamilies []*dto.MetricFamily, rm p
 			for _, fm := range family.GetMetric() {
 				dp := sum.DataPoints().AppendEmpty()
 				dp.SetTimestamp(now)
+				dp.SetStartTimestamp(now)
 				dp.SetDoubleValue(fm.GetCounter().GetValue())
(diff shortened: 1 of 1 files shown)


internal/receiver/lightprometheusreceiver/scraper.go
```diff
@@ -160,6 +160,7 @@ func (s *scraper) convertMetricFamilies(metricFamilies []*dto.MetricFamily, rm p
 			for _, fm := range family.GetMetric() {
 				dp := sum.DataPoints().AppendEmpty()
 				dp.SetTimestamp(now)
+				dp.SetStartTimestamp(now)
```
rmfitzpatrick, line 163: I think the start timestamp is used for [aggregation temporality](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L161) so should be the time of the first scrape instead of moving forward?

dmitryax: Right, start time for cumulative metrics is the time when scraping started or the cumulative got reset

atoulme: Thanks for the review, fixing

rmfitzpatrick (approved):

PR #4038, opened 2023-12-08
Update installation prefix for splunk-otel-js

jeffreyc-splunk:
- Update `NODE_OPTIONS` to `-r /usr/lib/splunk-instrumentation/splunk-otel-js/node_modules/@splunk/otel/instrument` in the ZC config files
- Update installer script to install `splunk-otel-js` with `cd /usr/lib/splunk-instrumentation/splunk-otel-js && npm install` instead of `--global`
- Replace the `--npm-command` installer script option with `--npm-path`
- Update docs and tests

### DIFF
--- a/internal/buildscripts/packaging/installer/install.sh
+++ b/internal/buildscripts/packaging/installer/install.sh
@@ -122,6 +122,7 @@ enable_metrics="false"
 java_zeroconfig_path="/etc/splunk/zeroconfig/java.conf"
 node_zeroconfig_path="/etc/splunk/zeroconfig/node.conf"
 node_package_path="/usr/lib/splunk-instrumentation/splunk-otel-js.tgz"
+node_install_prefix="/usr/lib/splunk-instrumentation/splunk-otel-js"
 
 repo_for_stage() {
   local repo_url=$1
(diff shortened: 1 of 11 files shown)


internal/buildscripts/packaging/installer/install.sh
```diff
@@ -333,8 +335,8 @@ ensure_not_installed() {
       echo "Please uninstall auto instrumentation, or try running this script with the '--uninstall' option." >&2
       exit 1
     fi
-    if command -v npm >/dev/null 2>&1 && npm ls --global @splunk/otel >/dev/null 2>&1; then
-      echo "The @splunk/otel npm package is already globally installed." >&2
+    if [ -n "$npm_path" ] && $npm_path ls --prefix $node_install_prefix @splunk/otel >/dev/null 2>&1; then
```
rmfitzpatrick, line 338: not an issue functionally but I think `--prefix` is discouraged: https://github.com/npm/cli/issues/1368#issuecomment-1241076700

rmfitzpatrick (approved):

internal/buildscripts/packaging/installer/install.sh
jeffreyc-splunk: Let me change it to use `cd` just to be safe in case they ever decide to deprecate the `--prefix` option.

jvoravong (approved):

PR #4039, opened 2023-12-08
deprecate signalfx-forwarder

atoulme:
**Description:**
Deprecate the signalfx-forwarder monitor, and remove it from default configuration.
Recommend using the OTLP receiver instead for forwarding signals.

**Documentation:**
This is a big change, which will have ripple effects in our documentation. This change removes mentions and examples using the signalfx-forwarder across the code and docs of the product.


rmfitzpatrick:
I believe https://github.com/signalfx/signalfx-php-tracing still relies on this and is still supported. Does it make sense to push removing from the default agent config to a later PR? (cc @akubik-splunk)

akubik-splunk:
We are going to deprecate php-signalfx soon so if we coordinate we can get both things done

atoulme:
Closing as we wait for updates to signalfx-php-tracing.

pjanotti (approved):

PR #4066, opened 2023-12-13
discoverybundler: Generate component files from components.go

rmfitzpatrick:
**Description:**
These changes add a new `components.go` file used to generate the previously manually updated bundle_gen and BundledFS parent files (which were renamed)*. They also remove implicit discoverybundler installation w/ `make bundle.d` calls.

**Testing:**
Existing tests ensure no component changes and added `make bundle.d` diff check in test job can help detect unexpected deviation.


internal/confmapprovider/discovery/bundle/bundle_gen.tmpl
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build bundle.d
+
+// These are the discovery config component generating statements.
+// In order to update run go generate -tags bundle.d ./...
+
+{{- range $ext := .Extensions }}
```
jvoravong, line 20: Asking a simple minded probing question here. Should there be a section for windows support?

jvoravong (approved):

internal/confmapprovider/discovery/bundle/bundle_gen.tmpl
rmfitzpatrick: We don't currently bundle discovery config for windows packages, but there'd be one when we do.

PR #4091, opened 2023-12-18
ecs-metadata: support `dimensionToUpdate` config field 

rmfitzpatrick:
These changes add a `dimensionToUpdate` config field to the ecs-metadata monitor to allow configuring dimension updates on either the `container_name` and `container_id` dimension. Since container_name's can be long lived, perpetuating entities to update becomes unnecessary and setting to the more ephemeral `container_id` is preferable. ~~My plan is to update the default value to `container_id` in a subsequent release but can understand wanting to here instead.~~

edit: updated to sync on `container_id` by default, w/ bw-compatibility option w/ new config field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 🛑 Breaking changes 🛑
+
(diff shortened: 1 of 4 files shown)


rmfitzpatrick:
@dloucasfx would appreciate your thoughts on if this is too conservative and we should simply update the dimension to sync to `container_id` without a transitional option and future default value change.

atoulme (approved):

PR #4092, opened 2023-12-20
Add tests for collectd/apache

atoulme:
Tests for collectd/apache. I used the golden file approach this time around as I would like to reuse the expected output for categorization tests.

### DIFF
--- a/tests/receivers/apache/apache_test.go
+++ b/tests/receivers/apache/apache_test.go
@@ -1,0 +1,69 @@
+// Copyright Splunk, Inc.
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build integration
+
+package tests
+
+import (
+	"path"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/golden"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/pdatatest/pmetrictest"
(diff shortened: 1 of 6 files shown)


tests/receivers/apache/apache_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	selected := tc.OTLPReceiverSink.AllMetrics()[len(tc.OTLPReceiverSink.AllMetrics())-1]
+
+	metricNames := []string{"apache_requests", "apache_bytes", "apache_connections", "apache_idle_workers", "apache_scoreboard.open"}
+	require.NoError(t, pmetrictest.CompareMetrics(expected, selected,
+		pmetrictest.IgnoreMetricDataPointsOrder(),
+		pmetrictest.IgnoreTimestamp(),
+		pmetrictest.IgnoreMetricsOrder(),
+		pmetrictest.IgnoreMetricValues(metricNames...),
+		pmetrictest.IgnoreMetricAttributeValue("host", metricNames...)))
```
rmfitzpatrick, line 68: I really don't see the value in introducing these kinds of checks to the integration suite where the resource metrics conventions have been established and enhanced over time. I'm obviously biased but generally see your desired transition as a step backward.

>I would like to reuse the expected output for categorization tests.

I strongly recommend you not introduce unnecessary dependencies between projects, especially for items and projects not publicly available.

tests/receivers/apache/apache_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/signalfx/splunk-otel-collector/tests/testutils"
+)
+
+var apache = []testutils.Container{testutils.NewContainer().WithContext(
+	path.Join(".", "testdata", "server"),
+).WithName("apache").WithExposedPorts("8000:80").WillWaitForLogs("Command line: 'httpd -D FOREGROUND'")}
+
+// This test ensures the collector can connect to an Apache Web Server, and properly get metrics. It's not intended to
+// test the receiver itself.
```
rmfitzpatrick, line 38: Why are you adding it if it's not intended to test the receiver itself? That's the main point of these tests thus far since they have effectively 0 test coverage otherwise.

rmfitzpatrick (changes requested):
There is existing coverage: https://github.com/signalfx/splunk-otel-collector/blob/main/tests/receivers/smartagent/collectd-apache/collectd_apache_test.go. If these tests are incompatible with your aims, please update or add using the existing fixtures.

atoulme:
Closing, we already have tests.

PR #4097, opened 2023-12-21
Convert loglevel to verbosity on logging exporter

atoulme:
Add a converter that takes loglevel and translates it over to verbosity.

### DIFF
--- a/internal/configconverter/loglevel_to_verbosity.go
+++ b/internal/configconverter/loglevel_to_verbosity.go
@@ -29,10 +29,10 @@ type LogLevelToVerbosity struct{}
 
 func (LogLevelToVerbosity) Convert(_ context.Context, in *confmap.Conf) error {
 	if in == nil {
-		return fmt.Errorf("cannot MoveHecTLS on nil *confmap.Conf")
+		return fmt.Errorf("cannot LogLevelToVerbosity on nil *confmap.Conf")
 	}
 
-	const expression = "exporters::logging(/\\w+)?::loglevel"
+	const expression = "exporters::logging(/.+)?::loglevel"
 	re := regexp.MustCompile(expression)
 	out := map[string]any{}
 	unsupportedKeyFound := false
--- a/internal/configconverter/loglevel_to_verbosity.go
+++ b/internal/configconverter/loglevel_to_verbosity.go
@@ -1,0 +1,93 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configconverter
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"regexp"
+
+	"go.opentelemetry.io/collector/config/configtelemetry"
+	"go.opentelemetry.io/collector/confmap"
+	"go.uber.org/zap/zapcore"
+)
+
+type LogLevelToVerbosity struct{}
+
+func (LogLevelToVerbosity) Convert(_ context.Context, in *confmap.Conf) error {
+	if in == nil {
+		return fmt.Errorf("cannot MoveHecTLS on nil *confmap.Conf")
+	}
+
+	const expression = "exporters::logging(/\\w+)?::loglevel"
+	re := regexp.MustCompile(expression)
+	out := map[string]any{}
+	unsupportedKeyFound := false
+	for _, k := range in.AllKeys() {
+		v := in.Get(k)
+		match := re.FindStringSubmatch(k)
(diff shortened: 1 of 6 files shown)


theletterf (approved):

theletterf:
Would this somehow unlock https://github.com/signalfx/splunk-otel-collector/pull/3768 ?

atoulme:
> Would this somehow unlock #3768 ?

No, but we could do a converter for it. It's just a bit harder to rename a receiver as we'd need to change its occurrences in the pipelines too.

internal/configconverter/loglevel_to_verbosity.go
```diff
@@ -0,0 +1,93 @@
...
+		v := in.Get(k)
+		match := re.FindStringSubmatch(k)
+		if match == nil {
+			out[k] = v
+		} else {
+			// check if verbosity is also set:
+			verbosityKey := fmt.Sprintf("exporters::logging%s::verbosity", match[1])
+			if in.Get(verbosityKey) == nil {
+				log.Printf("Deprecated key found: %s. Translating to %s\n", k, verbosityKey)
+				var l zapcore.Level
```
rmfitzpatrick, line 49: Why would it be necessary to use zap in addition to configtelemetry?

internal/configconverter/loglevel_to_verbosity.go
```diff
@@ -0,0 +1,93 @@
...
+	"go.opentelemetry.io/collector/config/configtelemetry"
+	"go.opentelemetry.io/collector/confmap"
+	"go.uber.org/zap/zapcore"
+)
+
+type LogLevelToVerbosity struct{}
+
+func (LogLevelToVerbosity) Convert(_ context.Context, in *confmap.Conf) error {
+	if in == nil {
+		return fmt.Errorf("cannot MoveHecTLS on nil *confmap.Conf")
```
rmfitzpatrick, line 32: cp remnant

internal/configconverter/loglevel_to_verbosity.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	const expression = "exporters::logging(/\\w+)?::loglevel"
```
rmfitzpatrick, line 35: I know this is used throughout but there's no word character restriction on component id name values.

internal/configconverter/loglevel_to_verbosity_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+func TestLogLevelToVerbosity(t *testing.T) {
+	cfgMap, err := confmaptest.LoadConf("testdata/logging_loglevel.yaml")
+	require.NoError(t, err)
+	require.NotNil(t, cfgMap)
+
+	err = LogLevelToVerbosity{}.Convert(context.Background(), cfgMap)
+	require.NoError(t, err)
+
+	assert.False(t, cfgMap.IsSet("exporters::logging::loglevel"))
```
rmfitzpatrick, line 34: Defining an expected config would be more robust than these specific checks. It's possible a bug could be dropping all values except verbosity and the test wouldn't find it.

internal/configconverter/loglevel_to_verbosity_test.go
```diff
@@ -0,0 +1,39 @@
...
+	assert.Equal(t, "Detailed", cfgMap.Get("exporters::logging::verbosity"))
```
rmfitzpatrick, line 35: We should preserve the specified value without parsing imo. If a user entered `false` it would be clearer for the exporter to validate rather than directing them to the converter.

internal/configconverter/loglevel_to_verbosity.go
atoulme: the loglevel is a string representation of a zapcore.Level, and we have to translate it over to verbosity levels using the logic from the logging exporter config.go file: https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/loggingexporter/config.go#L48

internal/configconverter/loglevel_to_verbosity_test.go
atoulme: ok sure will do.

internal/configconverter/loglevel_to_verbosity.go
atoulme: replaced with `.`

internal/configconverter/loglevel_to_verbosity.go
atoulme: fixed

internal/configconverter/loglevel_to_verbosity_test.go
atoulme: Done.

internal/configconverter/loglevel_to_verbosity_test.go
atoulme: added a test to make sure verbosity is preserved, please let me know if I understood you correctly

internal/configconverter/loglevel_to_verbosity.go
rmfitzpatrick: right, but that's in the exporter itself to relay to a new zap instance, where we are only obligated to work with the config value itself. Why validate it at all and not transparently forward?

rmfitzpatrick: Oh my bad, I forgot the field type changed in total and you're coercing it to the intermediary.

rmfitzpatrick (approved):

PR #4119, opened 2024-01-03
Add kong support

atoulme:
**Description:**
Add a test ensuring we get the expected metrics out of a kong image.

This is taken after this tutorial: https://docs.konghq.com/gateway/latest/install/docker/#install-kong-gateway-in-db-less-mode


tests/receivers/kong/testdata/all_metrics_config.yaml
```diff
@@ -0,0 +1,24 @@
+receivers:
+  prometheus:
```
rmfitzpatrick, line 2: since this uses the prometheus receiver, I think it belongs in that directory.

tests/receivers/kong/testdata/server/kong.yml
```diff
@@ -0,0 +1,13 @@
+_format_version: "3.0"
+_transform: true
+
+services:
+  - host: httpbin.org
```
rmfitzpatrick, line 5: no need to refer to an external resource (don't think it's even used?)

```suggestion
  - host: localhost
```

tests/receivers/kong/kong_test.go
```diff
@@ -0,0 +1,35 @@
+// Copyright Splunk, Inc.
+// Copyright The OpenTelemetry Authors
```
rmfitzpatrick, line 2:
```suggestion
```

tests/receivers/kong/testdata/expected.yaml
crobert-1, line 1: It looks like this file is empty and isn't referenced, we should remove it (unless I'm missing something.

atoulme: oh sorry, removing.

tests/receivers/kong/testdata/all_metrics_config.yaml
atoulme: moved

crobert-1 (approved):

tests/receivers/prometheus/testdata/resource_metrics/kong_metrics.yaml
```diff
@@ -0,0 +1,891 @@
...
+  - attributes:
+      http.scheme: http
+      net.host.port: "8001"
+      service.instance.id: localhost:8001
+      service.name: kong
+    scope_metrics:
+      - instrumentation_scope:
+          attributes: {}
+          name: otelcol/prometheusreceiver
+          version: v0.91.1
```
rmfitzpatrick, line 11:
```suggestion
          version: <VERSION_FROM_BUILD>
```

rmfitzpatrick (approved):

PR #4127, opened 2024-01-05
ansible: Support Node.js zero config for Linux

jeffreyc-splunk:


deployments/ansible/molecule/config/windows.yml
```diff
@@ -31,13 +31,13 @@ platforms:
     instance_raw_config_args: *vagrant_args
   - name: "2019"
     box: gusztavvargadr/windows-server-2019-standard
-    box_version: 1809.0.2305
+    box_version: 1809.0.2312
```
atoulme, line 34: maybe this needs to be in a different PR

deployments/ansible/molecule/config/windows.yml
```diff
@@ -31,13 +31,13 @@ platforms:
...
     cpus: 2
     memory: 4096
     instance_raw_config_args: *vagrant_args
   - name: "2022"
     box: gusztavvargadr/iis-windows-server
-    box_version: 10.2102.2306
+    box_version: 2102.0.2312
```
atoulme, line 40: same

deployments/ansible/molecule/config/windows.yml
pjanotti: I don't think this is an option: it seems that the version 2305 was removed https://app.vagrantup.com/gusztavvargadr/boxes/windows-server-2019-standard-core/versions/1809.0.2305 - it seems that the oldest available is 2307.

atoulme: ouch. But that can go into a separate PR right now to get main back to green?

atoulme (approved):
LGTM. Given that this tests Linux only, I'm inclined to disregard test failures on Windows as irrelevant.

PR #4146, opened 2024-01-09
Deprecate httpsinkexporter

atoulme:
httpsinkexporter is used for testing in our SDKs, but it should not be used by customers. It is no longer developed and will be kept around as deprecated until such time it is no longer used, and will then be removed.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -110,3 +110,9 @@ linters:
     - staticcheck
     - unconvert
     - unparam
+
+issues:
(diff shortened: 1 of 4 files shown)


jvoravong (approved):

crobert-1 (commented):
It would be good to add a note in the [README](https://github.com/signalfx/splunk-otel-collector/tree/main/internal/exporter/httpsinkexporter) as well.

PR #4202, opened 2024-01-26
update opentelemetry dependencies

jinja2:

- Update Opentelemetry dependencies to latest. 
- Updates mdatagen package to `go.opentelemetry.io/collector/cmd/mdatagen`.
- Remove the disabled feature gate `telemetry.useOtelForInternalMetrics` which is now enabled by default.
- Updates default configs to drop new internal metrics exposed by the collector. These are metrics tracking grpc and http request/responses.
- Updates to tests for above changes


dmitryax (approved):

tests/receivers/prometheus/testdata/resource_metrics/internal.yaml
```diff
@@ -4,6 +4,9 @@ resource_metrics:
       net.host.port: "8889"
       service.instance.id: localhost:8889
       service.name: otel-agent
+      service_instance_id: <ANY>
+      service_name: otelcol
+      service_version: <VERSION_FROM_BUILD>
```
jinja2, line 9: These are added to resource attributes due to `target_info` metric now being exposed. The prometheus receiver adds the [labels on target_info to resource attrs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/transaction.go#L323).

dmitryax: I think that's ok. Ideally this should've been configurable on the Prometheus receiver side

jinja2: Yes, just wanted to note that all internal metrics have 2 sets of `service name`, and `service instance id`. The [ones added by the receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/prom_to_otlp.go#L42-L48) in otel format and a 2nd set from target_info with different values, which might become confusing.

tests/receivers/discovery/testdata/host_observer_simple_prometheus_config.yaml
```diff
@@ -13,14 +13,14 @@ receivers:
     embed_receiver_config: true
     receivers:
       prometheus_simple:
-        rule: type == "hostport"
+        rule: type == "hostport" and command contains "otelcol"
```
jinja2, line 16: Without this change in rule, too many ports were being discovered and the test would timeout before it even tried the internal metrics port. I also updated the metric to match against for a successful status to get more reliable runs.

tests/receivers/prometheus/testdata/resource_metrics/internal.yaml
dmitryax: @jinja2 do they have different keys with the same values?

jinja2: Are you asking about attributes from target_info and the ones added by the receiver, right? I don't think I have seen any in the tests I have worked on.

PR #4209, opened 2024-01-28
Add default deployment.environment resource attribute as part of traces pipeline

atoulme:
**Description:**
Add default `deployment.environment` resource attribute as part of traces pipeline

### DIFF
--- a/cmd/otelcol/config/collector/agent_config.yaml
+++ b/cmd/otelcol/config/collector/agent_config.yaml
@@ -114,15 +114,14 @@ processors:
     detectors: [gcp, ecs, ec2, azure, system]
     override: true
 
-  # Optional: The following processor can be used to add a default "deployment.environment" attribute to the logs and 
+  # The following processor can be used to add a default "deployment.environment" attribute to the logs and
   # traces when it's not populated by instrumentation libraries.
-  # If enabled, make sure to enable this processor in a pipeline.
   # For more information, see https://docs.splunk.com/Observability/gdi/opentelemetry/components/resource-processor.html
-  #resource/add_environment:
-    #attributes:
-      #- action: insert
-        #value: staging/production/...
-        #key: deployment.environment
+  resource/add_environment:
+    attributes:
+      - action: insert
+        value: default
+        key: deployment.environment
 
 exporters:
   # Traces
@@ -172,7 +171,7 @@ service:
       - memory_limiter
       - batch
       - resourcedetection
-      #- resource/add_environment
+      - resource/add_environment
       exporters: [sapm, signalfx]
       # Use instead when sending to gateway
       #exporters: [otlp, signalfx]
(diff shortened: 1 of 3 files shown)


cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -114,15 +114,14 @@ processors:
     detectors: [gcp, ecs, ec2, azure, system]
     override: true
 
-  # Optional: The following processor can be used to add a default "deployment.environment" attribute to the logs and 
+  # The following processor can be used to add a default "deployment.environment" attribute to the logs and
```
crobert-1, line 117: I'm a bit confused by this setup. The comment wording is "can be used", but by default it's adding the value `default` as the resource attribute `deployment.environment`. If it's enabled by default should we word it differently, like "is used"? Do we want the default value to be `default`? Is this something that everyone installing the collector will be aware of to update, does it matter?

Lots of questions, I'm just unsure about the UX of this.

jinja2: Yeah, this is a bit confusing. Should the value be set from a variable which is now made required in different methods of installation? Having the value be `default` does not seem correct.

jvoravong: This change is for a specific Splunk distribution initiative.
- In order for traces to work within Splunk products, `deployment.environment` must be set. Product has currently decided the value of this attribute should be `default`.
- The `deployment.environment` in the OpenTelemetry community is marked as ["recommended"](https://opentelemetry.io/docs/specs/semconv/resource/deployment-environment/) For Splunk, it is a requirement for trace data and for some correlation of telemetry data.
- We are still have some convesations at Splunk for how we should influence the OpenTelemetry community and how internal Splunk should use the  `deployment.environment` attribute. As of right now, this PR best support the direction Splunk product wants to move forward with.

atoulme: I have made a clarification in the docs.

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -114,15 +114,14 @@ processors:
...
   # For more information, see https://docs.splunk.com/Observability/gdi/opentelemetry/components/resource-processor.html
-  #resource/add_environment:
-    #attributes:
-      #- action: insert
-        #value: staging/production/...
-        #key: deployment.environment
+  resource/add_environment:
+    attributes:
+      - action: insert
+        value: default
```
pjanotti, line 123: For me it feels like it will pull a default value from somewhere, not that the value will be `default`. That said I don't have a good alternative, even a different value like `default_environment` doesn't improve much.

pjanotti (approved):

cmd/otelcol/config/collector/agent_config.yaml
theletterf: "Default" might have a pre-existing meaning for customers, too.

Could we call it something else perhaps? Something that signals that the environment hasn't been defined by instrumentation libraries? 

Some ideas:

- `empty`
- `undefined` (Prolly not, as it has meanings in programming languages)
- `none`

atoulme:
Discussed with Dmitrii and Aunsh, and there is a better way Josh offered that would do the job with existing and new installations.

PR #4228, opened 2024-01-30
tidy all

hughesjj:
Ran `make tidy-all` on main


crobert-1 (changes requested):
I think this may be an issue of local go version being different/upgraded vs. the repo itself. I don't think we should commit this.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/signalfx/splunk-otel-collector
 
-go 1.20
+go 1.21
+
+toolchain go1.21.6
```
atoulme, line 5: please stick with 1.20 for now.

PR #4233, opened 2024-02-01
Update Python to 3.11.7

atoulme:


crobert-1 (approved):
Looks good to me, as long as tests pass.

Do we have a full list of all places the Python version is defined, and what all needs to be updated? I'm seeing a lot of references in CI actions as well, not sure if they should be updated with this or not.

pjanotti (approved):

jinja2 (approved):
LGTM. Since this is a patch version update, I assume this wouldn't impact any custom collects plugins 🤞

atoulme:
Our tests check collectd plugins, for the most part. Any gap in testing, we should make sure we catch it with more tests. Let me know if you see any gaps under https://github.com/signalfx/splunk-otel-collector/tree/main/tests/receivers/smartagent and we can add more tests :)

PR #4237, opened 2024-02-01
Update Ansible test tools

pjanotti:
**Description:**
Update tools used in Ansible tests, the main target being `molecule-vagrant` that was archived one year ago.

The new tools complained about the role "collector" not being found, [error message](https://github.com/pjanotti/splunk-otel-collector/actions/runs/7746705618/job/21125532013#step:6:119):
```terminal
  ERROR! the role 'collector' was not found in /home/runner/work/splunk-otel-collector/splunk-otel-collector/deployments/ansible/molecule/custom_vars/roles:/home/runner/.cache/molecule/ansible/custom_vars/roles:/home/runner/work/splunk-otel-collector/splunk-otel-collector/deployments:/home/runner/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:/home/runner/work/splunk-otel-collector/splunk-otel-collector/deployments/ansible/molecule/custom_vars
  
  The error appears to be in '/home/runner/work/splunk-otel-collector/splunk-otel-collector/deployments/ansible/molecule/custom_vars/converge.yml': line 26, column 15, but may
  be elsewhere in the file depending on the exact syntax problem.
  
  The offending line appears to be:
  
        include_role:
          name: "collector"
                ^ here

```
Reading our docs it seems that we need to use a full-qualified name. I considered if `ANSIBLE_ROLES_PATH` should be used, but, since the full-qualified name resolved and matches what is defined on the `galaxy.yml` I opted to use it.
 
There are also more warnings, which seems related to the docker package, being logged during the tests.

On the positive side it seems that is makes slightly less like for the Windows test to hit the `gurumeditation` state: it still does that for the majority of the tests but I got 4 passing on Windows.

**Link to Splunk idea:**
https://splunk.atlassian.net/browse/OTL-2555

**Testing:**
Windows tested locally on mac Intel.

**Documentation:**
N/A

### DIFF
--- a/deployments/ansible/molecule/custom_vars/converge.yml
+++ b/deployments/ansible/molecule/custom_vars/converge.yml
@@ -23,4 +23,4 @@
   tasks:
     - name: "Include signalfx.splunk_otel_collector.collector role"
       include_role:
-        name: "collector"
+        name: "signalfx.splunk_otel_collector.collector"
--- a/.github/workflows/ansible.yml
+++ b/.github/workflows/ansible.yml
@@ -93,10 +93,10 @@ jobs:
           path: "${{ github.workspace }}/requirements.txt"
           contents: |
             ${{ matrix.ansible }}
-            molecule==3.3.0
-            molecule-docker==0.2.4
-            docker==5.0.0
-            ansible-lint==5.4.0
+            ansible-lint==6.22.2
+            molecule==6.0.3
+            molecule-plugins[docker]==23.5.0
+            docker==7.0.0
             urllib3<2
 
       - name: Set up Python 3.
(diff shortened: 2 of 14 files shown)


deployments/ansible/molecule/custom_vars/converge.yml
```diff
@@ -23,4 +23,4 @@
   tasks:
     - name: "Include signalfx.splunk_otel_collector.collector role"
       include_role:
-        name: "collector"
+        name: "signalfx.splunk_otel_collector.collector"
```
jinja2, line 26: LGTM. Another option would be to add the path to roles explicitly like below under the [ansible provisioner config](https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/ansible/molecule/config/docker.yml#L21). Maybe add the path and keep the full name both to cover all bases?
```
  env:
    ANSIBLE_ROLES_PATH: ../../roles
 ```
 
Probably not related but another thing I noticed is the [role_name in galaxy_info](https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/ansible/roles/collector/meta/main.yml#L3) is incorrect. I wonder if this is just a leftover from a legacy role.

pjanotti: Giving a spin to `ANSIBLE_ROLES_PATH` it seems a good addition to make the tests consistent between the platforms - now I have to understand why there is a vagrant.yml...

Regarding role_name on the galaxy file, per schema at https://raw.githubusercontent.com/ansible/ansible-lint/main/src/ansiblelint/schemas/meta.json it doesn't allow the `.` so keeping as it is, at least for now.

pjanotti:
After 11 attempts it clears the Windows runs... not good, but, recording this info...

jinja2 (approved):

PR #4240, opened 2024-02-02
remove unused code for filtering as it is no longer in execution path

atoulme:
**Description:**
Remove code that is no longer in the execution path for dimension and metric filtering.


jinja2 (approved):

pjanotti (approved):
Nice!

PR #4243, opened 2024-02-02
Output pdata

atoulme:
**Description:**
Avoid the translation to signalfx.Datapoint and directly ingest data as pmetric.Metric data.


samiura:
All the commits are unsigned.

samiura (changes requested):
unsigned commits

PR #4262, opened 2024-02-06
inherited quay login

bmeller:
Move quay login from jobs to template, what fixes windows2022 access issue

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,7 +54,11 @@ fossa:
 
 .sign-docker:
   extends: .trigger-filter
+  allow_failure: true
   stage: sign-image
+  before_script:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -54,7 +54,11 @@ fossa:
 
 .sign-docker:
   extends: .trigger-filter
+  allow_failure: true
```
jvoravong, line 57: Do we need this allow_failure?

bmeller: Underlying signing service sometimes returns 500 randomly (already reported to its maintainers). I didn't want it to block deployments. On the other hand, we can always rety the job. I will remove it.

atoulme (approved):

PR #4270, opened 2024-02-07
move off sapm and use otlp by default

atoulme:
**Description:**
This PR removes sapm from the default configuration, and use otlp instead.
The PR also introduces a converter that will take the default sapm exporter configuration and replace it with an otlp exporter.


pjanotti (commented):
@atoulme the changes to the MSI added another location with `/v2/trace` when building the default value for `SPLUNK_TRACE_URL`. Could you, please, grep again for `/v2/trace` to cover those?

.github/workflows/scripts/win-test-services.ps1
```diff
@@ -44,7 +44,9 @@ $expected_svc_env_vars = @{
   "SPLUNK_REALM"            = "$realm";
   "SPLUNK_API_URL"          = "$api_url";
   "SPLUNK_INGEST_URL"       = "$ingest_url";
-  "SPLUNK_TRACE_URL"        = "${ingest_url}/v2/trace";
+  # We no longer test for this as we have moved to OTLP and it breaks upgrade tests.
+  # We have a configuration converter in place to catch old values as part of the collector startup.
+  #"SPLUNK_TRACE_URL"        = "${ingest_url}:443;
```
pjanotti, line 49: Hum, this can be changed in the test, as long as we track the first version doing the switch. Anyway, much better as a follow-up then in this PR.

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -89,9 +89,10 @@ processors:
 
 exporters:
   # Traces
-  sapm:
-    access_token: "${SPLUNK_ACCESS_TOKEN}"
-    endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace"
+  otlp:
+    endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com:443"
```
pjanotti, line 93: q: why in some places we explicitly have the default HTTPS port and others not? It doesn't matter since 443 the default anyway, just curious if this is intentional or not.

pjanotti (approved):
Most places have the port `:443` appended to the default value, it will be good if all of them had it - I mean that the default was consistent everywhere, in this case doesn't matter much if the port is included or not.

Other than that, LGTM.

.github/workflows/scripts/win-test-services.ps1
atoulme: Note 443 is not optional right now.

cmd/otelcol/config/collector/gateway_config.yaml
atoulme: See https://github.com/open-telemetry/opentelemetry-collector/issues/9505 for context: right now the code fails if no port is explicitly provided, which really is a bad outcome.

HTTPS means nothing for grpc endpoints, but is used to indicate that a TLS connection is necessary. See https://github.com/open-telemetry/opentelemetry-specification/pull/1729 for more information.

cmd/otelcol/config/collector/upstream_agent_config.yaml
```diff
@@ -164,30 +165,30 @@ service:
       - batch
       - resourcedetection
       #- resource/add_environment
-      exporters: [sapm, signalfx]
+      exporters: [otlp, signalfx]
       # Use instead when sending to gateway
-      #exporters: [otlp, signalfx]
+      #exporters: [otlp/gateway, signalfx]
```
dmitryax, line 170: Why do we stick with these comments in every pipeline? Won't it be clearer to remove `otlp/gateway` and put this comment just in one place: `exporters::otlp`?

tests/zeroconfig/windows/testdata/resource_traces/aspnetcore.yaml
```diff
@@ -2,7 +2,6 @@ resource_spans:
 - attributes:
     deployment.environment: zc-iis-test
     host.name: <ANY>
-    os.type: windows
```
dmitryax, line 5: Why is this changed? Is this what was set by the collector, and now we send data directly?

cmd/otelcol/config/collector/upstream_agent_config.yaml
dmitryax: Having two otlp exporters active but using only one of them doesn't seem right

dmitryax: Submitted https://github.com/signalfx/splunk-otel-collector/pull/4587

cmd/otelcol/config/collector/ecs_ec2_config.yaml
```diff
@@ -122,9 +122,10 @@ processors:
 
 exporters:
   # Traces
-  sapm:
-    access_token: "${SPLUNK_ACCESS_TOKEN}"
-    endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace"
+  otlp:
+    endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com"
```
pjanotti, line 126: @atoulme so places like this are broken

cmd/otelcol/config/collector/fargate_config.yaml
```diff
@@ -96,9 +96,10 @@ processors:
 
 exporters:
   # Traces
-  sapm:
-    access_token: "${SPLUNK_ACCESS_TOKEN}"
-    endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com/v2/trace"
+  otlp:
+    endpoint: "https://ingest.${SPLUNK_REALM}.signalfx.com"
```
pjanotti, line 100: @atoulme same here

tests/zeroconfig/windows/testdata/resource_traces/aspnetcore.yaml
pjanotti: The collector receiving the signals from the instrumentation is not using the default agent configuration anymore, the stripped down config used now doesn't have the `resourcedetection` processor that adds that attribute.

cmd/otelcol/config/collector/upstream_agent_config.yaml
atoulme: thanks

PR #4323, opened 2024-02-23
Add a resource attribute to internal metrics to track discovery usage

dmitryax:
OTL-2534

### DIFF
--- a/internal/configconverter/discovery.go
+++ b/internal/configconverter/discovery.go
@@ -69,8 +69,26 @@ func (Discovery) Convert(_ context.Context, in *confmap.Conf) error {
 		metricsPipeline["receivers"] = appendUnique(metricsReceivers, discoReceivers)
 	}
 
+	setAutoDiscoveryResourceAttribute(service)
+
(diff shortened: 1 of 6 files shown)


atoulme (approved):

internal/configconverter/discovery.go
```diff
@@ -69,10 +69,28 @@ func (Discovery) Convert(_ context.Context, in *confmap.Conf) error {
...
+	}
+	service["telemetry"] = telemetry
+
+	resAttrs := map[string]any{}
+	if ra, ok := telemetry["resource"]; ok {
+		resAttrs = ra.(map[string]any)
+	}
+	telemetry["resource"] = resAttrs
+
+	resAttrs["splunk_autodiscovery"] = "true"
```
jinja2, line 91: should we make it more inline with otel convention and have it be `splunk.autodiscovery`?

dmitryax: It's passed through the prometheus, so it'll be replaced with `splunk_autodiscovery` anyway.

samiura (approved):

PR #4343, opened 2024-02-26
Windows script to install Splunk distro for .NET instead of SFx.NET

pjanotti:
**Description:**
Change the Windows install script to install the [Splunk Distribution of OpenTelemetry .NET](https://docs.splunk.com/observability/en/gdi/get-data-in/application/otel-dotnet/get-started.html#instrument-net-applications-for-splunk-observability-cloud-opentelemetry)
  instead of the [SignalFx Instrumentation for .NET](https://docs.splunk.com/observability/en/gdi/get-data-in/application/otel-dotnet/sfx/sfx-instrumentation.html#signalfx-instrumentation-for-net-deprecated)
  when the parameter `-with_dotnet_instrumentation` is set to `$true`

**Link to Splunk idea:**
N/A

**Testing:**
Local and CI validation on fork.

**Documentation:**
Pending: update install script documentation.

### DIFF
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -69,7 +69,8 @@
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_fluentd $true
 .PARAMETER with_dotnet_instrumentation
-    (OPTIONAL) Whether to install and configure .NET tracing to forward .NET application traces to the local collector (default: $false)
+    (OPTIONAL) Whether to install and configure the Splunk Distribution of OpenTelemetry .NET to forward .NET application telemetry to the local collector (default: $false).
+    If the instrumentation is already installed the script will update it to the latest version - stop all instrumented applications before attempting to update the instrumentation.
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true
 .PARAMETER deployment_env
@@ -703,7 +704,14 @@ if ($with_dotnet_instrumentation) {
     if (-not $currentInstallVersion) {
         Install-OpenTelemetryCore
     } else {
-        Update-OpenTelemetryCore
+        Write-Warning "Splunk Distribution of OpenTelemetry .NET is already installed. Updating to the latest version..."
+        try {
+            Update-OpenTelemetryCore
+        }
+        catch {
+            Write-Error "Failed to update Splunk Distribution of OpenTelemetry .NET. Stop all instrumented applications and then try again."
+            throw $_
+        }
     }
 
     $installed_version = Get-OpenTelemetryInstallVersion
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -492,7 +492,7 @@ $tempdir = create_temp_dir -tempdir $tempdir
 
 if ($with_dotnet_instrumentation) {
     if ((msi_installed -name "SignalFx .NET Tracing 64-bit") -Or (msi_installed -name "SignalFx .NET Tracing 32-bit")) {
-        throw "SignalFx .NET Instrumentation is already installed. Uninstall SignalFx Instrumentation for .NET and rerun this script."
+        throw "SignalFx .NET Instrumentation is already installed. Stop all instrumented applications and uninstall SignalFx Instrumentation for .NET before running this script again."
     }
     echo "Downloading Splunk Distribution of OpenTelemetry .NET ..."
     $module_name = "Splunk.OTel.DotNet.psm1"
@@ -719,6 +719,7 @@ if ($otel_resource_attributes -ne "") {
     try {
         update_registry -path "$regkey" -name "OTEL_RESOURCE_ATTRIBUTES" -value "$otel_resource_attributes"
     } catch {
+        Write-Warning "Failed to set OTEL_RESOURCE_ATTRIBUTES environment variable."
         continue
     }
 }
--- a/internal/buildscripts/packaging/installer/install.ps1
+++ b/internal/buildscripts/packaging/installer/install.ps1
@@ -73,7 +73,7 @@
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true
 .PARAMETER deployment_env
-    (OPTIONAL) A system-wide SignalFx "environment" used by .NET instrumentation. Sets the SIGNALFX_ENV environment variable. Ignored if -with_dotnet_instrumentation is false.
+    (OPTIONAL) A system-wide "deployment.environment" set via the environment variable 'OTEL_RESOURCE_ATTRIBUTES' for the whole machine. Ignored if -with_dotnet_instrumentation is false.
     .EXAMPLE
     .\install.ps1 -access_token "ACCESSTOKEN" -with_dotnet_instrumentation $true -deployment_env staging
 .PARAMETER bundle_dir
@@ -414,7 +414,7 @@ function download_collector_package([string]$collector_version=$collector_versio
 }
 
 # check registry for the agent msi package
-function msi_installed([string]$name="Splunk OpenTelemetry Collector") {
+function msi_installed([string]$name) {
     return (Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Where { $_.DisplayName -eq $name }) -ne $null
 }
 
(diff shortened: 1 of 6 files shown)


samiura (approved):

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -689,28 +689,50 @@ restarted to apply the changes by restarting the system or running the following
...
-        echo "Setting SIGNALFX_ENV environment variable to $deployment_env ..."
-        update_registry -path "$regkey" -name "SIGNALFX_ENV" -value "$deployment_env"
+if ($with_dotnet_instrumentation) {
+    echo "Installing Splunk Distribution of OpenTelemetry .NET..."
+    $currentInstallVersion = Get-OpenTelemetryInstallVersion
+    if (-not $currentInstallVersion) {
+        Install-OpenTelemetryCore
     } else {
-        echo "SIGNALFX_ENV environment variable not set. Unless otherwise defined, will appear as 'unknown' in the UI."
+        Update-OpenTelemetryCore
```
RassK, line 706: Here is a risk of doing unwanted silent update?

internal/buildscripts/packaging/installer/install.ps1
```diff
@@ -689,28 +689,50 @@ restarted to apply the changes by restarting the system or running the following
...
+    $otel_resource_attributes += "splunk.zc.method=splunk-otel-dotnet-$installed_version"
+}
+
+if ($otel_resource_attributes -ne "") {
+    # The OTEL_RESOURCE_ATTRIBUTES environment variable must be set before restarting IIS.
+    $regkey = "HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment"
     try {
-        $dotnet_version = (Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Where { $_.DisplayName -eq "SignalFx .NET Tracing 64-bit" }).DisplayVersion
-        update_registry -path "$regkey" -name "SIGNALFX_GLOBAL_TAGS" "splunk.zc.method:signalfx-dotnet-tracing-${dotnet_version}"
+        update_registry -path "$regkey" -name "OTEL_RESOURCE_ATTRIBUTES" -value "$otel_resource_attributes"
```
RassK, line 720: Not sure what to do here, but `OTEL_RESOURCE_ATTRIBUTES` seems to be conflicting with instrumenting agent and could leak to an instrumented process when in a global scope.

pjanotti: I didn't want to change this on this PR but the main thing here to me is that the `splunk.zc.method` seems to need a different value since we could have multiple instrumentations on the same box. However, `deployment.environment` is something that we want for the whole machine. I will have a conversation with Rotel about it before changing it, the current code is in line with what mass deployments are doing.

RassK: I guess the conflict is getting even bigger then? ... since envs don't stack by default and `OTEL_RESOURCE_ATTRIBUTES` can be easily overwritten (due being a generic variable) by a client.

internal/buildscripts/packaging/installer/install.ps1
pjanotti: Good question. Most of the time the instrumented applications are running in IIS and the update will stop and recycle them which seems what the typical user wants. However, if there are instrumented .NET applications outside IIS then the update will fail when copying files which probably is not very helpful. Let me think a bit more about this case.

pjanotti: a283f827

RassK: I'm thinking maybe to fail with a specific message "... dotnet is already installed, run the install script again with -UpdateDotnet to update the agent'" 🤔

tests/zeroconfig/windows/testdata/resource_traces/aspnetcore.yaml
```diff
@@ -3,22 +3,32 @@ resource_spans:
     deployment.environment: zc-iis-test
     host.name: <ANY>
     os.type: windows
+    process.pid: <ANY>
+    process.runtime.description: .NET 6.0.27
+    process.runtime.name: .NET
+    process.runtime.version: 6.0.27
```
Kielek, line 9: Tests can start failing if you upgrade .NET version.
I am not sure if you can add regex here something like
```suggestion
    process.runtime.description: .NET \d.\d.\d+
    process.runtime.name: .NET
    process.runtime.version: \d.\d.\d+
```

Not necessary part of this PR.

Alternatively, you can disable `PROCESSRUNTIME` resource detector.

tests/zeroconfig/windows/testdata/resource_traces/aspnetfx.yaml
```diff
@@ -3,30 +3,30 @@ resource_spans:
     deployment.environment: zc-iis-test
     host.name: <ANY>
     os.type: windows
+    process.pid: <ANY>
+    process.runtime.description: .NET Framework 4.8.4645.0
+    process.runtime.name: .NET Framework
+    process.runtime.version: "4.8"
```
Kielek, line 9: Regexes, as for .NET?

Kielek (approved):

theletterf:
+CC @mbechtold-splunk

aunshc:
@pjanotti to better understand our recommendation for existing users moving forward, today we ask them to upgrade the Collector by manually pulling the new Windows MSI correct (https://docs.splunk.com/observability/en/gdi/opentelemetry/collector-windows/windows-upgrade.html#id1)?

internal/buildscripts/packaging/installer/install.ps1
pjanotti: We will have to investigate this separately, likely also changing the value that we use for OTel resource attributes.

tests/zeroconfig/windows/testdata/resource_traces/aspnetcore.yaml
pjanotti: I want to keep the default configuration and this seems a good check to keep in place.

The idea of using regular expressions is very good and the the `testutils` package support that via the `<RE2()>` tag, unfortunately it is not implemented for spans, only for metrics and logs. I will create a KTLO ticket to add it to spans and link to this comment so we can update this test when that is in place.

internal/buildscripts/packaging/installer/install.ps1
pjanotti: The warning message is `Splunk Distribution of OpenTelemetry .NET is already installed. Updating to the latest version...` and will be displayed before any error so I think it gives enough context in case of an error updating.

pjanotti:
> to better understand our recommendation for existing users moving forward, today we ask them to upgrade the Collector by manually pulling the new Windows MSI correct (https://docs.splunk.com/observability/en/gdi/opentelemetry/collector-windows/windows-upgrade.html#id1)?

Yes, the current recommendation for update is to use the MSI directly, not, the script. That said the script still needs to handle the case in which someone attempts to run it with the current version installed. This works for the collector because it is a single process that we need to stop. For the instrumentations the problem is a bit different because the user needs to stop the instrumented applications or alternatively upgrade and then reboot the machine to switch to the new version - the reboot is needed since files will be locked by the running process. Unfortunately this via reboot requires an MSI or code that we don't have in the current .NET Powershell script. Anyway, the recommendation should still be the same:

1. To upgrade the collector, download and run the MSI of the new version - caveat: I need to document when upgrading to the version that uses the service scope environment, since this is a special case. I will create a KTLO ticket for it.
2. To upgrade the .NET instrumentation OTel.NET to OTel.NET: stop all instrumented applications, download and run the upgrade function from powershell script for OTel .NET. Finally, restart all the applications. 
3. To upgrade the .NET instrumentation SFx.NET to OTel.NET: stop all instrumented applications, uninstall SFx.NET, reboot if needed (only if failed to stop all instrumented applications), install OTel.NET either via the powershell script (either the one from OTel.NET or the collector one).

internal/buildscripts/packaging/installer/install.ps1
pjanotti: After writing the answer to Aunsh I think it will make more sense to just bail out here: it will be consistent with what is done for the collector MSI: the script just bails out if there is already one version installed. We are also bailing out if SFx.NET is already installed.

tests/zeroconfig/windows/testdata/resource_traces/aspnetcore.yaml
pjanotti: https://splunk.atlassian.net/browse/OTL-2617

RassK (approved):

atoulme (approved):

PR #4352, opened 2024-02-29
MSI defaults to per machine install

pjanotti:
**Description:**
This avoids a confusing case in which different admins are installing/updating the collector on a Windows box. Installing per machine, i.e. with `ALLUSERS=1`, avoids the issue. 

**Link to Splunk support:**
https://splunk.atlassian.net/browse/SWAT-7073?focusedCommentId=14264894

**Testing:**
Manually tested the MSI and installed it with different administrators.

**Documentation:**
Follow-up to update Splunk doc mentions of ALLUSERS.

### DIFF
--- a/internal/buildscripts/packaging/msi/splunk-otel-collector.wxs
+++ b/internal/buildscripts/packaging/msi/splunk-otel-collector.wxs
@@ -1,6 +1,6 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
    <Product Id="*" UpgradeCode="fde3e4d9-9ca5-4c82-be7b-81445ab5b605" Name="Splunk OpenTelemetry Collector" Version="$(var.Version)" Manufacturer="Splunk, Inc." Language="1033">
-      <Package InstallerVersion="200" Compressed="yes" Comments="Windows Installer Package"/>
+      <Package InstallerVersion="200" Compressed="yes" Comments="Windows Installer Package" InstallPrivileges="elevated" InstallScope="perMachine" />
       <Media Id="1" Cabinet="product.cab" EmbedCab="yes"/>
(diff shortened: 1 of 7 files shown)


pjanotti:
Somehow my change broke choco uninstall...

internal/buildscripts/packaging/msi/splunk-otel-collector.wxs
```diff
@@ -1,6 +1,6 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
    <Product Id="*" UpgradeCode="fde3e4d9-9ca5-4c82-be7b-81445ab5b605" Name="Splunk OpenTelemetry Collector" Version="$(var.Version)" Manufacturer="Splunk, Inc." Language="1033">
-      <Package InstallerVersion="200" Compressed="yes" Comments="Windows Installer Package"/>
+      <Package InstallerVersion="200" Compressed="yes" Comments="Windows Installer Package" InstallPrivileges="elevated" InstallScope="perMachine" />
```
jinja2, line 3: Does this mean that it can only be installed/started by a machine admin? I am wondering if we should have the default set to the equivalent of `ALLUSERS=2` instead. From the docs, `The value ALLUSERS=2 enables the system to reset the value of ALLUSERS, and the installation context, dependent upon the user's privileges and the version of Windows.`. I gather this would mean that if an admin installs it, the scope will be per-machine, but a normal user install per-user which will make this work for both cases?

pjanotti: Good Q. Your reading is the same as mine, however, a non-admin user can't install a Windows service, so in practice the MSI already requires admin to install the collector, so there is no practical value in our case.

samiura (approved):

PR #4356, opened 2024-03-01
update nomad

atoulme:
**Description:**
Update nomad to test 1.6.2 and 1.7.5.

### DIFF
--- a/deployments/nomad/otel-agent.nomad
+++ b/deployments/nomad/otel-agent.nomad
@@ -5,7 +5,7 @@ job "otel-agent" {
   constraint {
     attribute = "${attr.nomad.version}"
     operator  = "semver"
-    value     = "< 1.3.0"
+    value     = "< 1.8.0"
   }
 
   group "otel-agent" {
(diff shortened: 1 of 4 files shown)


deployments/nomad/otel-agent.nomad
```diff
@@ -5,7 +5,7 @@ job "otel-agent" {
   constraint {
     attribute = "${attr.nomad.version}"
     operator  = "semver"
-    value     = "< 1.3.0"
+    value     = "< 1.8.0"
```
jinja2, line 8: should this be `>= 1.6.2, , < 1.8.0` instead? Not sure we are checking for version 1.8.0 when the readme say 1.6.2+

atoulme: Sorry, I had removed 1.6.2 from the test matrix. I was ambivalent as to whether we needed to have it on, but I guess we should.

atoulme:
@jinja2 ready for another look, thanks!

crobert-1 (approved):

PR #4365, opened 2024-03-02
[draft] removed memory balast and using GOMEMLIMIT. OTL-2488

samiura:
**Description:** Removed memory_ballast and using GOMEMLIMIT


internal/settings/settings.go
```diff
@@ -549,6 +532,22 @@ func setMemoryLimit(memTotalSizeMiB int) (int, error) {
 	return memLimit, nil
 }
 
+// Validate and set GOMEMLIMIT
```
pjanotti, line 535: `GOMEMLIMIT` is already validated by the runtime, if it is invalid the go executable won't start (unless it is an old version that didn't support `GOMEMLIMIT`).

samiura: good catch but we have more fundamental issue. More on stand up. for now, I am closing the PR.

PR #4390, opened 2024-03-04
Upgrade golang dependency to 1.21.5

crobert-1:

Update matches core/contrib upstream go version.

### DIFF
--- a/.github/workflows/auto-instrumentation.yml
+++ b/.github/workflows/auto-instrumentation.yml
@@ -21,7 +21,7 @@ env:
   PYTHON_VERSION: '3.11'
   PIP_VERSION: '22.0.4'
   REQUIREMENTS_PATH: "internal/buildscripts/packaging/tests/requirements.txt"
-  GO_VERSION: ~1.21.5
+  GO_VERSION: 1.21.5
 
 jobs:
   cross-compile:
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### 💡 Enhancements 💡
+
+- (Splunk) Update golang to 1.21.5 ()
```
crobert-1, line 7:
```suggestion
- (Splunk) Update golang to 1.21.5 ([#4390](https://github.com/signalfx/splunk-otel-collector/pull/4390))
```

.github/workflows/auto-instrumentation.yml
```diff
@@ -21,7 +21,7 @@ env:
   PYTHON_VERSION: '3.11'
   PIP_VERSION: '22.0.4'
   REQUIREMENTS_PATH: "internal/buildscripts/packaging/tests/requirements.txt"
-  GO_VERSION: 1.20.10
+  GO_VERSION: ~1.21.5
```
crobert-1, line 24: The `~` addition is not required here, it just allows more flexibility with which version of `go` is used, and now matches `contrib` and `core` more closely. I'm happy to revert if that's preferred.

crobert-1: I'm actually going to remove this for now, if we want it we can just open another PR.

PR #4392, opened 2024-03-05
Removed memory balast and setting soft limit 

samiura:
**Description:** Removed memory balast and setting soft limit if GOMMEMLIMIT  is absent.

**Link to Splunk idea:** 

**Testing:** Unit tests, integration tests and memory profiling using pprof

PR #4393, opened 2024-03-05
Remove fieldalignment tool and check

pjanotti:
**Description:** 
Noticed while checking the status of GH actions: this was being flagged as error on the "Checks" tab, but, the make file ignores the actual result. After digging the history my understanding is that the fieldalignment is disabled both in core and contrib:

* Tool seems to have been introduced via https://github.com/signalfx/splunk-otel-collector/commit/b392b5b734a7c8ef68f3896c58a4aee52f1226de
* https://github.com/open-telemetry/opentelemetry-collector/issues/2789#issuecomment-1867105194

Noise to be removed by this PR: the "2" issues for build-and-test https://github.com/signalfx/splunk-otel-collector/pull/4392/checks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,8 +107,6 @@ install-tools:
 	cd ./internal/tools && go install github.com/pavius/impi/cmd/impi
 	cd ./internal/tools && go install github.com/tcnksm/ghr
 	cd ./internal/tools && go install golang.org/x/tools/cmd/goimports
-	cd ./internal/tools && go install golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment
-
 
 .PHONY: generate-metrics
 generate-metrics:
--- a/Makefile.Common
+++ b/Makefile.Common
@@ -104,7 +104,6 @@ checklinks:
 fmt: addlicense misspell-correction
 	gofmt -w -s .
 	goimports -w  -local github.com/signalfx/splunk-otel-collector ./
-	fieldalignment -fix ./... || true
 
 .PHONY: lint
 lint: checklicense misspell impi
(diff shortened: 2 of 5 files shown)


crobert-1:
It looks like the goal in core and contrib is to [eventually enable this](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f27143fae73a4a0631db8c55e46f6821d6121f76/.golangci.yml#L78-L80), so I'd lean more towards keeping it and fixing any failures we bump into.

Also, it looks like the failure in `build-and-test` that you linked is coming from the `generate-metrics` job, isn't it? Or am I looking at something else?

pjanotti:
> It looks like the goal in core and contrib is to [eventually enable this](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f27143fae73a4a0631db8c55e46f6821d6121f76/.golangci.yml#L78-L80), so I'd lean more towards keeping it and fixing any failures we bump into.

I think contrib didn't catch to core yet:
https://github.com/open-telemetry/opentelemetry-collector/blob/420772e09f51951abc873fd5e5c1fc9103a25407/.golangci.yml#L74-L78

At minimum `fieldalignment` should skip generated files. Alternatively, if we think that it is worth the trouble we should fix the generator to create the data structures as preferred instead of having them repaired afterwards by the tool. 

> Also, it looks like the failure in build-and-test that you linked is coming from the generate-metrics job, isn't it? Or am I looking at something else?

🤦🏼 I forgot to add the generated files not aligned to the PR.

crobert-1:
Thanks for linking to core, I only checked contrib 👍 (Shouldn't have included core in my comment). The generated files use case makes sense to make as to why we should remove it. It would be nice though, as you pointed out, to only disable for specific files if possible, or even generate them properly.

pjanotti:
I'm going to close it for now and will get back to it later.

PR #4395, opened 2024-03-05

srv-gh-o11y-gdi:


crobert-1:
We need to upgrade to v0.95.0 first, then we'll upgrade to v0.96.0. I'll have to open this manually, and then hopefully we'll be back in sync with upstream so the automated PRs can work again 👍

PR #4404, opened 2024-03-07
Removed memory_ballast and making sure soft memory is set

samiura:
**Description:** Removed memory_ballast and making sure soft memory is set


tests/helm/helm_test.go
```diff
@@ -137,8 +137,8 @@ gateway:
   replicaCount: 1
   resources:
     limits:
-      cpu: 200m
-      memory: 128Mi
+      cpu: 1024m
+      memory: 512Mi
```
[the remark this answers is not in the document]
samiura: @dmitryax this is not needed at all. I was under the impression this will fix. Let me remove it.

internal/settings/settings.go
```diff
@@ -36,13 +37,13 @@ import (
 
 const (
 	APIURLEnvVar              = "SPLUNK_API_URL"
-	BallastEnvVar             = "SPLUNK_BALLAST_SIZE_MIB"
```
atoulme, line 39: Do we want to let people know to change this env var to GOMEMLIMIT?

samiura: My idea was GOMEMLIMIT env var do not have effect in the binary  or in other orders you cannot anything once the process start. The idea of declaring it here is just to check if the env is set already ( via helm chart or not ).  I think letting people know about this via Helm chart ( where it is effective ) should be more useful. We can update customer facing document for that.

internal/settings/settings_test.go
```diff
@@ -381,17 +354,22 @@ func TestSetDefaultFeatureGatesRespectsOverrides(t *testing.T) {
...
 	require.NoError(t, os.Setenv(MemTotalEnvVar, "200"))
-	require.NoError(t, os.Setenv(MemLimitMiBEnvVar, "150"))
-	require.NoError(t, os.Setenv(BallastEnvVar, "50"))
-
 	settings, err := New([]string{})
 	require.NoError(t, err)
 	require.NotNil(t, settings)
-	require.Equal(t, "50", os.Getenv(BallastEnvVar))
-	require.Equal(t, "150", os.Getenv(MemLimitMiBEnvVar))
+	require.Equal(t, int64(188743680), debug.SetMemoryLimit(100))
```
jinja2, line 365:
```suggestion
	require.Equal(t, int64(188743680), debug.SetMemoryLimit(-1))
```
If I am following the test correctly, it seems we just want to retrieve the currently set limit. Same for the test below.

jinja2:
Please add a changelog entry.

We are planning to update the default configs [here](https://github.com/signalfx/splunk-otel-collector/tree/main/cmd/otelcol/config/collector) in a separate PR? We'll also need to address ansible, chef, etc. deployments after this is available. They seem to set env var for ballast size.

internal/settings/settings.go
```diff
@@ -517,20 +516,12 @@ func envVarAsInt(env string) int {
...
-
-	_ = os.Setenv(BallastEnvVar, strconv.Itoa(ballastSize))
-	log.Printf("Set ballast to %d MiB", ballastSize)
-	return ballastSize
+// Check if the GOMEMLIMIT is specified via the env var, if not set the soft memory limit to 90% of the MemLimitMiBEnvVar
+func setSoftMemoryLimit(memTotalSizeMiB int) {
+	memLimit := int64(memTotalSizeMiB * DefaultMemoryLimitPercentage / 100)
+	// 1 MiB = 1048576 bytes
+	debug.SetMemoryLimit(memLimit * 1048576)
+	log.Printf("Set soft memory limit set to %d MiB", memLimit)
```
jinja2, line 524:
```suggestion
	log.Printf("Set soft memory limit for Go runtime to %d MiB", memLimit)
```

dmitryax:
As @jinja2 mentioned, we need to remove the `memory_balast` from all the configs like https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/agent_config.yaml. It should be done along with this change, probably the same PR. Please make sure `SPLUNK_BALLAST_SIZE_MIB` env var isn't referenced anymore across the repo.

internal/settings/settings_test.go
samiura: The way this test work bit differently, The idea of setting setting SetMemoryLimit is not to match the set memory limit ( 100 )  rather each time you use that API it returns the previously set memorylimit if it had been set.  Test just asserts if it was the soft limit was set or not. For more information -> https://pkg.go.dev/runtime/debug#SetMemoryLimit

jinja2: yes, `SetMemoryLimit returns the previously set memory limit. A negative input does not adjust the limit, and allows for retrieval of the currently set memory limit.` seems like you just want to retrieve it and passing -1 will do just that

samiura:
@jinja2  and @dmitryax , I will remove all the memory_ballast from configs. i was not sure if that was part of the scope of this PR.

internal/settings/settings_test.go
samiura: That works too!

deployments/salt/templates/agent_config.yaml
```diff
@@ -31,7 +29,7 @@ exporters:
     verbosity: normal
 
 service:
-  extensions: [memory_ballast]
+  extensions:
```
dmitryax, line 32: Not sure if leaving this empty is valid. It doesn't hurt to add `zpages` instead at least

internal/configconverter/remove_memory_ballast_key.go
```diff
@@ -0,0 +1,71 @@
...
+
+	const exp = "extensions::memory_ballast.*"
+	regExp := regexp.MustCompile(exp)
+
+	out := map[string]any{}
+	for _, k := range cfgMap.AllKeys() {
+		if regExp.MatchString(k) {
+			log.Println("[WARNING]  `memory_ballast` parameter in extensions is " +
+				"deprecated/removed. Please update the config according to the guideline: " +
+				"https://github.com/signalfx/splunk-otel-collector#from-0340-to-0350.")
```
dmitryax, line 54: Incorrect link. Maybe we don't even need instruction sections for this if it can fit this section

samiura: removed the instructions.

internal/configconverter/remove_memory_ballast_key.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	const exp = "extensions::memory_ballast.*"
+	regExp := regexp.MustCompile(exp)
+
+	out := map[string]any{}
+	for _, k := range cfgMap.AllKeys() {
+		if regExp.MatchString(k) {
+			log.Println("[WARNING]  `memory_ballast` parameter in extensions is " +
+				"deprecated/removed.")
```
dmitryax, line 53:
```suggestion
			log.Println("[WARNING]  `memory_ballast` parameter in extensions is deprecated. Please remove it from your configuration.")
```

internal/configconverter/remove_memory_ballast_key.go
```diff
@@ -0,0 +1,70 @@
...
+		} else {
```
dmitryax, line 54: This should have another condition: `k == service::extensions`. Otherwise, we will be removing `memory_ballast` from _any_ slice, which is very unlikely to be something else, but it's still better to ensure we avoid surprises.

samiura: done

internal/configconverter/remove_memory_ballast_key.go
samiura: added it.

internal/configconverter/remove_memory_ballast_key.go
```diff
@@ -0,0 +1,73 @@
...
+	const secondExp = "service::extensions"
+	secondRegExp := regexp.MustCompile(secondExp)
+
+	out := map[string]any{}
+	for _, k := range cfgMap.AllKeys() {
+		if firstRegExp.MatchString(k) {
+			log.Println("[WARNING]  `memory_ballast` parameter in extensions is deprecated. Please remove it from your configuration.")
+		} else {
+			out[k] = cfgMap.Get(k)
+			if secondRegExp.MatchString(k) {
```
dmitryax, line 57: No need to touch any other keys
```suggestion
		} else if secondRegExp.MatchString(k) {
```

internal/settings/settings.go
```diff
@@ -517,22 +523,29 @@ func envVarAsInt(env string) int {
 	return val
 }
 
-// Validate and set the memory ballast
+// Validate and set the memory ballast.
+// Note this will eventually be removed and here only for backwardcompatibility. Softlimit/GOMemlimit sets memory limit.
```
dmitryax, line 527:
```suggestion
// Note this will eventually be removed and here only for backward compatibility. Softlimit/GOMemlimit sets memory limit.
```

internal/settings/settings.go
```diff
@@ -517,22 +523,29 @@ func envVarAsInt(env string) int {
...
 func setMemoryBallast(memTotalSizeMiB int) int {
 	ballastSize := memTotalSizeMiB * DefaultMemoryBallastPercentage / 100
 	// Check if the memory ballast is specified via the env var, if so, validate and set properly.
 	if os.Getenv(BallastEnvVar) != "" {
 		ballastSize = envVarAsInt(BallastEnvVar)
-		if 33 > ballastSize {
-			log.Fatalf("Expected a number greater than 33 for %s env variable but got %d", BallastEnvVar, ballastSize)
+		if 10 > ballastSize {
+			log.Fatalf("Expected a number greater than 10 for %s env variable but got %d", BallastEnvVar, ballastSize)
```
dmitryax, line 534: Why is this change needed?

README.md
```diff
@@ -216,11 +216,8 @@ as a reference.
   Remove `ballast_size_mib` parameter from `memory_limiter` and make sure that it's added to
   `memory_ballast` extension as `size_mib` parameter instead:
 
-  ```yaml
-  extensions:
-    memory_ballast:
-      size_mib: ${SPLUNK_BALLAST_SIZE_MIB}
-  ```
+-  `memory_ballast` extension has been removed in favor of `GOMEMLIMIT` env var. If `GOMEMLIMIT` is not set, collector will enforce a soft limit automatically.
```
jinja2, line 219: Looks like this is being added to the upgrade guide's `### From 0.34.0 to 0.35.0` section. Were you trying to add it to some other section of the readme instead?

internal/settings/settings.go
```diff
@@ -56,7 +58,7 @@ const (
 	DefaultGatewayConfig           = "/etc/otel/collector/gateway_config.yaml"
 	DefaultOTLPLinuxConfig         = "/etc/otel/collector/otlp_config_linux.yaml"
 	DefaultConfigDir               = "/etc/otel/collector/config.d"
-	DefaultMemoryBallastPercentage = 33
+	DefaultMemoryBallastPercentage = 10
```
jinja2, line 61: Is there any reason to keep this var and `setMemoryBallast` around anymore?

samiura: backward compatibility, eventually it will go away. Look above to our other thread.

README.md
samiura: Well, I thought we should give a message but this is redundant. I will remove it.

internal/settings/settings.go
samiura: not at all needed!

internal/configconverter/remove_memory_ballast_key.go
```diff
@@ -0,0 +1,74 @@
...
+type RemoveMemoryBallastKey struct{}
+
+func removeMemoryBallast(strList []interface{}) []interface{} {
+	ret := make([]interface{}, 0)
+	for i, v := range strList {
+		if v == "memory_ballast" {
+			ret = append(ret, strList[:i]...)
+			return append(ret, strList[i+1:]...)
+		}
+	}
```
dmitryax, line 37: Seems like an extensions slice with no `memory_ballast` will be overridden to an empty slice. Please add a test case for that even if I'm wrong

samiura: I think you're correct about the method's inconsistency even though we are making sure to send slices with "memory_ballast" string. None the less, I have corrected the bug and added two new test cases. A test case just to test the helper function and an actual converter test without memory_ballast in the config

internal/configconverter/remove_memory_ballast_key_test.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func TestMemoryBallastConverter_Without_Memory_Ballast(t *testing.T) {
+	cfgMap, err := confmaptest.LoadConf("testdata/without_memory_ballast_config.yaml")
+	require.NoError(t, err)
+	require.NotNil(t, cfgMap)
+	pmp := RemoveMemoryBallastKey{}
+	assert.NoError(t, pmp.Convert(context.Background(), cfgMap))
+	assert.Equal(t, map[string]interface{}{"extensions": map[string]interface{}{"health_check": map[string]interface{}{"endpoint": "${SPLUNK_LISTEN_INTERFACE}:13133"}, "http_forwarder": map[string]interface{}{"egress": map[string]interface{}{"endpoint": "${SPLUNK_API_URL}"}, "ingress": map[string]interface{}{"endpoint": "${SPLUNK_LISTEN_INTERFACE}:6060"}}, "smartagent": map[string]interface{}{"bundleDir": "${SPLUNK_BUNDLE_DIR}", "collectd": map[string]interface{}{"configDir": "${SPLUNK_COLLECTD_DIR}"}}, "zpages": interface{}(nil)}, "service": map[string]interface{}{"extensions": []interface{}{"health_check", "http_forwarder", "zpages", "smartagent"}, "telemetry": map[string]interface{}{"metrics": map[string]interface{}{"address": "${SPLUNK_LISTEN_INTERFACE}:8888"}}}}, cfgMap.ToStringMap())
```
dmitryax, line 51: It's a bit hard to read and maintain. Can you read `testdata/without_memory_ballast_config.yaml` again instead?

internal/configconverter/testdata/without_memory_ballast_config.yaml
```diff
@@ -0,0 +1,23 @@
+extensions:
+  health_check:
+    endpoint: "${SPLUNK_LISTEN_INTERFACE}:13133"
+  http_forwarder:
+    ingress:
+      endpoint: "${SPLUNK_LISTEN_INTERFACE}:6060"
+    egress:
+      endpoint: "${SPLUNK_API_URL}"
+      # Use instead when sending to gateway
+      #endpoint: "${SPLUNK_GATEWAY_URL}"
```
dmitryax, line 10: Please remove comments from test configs

internal/settings/settings.go
dmitryax, line 61: Why is this changed?

samiura: removed. Not needed at all.

internal/configconverter/remove_memory_ballast_key_test.go
samiura: I agree, read from the same configmap and also for the other test create an yaml file and reading it to create a expected map.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+
+
```
dmitryax, line 6:
```suggestion
```

dmitryax (approved):
Please rebase

internal/configconverter/remove_memory_ballast_key.go
```diff
@@ -0,0 +1,74 @@
...
+			out[k] = cfgMap.Get(k)
+			temp := out[k]
+			switch temp := temp.(type) {
+			default:
+				continue
+			case []interface{}:
+				ret := removeMemoryBallastStrElementFromSlice(temp)
+				if len(ret) > 0 {
+					out[k] = ret
+				}
```
dmitryax, line 66: The condition is actually not needed. We need to be able to do `[ballast_extension]` -> `[]`. Please add a test :)

internal/configconverter/remove_memory_ballast_key.go
```diff
@@ -0,0 +1,74 @@
...
+
+	out := map[string]any{}
+	for _, k := range cfgMap.AllKeys() {
+		if firstRegExp.MatchString(k) {
+			log.Println("[WARNING]  `memory_ballast` parameter in extensions is deprecated. Please remove it from your configuration.")
+			continue
+		} else if secondRegExp.MatchString(k) {
+			out[k] = cfgMap.Get(k)
+			temp := out[k]
+			switch temp := temp.(type) {
```
dmitryax, line 59: `if extSlice, ok := out[k].([]any); ok` would be cleaner than `switch`

internal/configconverter/remove_memory_ballast_key.go
```diff
@@ -0,0 +1,74 @@
...
+	firstRegExp := regexp.MustCompile(firstExp)
+	const secondExp = "service::extensions"
+	secondRegExp := regexp.MustCompile(secondExp)
+
+	out := map[string]any{}
+	for _, k := range cfgMap.AllKeys() {
+		if firstRegExp.MatchString(k) {
+			log.Println("[WARNING]  `memory_ballast` parameter in extensions is deprecated. Please remove it from your configuration.")
+			continue
+		} else if secondRegExp.MatchString(k) {
```
dmitryax, line 56: No need for `else` if you have `continue` above
```suggestion
		} 
		if secondRegExp.MatchString(k) {
```

dmitryax:
@samiura are the failing integration tests related to the change?

CHANGELOG.md
samiura: sorry, should have removed it.

internal/configconverter/remove_memory_ballast_key.go
samiura: good catch! added a new test too.

internal/configconverter/remove_memory_ballast_key.go
samiura: liked the idea, much cleaner now!

internal/configconverter/remove_memory_ballast_key.go
samiura: absolutely!

samiura:
@dmitryax No, if you look at the history, they have been failing ever since start of the last week. I have manually deployed this branch on my machine, kind cluster and observed the memory/cpu usage, I do not see any issues thus far.

samiura:
@dmitryax  As we speak this is the state of integration test in the main branch.

https://github.com/signalfx/splunk-otel-collector/actions/runs/8425301171/job/23071388163

dmitryax (approved):

dmitryax:
LGTM. @samiura please proceed with the follow-up cleanup as we agreed

PR #4418, opened 2024-03-09
Remove docs related to internal/signalfxagent

atoulme:


crobert-1 (commented):
Why are we deleting the README for the agent? Would it make sense to keep until we can fully remove it?

atoulme:
I am not sure there is value to this README at all. Does it help anything, given that we have now modified the code underneath heavily? We might flatten that code a bit more moving forward to make it less nested.

samiura (approved):

PR #4422, opened 2024-03-09
Remove end to end tests: they currently do not run

atoulme:
**Description:**
Those tests check interactions with the backend and seem to test the validity of the collector data being sent ; those tests do not run as part of CI and require a token. Removing them at this time.


crobert-1:
Do we know why they weren't running? Is there a reason we're removing vs. fixing/running them?

atoulme:
They are meant to be run manually, not as part of CI, as best I can tell. I don't see a valid reason for us to run those tests as part of CI. We can bring them back if we want to, but only code that is executed should be in main.

PR #4468, opened 2024-03-14
do not run tests in parallel

atoulme:


crobert-1 (commented):
I assume the idea here is to reduce flakiness? Do we have a comparison on how long it takes to run in parallel vs. sequentially?

atoulme:
The idea is that with most tests now not needing to pull, build and run docker containers, they should be quick. This PR is trying to assess how long those tests run if sequential.

crobert-1 (approved):
_Looks good to me, once tests are passing_

samiura (approved):

jvoravong (approved):

PR #4478, opened 2024-03-15
Remove this folder, as it was moved under /docker/kafka

atoulme:
This was migrated to github action services, but the folder was left behind. Removing.


pjanotti (approved):
Removal always looks good :) I re-kicked the integration test failures, they were related to Kong metrics.

PR #4501, opened 2024-03-20
Migrate integration testing postgres Dockefile from TestUtils to github services

jvoravong:


atoulme:
you need to remove the container created inline at tests/receivers/postgresql/postgresql_test.go or it will clash on the port.

docker/postres/Dockerfile
atoulme, line 1: postres -> postgres

jvoravong: Yeah sorry, I had 3 of these image updates in one commit I'm splitting out.

PR #4528, opened 2024-03-25
Use golden files with haproxy test

atoulme:


tests/receivers/smartagent/haproxy/haproxy_test.go
```diff
@@ -17,13 +17,125 @@
...
+			"haproxy_connection_total",
+			"haproxy_request_total",
+			"haproxy_response_2xx",
+			"haproxy_session_current",
+			"haproxy_session_total",
+		),
 	)
 }
+
+func checkGoldenFile(t *testing.T, configFile string, expectedFilePath string, options ...pmetrictest.CompareMetricsOption) {
```
crobert-1, line 101: Would it make sense to move this function to some central location that can be reused? It looks like it's being defined for every component, but maybe there are differences I'm missing.

tests/receivers/smartagent/haproxy/haproxy_test.go
```diff
@@ -17,13 +17,125 @@
...
+		WithEnv(map[string]string{"OTLP_ENDPOINT": fmt.Sprintf("%s:%d", dockerHost, port)}).
+		Build()
+	require.NoError(t, err)
+	require.NoError(t, p.Start())
+	t.Cleanup(func() {
+		require.NoError(t, p.Shutdown())
+	})
+
+	expected, err := golden.ReadMetrics(filepath.Join("testdata", expectedFilePath))
+	//require.NoError(t, err)
```
crobert-1, line 131:
```suggestion
	require.NoError(t, err)
```

tests/receivers/smartagent/haproxy/haproxy_test.go
atoulme: yes, of course. I wanted my PRs to be independent, but we can refactor.

crobert-1 (approved):

PR #4541, opened 2024-03-26
Lint fixes

atoulme:
In preparation of a golangci-lint upgrade, this is a set of lint fixes that just remove unused function arguments.


crobert-1 (approved):
Awesome, thanks for fixing!

PR #4549, opened 2024-03-28
Remove cve-scan job, commented out in the build

atoulme:
**Description:**
This code was commented out as we moved Snyk outside of gitlab and are using it with GitHub now. This PR removes the code entirely.


crobert-1 (approved):
Thanks for cleaning this up!

PR #4550, opened 2024-03-28
Add a job to update with nightly dependencies

atoulme:
**Description:**
Add a nightly update build to update the otel dependencies to the latest on the main branch.

### DIFF
--- a/.gitlab/update-otel-deps.sh
+++ b/.gitlab/update-otel-deps.sh
@@ -5,6 +5,8 @@ set -euo pipefail
 # Based on https://github.com/signalfx/splunk-otel-java/blob/c9134906c84e9a32a974dec4b380453fe1757410/scripts/propagate-version.sh
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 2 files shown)


.gitlab/update-otel-deps.sh
```diff
@@ -15,18 +17,17 @@ cd "${ROOT_DIR}"
 create_collector_pr() {
   local repo="signalfx/splunk-otel-collector"
   local repo_url="https://srv-gh-o11y-gdi:${GITHUB_TOKEN}@github.com/${repo}.git"
-  local update_deps_branch="create-pull-request/update-deps"
-  local message="Update OpenTelemetry Dependencies to latest"
+  local update_deps_branch="$BRANCH"
```
crobert-1, line 20: Can we rename existing references to `update_deps_branch` to simply use `BRANCH` instead? Unless there's a specific reason that I'm missing, it would make more sense to me to use one variable for everything instead of copying the value here and using a different one.

atoulme: sure, done

crobert-1 (approved):

jvoravong:
+1

PR #4557, opened 2024-03-29
Move Redis test to use a golden file.

atoulme:


jvoravong (approved):

atoulme:
this is weird. The receiver's scraping is a bit odd. I'll try one more thing with batching for a longer period of time, but if that doesn't work, I will close this PR.

PR #4558, opened 2024-03-29
Relax assertion to avoid flaky tests

atoulme:


tests/general/discoverymode/docker_observer_discovery_test.go
```diff
@@ -314,7 +314,7 @@ service:
 `, stdout)
 	require.Contains(
 		t, stderr,
-		fmt.Sprintf(`Successfully discovered "prometheus_simple" using "docker_observer" endpoint "%s:32781".`, prometheus.GetContainerID()),
+		fmt.Sprintf(`Successfully discovered "prometheus_simple" using "docker_observer" endpoint "%s:".`, prometheus.GetContainerID()),
```
crobert-1, line 317:
```suggestion
		fmt.Sprintf(`Successfully discovered "prometheus_simple" using "docker_observer" endpoint "%s:`, prometheus.GetContainerID()),
```

I might be missing something, but wouldn't including the `".` always fail when a port is included in the `stderr` output?

PR #4582, opened 2024-04-01
remove small remnants of memory_ballast from Puppet. OTL-2703

samiura:
**Description:**  Removed memory ballast from puppet deployment.


deployments/puppet/manifests/collector_win_registry.pp
```diff
@@ -16,7 +16,10 @@
     "SPLUNK_REALM=${splunk_otel_collector::splunk_realm}",
     "SPLUNK_TRACE_URL=${splunk_otel_collector::splunk_trace_url}",
   ]
-  + if !$splunk_otel_collector::splunk_ballast_size_mib.strip().empty() {
+  + if ($splunk_otel_collector::params::collector_version != 'latest' or
+  versioncmp($splunk_otel_collector::params::collector_version, '0.97.0') < 0 ) and
+  !$splunk_otel_collector::splunk_ballast_size_mib.strip().empty()
```
dmitryax, line 21: Do we need to worry about running a new puppet release with an old collector? Maybe just remove it and document that new puppet release requires 0.97.0+?

dmitryax: Aren't we use latest for Windows as well? @jeffreyc-splunk WDYT?

samiura: @dmitryax  FYI-> https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/puppet/manifests/params.pp#L30

samiura: We can make it "latest" but not sure why the default is an empty string.

deployments/puppet/templates/splunk-otel-collector.conf.erb
```diff
@@ -13,6 +13,7 @@ SPLUNK_LISTEN_INTERFACE=<%= @splunk_listen_interface %>
 <% end -%>
 SPLUNK_REALM=<%= @splunk_realm %>
 SPLUNK_TRACE_URL=<%= @splunk_trace_url %>
+GOMEMLIMIT=<%= @gomemlimit %>
```
jeffreyc-splunk, line 16: The `gomemlimit` parameter should be added to init.pp and documented in the readme.

deployments/puppet/manifests/collector_win_registry.pp
jeffreyc-splunk: For windows, the `collector_version` parameter is required in the user's manifest and does not currently support `latest` (see https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/puppet/README.md#collector).

deployments/puppet/manifests/init.pp
```diff
@@ -10,7 +10,6 @@
   $splunk_bundle_dir       = $splunk_otel_collector::params::splunk_bundle_dir,
   $splunk_collectd_dir     = $splunk_otel_collector::params::splunk_collectd_dir,
   $splunk_memory_total_mib = '512',
-  $splunk_ballast_size_mib = '',
```
jeffreyc-splunk, line 13: Instead of removing the `splunk_ballast_size_mib` parameter, maybe we should just add a new parameter for configuring `GOMEMLIMIT`, and add some logic to determine which to configure based on the `collector_version` parameter?  We can then just document that `splunk_ballast_size_mib` is only applicable for versions < 0.97.0.

deployments/puppet/manifests/collector_win_registry.pp
jeffreyc-splunk: `GOMEMLIMIT` will need to be added to the windows registry if it is non-empty and `collector_version` >= 0.97.0.

deployments/puppet/manifests/init.pp
samiura: GOMEMLimit is acceptable for all version. In the code, we have 90% of the total memory limit assigned if GOMEMLIMIT is not set. We can do the same.

deployments/puppet/manifests/collector_win_registry.pp
dmitryax: Ok, sounds good. We should keep it then. @jeffreyc-splunk is there a way to have a default collector version for Windows going forward?

deployments/puppet/templates/splunk-otel-collector.conf.erb
samiura: okay

deployments/puppet/manifests/collector_win_registry.pp
jeffreyc-splunk: > Ok, sounds good. We should keep it then. @jeffreyc-splunk is there a way to have a default collector version for Windows going forward?

We could simply set the default version to something specific, e.g. `0.97.0`, but it may require us to bump this version and release the puppet module for each collector release.  And setting the default version for windows to `latest` would be more complex since there isn't a repo manager like yum or apt for MSI's that can handle determining the latest release.

Also, to help maintain idempotence and prevent inadvertent upgrades in production environments, it is recommended for the user to pin the desired version in their manifests.

deployments/puppet/manifests/init.pp
samiura: But, I have added the logic to have` GOMEMLIMIT` only for 0.97.0 or greater.

deployments/puppet/manifests/init.pp
```diff
@@ -32,6 +32,7 @@
   $fluentd_config_dest     = $splunk_otel_collector::params::fluentd_config_dest,
   $fluentd_capng_c_version = '<=0.2.2',  # linux only
   $fluentd_systemd_version = '<=1.0.2',  # linux only
+  $gomemlimit              = '460',
```
jeffreyc-splunk, line 35: 1. Should the default be `''` instead?
2. Please re-add this to the [template](https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/puppet/templates/splunk-otel-collector.conf.erb) for the linux service if the value is non-empty.
3. Please document this new parameter in the [readme](https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/puppet/README.md).

deployments/puppet/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+## puppet-v0.14.0
```
jeffreyc-splunk, line 5: Please bump the version in [metadata.json](https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/puppet/metadata.json#L3) so that it'll automatically trigger the release when this is merged.

deployments/puppet/manifests/collector_win_registry.pp
dmitryax: I think setting the default value to the latest release is fine if we have a release automation that would take care of this. Will the default values be applied to both Windows and Linux?

deployments/puppet/manifests/init.pp
dmitryax: We should not hardcode GOMEMLIMIT by default.

deployments/puppet/manifests/collector_win_registry.pp
samiura: per discussion offline, we have decided to have this change in a different scope. It is beyond this PR's scope for now.

deployments/puppet/README.md
```diff
@@ -73,6 +73,7 @@ This class accepts the following parameters:
...
+| `gomemlimit`         | GOMEMLIMIT env var is introduced for the Splunk Otel Collector version >=0.97.0, allowing the limitation of memory usage in the Go runtime. This feature can help enhance GC (Garbage Collection) related performance and prevent GC related Out of Memory (OOM) situations. **Note:**  By default 90% of the total memory is recommended. To increase frequency of garbage collections, set GOMEMLIMIT to a higher value than the default 90% of total memory and to reduce garbage collections, set GOMEMLIMIT to a lower value than the default 90% of the total memory. | `true`                                                                                                                        |
```
jeffreyc-splunk, line 76:
```suggestion
| `gomemlimit`         | GOMEMLIMIT env var is introduced for the Splunk Otel Collector version >=0.97.0, allowing the limitation of memory usage in the Go runtime. This feature can help enhance GC (Garbage Collection) related performance and prevent GC related Out of Memory (OOM) situations. **Note:**  By default 90% of the total memory is recommended. To increase frequency of garbage collections, set GOMEMLIMIT to a higher value than the default 90% of total memory and to reduce garbage collections, set GOMEMLIMIT to a lower value than the default 90% of the total memory. | None                                                                                                                        |
```

internal/buildscripts/packaging/tests/deployments/puppet/puppet_test.py
```diff
@@ -165,6 +165,7 @@ class {{ splunk_otel_collector:
     + [pytest.param(distro, marks=pytest.mark.rpm) for distro in RPM_DISTROS],
 )
 @pytest.mark.parametrize("puppet_release", PUPPET_RELEASE)
+@pytest.mark.skipif(sys.platform != "amazonlinux-2023", reason="only runs on windows")
```
jeffreyc-splunk, line 168:
```suggestion
```

deployments/puppet/manifests/collector_win_registry.pp
```diff
@@ -16,11 +16,21 @@
...
+  {
     ["SPLUNK_BALLAST_SIZE_MIB=${splunk_otel_collector::splunk_ballast_size_mib}"]
   } else {
     []
   }
+  + if ($splunk_otel_collector::params::collector_version == 'latest' or
+  versioncmp($splunk_otel_collector::params::collector_version, '0.97.0') >= 0 )
```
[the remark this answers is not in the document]
jeffreyc-splunk: We should only set `GOMEMLIMIT` if the `gomemlimit` parameter is not empty.

deployments/puppet/manifests/init.pp
```diff
@@ -10,7 +10,7 @@
   $splunk_bundle_dir       = $splunk_otel_collector::params::splunk_bundle_dir,
   $splunk_collectd_dir     = $splunk_otel_collector::params::splunk_collectd_dir,
   $splunk_memory_total_mib = '512',
-  $splunk_ballast_size_mib = '',
+  $splunk_ballast_size_mib  = '',
```
jeffreyc-splunk, line 13: nit
```suggestion
  $splunk_ballast_size_mib = '',
```

dmitryax:
Can we remove splunk_ballast_size_mib option?

dmitryax:
> @dmitryax I believe Jeff brought this that if we remove and this new Puppet module is used for version older than 0.97.0 will not work.

Why? The ballast env var is not required even for old versions. They will use the default value, which is okay, I guess.

dmitryax:
At least this should be removed from the documentation

samiura:
Why? The ballast env var is not required even for old versions. They will use the default value, which is okay, I guess.

> > @dmitryax I believe Jeff brought this that if we remove and this new Puppet module is used for version older than 0.97.0 will not work.
> 
> Why? The ballast env var is not required even for old versions. They will use the default value, which is okay, I guess.

I think they are present in the flags of installer scripts which @jeffreyc-splunk  had shown me and suggested not to remove totally rather mark it deprecated. 

The installer scripts removal is a "breaking change" per Jeff.

deployments/puppet/README.md
```diff
@@ -66,13 +66,14 @@ This class accepts the following parameters:
...
+|**DEPRECATED** `splunk_ballast_size_mib`            | Set the ballast size for the collector explicitly instead of the value calculated from the `$splunk_memory_total_mib` parameter.  This should be set to 1/3 to 1/2 of configured memory.  The `SPLUNK_BALLAST_SIZE_MIB` environment variable will be set with this value for the collector service.                                                                                                                                                                                                                                                                                                                                          | None                                                                                                                          |
```
jeffreyc-splunk, line 69:
```suggestion
|**DEPRECATED** `splunk_ballast_size_mib`            | Set the ballast size for the collector explicitly instead of the value calculated from the `$splunk_memory_total_mib` parameter.  This should be set to 1/3 to 1/2 of configured memory.  The `SPLUNK_BALLAST_SIZE_MIB` environment variable will be set with this value for the collector service. Only applicable if `collector_version` is < 0.97.0.                                                                                                                                                                                                                                                                                                                                         | None                                                                                                                          |
```

deployments/puppet/manifests/collector_win_registry.pp
```diff
@@ -16,11 +16,22 @@
...
-  + if !$splunk_otel_collector::splunk_ballast_size_mib.strip().empty() {
+  + if ($splunk_otel_collector::params::collector_version != 'latest' and
+  versioncmp($splunk_otel_collector::params::collector_version, '0.97.0') < 0 ) and
+  !$splunk_otel_collector::splunk_ballast_size_mib.strip().empty()
+  {
     ["SPLUNK_BALLAST_SIZE_MIB=${splunk_otel_collector::splunk_ballast_size_mib}"]
   } else {
     []
   }
+  + if ($splunk_otel_collector::params::collector_version == 'latest' and
```
jeffreyc-splunk, line 27:
```suggestion
  + if ($splunk_otel_collector::params::collector_version == 'latest' or
```

jeffreyc-splunk (commented):
@samiura Let me try to update the tests for some simple checks to verify that `GOMEMLIMIT` is configured correctly.

dmitryax:
I'm unsure if we need to expose GOMEMLIMIT env var with another config option. It seems unnecessary, given the internal discussion we had today about how to move forward. However, I'm not going to block this PR, it's up to you, folks

jeffreyc-splunk:
> I'm unsure if we need to expose GOMEMLIMIT env var with another config option. It seems unnecessary, given the internal discussion we had today about how to move forward. However, I'm not going to block this PR, it's up to you, folks

Yeah, it should be possible for the user to just include `GOMEMLIMIT => 'somevalue'` in the existing `collector_additional_env_vars` parameter within their manifest, instead of adding the new `gomemlimit` parameter.  WDYT @samiura?

deployments/puppet/manifests/collector_win_registry.pp
```diff
@@ -16,11 +16,22 @@
     "SPLUNK_REALM=${splunk_otel_collector::splunk_realm}",
     "SPLUNK_TRACE_URL=${splunk_otel_collector::splunk_trace_url}",
   ]
-  + if !$splunk_otel_collector::splunk_ballast_size_mib.strip().empty() {
+  + if ($splunk_otel_collector::params::collector_version != 'latest' and
+  versioncmp($splunk_otel_collector::params::collector_version, '0.97.0') < 0 ) and
```
jeffreyc-splunk, line 20:
```suggestion
  + if ($splunk_otel_collector::collector_version != 'latest' and
  versioncmp($splunk_otel_collector::collector_version, '0.97.0') < 0 ) and
```

deployments/puppet/manifests/collector_win_registry.pp
```diff
@@ -16,11 +16,22 @@
...
+  !$splunk_otel_collector::splunk_ballast_size_mib.strip().empty()
+  {
     ["SPLUNK_BALLAST_SIZE_MIB=${splunk_otel_collector::splunk_ballast_size_mib}"]
   } else {
     []
   }
+  + if ($splunk_otel_collector::params::collector_version == 'latest' and
+  versioncmp($splunk_otel_collector::params::collector_version, '0.97.0') >= 0 ) and
```
jeffreyc-splunk, line 28:
```suggestion
  + if ($splunk_otel_collector::collector_version == 'latest' and
  versioncmp($splunk_otel_collector::collector_version, '0.97.0') >= 0 ) and
```

samiura:
I think we should remove it given the drawbacks.

deployments/puppet/manifests/collector_win_registry.pp
```diff
@@ -16,11 +16,22 @@
...
-  + if !$splunk_otel_collector::splunk_ballast_size_mib.strip().empty() {
+  + if ($splunk_otel_collector::collector_version != 'latest' and
+  versioncmp($splunk_otel_collector::collector_version, '0.97.0') < 0 ) and
+  !$splunk_otel_collector::splunk_ballast_size_mib.strip().empty()
+  {
     ["SPLUNK_BALLAST_SIZE_MIB=${splunk_otel_collector::splunk_ballast_size_mib}"]
   } else {
     []
   }
+  + if ($splunk_otel_collector::collector_version == 'latest' and
```
jeffreyc-splunk, line 27:
```suggestion
  + if ($splunk_otel_collector::collector_version == 'latest' or
```

jeffreyc-splunk (approved):
LGTM. Thanks @samiura!

PR #4583, opened 2024-04-01
[receiver/discovery] Remove severity_text from log evaluation statement config

dmitryax:
Remove severity_text field from statements config and emitted logs.

This is needed for moving from discovery logs to entity events. The value from the severity_text field is just added as severity text to emitted log message which is not being used by the discovery config provider. Even if the evaluated log record has it's own log level, it will be overridden by the severity_text value set in the statement. And all the discovery statements have "info" values set in the severity_text field.

Later, we might want to add a `level` field in the statement that would be used to filter logs by a specific level, but it's not how the severity_text used right now.


dmitryax:
linkChecked fix https://github.com/signalfx/splunk-otel-collector/pull/4584

samiura (approved):

PR #4586, opened 2024-04-02
[discovery] Emit only one log record per matched endpoint

dmitryax:
This is another step towards migrating from the discovery events to entity events. Entity batch has to have only one entity with a unique identifier.

There is no reason to send a log record for every match. The matching rules should not conflict with each other. We can document and assume that.

The receiving part (discovery config provider) don't need all the matching states. It only maintains the latest state. And the successful state always wins. With this change, in case if we had conflicting matching rules across failed, partial and success sections, the result would be non-deterministic. However, given that we don't have any conflicting matching rules in the default discovery receiver configuration, it's safe to proceed.

As the next step, we will need to restructure the discovery rules schema to avoid any possible conflicts across different statuses.


internal/confmapprovider/discovery/discoverer.go
```diff
@@ -733,32 +733,26 @@ func (d *discoverer) ConsumeLogs(_ context.Context, ld plog.Logs) error {
...
-						d.discoveredReceivers[receiverID] = receiverStatus
-						d.discoveredObservers[observerID] = determineCurrentStatus(currentObserverStatus, rStatus)
-					}
+		// We assume that every resource log has a single log record as per the current implementation of the discovery receiver.
+		lr := rlog.ScopeLogs().At(0).LogRecords().At(0)
+		if currentReceiverStatus != discovery.Successful || currentObserverStatus != discovery.Successful {
+			if rStatusAttr, ok := lr.Attributes().Get(discovery.StatusAttr); ok {
+				rStatus := discovery.StatusType(rStatusAttr.Str())
+				if valid, e := discovery.IsValidStatus(rStatus); !valid {
+					d.logger.Debug("invalid status from log record", zap.Error(e), zap.Any("lr", lr.Body().AsRaw()))
```
samiura, line 742: This is much more readable 👍🏽

atoulme (approved):

PR #4587, opened 2024-04-02
Update default config to have only one otlp exporter

dmitryax:
Remove an exporter that is not being used. Having one point for switching to gateway seems more convenient.


cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -131,6 +131,10 @@ exporters:
     endpoint: "${SPLUNK_TRACE_URL}"
     headers:
       "X-SF-Token": "${SPLUNK_ACCESS_TOKEN}"
+    # Use instead when sending to gateway
```
pjanotti, line 134:
```suggestion
    # When sending to gateway use
```

pjanotti (approved):
Minor wording suggestion

PR #4588, opened 2024-04-02
[receiver/discovery] Combine matching conditions with different statuses

dmitryax:
Define the status of match conditions as a field instead of a map key. This way, we make the status evaluation deterministic. The first matching rule defines the status of the discovery endpoint. Without this reorganization, it's unclear which status wins if matching rules with different statuses overlap.

For example the following configuration:
```
    receivers:
      prometheus_simple:
        rule: type == "hostport" and command contains "otelcol"
        resource_attributes:
          one.key: one.value
          two.key: two.value
        status:
          metrics:
            successful:
            - regexp: ^otelcol_process_uptime$
              first_only: true
              log_record:
                body: Successfully connected to prometheus server
```
must be changed to
```
    receivers:
      prometheus_simple:
        rule: type == "hostport" and command contains "otelcol"
        resource_attributes:
          one.key: one.value
          two.key: two.value
        status:
          metrics:
            - status: successful
              regexp: ^otelcol_process_uptime$
              first_only: true
              log_record:
                body: Successfully connected to prometheus server
```

### DIFF
--- a/cmd/otelcol/config/collector/config.d.linux/receivers/mysql.discovery.yaml
+++ b/cmd/otelcol/config/collector/config.d.linux/receivers/mysql.discovery.yaml
@@ -18,26 +18,26 @@
 #       password: splunk.discovery.default
 #   status:
 #     metrics:
-#       successful:
-#         - strict: mysql.locks
-#           first_only: true
-#           log_record:
-#             body: Mysql receiver is working!
+#       - status: successful
+#         strict: mysql.locks
+#         first_only: true
+#         log_record:
+#           body: Mysql receiver is working!
 #     statements:
-#       failed:
-#         - regexp: "Can't connect to MySQL server on .* [(]111[)]"
-#           first_only: true
-#           log_record:
-#             append_pattern: true
-#             body:  The container cannot be reached by the Collector. The container is refusing MySQL connections.
-#       partial:
-#         - regexp: 'Access denied for user'
-#           first_only: true
-#           log_record:
-#             append_pattern: true
-#             body: >-
-#               Make sure your user credentials are correctly specified using the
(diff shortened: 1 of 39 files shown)


cmd/otelcol/config/collector/config.d.linux/receivers/mysql.discovery.yaml
```diff
@@ -18,26 +18,26 @@
 #       password: splunk.discovery.default
 #   status:
```
dmitryax, line 19: This can be changed to something like `status_detection` as the next step to avoid different `status` fields

samiura: Is this change absolutely necessary for entity events?

dmitryax: It's unnecessary, but having two status fields is confusing. The higher level `status` doesn't seem like a good name for what it's defining. Do you have any concerns with renaming it?

dmitryax: Giving that we are significantly changing this interface, I don't see why we need to restrain any further changes

samiura (approved):

cmd/otelcol/config/collector/config.d.linux/receivers/mysql.discovery.yaml
samiura: Understood!

samiura: I think I would go with your change as it makes things more deterministic.

jinja2 (approved):

PR #4622, opened 2024-04-05
Add dependabot check for MSI custom action

pjanotti:
Add dependabot for MSI custom action projects. The package system in this case is NuGet.

### DIFF
--- a/internal/buildscripts/packaging/msi/SplunkCustomActions/src/SplunkCustomActions.csproj
+++ b/internal/buildscripts/packaging/msi/SplunkCustomActions/src/SplunkCustomActions.csproj
@@ -7,7 +7,8 @@
     <RootNamespace>SplunkCustomActions</RootNamespace>
     <TargetFramework>net462</TargetFramework>
     <LangVersion>Latest</LangVersion>
-    <WindowsInstallerAssembliesDll Condition="'$(OS)' != 'Windows_NT'">$(HOME)/.nuget/packages/wix/3.14.1/tools/sdk</WindowsInstallerAssembliesDll>
+    <WiXPackageVersion>3.14.1</WiXPackageVersion>
+    <WindowsInstallerAssembliesDll Condition="'$(OS)' != 'Windows_NT'">$(HOME)/.nuget/packages/wix/$(WiXPackageVersion)/tools/sdk</WindowsInstallerAssembliesDll>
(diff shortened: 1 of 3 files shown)


samiura (approved):

internal/buildscripts/packaging/msi/SplunkCustomActions/src/SplunkCustomActions.csproj
```diff
@@ -7,7 +7,8 @@
     <RootNamespace>SplunkCustomActions</RootNamespace>
     <TargetFramework>net462</TargetFramework>
     <LangVersion>Latest</LangVersion>
-    <WindowsInstallerAssembliesDll Condition="'$(OS)' != 'Windows_NT'">$(HOME)/.nuget/packages/wix/3.14.1/tools/sdk</WindowsInstallerAssembliesDll>
+    <WiXPackageVersion>3.14.1</WiXPackageVersion>
+    <WindowsInstallerAssembliesDll Condition="'$(OS)' != 'Windows_NT'">$(HOME)/.nuget/packages/wix/$(WiXPackageVersion)/tools/sdk</WindowsInstallerAssembliesDll>
     <WindowsInstallerAssembliesDll Condition="'$(OS)' == 'Windows_NT'">$(MSBuildProgramFiles32)\Wix Toolset v3.14\SDK</WindowsInstallerAssembliesDll>
```
jinja2, line 12: could the path referenced here also use $(WiXPackageVersion)? or we need to update it when the version is bumped?

jeffreyc-splunk:
https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/scripts/gendependabot.sh will need to be updated for nuget.

internal/buildscripts/packaging/msi/SplunkCustomActions/src/SplunkCustomActions.csproj
pjanotti: No, the version here doesn't move. For Windows it is harder to use the nuget folder because dev box typically is the users AppData but GH runner has under a different location. I could make it conditional on some env var saying that is a GH runner 🤔 not sure if it is worth the trouble.

pjanotti: or I can do something simpler...

jinja2 (approved):

internal/buildscripts/packaging/msi/SplunkCustomActions/src/SplunkCustomActions.csproj
pjanotti: skipping this for the time being...

PR #4638, opened 2024-04-08
[discovery] Apply entity events schema the discovery receiver logs

dmitryax:
This change updates the schema of the logs emitted by the discovery receiver to represent entity events. No information is dropped, but it was reshuffled to adopt for the new schema:
- Entity ID is made of one key/value pair moved from `discovery.endpoint.id` resource attribute
- All other resource attributes are moved to the entity state event attributes
- Log body was moved to the `discovery.event.message` entity state attribute

The Discovery config provider is updated to use the entity events.

The keys used in the entity ID and entity attributes are not finalized and are expected to be changed in the following PRs.

### DIFF
--- a/tests/receivers/discovery/testdata/host_observer_simple_prometheus_config.yaml
+++ b/tests/receivers/discovery/testdata/host_observer_simple_prometheus_config.yaml
@@ -1,5 +1,7 @@
 extensions:
   host_observer:
+  host_observer/with_name:
+  host_observer/with/another/name:
 
 receivers:
   # set up otlp receiver to use its endpoints as assertion material
@@ -38,6 +40,8 @@ receivers:
 
     watch_observers:
       - host_observer
+      - host_observer/with_name
+      - host_observer/with/another/name
 
 # drop scrape_timestamp attributes until we can accept arbitrary values
 processors:
@@ -60,6 +64,8 @@ service:
       level: debug
   extensions:
     - host_observer
+    - host_observer/with_name
+    - host_observer/with/another/name
   pipelines:
     metrics:
       receivers: [otlp]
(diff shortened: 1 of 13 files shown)


samiura:
TestDiscoveryReceiverWithHostObserverAndSimplePrometheusReceiverProvideStatusLogs -> failing.

samiura:
> TestDiscoveryReceiverWithHostObserverAndSimplePrometheusReceiverProvideStatusLogs -> failing.

Rebasing....

tests/receivers/discovery/testdata/host_observer_endpoints_config.yaml
```diff
@@ -1,15 +1,11 @@
 extensions:
   host_observer:
-  host_observer/with_name:
-  host_observer/with/another/name:
```
dmitryax, line 4: Had to remove these observers because the testing framework doesn't work with several resource metrics having empty resource attributes. We could've fixed the framework, but I'm not sure these extra observers provide any value to the integration test. Also, the framework will be replaced with the `pdatatest` from contrib

tests/receivers/discovery/testdata/host_observer_simple_prometheus_config.yaml
```diff
@@ -1,7 +1,5 @@
 extensions:
   host_observer:
-  host_observer/with_name:
-  host_observer/with/another/name:
```
dmitryax, line 4: Had to remove these observers because the testing framework doesn't work with several resource metrics having empty resource attributes. We could've fixed the framework, but I'm not sure these extra observers provide any value to the integration test. Also, the framework will be replaced with the pdatatest from contrib

samiura (approved):

PR #4641, opened 2024-04-08
Build images for integration tests

jeffreyc-splunk:
Since the remote `quay.io/splunko11ytest` images may not be up-to-date with the changes in their respective `docker/<service>` directory, always build the images in CI using the remote images as cache (currently takes ~1m30s to build all images).

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -110,22 +110,17 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Changed service files
-        id: changed-services
-        uses: tj-actions/changed-files@v44
-        with:
-          files: docker/*/**
-      - name: Build changed services
-        if: steps.changed-services.outputs.any_changed == 'true'
-        env:
-          CHANGED_SERVICES: ${{ steps.changed-services.outputs.all_changed_files }}
+      - name: Build ${{ matrix.PROFILE }} service images
         run: |
-          for i in $CHANGED_SERVICES; do
-            service=$( echo $i | cut -d '/' -f2 )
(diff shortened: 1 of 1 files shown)


pjanotti (approved):

.github/workflows/integration-test.yml
```diff
@@ -110,22 +110,17 @@ jobs:
...
         run: |
-          for i in $CHANGED_SERVICES; do
-            service=$( echo $i | cut -d '/' -f2 )
-            if [[ -f docker/${service}/Dockerfile ]] && grep -q "image: quay.io/splunko11ytest/${service}:latest" docker/docker-compose.yml; then
-              docker build -t quay.io/splunko11ytest/${service}:latest docker/${service}
+          images=$(yq '.services[] | select(.profiles[] | contains("${{ matrix.PROFILE }}")) | .image' docker/docker-compose.yml | grep "quay.io/splunko11ytest/" | sort -u)
+          for image in $images; do
+            service=$(echo $image | sed 's|quay.io/splunko11ytest/\(.*\):latest|\1|')
+            if [[ -f docker/${service}/Dockerfile ]]; then
+              docker build --cache-from=quay.io/splunko11ytest/${service}:latest -t quay.io/splunko11ytest/${service}:latest docker/${service}
```
atoulme, line 119: Can you build for both linux arm and amd please?

jeffreyc-splunk: > Can you build for both linux arm and amd please?

Do the test services themselves need to run for arm64? Historically, these integration tests only required the collector to run for amd64/arm64, and the services ran only for amd64.  Also, the `quay.io/splunko11ytest` images are currently only built for amd64, so the arm64 images would get built from scratch every time without caches.

atoulme: just for laptops

atoulme: not a blocker, you're right. We can go with this.

PR #4653, opened 2024-04-11
[ktlo] Reduce CI test flakiness on Windows

pjanotti:
Ensure that the required port on the Windows dynamic range is available prior to running the collector.

Many of the failures related to starting the collector on CI are due to the fact that a port in the dynamic range is taken by some applications on Windows. This change ensures that the required port is available for the Windows tests that start the collector with the default configuration.


pjanotti:
As expected the runner bailing out occasionally when uninstalling the collector keeps happening.

jinja2 (approved):

PR #4655, opened 2024-04-11
add splunk_otlp_histograms resource attr

jinja2:

Adds the resource attribute `splunk_otlp_histograms: true` to the `service::telemetry::resource` if the collector detects any in use signalfx metrics exporter with the flag `send_otlp_histograms` enabled.

Added tests

### DIFF
--- a/internal/configconverter/otlp_histogram_attr.go
+++ b/internal/configconverter/otlp_histogram_attr.go
@@ -1,0 +1,82 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configconverter
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+
+	"go.opentelemetry.io/collector/confmap"
+)
+
+type AddOTLPHistogramAttr struct{}
+
+// Convert updates the service::telemetry::resource to add the attribute splunk.otlp.histograms=true.
+// This additional resource attr is only added if we see any signalfx exporter in use with the config
+// send_otlp_histograms set to true.
+func (AddOTLPHistogramAttr) Convert(_ context.Context, cfgMap *confmap.Conf) error {
+	if cfgMap == nil {
+		return nil
+	}
+
+	sfxExporters := map[string]bool{}
+	metricsPlRe := regexp.MustCompile(`^metrics($|/.*$)`)
+	signalfxExpRe := regexp.MustCompile(`^signalfx($|/.*$)`)
+
+	exp, err := cfgMap.Sub("exporters")
+	if err != nil {
+		return nil // Ignore invalid config. Rely on the config validation to catch this.
+	}
+
+	// get signalfx exporter names which have send_otlp_histograms enabled
+	for expName := range exp.ToStringMap() {
+		if signalfxExpRe.MatchString(expName) {
+			otlpHistoEnabled := cfgMap.Get(fmt.Sprintf("exporters::%s::send_otlp_histograms", expName))
+			if otlpHistoEnabled != nil && otlpHistoEnabled.(bool) {
+				sfxExporters[expName] = true
+			}
+		}
+	}
+
+	// no sfx exporters with send_otlp_histograms enabled
+	if len(sfxExporters) == 0 {
+		return nil
+	}
+
+	pl, err := cfgMap.Sub("service::pipelines")
+	if err != nil {
+		return nil
+	}
+
+	// check if metrics pipelines use any of the signalfx exporters which have send_otlp_histograms enabled
+	for pipelineName := range pl.ToStringMap() {
+		if metricsPlRe.MatchString(pipelineName) {
+			mExps := cfgMap.Get(fmt.Sprintf("service::pipelines::%s::exporters", pipelineName))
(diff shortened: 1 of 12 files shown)


internal/configconverter/otlp_histogram_attr.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Convert updates the service::telemetry::resource to add the attribute splunk.otlp.histograms=true.
+// This additional resource attr is only added if we see any signalfx exporter in use with the config
+// send_otlp_histograms set to true.
+func (AddOTLPHistogramAttr) Convert(_ context.Context, cfgMap *confmap.Conf) error {
+	if cfgMap == nil {
+		return nil
+	}
+
+	sfxExporters := map[string]bool{}
```
pjanotti, line 35: It won't make a real difference, but, since this is used only as a presence test I think that a `map[string]struct{}` will read better since it makes clear that nothing will be stored for each key.

internal/configconverter/otlp_histogram_attr.go
```diff
@@ -0,0 +1,82 @@
...
+	metricsPlRe := regexp.MustCompile(`^metrics($|/.*$)`)
+	signalfxExpRe := regexp.MustCompile(`^signalfx($|/.*$)`)
+
+	exp, err := cfgMap.Sub("exporters")
+	if err != nil {
+		return nil // Ignore invalid config. Rely on the config validation to catch this.
```
pjanotti, line 41: Q.: what would happen if you return the error here? I ask because a possible future re-organization of the code can be surprised by this not returning error.

internal/configconverter/otlp_histogram_attr.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	// get signalfx exporter names which have send_otlp_histograms enabled
+	for expName := range exp.ToStringMap() {
+		if signalfxExpRe.MatchString(expName) {
+			otlpHistoEnabled := cfgMap.Get(fmt.Sprintf("exporters::%s::send_otlp_histograms", expName))
+			if otlpHistoEnabled != nil && otlpHistoEnabled.(bool) {
```
pjanotti, line 48: Does the `confmap.Conf` object ensure that this cast always succeed?

internal/configconverter/otlp_histogram_attr_test.go
```diff
@@ -0,0 +1,70 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, cfgMap)
+	pmp := AddOTLPHistogramAttr{}
+	assert.NoError(t, pmp.Convert(context.Background(), cfgMap))
+	cfgMapExpected, err := confmaptest.LoadConf("testdata/otlp_histograms_attr/send_otlp_histrograms_disabled_expected.yaml")
+	require.NoError(t, err)
+	assert.Equal(t, cfgMapExpected.ToStringMap(), cfgMap.ToStringMap())
+}
+
+func TestOTLPHistograms_DisabledExporterNotInUse(t *testing.T) {
```
pjanotti, line 61: Except for the testdata files I think the tests have the same code. Why not create either a helper or use the `t.Run` pattern with multiple data files?

internal/configconverter/otlp_histogram_attr.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil
+	}
+
+	// check if metrics pipelines use any of the signalfx exporters which have send_otlp_histograms enabled
+	for pipelineName := range pl.ToStringMap() {
+		if metricsPlRe.MatchString(pipelineName) {
+			mExps := cfgMap.Get(fmt.Sprintf("service::pipelines::%s::exporters", pipelineName))
+			if metricsExps, ok := mExps.([]any); ok {
+				for _, expName := range metricsExps {
+					if _, ok := sfxExporters[expName.(string)]; ok {
```
pjanotti, line 70: Is this cast guaranteed to succeed given the `cfgMap.Get` returned data?

internal/configconverter/otlp_histogram_attr.go
jinja2: Any errors returned by the converter are treated as fatal. I was referring to the existing config converters for guidance on this and they are relying on the config parsing failing since exporters and pipelines are required. Returning the error here should be okay since decoding would also fail

internal/configconverter/otlp_histogram_attr.go
jinja2: it should or initial parsing will fail for the pipeline, but let me update this

internal/configconverter/otlp_histogram_attr.go
```diff
@@ -0,0 +1,86 @@
...
+		if metricsPlRe.MatchString(pipelineName) {
+			mExps := cfgMap.Get(fmt.Sprintf("service::pipelines::%s::exporters", pipelineName))
+			if metricsExps, ok := mExps.([]any); ok {
+				for _, expName := range metricsExps {
+					if expNameStr, ok := expName.(string); ok {
+						if _, ok := sfxExporters[expNameStr]; ok {
+							resAttrKey := "service::telemetry::resource::splunk_otlp_histograms"
+							if err = cfgMap.Merge(confmap.NewFromStringMap(map[string]any{resAttrKey: "true"})); err != nil {
+								return err
+							}
```
dmitryax, line 78: given that we don't need to iterate anymore
```suggestion
							return cfgMap.Merge(confmap.NewFromStringMap(map[string]any{resAttrKey: "true"}))
```

dmitryax (approved):

theletterf:
@jinja2 From an end-user perspective, enabling histograms in SFx exporter would still be the same? Is this transparent to the end users configuring the Collector for histogram export?

internal/configconverter/otlp_histogram_attr.go
pjanotti:
```suggestion
		return err // In practice this should never happen: config parsing should have caught the error prior to here.
```

pjanotti (approved):

jinja2:
@theletterf yes, this change is transparent to the user. We want to use this attribute to track usage of the feature on our end.

PR #4657, opened 2024-04-12
Remove redis test already present upstream

atoulme:
This test is already present with this golden file: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/redisreceiver/testdata/integration/expected-cluster.yaml


crobert-1 (approved):
Thanks for including the contrib test reference! 👍

PR #4660, opened 2024-04-14
[receiver/discovery] Track only endpoints matching a receiver rule

dmitryax:
This way, we can start tracking target endpoints in the correlation store and emit entity events on regular basis.

Also, this requires limiting discovered endpoint to only one receiver. If an overlapping rule is found in multiple receivers, that endpoint will be ignored with a warning. If we keep allowing multiple receivers per endpoint, it'll be unclear how to track successful discovery of endpoint/service because different receivers may provide conflicting results.

This change required pulling code from the receivercreator, which will be reused once the discovery receiver is upstreamed.

### DIFF
--- a/internal/receiver/discoveryreceiver/config_test.go
+++ b/internal/receiver/discoveryreceiver/config_test.go
@@ -40,7 +40,7 @@ func TestValidConfig(t *testing.T) {
 
 	assert.Equal(t, 1, len(configs.ToStringMap()))
 
-	cm, err := configs.Sub("discovery/discovery-name")
+	cm, err := configs.Sub("discovery")
 	require.NoError(t, err)
 	cfg := createDefaultConfig().(*Config)
 	err = component.UnmarshalConfig(cm, cfg)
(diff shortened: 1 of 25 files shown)


internal/receiver/discoveryreceiver/correlation.go
```diff
@@ -94,61 +87,43 @@ func newCorrelationStore(logger *zap.Logger, ttl time.Duration) correlationStore
 	}
 }
 
-// UpdateEndpoint will update all existing correlation timestamps and states by endpoint.ID, or
-// creates a new no-type ~singleton w/ the initial correlation info for later use in correlation creation.
-func (s *store) UpdateEndpoint(endpoint observer.Endpoint, state endpointState, observerID component.ID) {
-	defer s.endpointLocks.Lock(endpoint.ID)()
```
dmitryax, line 100: Fix the map-locking mechanism along the way. This line has to be broken down to be effective. Otherwise the whole `s.endpointLocks.Lock(endpoint.ID)()` line is deffered

internal/receiver/discoveryreceiver/config_test.go
```diff
@@ -40,7 +40,7 @@ func TestValidConfig(t *testing.T) {
 
 	assert.Equal(t, 1, len(configs.ToStringMap()))
 
-	cm, err := configs.Sub("discovery/discovery-name")
+	cm, err := configs.Sub("discovery")
```
samiura, line 43: much cleaner.. always wondered why we named it like that on the config that was overwritten!

samiura (approved):

PR #4679, opened 2024-04-16
Include splunk-otel-dotnet in auto-instrumentation deb/rpm packages

jeffreyc-splunk:
- Similar to `splunk-otel-java` and `splunk-otel-js`, include `splunk-otel-dotnet` in the auto-instrumentation deb/rpm packages (amd64/x86_64 only)
- Add default config files for systemd and libsplunk with the required env vars
- Update tests to verify manual installation and basic instrumentation with a sample .NET app
- Add gitlab job to update `splunk-otel-dotnet` to the latest release (similar to how we already update `splunk-otel-java` and `splunk-otel-js`)

TODO: Update installer script and deployment modules after these changes are released.


instrumentation/README.md
```diff
@@ -73,6 +76,17 @@ The following methods are supported to manually activate and configure Auto Inst
      ```
      NODE_OPTIONS=-r /usr/lib/splunk-instrumentation/splunk-otel-js/node_modules/@splunk/otel/instrument
      ```
+   - `/etc/splunk/zeroconfig/dotnet.conf`:
+     ```
+     CORECLR_ENABLE_PROFILING=1
+     CORECLR_PROFILER={918728DD-259F-4A6A-AC2B-B85E1B658318}
+     CORECLR_PROFILER_PATH=/usr/lib/splunk-instrumentation/splunk-otel-dotnet/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so
```
pjanotti, line 83: Just to be sure that you are aware: since this is for deb/rpm packages `linux-x64` is correct, when using arm64 the path will change and also for Alpine.

pjanotti (approved):
Q.: why is it necessary to add `libicu`?

jeffreyc-splunk:
> Q.: why is it necessary to add `libicu`?

`libicu` seems to be required by the sample .NET app used in these tests. Otherwise, it fails with:
```
Couldn't find a valid ICU package installed on the system. Please install libicu using your package manager and try again. Alternatively you can set the configuration flag System.Globalization.Invariant to true if you want to run with no globalization support. Please see https://aka.ms/dotnet-missing-libicu for more information.
```

So I just install `libicu` in the test images rather than mucking with the sample app.

instrumentation/README.md
jeffreyc-splunk: Yes, we can update the paths accordingly when `splunk-otel-dotnet` supports arm64 or when we need to support zero config for alpine.

PR #4683, opened 2024-04-17
[ktlo] Try to be less impactful on MSI uninstall

pjanotti:
Unfortunately the GH runner is still frequently bailing out during the MSI uninstall. This change makes the uninstall a bit less impactful, removing logs and ensuring that no restart is required, in an attempt to reduce the flakiness of this step.


.github/workflows/win-package-test.yml
```diff
@@ -305,10 +305,6 @@ jobs:
             throw "MSI uninstall failed"
           }
 
-      - name: "In case of failure: display the uninstall logs"
```
crobert-1, line 308: Will this result in less information being available for debugging failures? (If so, is that acceptable in this case?)

pjanotti: Yes, but, that will be in the case that the uninstall fails but the runner is alive :) right now the runner frequently dies in the middle of the uninstall and there are no logs anyway.

crobert-1 (approved):

PR #4686, opened 2024-04-18
[receiver/discovery] Emit active endpoints as entity events periodically

dmitryax:
Only applicable if log_endpoints: true

### DIFF
--- a/internal/receiver/discoveryreceiver/endpoint_tracker.go
+++ b/internal/receiver/discoveryreceiver/endpoint_tracker.go
@@ -42,12 +42,16 @@ var (
 )
 
 type endpointTracker struct {
+	correlations correlationStore
 	config       *Config
 	logger       *zap.Logger
 	pLogs        chan plog.Logs
 	observables  map[component.ID]observer.Observable
-	correlations correlationStore
+	stopCh       chan struct{}
 	notifies     []*notify
+	// emitInterval defines an interval for emitting entity state events.
+	// Potentially can be exposed as a user config option if there is a need.
+	emitInterval time.Duration
 }
 
 type notify struct {
@@ -66,6 +70,13 @@ func newEndpointTracker(
 		logger:       logger,
 		pLogs:        pLogs,
 		correlations: correlations,
+		// 15 minutes is a reasonable default for emitting entity state events given the 1 hour TTL in the inventory
+		// service. Potentially we could expose it as a user config, but only if there is a need.
+		// Note that we emit entity state events on every entity change. Entities that were changed in the last
+		// 15 minutes are not emitted again. So the actual interval of emitting entity state events can be more than 15
+		// minutes but always less than 30 minutes.
+		emitInterval: 15 * time.Minute,
+		stopCh:       make(chan struct{}),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


internal/receiver/discoveryreceiver/endpoint_tracker.go
```diff
@@ -66,6 +70,13 @@ func newEndpointTracker(
 		logger:       logger,
 		pLogs:        pLogs,
 		correlations: correlations,
+		// 15 minutes is a reasonable default for emitting entity state events given the 1 hour TTL in the inventory
+		// service. Potentially we could expose it as a user config, but only if there is a need.
+		// Note that we emit entity state events on every entity change. Entities that were changed in the last
+		// 15 minutes are not emitted again. So the actual interval of emitting entity state events can more than 15
+		// minutes but always less than 30 minutes.
+		emitInterval: 15 * time.Minute,
```
samiura, line 78: Thank you for the comments.

samiura (approved):

internal/receiver/discoveryreceiver/endpoint_tracker.go
```diff
@@ -42,12 +42,16 @@ var (
...
+	correlations correlationStore
 	config       *Config
 	logger       *zap.Logger
 	pLogs        chan plog.Logs
 	observables  map[component.ID]observer.Observable
-	correlations correlationStore
+	stopCh       chan struct{}
 	notifies     []*notify
+	// emitInterval defines an interval for emitting entity state events.
+	// Potentially can be exposed as a user config option is there is a need.
```
crobert-1, line 53:
```suggestion
	// Potentially can be exposed as a user config option if there is a need.
```

internal/receiver/discoveryreceiver/endpoint_tracker.go
```diff
@@ -66,6 +70,13 @@ func newEndpointTracker(
 		logger:       logger,
 		pLogs:        pLogs,
 		correlations: correlations,
+		// 15 minutes is a reasonable default for emitting entity state events given the 1 hour TTL in the inventory
+		// service. Potentially we could expose it as a user config, but only if there is a need.
+		// Note that we emit entity state events on every entity change. Entities that were changed in the last
+		// 15 minutes are not emitted again. So the actual interval of emitting entity state events can more than 15
```
crobert-1, line 76:
```suggestion
		// 15 minutes are not emitted again. So the actual interval of emitting entity state events can be more than 15
```

internal/receiver/discoveryreceiver/endpoint_tracker.go
```diff
@@ -82,6 +93,26 @@ func (et *endpointTracker) start() {
 		go observable.ListAndWatch(n)
 	}
 	et.correlations.Start()
+	go et.startEmitLoop()
+}
+
+func (et *endpointTracker) startEmitLoop() {
+	go func() {
```
crobert-1, line 100: I don't think this function should start its own goroutine, since a goroutine is already started when calling this method, right?

internal/receiver/discoveryreceiver/endpoint_tracker_test.go
```diff
@@ -16,6 +16,7 @@ package discoveryreceiver
 
 import (
 	"path"
+	"sync"
```
crobert-1, line 19: With the addition of more goroutine logic, we should add `goleak` to this package.

crobert-1 (commented):
I'm not very familiar with discovery mode, so hopefully I'm not too far off track here. _I didn't get a chance to review tests yet._

internal/receiver/discoveryreceiver/endpoint_tracker.go
dmitryax: good catch! 👍

internal/receiver/discoveryreceiver/endpoint_tracker_test.go
dmitryax: added

PR #4708, opened 2024-04-22
Improve service start stop on install ps1

pjanotti:
**Description:**
Better leverage the Powershell Service cmdlets to reduce code and to not stop the service if it didn't start in the given starting timeout (there is no point: if the service is taking long to start, stopping it will just ensure that it will start from zero again).


internal/buildscripts/packaging/choco/splunk-otel-collector/tools/chocolateyinstall.ps1
```diff
@@ -210,7 +206,7 @@ if (!$env_vars["SPLUNK_ACCESS_TOKEN"]) {
             # The fluentd service is automatically started after msi installation.
             # Wait for it to be running before trying to restart it with our custom config.
             Write-Host "Restarting $fluentd_service_name service..."
-            wait_for_service -name "$fluentd_service_name"
+            start_service -name "$fluentd_service_name"
```
jinja2, line 209: the comment above indicates that fluentd service will be started automatically. should the comment be updated or is this start_service safe to run in place of just waiting?

jinja2 (approved):

PR #4717, opened 2024-04-23
[receiver/discovery] Simplify config reporting

dmitryax:
This change removes the complicated logic of passing the encoded config by metrics receivers as part of a resource attribute and intercepting it in the discovery receiver. Instead, we compute the embedded config on the fly for every emitted entity. In the short run, this hits the performance of the discovery receiver, but once we move the `metric.match` and `statement.match` discovery events to entity states in the following PRs, it won't be an issue.

Except for the performance implications, there are no functional changes in this PR.

### DIFF
--- a/internal/receiver/discoveryreceiver/evaluator.go
+++ b/internal/receiver/discoveryreceiver/evaluator.go
@@ -116,79 +116,30 @@ func (e *evaluator) evaluateMatch(match Match, pattern string, status discovery.
 	return shouldLog, nil
 }
 
-// correlateResourceAttributes will copy all `from` attributes to `to` in addition to
-// updating embedded base64 config content, if configured, to include the correlated observer ID
-// that is otherwise unavailable to status sources.
-func (e *evaluator) correlateResourceAttributes(from, to pcommon.Map, corr correlation) {
-	receiverType := corr.receiverID.Type().String()
-	receiverName := corr.receiverID.Name()
-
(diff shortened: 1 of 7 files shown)


internal/receiver/discoveryreceiver/evaluator.go
```diff
@@ -116,79 +116,30 @@ func (e *evaluator) evaluateMatch(match Match, pattern string, status discovery.
...
-	if err = yaml.Unmarshal(dBytes, &cfg); err != nil {
-		return "", err
-	}
-
-	cfg["watch_observers"] = []string{observerID}
-
-	var cfgYaml []byte
-	if cfgYaml, err = yaml.Marshal(cfg); err != nil {
-		return "", fmt.Errorf("failed embedding receiver config to include %q: %w", observerID, err)
+		to.PutStr(discovery.ReceiverConfigAttr, base64.StdEncoding.EncodeToString(cfgYaml))
```
crobert-1, line 143: Why don't we have to update the correlation's attribute value here, like we used to when this logic was in `receiverCreatorReceiversConfig`?

dmitryax: Because it's not being used anymore

dmitryax: It'll eventually go to attributes once we stop emitting entities here. But I need to make some more changes before that

crobert-1 (approved):

PR #4722, opened 2024-04-24
added mongodb receiver into discovery mode  

samiura:
**Description:** Added mechanism to auto detect MongoDB instan es


cmd/otelcol/config/collector/config.d.linux/receivers/mongodb.discovery.yaml
```diff
@@ -0,0 +1,62 @@
...
+#           body: >-
+#             Make sure the user has been given access admin db with full root access.
+#             `db.grantRolesToUser('admin', [{ role: 'readWrite', db: 'admin' }])`
+#       - status: partial
+#         regexp: '.* error: invalid configuration: receivers::mongodb: invalid transport type .*'
+#         log_record:
+#           append_pattern: true
+#           body: >-
+#             Make sure to have a valid transport type inside the otel configuration. For more information check the documentation carefully.
+#             `transport: tcp`
```
dmitryax, line 62: This one is unclear on how a user should resolve it. We don't provide a config interface for particular components with the discovery mode

samiura: I think we can remove it. With the default configuration, we cannot hit this scenario. Only possible if someone is manual configuring but not in discovery mode.

internal/confmapprovider/discovery/bundle/bundle.d/receivers/mongodb.discovery.yaml
```diff
@@ -0,0 +1,51 @@
...
+              `--set splunk.discovery.receivers.mongodb.config.password="<password>"` or
+              `SPLUNK_DISCOVERY_RECEIVERS_mongodb_CONFIG_username="<username>"` and
+              `SPLUNK_DISCOVERY_RECEIVERS_mongodb_CONFIG_password="<password>"` environment variables.
+      - status: partial
+        regexp: '.* failed to fetch index stats metrics: (Unauthorized) not authorized on admin to execute command .*'
+        log_record:
+          append_pattern: true
+          body: >-
+            Make sure the user has been given access admin db with full root access.
+            `db.grantRolesToUser('admin', [{ role: 'readWrite', db: 'admin' }])`
```
dmitryax, line 51: _full root access_ sounds too much. `readWrite` doesn't sound like reading https://www.mongodb.com/docs/manual/reference/built-in-roles/#mongodb-authrole-readWrite. 

Also, why is `write` needed to fetch the index? Sounds like reading should be enough?

internal/confmapprovider/discovery/bundle/bundle.d/receivers/mongodb.discovery.yaml
```diff
@@ -0,0 +1,51 @@
...
+        regexp: 'connect: connection refused'
+        log_record:
+          append_pattern: true
+          body: The container is refusing mongodb connections.
+      - status: partial
+        regexp: '.* unable to authenticate using mechanism .*'
+        log_record:
+          append_pattern: true
+          body: >-
+              Please ensure your user credentials are correctly specified with
```
dmitryax, line 40: Should we specify which role we need credentials for, given that we mention `admin` in the other rule

internal/confmapprovider/discovery/bundle/bundle.d/receivers/mongodb.discovery.yaml
samiura: I think you're correct on readWrite. I have just given that permission and it works. Let me see if read is enough to do the job or not?

internal/confmapprovider/discovery/bundle/bundle.d/receivers/mongodb.discovery.yaml
samiura: Admin is not necessary to perform these action but the 'user' needs to have readWrite access to perform these actions.

internal/confmapprovider/discovery/bundle/bundle.d/receivers/mongodb.discovery.yaml
samiura: I have just verified, it works with `role: read`. I will shortly update the wording.


Successfully added user: {
        "user" : "testUser",
        "roles" : [
                {
                        "role" : "read",
                        "db" : "admin"
                }
        ]
}

samiura: I should read more carefully. Upstream has already mention the minimum requirements. It works well.
All we need is clustermonitor role.

https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/mongodbreceiver#prerequisites

samiura: Also, we have to tell BITS and prospective customers the upstream version supported are 4.0+ and 5.0. Both of which I have tested.

samiura: tested successfully with this user below.

db.createUser(
    {
      user: "testUser1",
      pwd: "testPass1",
      roles: [
         { role: "clusterMonitor", db: "admin" },
      ]
    }
)

dmitryax (approved):

PR #4723, opened 2024-04-24
[chore] Upgrade pymongo dep version to 4.6.3

crobert-1:


atoulme (approved):

internal/signalfx-agent/bundle/collectd-plugins.yaml
```diff
@@ -31,7 +31,7 @@
   version: v1.3.7
   repo: signalfx/collectd-mongodb
   pip_packages:
-    - pymongo[tls]==3.13.0
+    - pymongo[tls]==4.6.3
```
jinja2, line 34: This major version update might be breaking for the collectd-mongodb plugin. For e.g., [v4.x replaces the parameter names for ssl args](https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#renamed-uri-options) which we seem to be using in [our plugin](https://github.com/signalfx/collectd-mongodb/blob/master/mongodb.py#L88-L95).

atoulme: ah, I assumed we had an integration test covering this.

crobert-1: Good catch! I'll look into this 👍

crobert-1:
Superseded by https://github.com/signalfx/splunk-otel-collector/pull/4731

PR #4725, opened 2024-04-24
[receiver/discovery] Assign correct resource in the metric evaluation

dmitryax:
Metric evaluation emits entities with attributes from the first resource in the metric batch regardless of where the matching metrics were found. This change fixes the bug. The resource attributes are now assigned from the resource where the metrics are found.

It doesn't make sense to mention it in the changelog because we don't have any pre-configured receiver to run into this bug.

### DIFF
--- a/internal/receiver/discoveryreceiver/metric_evaluator.go
+++ b/internal/receiver/discoveryreceiver/metric_evaluator.go
@@ -18,7 +18,9 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/observer"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/experimentalmetricmetadata"
+	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/pdata/pcommon"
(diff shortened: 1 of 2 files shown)


samiura (approved):

internal/receiver/discoveryreceiver/metric_evaluator.go
```diff
@@ -100,66 +102,69 @@ func (m *metricEvaluator) evaluateMetrics(md pmetric.Metrics) plog.Logs {
...
 				} else if !shouldLog {
 					continue
 				}
-
-				entityEvents := experimentalmetricmetadata.NewEntityEventsSlice()
-				entityEvent := entityEvents.AppendEmpty()
-				entityEvent.ID().PutStr(discovery.EndpointIDAttr, string(endpointID))
-				entityState := entityEvent.SetEntityState()
-
-				md.ResourceMetrics().At(0).Resource().Attributes().CopyTo(entityState.Attributes())
```
dmitryax, line 130: The core of the bug fix is replacing this line with line 116

crobert-1: Should there be a changelog for this, or was this bug introduced since the last release?

atoulme (approved):

crobert-1 (approved):

PR #4728, opened 2024-04-25
[receiver/discovery] Emit evaluation entities by the endpoint tracker

dmitryax:
This change consolidates the entity emitting in one place: the endpoint tracker. Once the evaluation succeeds, the endpoint attributes are updated in the correlation store, and the new entity state is emitted immediately. Every subsequent state is emitted with all the added attributes.

Currently, we emit a superset of all the attributes that were previously emitted on a regular basis or during evaluation. Later, we will reduce the set and remove redundant attributes.

The tests are kept with minimal changes to ensure no attributes are lost in the emitted entities.

### DIFF
--- a/internal/receiver/discoveryreceiver/correlation.go
+++ b/internal/receiver/discoveryreceiver/correlation.go
@@ -50,6 +50,7 @@ type correlationStore interface {
 	GetOrCreate(endpointID observer.EndpointID, receiverID component.ID) correlation
 	Attrs(endpointID observer.EndpointID) map[string]string
 	UpdateAttrs(endpointID observer.EndpointID, attrs map[string]string)
+	EmitCh() chan correlation
 	Endpoints(updatedBefore time.Time) []observer.Endpoint
 	// Start the reaping loop to prevent unnecessary endpoint buildup
 	Start()
@@ -71,6 +72,7 @@ type store struct {
 	attrs         *sync.Map
 	// sentinel for terminating reaper loop
 	sentinel     chan struct{}
+	emitCh       chan correlation
 	reapInterval time.Duration
 	ttl          time.Duration
 }
@@ -84,6 +86,7 @@ func newCorrelationStore(logger *zap.Logger, ttl time.Duration) correlationStore
 		reapInterval:  30 * time.Second,
 		ttl:           ttl,
 		sentinel:      make(chan struct{}, 1),
+		emitCh:        make(chan correlation),
 	}
 }
 
@@ -116,6 +119,11 @@ func (s *store) MarkStale(endpointID observer.EndpointID) {
 		corr := c.(*correlation)
 		corr.stale = true
 	}
+}
+
+// EmitCh returns a channel to emit endpoints immediately.
+func (s *store) EmitCh() chan correlation {
+	return s.emitCh
 }
 
 // Endpoints returns all active endpoints that have not been updated since the provided time.
--- a/internal/receiver/discoveryreceiver/statement_evaluator_test.go
+++ b/internal/receiver/discoveryreceiver/statement_evaluator_test.go
@@ -25,7 +25,6 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/observer"
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/collector/component"
-	"go.opentelemetry.io/collector/pdata/plog"
 	"go.uber.org/zap"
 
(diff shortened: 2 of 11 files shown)


crobert-1:
Should this include a changelog entry?

dmitryax:
> Should this include a changelog entry?

I'm not sure about this. The only difference with this change is that we are sending more attributes (a superset across endpoint states and evaluations). We will need to significantly change the set of attributes later on, so I would postpone that changelog item to avoid polluting it.

crobert-1:
> I'm not sure about this. The only difference with this change is that we are sending more attributes (a superset across endpoint states and evaluations). We will need to significantly change the set of attributes later on, so I would postpone that changelog item to avoid polluting it. Let me know WDYT

If this affects billing by sending more attributes then we should log it, otherwise it's okay with me to skip.

internal/receiver/discoveryreceiver/correlation.go
```diff
@@ -50,6 +50,7 @@ type correlationStore interface {
 	GetOrCreate(endpointID observer.EndpointID, receiverID component.ID) correlation
 	Attrs(endpointID observer.EndpointID) map[string]string
 	UpdateAttrs(endpointID observer.EndpointID, attrs map[string]string)
+	EmitCh() chan correlation
```
crobert-1, line 53: Is it useful to have a public getter method for the channel here? Do we foresee a case where we can't just directly use the channel?

dmitryax:
All the entity events at this point are not expected to go outside of the collector. They are either consumed by the the discovery config provider or logged out with debug exporter

internal/receiver/discoveryreceiver/statement_evaluator_test.go
```diff
@@ -66,31 +64,26 @@ func TestStatementEvaluation(t *testing.T) {
...
 
-							plogs := make(chan plog.Logs)
-
 							// If debugging tests, replace the Nop Logger with a test instance to see
 							// all statements. Not in regular use to avoid spamming output.
 							// logger := zaptest.NewLogger(t)
 							logger := zap.NewNop()
 							cStore := newCorrelationStore(logger, time.Hour)
+
+							// Change the emit channel to a buffered channel to make the test synchronous
```
crobert-1, line 73: Wouldn't it be a slightly more realistic test to run the statement evaluator in its own goroutine, and then not override the channel with a buffered channel and just block until it receives?

internal/receiver/discoveryreceiver/statement_evaluator_test.go
```diff
@@ -105,8 +98,12 @@ func TestStatementEvaluation(t *testing.T) {
...
 							}
 
-							wg.Wait()
-							close(plogs)
+							corr := <-emitCh
+							entityEvents, numFailed, err := entityStateEvents(corr.observerID,
+								[]observer.Endpoint{corr.endpoint}, cStore, time.Now())
+							require.NoError(t, err)
+							require.Equal(t, 0, numFailed)
+							emitted := entityEvents.ConvertAndMoveToLogs()
```
crobert-1, line 106: Do we still want to call this in the test, even though we've removed the call in the statement evaluation itself?

internal/receiver/discoveryreceiver/correlation.go
dmitryax: I'm just following the pattern of how this interface is used. If I want to use it directly, I'd need to convert itto the underlying struct. I don't like having interfaces with one implementation used in the same package. I would like to remove the interface altogether, but it's out of the scope

internal/receiver/discoveryreceiver/statement_evaluator_test.go
dmitryax: Maybe, but why?

dmitryax: I can apply this if you want

internal/receiver/discoveryreceiver/statement_evaluator_test.go
dmitryax: As I mentioned in the description, I want to keep the test with minimal changes to showcase the diff. I want to simplify this test as follow up, there is no need to parse the entity event here anymore

internal/receiver/discoveryreceiver/statement_evaluator_test.go
dmitryax: Made it async

crobert-1 (approved):

PR #4730, opened 2024-04-25
snyk: fail only for upgradable vulns

jeffreyc-splunk:
Similar to our other scans, only fail snyk if the detected vulnerabilities are upgradable.


crobert-1:
I'd rather have it fail and let us know that something is going on, even if we can't fix it, so tha we can be aware ahead of time and potentially keep an eye on it ourselves. I'm interested to hear more opinions though.

atoulme:
Asking for guidance internally.

samiura (approved):

atoulme:
No guidance offered yet.

atoulme:
Guidance is to scan everything :)

PR #4738, opened 2024-04-26
Linux installer script: support .NET zero config

jeffreyc-splunk:
- Install and configure env vars for `splunk-otel-dotnet` (x86_64/amd64 only)
- Refactored handling of installer script options for auto instrumentation
- Removed support for the "original" `libsplunk.so` (requires v0.87.0 or greater)
- Updated tests to verify preload and systemd configs

### DIFF
--- a/internal/buildscripts/packaging/installer/install.sh
+++ b/internal/buildscripts/packaging/installer/install.sh
@@ -109,20 +109,23 @@ fi
 preload_path="/etc/ld.so.preload"
 default_instrumentation_version="latest"
 default_deployment_environment=""
-instrumentation_config_path="/usr/lib/splunk-instrumentation/instrumentation.conf"
 instrumentation_so_path="/usr/lib/splunk-instrumentation/libsplunk.so"
 instrumentation_jar_path="/usr/lib/splunk-instrumentation/splunk-otel-javaagent.jar"
-generate_service_name="true"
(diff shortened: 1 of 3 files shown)


internal/buildscripts/packaging/installer/install.sh
```diff
@@ -599,10 +586,42 @@ EOH
...
+    resource_attributes="${resource_attributes},deployment.environment=${deployment_environment}"
+  fi
+
+  backup_file "$dotnet_zeroconfig_path"
+
+  echo "Creating ${dotnet_zeroconfig_path}"
+  cat <<EOH > $dotnet_zeroconfig_path
+CORECLR_ENABLE_PROFILING=1
+CORECLR_PROFILER={918728DD-259F-4A6A-AC2B-B85E1B658318}
+CORECLR_PROFILER_PATH=${dotnet_install_dir}/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so
```
pjanotti, line 604: I guess this script doesn't support arm nor Alpine, is that correct?

pjanotti (approved):

internal/buildscripts/packaging/installer/install.sh
jeffreyc-splunk: > I guess this script doesn't support arm nor Alpine, is that correct?

Correct.  The installer script only supports deb/rpm distros, and `splunk-otel-dotnet` does not currently support arm64.

PR #4739, opened 2024-04-26
[receiver/discovery] Update entity events ID fields

dmitryax:
Update entity events ID fields according to the latest contract with the inventory service:
- `discovery.endpoint.id` is moved from identifying to regular attributes
- The following attributes now marked as identifying instead:
  - `k8s.pod.uid`
  - `k8s.node.uid`
  - `container.id`
  - `source.port`
  - `host.id`

### DIFF
--- a/internal/receiver/discoveryreceiver/endpoint_tracker.go
+++ b/internal/receiver/discoveryreceiver/endpoint_tracker.go
@@ -23,15 +23,25 @@ import (
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/plog"
+	semconv "go.opentelemetry.io/collector/semconv/v1.22.0"
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
 
 	"github.com/signalfx/splunk-otel-collector/internal/common/discovery"
 )
 
-var (
-	_ observer.Notify = (*notify)(nil)
-)
+const sourcePortAttr = "source.port"
+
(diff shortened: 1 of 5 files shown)


internal/receiver/discoveryreceiver/endpoint_tracker.go
```diff
@@ -214,40 +227,56 @@ func (n *notify) OnChange(changed []observer.Endpoint) {
...
 		for k, v := range correlations.Attrs(endpoint.ID) {
 			attrs.PutStr(k, v)
 		}
+		extractIdentifyingAttrs(attrs, entityEvent.ID())
 	}
 	return entityEvents, failed, err
 }
 
-func entityDeleteEvents(endpoints []observer.Endpoint, ts time.Time) experimentalmetricmetadata.EntityEventsSlice {
+func entityDeleteEvents(endpoints []observer.Endpoint, ts time.Time) (ees experimentalmetricmetadata.EntityEventsSlice, failed int, err error) {
```
samiura, line 260: Why we need the failed (int) value? Is it a return code?

samiura (approved):
Overall LGTM! Just a question left there to understand.

internal/receiver/discoveryreceiver/endpoint_tracker.go
dmitryax: This is just for logging. We do the same for `emitEntityStateEvents`, so I kept it the same, but I agree it's not ideal. We can probably move the logging inside the funcs. I can address it separately

dmitryax:
Test failures are unrelated and fixed in https://github.com/signalfx/splunk-otel-collector/pull/4740

PR #4755, opened 2024-04-30
Use latest MSI capabilities instead of custom code in

pjanotti:
**Description:**
Leverage the latest MSI capabilities so we don't rely on custom Chocolatey code to install the collector on Windows. When version 0.99.0 and lower are retired we can remove the code not used anymore.

Testing: CI

Documentation:


internal/buildscripts/packaging/choco/splunk-otel-collector/tools/chocolateyinstall.ps1
```diff
@@ -149,24 +149,43 @@ $packageArgs = @{
...
+    }
+    elseif ($MODE -eq "gateway"){
+        $config_path = "$program_data_path\gateway_config.yaml"
+        if (-NOT (Test-Path -Path "$config_path")) {
+            write-host "Copying gateway_config.yaml to $config_path"
+            Copy-Item "$installation_path\gateway_config.yaml" "$config_path"
+        }
     }
+
+    $env_vars["SPLUNK_CONFIG"] = "$config_path"
```
dmitryax, line 186: Shouldn't this still be outside so it's set regardless of `$use_msi_properties`?

internal/buildscripts/packaging/choco/splunk-otel-collector/tools/chocolateyinstall.ps1
```diff
@@ -149,24 +149,43 @@ $packageArgs = @{
     validExitCodes = @(0)
 }
 
-Install-ChocolateyInstallPackage @packageArgs
+$use_msi_properties = [Version]$env:chocolateyPackageVersion -ge [Version]"0.99.0"
```
dmitryax, line 152: Is this condition needed so users can use old collector versions with the new chocolatey release? Do we need to support that?

internal/buildscripts/packaging/choco/splunk-otel-collector/tools/chocolateyinstall.ps1
pjanotti: No, the MSI handles that automatically.

internal/buildscripts/packaging/choco/splunk-otel-collector/tools/chocolateyinstall.ps1
jinja2: Why are we comparing for >= "0.99"? 0.99 version of the chocolatey package was already released. Do we want to only use the msi properties code from new version of package going forward? Is this script not part of the packaging, as in, won't this change only go out in the next release of the package?

pjanotti: Unfortunately yes: all the "scripts for install" are made this way and their version is not tied to the collector so for now it keeps the old code around until we deprecate those versions... some time in the future there will be a lot of code deletion :)

pjanotti: > Is this script not part of the packaging, as in, won't this change only go out in the next release of the package?

Yes, this is correct only in the next published version. 

> Why are we comparing for >= "0.99"? 0.99 version of the chocolatey package was already released.

The chocolatey package has the same version as the MSI, they are packaged together - `0.99.0` is the first version of the MSI that supports the features needed for the install.

pjanotti: > It may be best to keep support in place until we get sign off on the explicit support plan w.r.t versions

The choco package is tied to the collector version, so in principle we can remove the old code right now. 🤔 I'm not sure yet which way to go...

pjanotti: Going with remove: it is just a small piece of code https://github.com/signalfx/splunk-otel-collector/pull/4755/commits/5d54c7e99431c97bef37cd298288b5a5b32b5050

jinja2 (approved):

PR #4757, opened 2024-04-30
Upgrades version of patchelf to 0.18.0

hughesjj:
They're no longer using the `releases.nixos.org` endpoint.

Personally, this won't build on my laptop without [`this patch`](https://github.com/signalfx/splunk-otel-collector/pull/4756/files), but I'm not sure that applies for others.


jeffreyc-splunk:
The tests are failing because `agent-bundle/jre/bin/java` seg faults with this version.

samiura (approved):

samiura:
@hughesjj few of the tests are failing, may be rebase and see if that alleviates the issues

PR #4763, opened 2024-05-02
chef: support .NET zero config on linux

jeffreyc-splunk:
- Activate .NET auto instrumentation by default if arch is amd64/x86_64 and version >= 0.99.0 for either preload or systemd.
- Update kitchen tests to verify env vars for preload and systemd configs.


deployments/chef/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## chef-v0.14.0
+
+- Initial support for [Splunk OpenTelemetry for .NET](https://github.com/signalfx/splunk-otel-dotnet) Auto
+  Instrumentation on Linux (x86_64/amd64 only):
+  - The .NET SDK is activated by default if the `with_auto_instrumentation` option is set to `true` and
+    `auto_instrumentation_version` is `latest` or >= `0.99.0`.
+  - Set the `with_auto_instrumentation_sdks` option to `java` and/or `nodejs` to skip .NET auto instrumentation.
```
jvoravong, line 9: I might suggest we make points like this more abstract now or later as we will keep adding more supported instrumentation instrumentation languages.

deployments/chef/recipes/auto_instrumentation.rb
```diff
@@ -5,6 +5,8 @@
 with_systemd = node['splunk_otel_collector']['auto_instrumentation_systemd'].to_s.downcase == 'true'
 with_java = node['splunk_otel_collector']['with_auto_instrumentation_sdks'].include?('java')
 with_nodejs = node['splunk_otel_collector']['with_auto_instrumentation_sdks'].include?('nodejs') && with_new_instrumentation
+dotnet_supported = %w(x86_64 amd64).include?(node['cpu']['architecture']) && (node['splunk_otel_collector']['auto_instrumentation_version'] == 'latest' || Gem::Version.new(node['splunk_otel_collector']['auto_instrumentation_version']) >= Gem::Version.new('0.99.0'))
```
jvoravong, line 8: arm support was just added with the recent https://github.com/signalfx/splunk-otel-dotnet/releases/tag/v1.5.0 release. Should we account for this in this PR?

jeffreyc-splunk: We'll first need to update and release our `splunk-otel-auto-instrumentation` deb/rpm packages to support arm64 before we can add arm64 support in our installers.

deployments/chef/CHANGELOG.md
jeffreyc-splunk: Something like?

```suggestion
  - To skip .NET auto instrumentation, configure the `with_auto_instrumentation_sdks` option without `dotnet`.
```

pjanotti (approved):

PR #4769, opened 2024-05-04
Remove test skip

atoulme:
This test passes locally ; trying to remove skip and see if it passes CI.


atoulme:
I don't get why, it works on my machine. I will close this PR...

PR #4772, opened 2024-05-05
Move off testutils for signalfx exporter translation tests

atoulme:


tests/exporters/signalfx/signalfx_translation_test.go
```diff
@@ -36,7 +36,12 @@ func TestSignalFxExporterTranslatesOTelCPUMetrics(t *testing.T) {
...
 func TestSignalFxExporterTranslatesOTelMemoryMetrics(t *testing.T) {
-	testutils.AssertAllMetricsReceived(
-		t, "memory_translations.yaml", "memory_translations_config.yaml", nil, nil,
+	testutils.CheckGoldenFile(t, "memory_translations_config.yaml", "memory_translations_expected.yaml",
+		pmetrictest.IgnoreTimestamp(),
+		pmetrictest.IgnoreMetricAttributeValue("host.name"),
+		pmetrictest.IgnoreMetricValues(),
+		pmetrictest.IgnoreMetricsOrder(),
+		pmetrictest.IgnoreMetricDataPointsOrder(),
+		pmetrictest.IgnoreSubsequentDataPoints(),
```
crobert-1, line 45: Why do we want to ignore subsequent data points? (I see that we're doing that for the `OTelCPUMetrics` test as well, but also not sure for that one)

crobert-1 (approved):

tests/exporters/signalfx/signalfx_translation_test.go
atoulme: My machine has more CPUs, so the datapoint count didn't match.

PR #4791, opened 2024-05-06
puppet: support .NET zero config on linux

jeffreyc-splunk:
- Activate .NET auto instrumentation by default if arch is amd64/x86_64 and version >= 0.99.0 for either preload or systemd.
- Update tests to verify env vars for preload and systemd configs.


theletterf (approved):

deployments/puppet/templates/dotnet.conf.erb
```diff
@@ -0,0 +1,20 @@
...
+<% else -%>
+OTEL_RESOURCE_ATTRIBUTES=splunk.zc.method=splunk-otel-auto-instrumentation-<%= @auto_instrumentation_version %>
+<% end -%>
+<% if defined?(@auto_instrumentation_service_name) && @auto_instrumentation_service_name != "" -%>
+OTEL_SERVICE_NAME=<%= @auto_instrumentation_service_name %>
+<% end -%>
+SPLUNK_PROFILER_ENABLED=<%= @auto_instrumentation_enable_profiler.to_s.downcase %>
+SPLUNK_PROFILER_MEMORY_ENABLED=<%= @auto_instrumentation_enable_profiler_memory.to_s.downcase %>
+SPLUNK_METRICS_ENABLED=<%= @auto_instrumentation_enable_metrics.to_s.downcase %>
+OTEL_EXPORTER_OTLP_ENDPOINT=<%= @auto_instrumentation_otlp_endpoint %>
```
pjanotti, line 20: If this end point is the gRPC one we need to set `OTEL_EXPORTER_OTLP_PROTOCOL` to `grpc`. The issue is that .NET defaults to `http/protobuf` instead of `grpc`.

pjanotti (approved):
The only potential issue is regarding the fact that .NET auto-defaults to `http/protobuf` instead of `grpc`. See comment below.

deployments/puppet/templates/dotnet.conf.erb
jeffreyc-splunk: > If this end point is the gRPC one we need to set `OTEL_EXPORTER_OTLP_PROTOCOL` to `grpc`. The issue is that .NET defaults to `http/protobuf` instead of `grpc`.

Hmm... `libsplunk.so` doesn't support `OTEL_EXPORTER_OTLP_PROTOCOL`: https://github.com/signalfx/splunk-otel-collector/blob/main/instrumentation/src/main.c#L9.

PR #4835, opened 2024-05-14
[receiver/discovery] Remove discovery.event.type and metric.name attrs

dmitryax:
Stop emitting `discovery.event.type` and `metric.name` entity attributes. They are not relevant anymore since the entity events are emitted constantly from the latest state


crobert-1 (commented):
It looks like there are still some references to these attributes in the receiver's [README](https://github.com/signalfx/splunk-otel-collector/blob/main/internal/receiver/discoveryreceiver/README.md), I believe we should update there too.

PR #4844, opened 2024-05-16
working integration tests for discovery

samiura:
**Description:** Using otlp sink to test mongodb discovery integration test.

### DIFF
--- a/tests/receivers/mongodb/mongodb_discovery_test.go
+++ b/tests/receivers/mongodb/mongodb_discovery_test.go
@@ -33,7 +33,6 @@ import (
 	"github.com/testcontainers/testcontainers-go"
 	"go.opentelemetry.io/collector/component/componenttest"
 	"go.opentelemetry.io/collector/consumer/consumertest"
-	"go.opentelemetry.io/collector/pdata/plog"
 	"go.opentelemetry.io/collector/receiver/otlpreceiver"
 	"go.opentelemetry.io/collector/receiver/receivertest"
 )
@@ -48,9 +47,18 @@ const (
 	OtelEntityAttributesAttr = "otel.entity.attributes"
 )
 
-func expectedPLogs() plog.Logs {
-	plogs := plog.NewLogs()
-	return plogs
+func getDockerGID() (string, error) {
+	finfo, err := os.Stat("/var/run/docker.sock")
+	if err != nil {
+		return "", err
+	}
+	fsys := finfo.Sys()
+	stat, ok := fsys.(*syscall.Stat_t)
+	if !ok {
+		return "", fmt.Errorf("OS error occurred while trying to get GID ")
+	}
+	dockerGID := fmt.Sprintf("%d", stat.Gid)
+	return dockerGID, nil
 }
 
 func mongoDBAutoDiscoveryHelper(t *testing.T, ctx context.Context, configFile string, logMessageToAssert string) (*otelContainer, error) {
@@ -67,16 +75,9 @@ func mongoDBAutoDiscoveryHelper(t *testing.T, ctx context.Context, configFile st
 		require.NoError(t, receiver.Shutdown(context.Background()))
 	})
 
-	finfo, err := os.Stat("/var/run/docker.sock")
-	if err != nil {
-		return nil, err
-	}
-	fsys := finfo.Sys()
-	stat, ok := fsys.(*syscall.Stat_t)
-	if !ok {
-		return nil, fmt.Errorf("OS error occurred while trying to get GID ")
-	}
-	dockerGID := fmt.Sprintf("%d", stat.Gid)
+	dockerGID, err := getDockerGID()
+	require.NoError(t, err)
+
 	otelConfigPath, err := filepath.Abs(filepath.Join(".", "testdata", configFile))
 	if err != nil {
 		return nil, err
@@ -133,12 +134,13 @@ func mongoDBAutoDiscoveryHelper(t *testing.T, ctx context.Context, configFile st
 			assert.Fail(tt, "No logs collected")
 			return
 		}
-		assert.NoError(tt, err)
+		countAtleastOneGoodLogAttr := 0
 		for i := 0; i < len(sink.AllLogs()); i++ {
 			plogs := sink.AllLogs()[i]
 			lr := plogs.ResourceLogs().At(0).ScopeLogs().At(0).LogRecords().At(0)
 			attrMap, ok := lr.Attributes().Get(OtelEntityAttributesAttr)
 			if ok {
+				countAtleastOneGoodLogAttr++
 				m := attrMap.Map()
 				discoveryMsg, ok := m.Get(MessageAttr)
 				if ok {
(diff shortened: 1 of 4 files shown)


tests/receivers/mongodb/mongodb_discovery_test.go
```diff
@@ -24,18 +24,49 @@ import (
...
 
 type otelContainer struct {
 	testcontainers.Container
 }
 
-func mongoDBAutoDiscoveryHelper(ctx context.Context, configFile string, logMessageToAssert string) (*otelContainer, error) {
+const (
+	ReceiverTypeAttr         = "discovery.receiver.type"
+	MessageAttr              = "discovery.message"
+	OtelEntityAttributesAttr = "otel.entity.attributes"
```
dmitryax, line 48: Can't we use constants from `internal/common/discovery`?

tests/receivers/mongodb/mongodb_discovery_test.go
```diff
@@ -97,52 +128,74 @@ func mongoDBAutoDiscoveryHelper(ctx context.Context, configFile string, logMessa
...
+			return
+		}
+		assert.NoError(tt, err)
+		for i := 0; i < len(sink.AllLogs()); i++ {
+			plogs := sink.AllLogs()[i]
+			lr := plogs.ResourceLogs().At(0).ScopeLogs().At(0).LogRecords().At(0)
+			attrMap, ok := lr.Attributes().Get(OtelEntityAttributesAttr)
+			if ok {
+				m := attrMap.Map()
+				discoveryMsg, ok := m.Get(MessageAttr)
```
dmitryax, line 143: We need to check that there is at least on `ok` in `sink.AllLogs()`. Otherwise, this test doesn't validate presence of the attributes

tests/receivers/mongodb/mongodb_discovery_test.go
```diff
@@ -24,18 +24,49 @@ import (
...
+const (
+	ReceiverTypeAttr         = "discovery.receiver.type"
+	MessageAttr              = "discovery.message"
+	OtelEntityAttributesAttr = "otel.entity.attributes"
+)
+
+func expectedPLogs() plog.Logs {
+	plogs := plog.NewLogs()
+	return plogs
+}
```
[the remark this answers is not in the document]
samiura: removed, I had earlier designs for it but no longer needed!

tests/receivers/mongodb/mongodb_discovery_test.go
```diff
@@ -24,18 +24,49 @@ import (
...
+
+func mongoDBAutoDiscoveryHelper(t *testing.T, ctx context.Context, configFile string, logMessageToAssert string) (*otelContainer, error) {
+	factory := otlpreceiver.NewFactory()
+	port := 16745
+	c := factory.CreateDefaultConfig().(*otlpreceiver.Config)
+	c.GRPC.NetAddr.Endpoint = fmt.Sprintf("localhost:%d", port)
+	endpoint := c.GRPC.NetAddr.Endpoint
```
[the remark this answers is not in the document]
samiura: I do use the endpoint later down [here](https://github.com/signalfx/splunk-otel-collector/pull/4844/files#diff-e8addc535a1908253acda7263b71bc2747195bb5b9beebbc432595d2b511a9cfR104)

tests/receivers/mongodb/mongodb_discovery_test.go
```diff
@@ -97,52 +128,74 @@ func mongoDBAutoDiscoveryHelper(ctx context.Context, configFile string, logMessa
 		return nil, err
 	}
 
+	assert.EventuallyWithT(t, func(tt *assert.CollectT) {
+		if len(sink.AllLogs()) == 0 {
+			assert.Fail(tt, "No logs collected")
+			return
+		}
+		assert.NoError(tt, err)
```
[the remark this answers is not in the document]
samiura: absolutely not needed, not sure what is it doing here and how? I have removed it.

tests/receivers/mongodb/mongodb_discovery_test.go
samiura: @dmitryax This is the same reason, we are trying to avoid not using TestUtils. When we bring this module underneath tests/go.mod it starts a circular dependency issues. I propose eventually, when we get rid of testutils  all together we can use it or may be move this somewhere inside tests package to reuse.

tests/receivers/mongodb/mongodb_discovery_test.go
samiura: I agree, I missed it.

tests/receivers/mongodb/mongodb_discovery_test.go
dmitryax: Ok. I believe we should move testutils from tests to `internal` even before it can be removed. It should not be exposed outside anyway. Then, we can have tests->internal dependencies

dmitryax: Not needed in this PR. Maybe as a follow up if you have time

tests/receivers/mongodb/mongodb_discovery_test.go
dmitryax: I meant checking both `MessageAttr` and `ReceiverTypeAttr`. If we change those attributes in the future, these tests will become a no-op. You can use booleans like `seenMessageAttr` and `seenReceiverTypeAttr`

samiura: Oh :), I was thinking for the case where you do not see any items because you ran the tests without a Mongodb Instance and still the test will pass still. Got it, I working on it.

tests/receivers/mongodb/mongodb_discovery_test.go
samiura: I will do it after this gets merged when I removed TestDockerObsever.

tests/receivers/mongodb/mongodb_discovery_test.go
samiura: By the way, I had to increase the assertEventuallyT tests timesteps, 1 second was not good enough sometimes for sink to absorb all the entity events.

tests/receivers/mongodb/mongodb_discovery_test.go
```diff
@@ -97,52 +129,79 @@ func mongoDBAutoDiscoveryHelper(ctx context.Context, configFile string, logMessa
...
+				}
+			}
+		}
+		assert.True(t, seenReceiverTypeAttr > 0)
+		assert.True(t, seenReceiverTypeAttr > 0)
+	}, 60*time.Second, 5*time.Second)
+
 	return &otelContainer{Container: container}, nil
 }
-
```
dmitryax, line 102: Why?

tests/receivers/mongodb/mongodb_discovery_test.go
```diff
@@ -97,52 +129,79 @@ func mongoDBAutoDiscoveryHelper(ctx context.Context, configFile string, logMessa
...
 		logMessageToAssert string
 		expected           error
 	}{
-		"Fully Successful Discovery test": {
-			ctx:                ctx,
-			configFileName:     "docker_observer_without_ssl_mongodb_config.yaml",
-			logMessageToAssert: successfulDiscoveryMsg,
-			expected:           nil,
-		},
+
```
dmitryax, line 178:
```suggestion
```

dmitryax (approved):

tests/receivers/mongodb/mongodb_discovery_test.go
samiura: My bad... removing it.

PR #4871, opened 2024-05-22
[OTL-2836] Set hosts in MongoDB config from autodiscover endpoint

crobert-1:

`endpoint` is coming from receiver creator when it matches an endpoint provided by an observer extension. The `mongodb` receiver uses the `endpoint` in a slightly different way than expected, as an item in a `hosts` map, rather than directly as an `endpoint` config option. This updates the receiver creator's way of creating `mongodb` as the target receiver, setting the endpoint properly in the `hosts` config option.

I was able to see that the option is being picked up properly, and the MongoDB receiver's config is setup as expected, but discovery didn't succeed due to docker's IP configuration on my local machine. Posting PR as draft while I investigate locally.


samiura (approved):

dmitryax:
LGTM. @crobert-1 can we turn it from draft?

PR #4911, opened 2024-05-30
[discovery] Fix crashing collector if discovered mongodb isn't reachable

dmitryax:
Reduce the connection timeout to 5s seconds. Otherwise, it's the 60s for every Mongodb query, which accumulates to a significant delay and makes the collector killed in k8s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


atoulme (approved):

crobert-1 (approved):

samiura (approved):

samiura:
@dmitryax  thank you!

PR #4915, opened 2024-05-31
[discovery] Update redis discovery instructions

dmitryax:
Remove duplicate password field from the instructions

### DIFF
--- a/cmd/otelcol/config/collector/config.d.linux/receivers/redis.discovery.yaml
+++ b/cmd/otelcol/config/collector/config.d.linux/receivers/redis.discovery.yaml
@@ -38,18 +38,16 @@
 #         log_record:
 #           body: >-
 #             Make sure your user credentials are correctly specified using the
-#             `SPLUNK_DISCOVERY_RECEIVERS_redis_CONFIG_password="<password>"` and
-#             `SPLUNK_DISCOVERY_RECEIVERS_redis_CONFIG_password="<password>"` environment variables.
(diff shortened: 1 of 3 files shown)


cmd/otelcol/config/collector/config.d.linux/receivers/redis.discovery.yaml
```diff
@@ -38,7 +38,6 @@
 #         log_record:
 #           body: >-
 #             Make sure your user credentials are correctly specified using the
-#             `SPLUNK_DISCOVERY_RECEIVERS_redis_CONFIG_password="<password>"` and
```
crobert-1, line 41: Should we be adding the username config option here to replace the duplicate?

dmitryax: There is no `username` config option for redis receiver. It needs password only

internal/confmapprovider/discovery/bundle/bundle.d/receivers/redis.discovery.yaml.tmpl
```diff
@@ -30,7 +30,6 @@
         log_record:
           body: >-
             Make sure your user credentials are correctly specified using the
-            `{{ configPropertyEnvVar "password" "<password>" }}` and
             `{{ configPropertyEnvVar "password" "<password>" }}` environment variables.
```
crobert-1, line 33:
```suggestion
            `{{ configPropertyEnvVar "password" "<password>" }}` environment variable.
```
We should remove the pluralization on each of these messages.

crobert-1 (approved):

PR #4949, opened 2024-06-07
Use msi capabilities for chef

pjanotti:
**Description:**
Leverage the latest MSI capabilities so we don't rely on custom Chef code to install the collector on Windows. At this point if the user wants to install only the collector on Windows they can directly use https://docs.chef.io/resources/windows_package/

The change itself move the handling of the configuration options (env. vars) to a step before installing the MSI. The MSI properties are passed on all versions since legacy versions would simply ignore the unknown properties. The code to setup the registry is only invoked when installing a MSI version that doesn't support the install properties.

**Testing:** CI

**Documentation:**

### DIFF
--- a/deployments/chef/recipes/collector_win_config_options.rb
+++ b/deployments/chef/recipes/collector_win_config_options.rb
@@ -1,0 +1,29 @@
+# Cookbook:: splunk_otel_collector
+# Recipe:: collector_win_config_options
+
+collector_env_vars = [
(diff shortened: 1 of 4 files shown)


samiura (approved):

jeffreyc-splunk (approved):
If you want to release these changes, please bump the version in https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/chef/metadata.rb#L6 and update https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/chef/CHANGELOG.md.

pjanotti:
> If you want to release these changes...

I will release all 3 mass deployments after Puppet is done. Thanks for the reminder about that.

PR #4975, opened 2024-06-12
Check for token before running snyk scans

jeffreyc-splunk:
Run snyk scan jobs only if the token is available. The snyk token secret is unavailable for PRs from external contributors, including dependabot, which causes the workflow to fail.

Also:
- Run the vulnerability scan workflow every weekday at midnight
- Fix snyk scan for arm64 collector images

### DIFF
--- a/.github/workflows/vuln-scans.yml
+++ b/.github/workflows/vuln-scans.yml
@@ -10,6 +10,8 @@ on:
       - '.grype.yaml'
       - '.trivyignore'
       - '.snyk'
+  schedule:
+    - cron: '0 0 * * 1-5' # Every weekday at midnight UTC
 
 env:
   GO_VERSION: '1.21.11'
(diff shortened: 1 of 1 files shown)


.github/workflows/vuln-scans.yml
```diff
@@ -187,8 +189,24 @@ jobs:
...
+          if [ -n "${{ secrets.SNYK_TOKEN }}" ]; then
+            echo "defined=true" >> $GITHUB_OUTPUT
+          else
+            echo "defined=false" >> $GITHUB_OUTPUT
+          fi
+
   snyk-fs-scan:
     runs-on: ubuntu-latest
+    needs: check-snyk-token
+    if: ${{ needs.check-snyk-token.outputs.has-snyk-token == 'true' }}
```
pjanotti, line 209: Nice! Quiet forks :)

.github/workflows/vuln-scans.yml
```diff
@@ -206,24 +224,31 @@ jobs:
...
       matrix:
         ARCH: [ "amd64", "arm64" ]
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
+      - uses: docker/setup-qemu-action@v3
+        if: ${{ matrix.ARCH != 'amd64' }}
+        with:
+          platforms: ${{ matrix.ARCH }}
+          image: tonistiigi/binfmt:qemu-v7.0.0
```
pjanotti, line 239: Is this image something that we can rely will be around for the long haul?

pjanotti (approved):

.github/workflows/vuln-scans.yml
jeffreyc-splunk: > Is this image something that we can rely will be around for the long haul?

I hope so.  It is the default image for this action (https://github.com/marketplace/actions/docker-setup-qemu#inputs).  We're just using a newer tag (`latest` is currently v6.2.0).

crobert-1 (approved):

PR #4977, opened 2024-06-13
integration-test workflow updates

jeffreyc-splunk:
- Move binary build to a separate job
- Add new `integration-test-binary` job to install and test the otelcol binary and agent-bundle directly on ubuntu 20.04, 22.04, and 24.04 runners, in addition to the existing job that only tests the collector docker image

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -54,11 +54,31 @@ jobs:
           name: agent-bundle-linux-${{ matrix.ARCH }}
           path: ./dist/agent-bundle_linux_${{ matrix.ARCH }}.tar.gz
 
+  otelcol:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        ARCH: [ "amd64", "arm64", "ppc64le" ]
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-test.yml
```diff
@@ -54,11 +54,31 @@ jobs:
           name: agent-bundle-linux-${{ matrix.ARCH }}
           path: ./dist/agent-bundle_linux_${{ matrix.ARCH }}.tar.gz
 
+  otelcol:
+    runs-on: ubuntu-20.04
```
pjanotti, line 58: Is it possible to use a more recent ubuntu?

jeffreyc-splunk: I think so.  Our compile jobs (which this job is based on) currently run on `ubuntu-20.04`, for example https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/build-and-test.yml#L214.  So we can try updating them to a newer runner in a separate PR.

pjanotti (approved):

PR #4978, opened 2024-06-14
Use msi capabilities for puppet

pjanotti:
**Description:**
Leverage the latest MSI capabilities so we don't rely on custom Puppet code to install the collector on Windows. At this point if the user wants to install only the collector on Windows they can directly use https://www.puppet.com/docs/puppet/8/types/package.html

The change itself move the handling of the configuration options (env. vars) to a step before installing the MSI. The MSI properties are passed on all versions since legacy versions would simply ignore the unknown properties. The code to setup the registry is only invoked when installing a MSI version that doesn't support the install properties or additional environment variables were specified for the setup (it is easy to setup everything in the registry than read the registry and merge the additional environment variables).

**Testing:**
Added tests in Windows to ensure that the latest version is being tested anyway.

**Documentation:**

### DIFF
--- a/.github/workflows/puppet-test.yml
+++ b/.github/workflows/puppet-test.yml
@@ -109,6 +109,8 @@ jobs:
       - name: Test puppet deployment
         id: pytest
         continue-on-error: true
+        env:
+          PUPPET_RELEASE: "${{ matrix.PUPPET_RELEASE }}"
         run: |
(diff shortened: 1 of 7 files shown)


pjanotti:
Some changes here for Puppet call my attention to the additional env vars support that I need to revisit for Ansible and Chef.

jeffreyc-splunk (approved):

PR #4996, opened 2024-06-19
Fix osfamily Fact

bran1501:
**Description:** 

Team, 

I am configuring the Splunk OTEL Collector using the puppet module and we encountered the issue that the osfamily was not part of the supported family. 

This is a windows box running Puppet 8.4.0, in this version puppet removed osfamily as mentioned here https://www.puppet.com/docs/puppet/8/core_facts.html#osfamily, it is currently a legacy fact. 

However, there is another server running a Puppet 6.17.0 and the osfamily exists there. 

Based on the forge https://forge.puppet.com/modules/signalfx/splunk_otel_collector/readme We support 6+

**Link to Splunk idea:** None

**Testing:** Explained above. Screenshots attached

**Documentation:** 
https://www.puppet.com/docs/puppet/8/core_facts.html#osfamily
https://forge.puppet.com/modules/signalfx/splunk_otel_collector/readme

Left side: Latest version
Right Side: Old version

image
image

### DIFF
--- a/deployments/puppet/manifests/params.pp
+++ b/deployments/puppet/manifests/params.pp
@@ -26,7 +26,7 @@ class splunk_otel_collector::params {
     $auto_instrumentation_version = 'latest'
     $auto_instrumentation_java_agent_jar = '/usr/lib/splunk-instrumentation/splunk-otel-javaagent.jar'
 
-  } elsif $::osfamily == 'windows' {
+  } elsif $::osfamily == 'windows' or $::family == 'windows' {
     $collector_version = ''
     $collector_install_dir = "${::win_programfiles}\\Splunk\\OpenTelemetry Collector"
     $collector_config_dir = "${::win_programdata}\\Splunk\\OpenTelemetry Collector"


deployments/puppet/manifests/params.pp
```diff
@@ -26,7 +26,7 @@
     $auto_instrumentation_version = 'latest'
     $auto_instrumentation_java_agent_jar = '/usr/lib/splunk-instrumentation/splunk-otel-javaagent.jar'
 
-  } elsif $::osfamily == 'windows' {
+  } elsif $::osfamily == 'windows' or $::family == 'windows' {
```
jeffreyc-splunk, line 29: - `$::osfamily` is a legacy fact and should still be available to use in manifests, but are just hidden in the facter output by default.  From https://www.puppet.com/docs/puppet/8/core_facts.html:

  > Legacy Facts Note: As of Facter 3, legacy facts such as architecture are hidden by default to reduce noise in Facter's default command-line output. These older facts are now part of more useful structured facts; for example, architecture is now part of the os fact and accessible as os.architecture. You can still use these legacy facts in Puppet manifests ($architecture), request them on the command line (facter architecture), and view them alongside structured facts (facter --show-legacy). Legacy facts are excluded by default.

- I'm not sure if `$::family` is a valid fact. To use the non-legacy structured fact, the correct syntax should be `$facts['os']['family']` (see https://www.puppet.com/docs/puppet/8/lang_facts_accessing).  We can try changing all references to `$::osfamily` to `$facts['os']['family']` in our manifests to see if that helps.

jeffreyc-splunk:
Superseded by #5003

PR #5001, opened 2024-06-20

srv-gh-o11y-gdi:


internal/receiver/signalfxgatewayprometheusremotewritereceiver/server.go
```diff
@@ -38,13 +38,13 @@ type prometheusRemoteWriteServer struct {
...
 
 type serverConfig struct {
-	confighttp.ServerConfig
 	component.TelemetrySettings
 	Reporter reporter
 	component.Host
 	Mc     chan<- pmetric.Metrics
 	Parser *prometheusRemoteOtelParser
 	Path   string
+	confighttp.ServerConfig
```
crobert-1, line 47: Not a problem at all, but curious as to why this line was moved to the end of the struct?

crobert-1 (approved):

crobert-1:
Thanks for all your work here @jeffreyc-splunk!

internal/receiver/signalfxgatewayprometheusremotewritereceiver/server.go
jeffreyc-splunk: > Not a problem at all, but curious as to why this line was moved to the end of the struct?

Not sure. It was just the output from `make generate-metrics`: https://github.com/signalfx/splunk-otel-collector/pull/5001/commits/bc8fa8ce02010c2941f18c74580c2be461b895bb

crobert-1:
Looks good to me. Looks like the main changes were some simple renames:
```
otelcol.NewCommand -> otelcol.NewCommandMustProviderSettings
CreateSettings -> Settings
```
And then a bit more of a complicated one:
```
UnmarshalConfig(conf *confmap.Conf, intoCfg Config) -> conf.Unmarshal(&intoCfg)
```
I didn't check every instance of the last change, but they all looked good from what I could see :+1:

PR #5025, opened 2024-06-25
Added Mongodb versions 6.0 and 7.0 to the list of auto discovery inte…

samiura:
**Description:**  Updated integration tests to cover mongo versions 6.0 and 7.0. Also updated Mongo discovery template.

mongo_test


cmd/otelcol/config/collector/config.d.linux/receivers/mongodb.discovery.yaml
```diff
@@ -21,7 +21,6 @@
 #         insecure: false
 #       hosts:
 #         - endpoint: '`endpoint`'
-#           transport: tcp
```
samiura, line 24: This is no longer needed for discovery due to upstream change and transport key is no longer recognized either.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/mongodbreceiver/README.md#example-configuration

atoulme (approved):

docker/mongodb/scripts/run.sh
```diff
@@ -1,22 +1,35 @@
...
   RET=$?
   sleep 2
 done
 
-bash /home/mongodb/scripts/setup_user.sh
-
+if (( $(echo "${MONGO_MAJOR} > 5.9" | bc -l) )); then
+    bash /home/mongodb/scripts/setup_user_newer_ver.sh
+else
+    bash /home/mongodb/scripts/setup_user.sh
```
dmitryax, line 32: So the only difference is mongo cli? Can we just define it as a var once and avoid duplicating the scripts?

```
MONGO=$((echo "${MONGO_MAJOR} > 5.9" | bc -l) && echo "mongo" || echo "mongosh")
```

Then use ${MONGODB} in other places

samiura: working on it.

samiura: I tried the expression above. For the truth case, meaning version 6.0 and above it returns "1 mongosh" hence, I am simplifying the expression and making it a bit more verbose.

docker/mongodb/scripts/run.sh
```diff
@@ -1,22 +1,31 @@
 #!/bin/bash
 
+
+run_mongo_commandline () {
+if (( $(echo "$MONGO_MAJOR > 5.9" | bc -l) )); then
```
jeffreyc-splunk, line 5: Instead of depending on `bc`, how about just doing a numeric comparison of the first digit?  Something like:
```
major=$(echo "$MAJOR_VERSION" | cut -d '.' -f1)
if [ $major -gt 5 ]; then
...
```

docker/mongodb/Dockerfile
```diff
@@ -1,6 +1,10 @@
 ARG MONGODB_VERSION=4.0
 FROM mongo:${MONGODB_VERSION}
 
+RUN if [ "$MONGO_MAJOR" = 6.0 ] || [ "$MONGO_MAJOR" = 7.0 ]; then \
```
dmitryax, line 4: Installing `bc` is cheap. Let's just add it unconditionally instead of keep updating this line going forward

docker/mongodb/scripts/setup_user.sh
```diff
@@ -6,8 +6,13 @@ echo "************************************************************"
 echo "Setting up users..."
 echo "************************************************************"
 
+MONGO="mongo"
+if (( $(echo "$MONGO_MAJOR > 5.9" | bc -l) )); then
+    MONGO="mongosh"
+fi
```
dmitryax, line 12: Can you pass it to the script instead of duplicating? Something like 
```
MONGO=$MONGO /home/mongodb/scripts/setup_user.sh
```

or just export $MONGO=...

docker/mongodb/scripts/run.sh
samiura: easier solution. Also bc for some reason is not getting installed in the 4.4 version.

docker/mongodb/scripts/setup_user.sh
samiura: export!!

docker/mongodb/scripts/run.sh
```diff
@@ -1,22 +1,35 @@
 #!/bin/bash
 
+
+
+export MONGO="mongo"
+major=$(echo "$MONGO_MAJOR" | cut -d '.' -f1)
+if [ $major -gt 5 ]; then
+    MONGO="mongosh"
```
dmitryax, line 8: Shouldn't this be export as well?

samiura: good catch!!

dmitryax (approved):

PR #5029, opened 2024-06-26
Update splunk-otel-javaagent to latest

srv-gh-o11y-gdi:
- Update splunk-otel-javaagent to v2.5.0
- Update `libsplunk.so` to support `OTEL_EXPORTER_OTLP_PROTOCOL` and `OTEL_METRICS_EXPORTER`
- Update linux installer script:
  - Change default for `--otlp-endpoint` to empty and sets `OTEL_EXPORTER_OTLP_ENDPOINT` only if non-empty
  - Add `--otlp-endpoint-protocol` option (default empty) and sets `OTEL_EXPORTER_OTLP_PROTOCOL` only if non-empty and `--otlp-endpoint` is also specified
  - Add `--metrics-exporter` option (default empty) and sets `OTEL_METRICS_EXPORTER` only if non-empty


jeffreyc-splunk:
v2 of the java agent changed the default otlp protocol from `grpc` to `http/protobuf` (see https://github.com/signalfx/splunk-otel-java/releases/tag/v2.0.0-alpha).  This will break our installer script and deployment modules since they configure the otlp endpoint as `http://127.0.0.1:4317` (grpc) by default.  We'll need to add support for the `OTEL_EXPORTER_OTLP_PROTOCOL` env var to the installer script, deployment modules, and `libsplunk.so` to align with the configured endpoint.

jeffreyc-splunk:
Basically, the installer script will set `OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_PROTOCOL`, and `OTEL_METRICS_EXPORTER` only if the respective command-line options are explicitly specified, otherwise we'll leave them unset and defer to the default values of each activated SDK.

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@
   report higher memory usage, but it will be bound to the same configured limits. If you want to revert to the previous
   behavior, set the `GOGC` environment variable to `100`.
 - (Splunk) Upgrade to golang 1.21.12
+- (Splunk) Auto Discovery for Linux:
+  - Update `splunk-otel-java` to v2.5.0 for the `splunk-otel-auto-instrumentation` deb/rpm packages. Check the [release notes](https://github.com/signalfx/splunk-otel-java/releases/tag/v2.5.0) for possible breaking changes.
```
breedx-splk, line 15:
```suggestion
  - Update `splunk-otel-java` to v2.5.0 for the `splunk-otel-auto-instrumentation` deb/rpm packages. This is a major version bump that includes breaking changes. Check the [release notes](https://github.com/signalfx/splunk-otel-java/releases/tag/v2.5.0) for details about breaking changes.
```

instrumentation/packaging/java-agent-release.txt
```diff
@@ -1 +1 @@
-v1.32.2
+v2.5.0
```
breedx-splk, line 1: 🚀

internal/buildscripts/packaging/installer/install.sh
```diff
@@ -910,48 +949,54 @@ Auto Instrumentation:
...
+                                        (default: empty, i.e. defer to the default OTEL_EXPORTER_OTLP_ENDPOINT value for
+                                        each activated SDK)
+  --otlp-endpoint-protocol <protocol>   Set the protocol for the configured OTLP endpoint, for example "grpc" or
+                                        "http/protobuf". The value will be set to the OTEL_EXPORTER_OTLP_PROTOCOL
+                                        environment variable.
+                                        Only applicable if the "--otlp-endpoint <host:port>" option is also specified.
+                                        (default: empty, i.e. defer to the default OTEL_EXPORTER_OTLP_PROTOCOL value for
+                                        each activated SDK)
+  --metrics-exporter <exporters>        Comma-separated list of exporters for collected metrics by all activated SDKs,
+                                        for example "otlp,zipkin". Set the value to "none" to disable collection and
```
jinja2, line 985: small nit: zipkin would be traces only 
```suggestion
                                        for example "otlp,prometheus". Set the value to "none" to disable collection and
```

breedx-splk (approved):
Looks good to me.

aunshc:
Quick question: I was unable to find the OTEL_METRICS_EXPORTER exporter configuration for .NET instrumentation here https://docs.splunk.com/observability/en/gdi/get-data-in/application/otel-dotnet/configuration/advanced-dotnet-configuration.html. In this case, would the variable be set to `none`? I don't expect that to be an issue but want to confirm cc @pjanotti.

jeffreyc-splunk:
> Quick question: I was unable to find the OTEL_METRICS_EXPORTER exporter configuration for .NET instrumentation here https://docs.splunk.com/observability/en/gdi/get-data-in/application/otel-dotnet/configuration/advanced-dotnet-configuration.html. In this case, would the variable be set to `none`? I don't expect that to be an issue but want to confirm cc @pjanotti.

`OTEL_METRICS_EXPORTER` will only be set to `none` if the user explicitly specifies the `--metrics-exporter none` installer script option.

jeffreyc-splunk:
> Quick question: I was unable to find the OTEL_METRICS_EXPORTER exporter configuration for .NET instrumentation here https://docs.splunk.com/observability/en/gdi/get-data-in/application/otel-dotnet/configuration/advanced-dotnet-configuration.html. In this case, would the variable be set to `none`? I don't expect that to be an issue but want to confirm cc @pjanotti.

According to the [upstream doc](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/main/docs/config.md#exporters), the default for `OTEL_METRICS_EXPORTER ` is `otlp`. I assume that this is carried over to our distro?

jinja2:
Discussed internally, this is good to go.

PR #5032, opened 2024-06-27
[receiver/discovery] Do not emit endpoint that are not evaluated yet

dmitryax:
This change fixes the issue when "unknown" service can be shown in the inventory service right at the collector startup. We don't want to send entity events for endpoints that are not evaluated yet


crobert-1 (commented):
Should we add a bug fix changelog entry?

dmitryax:
> Should we add a bug fix changelog entry?

Added

dmitryax:
Sure. Updated `TestEntityEmittingLifecycle

crobert-1 (approved):

PR #5034, opened 2024-06-28
Set Go garbage collection target percentage to 400%

dmitryax:
After the removal of the memory_ballast extension in v0.97.0, the Go garbage collection is running more aggressively, which increases CPU usage and leads to reduced throughput of the collector. This change reduces the frequency of garbage collection cycles to improve the collector's performance for typical workloads. As a result, the collector will report higher memory usage, but it will be bound to the same configured limits. If you want to revert to the previous behavior, set the `GOGC` environment variable to `100`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+### 💡 Enhancements 💡
+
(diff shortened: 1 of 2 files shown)


pjanotti (approved):
LGTM

PR #5106, opened 2024-07-16
Move MSI tests to golang

pjanotti:
Moving the MSI tests to `golang` so it will be easier to run them on a dev box. In order to avoid opaque errors in the GH runners the test needs to run a large GH Windows runner.


pjanotti:
Failures are unrelated:
```terminal
FAIL: TestExpandedDollarSignsViaStandardEnvVar
FAIL: TestScriptReceiverTop 
FAIL: TestIntegrationMongoDBAutoDiscovery (122.11s)
    FAIL: TestIntegrationMongoDBAutoDiscovery/Successful_Discovery_test (61.93s)
    FAIL: TestIntegrationMongoDBAutoDiscovery/Partial_Discovery_test (60.18s)
```
Re-running tests to look for consistent failures.

pjanotti:
On retry only:
```terminal
FAIL: TestExpandedDollarSignsViaStandardEnvVar
```

jvoravong (approved):

.github/workflows/win-package-test.yml
```diff
@@ -321,88 +223,18 @@ jobs:
...
-        if: ${{ failure() }}
-        run: Get-Content msi-install.log
+          "MSI_COLLECTOR_PATH=$msi_path" | Out-File -FilePath $env:GITHUB_ENV -Append
 
-      - name: Start the collector service
+      - name: Run the MSI tests
         run: |
-          $sc = Start-Service splunk-otel-collector -PassThru
-          $sc.WaitForStatus('Running', '00:00:30')
+          go test -timeout 15m -v github.com/signalfx/splunk-otel-collector/tests/msi
```
jeffreyc-splunk, line 235: Should we use the `setup-go` action, e.g.
https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/windows-test.yml#L43-L47, instead of the default installed on the runner?

pjanotti: The test doesn't relate to the go version used to build the collector. I opted to use the one that comes on board with the runner to save any possible download and install time. It is technically possible that we can use a feature on the test code not present on the default runner version, but, unless that actually happens any download and install cost seems unjustified to me.

BTW, thanks for looking even after merge, I appreciate that!

PR #5132, opened 2024-07-23
fix: add loop control for options list task in otelwinreg

alanbty:
**Description:** 
Fixing an issue -  Add a loop control for loop task "Get Splunk OpenTelemetry Collector options list" in `otel_win_reg.yml`.
This fix prevents a potentially sensitive value from being displayed.

**Testing:**
```
TASK [signalfx.splunk_otel_collector.collector : Get Splunk OpenTelemetry Collector options list]
**********************************************************************************************************************************
skipping: [hostA] => (item=KEYOFMYSECRETVALUE) 
skipping: [hostA]
```


alanbty:

pjanotti (approved):

jeffreyc-splunk:
I'm not too familiar, but the output from our CI tests still shows the values:
```
TASK [signalfx.splunk_otel_collector.collector : Get Splunk OpenTelemetry Collector options list] ***
ok: [2016] => (item=SPLUNK_ACCESS_TOKEN) => {"ansible_facts": {"splunk_otel_collector_options_list": ["SPLUNK_ACCESS_TOKEN=fake-token"]}, "ansible_loop_var": "item", "changed": false, "item": {"key": "SPLUNK_ACCESS_TOKEN", "value": "fake-token"}}
...
```
According to the note for [label](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#limiting-loop-output-with-label):
```
This is for making console output more readable, not protecting sensitive data. If there is sensitive data in `loop`, set `no_log: true` on the task to prevent disclosure.
```
So if the goal is hide the sensitive data, then maybe we should try `no_log: true` instead of `loop_control`?

alanbty:
> I'm not too familiar, but the output from our CI tests still shows the values:
> 
> ```
> TASK [signalfx.splunk_otel_collector.collector : Get Splunk OpenTelemetry Collector options list] ***
> ok: [2016] => (item=SPLUNK_ACCESS_TOKEN) => {"ansible_facts": {"splunk_otel_collector_options_list": ["SPLUNK_ACCESS_TOKEN=fake-token"]}, "ansible_loop_var": "item", "changed": false, "item": {"key": "SPLUNK_ACCESS_TOKEN", "value": "fake-token"}}
> ...
> ```
> 
> According to the note for [label](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#limiting-loop-output-with-label):
> 
> ```
> This is for making console output more readable, not protecting sensitive data. If there is sensitive data in `loop`, set `no_log: true` on the task to prevent disclosure.
> ```
> 
> So if the goal is hide the sensitive data, then maybe we should try `no_log: true` instead of `loop_control`?

Hello @jeffreyc-splunk,

Sorry for the late response.

I think your tests have the Ansible verbosity parameter enabled, that's why the pipeline shows values. 

You're right, `no log: true` is the best solution to hide sensitive values. But it also remove error messages if there is an issue on this task. 
So, as you wish, either you want to avoid displaying sensitive values and use `no log: true`, or you use `loop_control` and allow sensitive values to be displayed ONLY with Ansible verbose parameter.

pjanotti:
@hughesjj I'm closing this one due to the lack of activity - we can always reopen as needed

PR #5134, opened 2024-07-23
Disable $$ expansion for backward compatibility with old versions

dmitryax:
Stop expanding environment variables starting with $$ in configuration files. This behavior was [introduced in v0.42.0](https://github.com/signalfx/splunk-otel-collector/pull/1099) to support a bug causing double expansion.

`SPLUNK_DOUBLE_DOLLAR_CONFIG_SOURCE_COMPATIBLE` controlling this behavior is removed now.

`$$` is now treated as an escape sequence representing a literal $ character.


jvoravong (approved):
These changes LGTM. I know we still have some failing TestExpandedDollarSignsViaStandardEnvVar tests but this is known and can be addressed in a separate PR.

PR #5136, opened 2024-07-24
[chore] Upgrade core dependencies

dmitryax:
Update to 0.105.0 version + cherry pick https://github.com/open-telemetry/opentelemetry-collector/pull/10716

Also, disabling both confmap upstream feature gates for now. Enabling `confmap.unifyEnvVarExpansion` doesn't have any effect other than introducing a bug. The warning messages that are expected to show with the FG are ignored in our repo

We have duplicated functionality in our distro which needs to be deprecated first. Then we can remove it and align with upstream.


tests/general/testdata/incompat_env_config_source_labels_expected.yaml
```diff
@@ -30,13 +30,13 @@ resourceMetrics:
                         stringValue: prefix-$$$$an-envvar-value
                     - key: octuple-dollar
                       value:
-                        stringValue: $$$an-envvar-value-suffix
+                        stringValue: $$$${env:AN_ENVVAR}-suffix
```
dmitryax, line 33: Apparently, we have been expanding/resolving such strings twice after https://github.com/signalfx/splunk-otel-collector/pull/2893. And this test validates wrong behavior. The issue isn't fixed by this PR. However, https://github.com/open-telemetry/opentelemetry-collector/pull/10560 blocks the second expansion if we have an even number of $ after the first expansion. 

I'm updating the tests because it's a step in the right direction. In the meantime, I'm looking into the bigger problem.

atoulme (approved):

tests/general/testdata/incompat_env_config_source_labels_expected.yaml
jvoravong: Good work finding this.

PR #5137, opened 2024-07-24
gitlab: scan docker image

jeffreyc-splunk:
Add vulnerability scanning to the gitlab pipeline for the collector docker image (the `/prodsec/.container-scan.yml` template currently only supports `linux/amd64` images).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,6 +26,7 @@ include:
     file:
       - '/prodsec/.oss-scan.yml'
       - '/prodsec/.binary-scan.yml'
+      - '/prodsec/.container-scan.yml'
   - project: 'core-ee/signing/api-integration'
     ref: develop
     file: '/templates/.sign-client.yml'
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -1271,6 +1272,26 @@ sign-multiarch-manifest:
...
+xray-scan-docker:
+  only:
+    - main
+  except:
+    - schedules
+  extends: .container-scan
+  stage: xray-scan
+  parallel:
+    matrix:
+      # - CHILD: ["amd64", "arm64", "ppc64le", "2019", "2022"]
```
pjanotti, line 1284: nit: I think the comment on the actual like is enough

pjanotti (approved):
LGTM - only an optional suggestion

.gitlab-ci.yml
```diff
@@ -1271,6 +1272,26 @@ sign-multiarch-manifest:
...
+      - CHILD: ["amd64"]  # the .container-scan template currently only supports linux/amd64 images
+  needs:
+    - push-multiarch-manifest
+    - sign-multiarch-manifest
+  variables:
+    create_jira: "true"
```
jinja2, line 1290: will the created jira be auto-assigned to our project?

.gitlab-ci.yml
jeffreyc-splunk: > nit: I think the comment on the line below is enough

Just wanted to explicitly show all child tags that we publish that we should scan whenever prodsec's template support them.

.gitlab-ci.yml
jeffreyc-splunk: > will the created jira be auto-assigned to our project?

The jira creation is handled by prodsec's template, but I believe this will be similar to the jira's created by the other scan jobs.

atoulme (approved):

PR #5140, opened 2024-07-26
Fix Collector methods in the tests utils

dmitryax:
When a receiver argument is passed by value to a method, mutating the argument doesn't affect the original copy. So, most of the methods like WithArgs don't actually work. This change fixes it, making the receiver argument passed by its pointer.


dmitryax:
This is not what broke for me actually. Submitted another PR https://github.com/signalfx/splunk-otel-collector/pull/5141

PR #5162, opened 2024-07-31
Fix default discovery crash on windows

pjanotti:
**Description:**
A bad default on the OTel `docker_host` (to be fixed upstream) exposed that if a observer extension fails to `Start` the discovery code still calls `ListAndWatch` what causes a panic.

Details: the Start of the observer is called via internal/confmapprovider/discovery, but if Start fails it logs a warning and proceeds with the execution. Later internal/receiver/discoveryreceiver uses the host.GetExtensions()  and the observer is there since it was created successfully (it did fail to start in a step that the host doesn't have visibility). The extension is cast to an observer.Observable and a go routine calling ListAndWatch is launched, but, then the private that it needs to run are nil because the start failed and then panic.

Right now I am removing the `docker_observer` from Windows, but, it would be re-enabled as soon as we have it fixed on upstream.

**Jira:** 
https://splunk.atlassian.net/browse/OTL-2883

**Testing:**
Added 2 new tests and validated manually.

**Documentation:**
N/A


pjanotti:
Perhaps that some intermittent error that we got on Windows runners when some ports that we use are already taken? Not sure at this point... kicking it again: https://github.com/signalfx/splunk-otel-collector/actions/runs/10184803130/job/28172948607?pr=5162#step:5:386

pjanotti:
The port failure is not a Windows only thing it also affects mac, fixing via https://github.com/signalfx/splunk-otel-collector/pull/5168

pjanotti:
https://github.com/signalfx/splunk-otel-collector/pull/5168 was not enough... also failed on Ubuntu, it may need a refactor of the actual type.

cmd/otelcol/main.go
```diff
@@ -80,14 +85,21 @@ func main() {
 		},
 	}
 
-	os.Args = append(os.Args[:1], collectorSettings.ColCoreArgs()...)
+	allArgs := args[:1]
+	allArgs = append(allArgs, collectorSettings.ColCoreArgs()...)
+	os.Args = allArgs
```
dmitryax, line 90: why this change? can't see a difference

cmd/otelcol/main.go
```diff
@@ -36,10 +37,14 @@ import (
 )
 
 func main() {
+	runFromCmdLine(os.Args)
+}
+
+func runFromCmdLine(args []string) {
```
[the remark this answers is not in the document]
pjanotti: Testability, this way I can call `runFromCmdLine` directly from tests.

cmd/otelcol/main.go
pjanotti: Lint doesn't like `append` that doesn't assign resulting slice to the first parameter so logically what I want is `os.Args = append(args[:1], collectorSettings.ColCoreArgs()...)`, but, lint is not happy that `args` is not on the left-hand side.

pjanotti:
Working to fix the test failure via https://github.com/signalfx/splunk-otel-collector/pull/5189

pjanotti:
Still having issues with the new test affecting other tests on the same test executable, switching to draft until that is resolved.

pjanotti:
I will create a new PR to do the actual fix after https://github.com/signalfx/splunk-otel-collector/pull/5201 is merged. This will reduce the review scope and will make the history better.

PR #5166, opened 2024-08-01
Deprecate Apache monitor

atoulme:
**Description:**
Mark the apache monitor deprecated and to be removed in a future release. The apache receiver should be used instead.


internal/signalfx-agent/pkg/monitors/collectd/apache/apache.go
```diff
@@ -6,6 +6,7 @@ package apache
 //go:generate ../../../../scripts/collectd-template-to-go apache.tmpl
 
 import (
+	"fmt"
```
atoulme, line 9:
```suggestion
	"fmt"

```

crobert-1 (approved):

PR #5175, opened 2024-08-02
Fix config_server_test flakiness

pjanotti:
The test flakiness is due to the release of the port used by the ConfigServer not being deterministic since it happens in a goroutine, not during the calls to OnShutdown.

Made the test reliable by actually checking if the port is available.


pjanotti:
I'm opting to change the test to match the actual behavior of the type, but, in general we should make resource release deterministic. At least now this non-determinism is being documented.

atoulme (approved):

PR #5191, opened 2024-08-06
[recveiver/discovery] [chore] Remove redundant config.Validate() call

dmitryax:
The `Validate` method is implicitly called by the config resolver, no need to call it in the constructor.

### DIFF
--- a/internal/receiver/discoveryreceiver/factory_test.go
+++ b/internal/receiver/discoveryreceiver/factory_test.go
@@ -23,7 +23,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/collector/component/componenttest"
 	"go.opentelemetry.io/collector/consumer/consumertest"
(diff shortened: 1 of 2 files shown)


internal/receiver/discoveryreceiver/factory_test.go
```diff
@@ -41,19 +41,18 @@ func TestCreateLogsReceiver(t *testing.T) {
 	factory := NewFactory()
 	cfg := factory.CreateDefaultConfig()
 
-	params := otelcolreceiver.Settings{}
+	params := receivertest.NewNopSettings()
 	receiver, err := factory.CreateLogsReceiver(context.Background(), params, cfg, consumertest.NewNop())
-	assert.Error(t, err)
-	assert.EqualError(t, err, "`watch_observers` must be defined and include at least one configured observer extension")
```
dmitryax, line 47: This case is extensively tested in `config_test.go`

pjanotti (approved):

atoulme (approved):

PR #5198, opened 2024-08-07
change template to use long form of discovery

hughesjj:
I always forget what the short hand identifiers mean


pjanotti (approved):

atoulme (approved):

atoulme:
> The base branch requires all commits to be signed

PR #5201, opened 2024-08-08
Make ConfigServer OnShutdown close the HTTP server deterministically

pjanotti:
Although not strictly necessarily a bug, making the release of the HTTP server to happen during the last call to `OnShutdown` makes any reasoning about availability of the corresponding port in other tests much easier. See all the struggles and debug at #5162

Made the respective tests check that the port is released after the "last" call to `OnShutdown`.

Also added the default `ConfigServer` port to the set of ports that shouldn't be blocked by Windows (hit this a few times oh Windows GH runners).


internal/configconverter/config_server.go
```diff
@@ -156,26 +159,21 @@ func (cs *ConfigServer) start() {
 
 			go func() {
 				httpErr := cs.server.Serve(listener)
+				defer cs.serverShutdown.Done()
 				if httpErr != http.ErrServerClosed {
```
atoulme, line 163: can httpErr be nil?

atoulme (approved):

internal/configconverter/config_server.go
pjanotti: From the docs: `Serve` always returns a non-nil error and closes l. After Shutdown or Close, the returned error is `ErrServerClosed`.

PR #5202, opened 2024-08-08
Fix default discovery crash on Windows

pjanotti:
Replaces #5162

**Description:**
A bad default on the OTel `docker_host` (just fixed upstream, see https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34358) exposed that if a observer extension fails to `Start` the discovery code still calls `ListAndWatch` what causes a panic.

Details: the Start of the observer is called via internal/confmapprovider/discovery, but if Start fails it logs a warning and proceeds with the execution. Later internal/receiver/discoveryreceiver uses the host.GetExtensions()  and the observer is there since it was created successfully (it did fail to start in a step that the host doesn't have visibility). The extension is cast to an observer.Observable and a go routine calling ListAndWatch is launched, but, then the private that it needs to run are nil because the start failed and then panic.

Right now I am removing the `docker_observer` from Windows, but, it would be re-enabled as soon as we have it fixed on upstream.

**Jira:** 
https://splunk.atlassian.net/browse/OTL-2883

**Testing:**
Added 2 new tests and validated manually.

**Documentation:**
N/A


cmd/otelcol/main.go
```diff
@@ -36,10 +37,14 @@ import (
 )
 
 func main() {
+	runFromCmdLine(os.Args)
```
pjanotti, line 40: This extra function is to make the code easily callable from tests without having to touch `os.Args` too much, notice that the actual code in fact already changes `os.Args`, but, I don't want to do that here when we can pass the arguments as a slice in the tests.

samiura: sounds good

cmd/otelcol/main_test.go
```diff
@@ -14,3 +14,111 @@
...
+			name:    "gateway",
+			args:    []string{"otelcol", "--config=config/collector/gateway_config.yaml"},
+			timeout: 15 * time.Second,
+		},
+		{
+			name:    "default_discovery",
+			args:    []string{"otelcol", "--discovery", "--config=config/collector/agent_config.yaml"},
+			timeout: 30 * time.Second,
+		},
+		// Running the discovery with --dry-run in CI is not desirable because of the use of os.Exit(0) to end the execution.
```
samiura, line 53: --dry-run was originally intended as a helper/utility  for the field/sales to capture what services are available  but design requirements have changed. I wonder we need this test at all. But we can keep it like this. I do not have strong feelings either way.

pjanotti: Got it, I will ask around to see if the need/purpose is clear.

cmd/otelcol/main_test.go
```diff
@@ -14,3 +14,111 @@
...
+			testCtx, cancel := context.WithTimeout(context.Background(), tt.timeout)
+			defer cancel()
+
+			otelcolCmdTestCtx = testCtx
+			defer func() {
+				otelcolCmdTestCtx = nil
+			}()
+
+			// Wait for the ConfigServer to be down after the test.
+			defer waitForPort(t, "55554")
```
samiura, line 102: excellent idea to check we recovered ports after or not. One caution though, sometimes it take a bit of time, not sure if we should sleep  a bit on the helper function to give it some time to recover.

internal/confmapprovider/discovery/discoverer.go
```diff
@@ -449,11 +460,15 @@ func (d *discoverer) updateReceiverForObserver(receiverID component.ID, receiver
 
 func factoryForObserverType(extType component.Type) (otelcolextension.Factory, error) {
 	factories := map[component.Type]otelcolextension.Factory{
-		component.MustNewType("docker_observer"):   dockerobserver.NewFactory(),
```
samiura, line 452: Just curious why are we removing it?

pjanotti: It is not being added by default on Windows, this is temporary until next OTel deps upgrade, since then we will have the docker observer fixed for Windows.

cmd/otelcol/main_test.go
pjanotti: `waitForPort` by default waits 60 seconds for the port to be available, see code below.

pjanotti: That said, with #5201 the port is expected to be released whenever run finished (except for `--dry-run` as commented in the test). Let me update that to put my fix to the test 😁

pjanotti: Actually, I shouldn't do that because other tests can take the port and the check will fail and this will become flaky.

samiura: sounds good. Let me know when you're done.

internal/confmapprovider/discovery/discoverer.go
samiura: Oh, I forgot the context! These changes are only for Windows binary, correct?

samiura: Then it is fine. I can go ahead with LGTM

samiura (approved):

cmd/otelcol/main_test.go
pjanotti: Keeping as it is.

PR #5206, opened 2024-08-09
Rely on upstream confmap implementation to expand env variables

dmitryax:
After [recent changes](https://github.com/open-telemetry/opentelemetry-collector/pull/10508) in the collector core, we can rely on the upstream to expand environmental variables along with $$ escaping. Unfortunately, we cannot do that for the `config_sources` section as it's custom for our distro. So we need to split the behavior for a while to unblock the 0.107.0 release.

This change also fixes the `$$` expansion bug. See the corrected test data for `TestExpandedDollarSignsViaEnvConfigSource`, which previously validate invalid behavior.

Later, we can refactor it further and migrate the config sources to the config providers to be handled upstream, significantly reducing the amount of code we maintain.


internal/configsource/source_test.go
```diff
@@ -367,46 +367,6 @@ func TestConfigSourceManagerMultipleWatchForUpdate(t *testing.T) {
...
-	require.NoError(t, err)
-
-	res, closeFunc, err := ResolveWithConfigSources(context.Background(), map[string]ConfigSource{"tstcfgsrc": &tstCfgSrc}, nil, cp, func(_ *confmap.ChangeEvent) {
-		panic("must not be called")
-	})
-	require.NoError(t, err)
-	assert.Equal(t, expectedParser.ToStringMap(), res.ToStringMap())
-	assert.NoError(t, callClose(closeFunc))
-}
-
```
dmitryax, line 409: The test has been removed since the expansion was disabled in favor of the upstream implementation. e2e functionality is covered in the integration tests

samiura (approved):

PR #5246, opened 2024-08-16
Update core deps to include config parsing bug fix

crobert-1:


crobert-1:
We're going to need to update `contrib` as well, I'll do this in another broader PR.

PR #5272, opened 2024-08-21
Remove redundant check

atoulme:
Fix one of the issues reported in https://github.com/signalfx/splunk-otel-collector/pull/5268

### DIFF
--- a/internal/receiver/discoveryreceiver/config.go
+++ b/internal/receiver/discoveryreceiver/config.go
@@ -103,7 +103,7 @@ func (cfg *Config) Validate() error {
 		}
 	}
 
-	if cfg.WatchObservers == nil {
+	if len(cfg.WatchObservers) == 0 {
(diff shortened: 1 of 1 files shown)


internal/receiver/discoveryreceiver/config.go
```diff
@@ -103,7 +103,7 @@ func (cfg *Config) Validate() error {
 		}
 	}
 
-	if cfg.WatchObservers == nil || len(cfg.WatchObservers) == 0 {
+	if cfg.WatchObservers == nil {
```
crobert-1, line 106:
```suggestion
	if len(cfg.WatchObservers) == 0 {
```

From my understanding, I don't _think_ it matters here which condition we take, but I'd slightly favor the length check more than nil, which is also what has been [done upstream](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34724/files). I'd be interested to know though if there's a reason/preference to do the `nil` check instead.

crobert-1 (approved):
Nit, feel free to drop.

PR #5278, opened 2024-08-23
Remove golang lint suppression for int overflow

pjanotti:
Removing lint suppression for G115: int overflow.

Fixed or suppressed the 3 locations outside `internal/signalfx-agent/pkg/monitors`, I will call these out via PR comments. For the `internal/signalfx-agent/pkg/monitors` I'm suppressing them line by line per Slack conversation, see https://splunk.slack.com/archives/G01L5S7BKLY/p1724431826011759?thread_ts=1724358192.342099&cid=G01L5S7BKLY

### DIFF
--- a/internal/confmapprovider/discovery/properties/property.go
+++ b/internal/confmapprovider/discovery/properties/property.go
@@ -222,7 +222,7 @@ func unwordify(text string) (string, error) {
 		for len(decoded) < 4 {
 			decoded = append([]byte{0}, decoded...)
 		}
-		r := int32(binary.BigEndian.Uint32(decoded))
+		r := binary.BigEndian.Uint32(decoded)
 		return fmt.Sprintf("%c", r)
 	})
 	if err != nil {
--- a/internal/receiver/signalfxgatewayprometheusremotewritereceiver/mock_reporter_test.go
+++ b/internal/receiver/signalfxgatewayprometheusremotewritereceiver/mock_reporter_test.go
@@ -73,7 +73,7 @@ func (m *mockReporter) OnError(_ context.Context, errorLocation string, err erro
 }
 
 func (m *mockReporter) OnMetricsProcessed(_ context.Context, numReceivedMessages int, _ error) {
-	m.TotalSuccessMetrics.Add(int32(numReceivedMessages))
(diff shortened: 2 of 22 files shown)


internal/confmapprovider/discovery/properties/property.go
```diff
@@ -222,7 +222,7 @@ func unwordify(text string) (string, error) {
 		for len(decoded) < 4 {
 			decoded = append([]byte{0}, decoded...)
 		}
-		r := int32(binary.BigEndian.Uint32(decoded))
+		r := binary.BigEndian.Uint32(decoded)
```
pjanotti, line 225: Internally `"%c"` casts everything to `uint64` and doesn't care for signal, so it is safe to remove the cast to `int32`.

internal/receiver/signalfxgatewayprometheusremotewritereceiver/mock_reporter_test.go
```diff
@@ -73,7 +73,7 @@ func (m *mockReporter) OnError(_ context.Context, errorLocation string, err erro
 }
 
 func (m *mockReporter) OnMetricsProcessed(_ context.Context, numReceivedMessages int, _ error) {
-	m.TotalSuccessMetrics.Add(int32(numReceivedMessages))
+	m.TotalSuccessMetrics.Add(int32(numReceivedMessages)) //nolint:gosec
```
pjanotti, line 76: This is in test code.

@hughesjj the field is not actually used anywhere I can remove it, in a separate PR, unless you have some reason to keep it.

tests/testutils/endpoint.go
```diff
@@ -78,8 +79,9 @@ func GetAvailablePort(t testing.TB) uint16 {
 
 	portInt, err := strconv.Atoi(port)
 	require.NoError(t, err)
-
-	return uint16(portInt)
+	require.Greater(t, portInt, -1)
+	require.Less(t, portInt, math.MaxUint16+1)
+	return uint16(portInt) //nolint:gosec
```
pjanotti, line 84: This is in test code. Made safe to ignore the lint on this line by checking if the `int` is on the range of a "good" cast to `uint64`.

crobert-1 (approved):

atoulme (approved):

PR #5316, opened 2024-09-02
[chore] Skip flaky mongodb discovery tests

dmitryax:
Tests for mongo 6 and 7 are flaky. Skipping for now. Reported a ticket to fix later


atoulme:
Some of the docker repositories were not public, so the image would be rebuilt and subject to docker hub limits every time. It should be resolved.

atoulme (approved):
Still flaky, approved

PR #5324, opened 2024-09-05
Disable OOTB Apache discovery

pjanotti:
**Description:** 
Disabling it since it requires configuration and doesn't work OOTB.

**Link to Splunk idea:**
https://splunk.atlassian.net/browse/OTL-2992

**Testing:**
Manually validated that it is still possible to perform Apache discovery given a correct configuration.

**Documentation:**
Not listing on changelog since it wasn't work anyway.


crobert-1 (approved):

dmitryax:
Can we update the default config to avoid crashing and make the discovery work for at least partially?

pjanotti:
> Can we update the default config to avoid crashing and make the discovery work at least partially?

Let's do that in a separate PR - I still have to try the other observers to see the proper configuration in this case.

dmitryax:
Sounds good

dmitryax (approved):

PR #5327, opened 2024-09-05
DNM: Test the need on CI for use_host_bindings

pjanotti:
Testing this settings since my local run is not really equivalent to the test itself.


pjanotti:
Test passed as expected, keeping the code as it is for the time being. Related to #5326

PR #5353, opened 2024-09-09
Re-enable OOTB discovery for Apache Web Server

pjanotti:
**Description:**
Re-enable OOTB discovery for Apache Web Server with a default endpoint that matches the one from the OpenTelemetry `apachereceiver`: ```http://`endpoint`/server-status?auto```

_Collateral changes:_
- Consolidate Apache docker test image. There is no need to use 2 separate images for testing the legacy SmartAgent monitor and the `apachereceiver` discovery.
- Renamed some files and folders that were using `apachewebserver` to `apache`. While the former is more descriptive it doesn't follow the pattern of other receivers from OpenTelemetry that can be discovered.

**Link to Splunk idea:**
https://splunk.atlassian.net/browse/OTL-2992

**Testing:**
Manually verified on Windows.

**Documentation:**
Updated changelog - need to follow-up with docs team if this needs to be updated on docs website.


CHANGELOG.md
```diff
@@ -10,6 +10,11 @@
 
 - (Splunk) Add `apachespark` receiver ([#5318](https://github.com/signalfx/splunk-otel-collector/pull/5318))
 
+### 💡 Enhancements 💡
+
+- (Splunk) Apache Web Server Auto Discovery: set the default discovered endpoint to match the OpenTelemetry `apachereceiver` default: `http://`endpoint`/server-status?auto` ([#5353](https://github.com/signalfx/splunk-otel-collector/pull/5353))
+  If the collector is running as a process on the host OS and the Apache Web Server is in a Docker container add `--set=splunk.discovery.extensions.docker_observer.config.use_host_bindings=true` to the command-line arguments for the discovery to create the correct endpoint.
```
samiura, line 16: These are already know facts [here](https://github.com/signalfx/splunk-otel-collector/tree/main/internal/confmapprovider/discovery). I meant to say it holds for all discovery apps/services/dbs. May be we can skip linke16

docker/apache/status.conf
```diff
@@ -1,4 +1,4 @@
 ExtendedStatus on
-<Location /mod_status>
```
samiura, line 2: not sure what ` mod_status` as have not played with Apache web server for a while but will go with your way.

samiura (approved):

docker/apache/status.conf
pjanotti: That is the path that one would use on the endpoint for the Apache receiver. The OTel receiver and most of the samples defaults to `server-status`, confusingly the name of the module that provides the functionality is actually `mod_status`, see https://httpd.apache.org/docs/2.4/mod/mod_status.html

CHANGELOG.md
pjanotti: Good point, unfortunately, `use_host_bindings` is not explicitly listed there. To be fair I didn't read that doc carefully, but, I will add a task to review and update the doc, I think that there are a few things that we should be more explicit about.

PR #5355, opened 2024-09-10
Add nop receiver and exporter

atoulme:
**Description:**
Adds nop exporter and receiver to the distribution.


crobert-1 (approved):

atoulme:
Looks like updating to 0.109.0 first is necessary.

crobert-1:
> Looks like updating to 0.109.0 first is necessary.

Was there a reason some of the dependencies were updated to `v0.109.0` in this PR already? Was there some failure in leaving imports at `v0.108.0`? 

I'm seeing some panics happening that look related to the upgrade you're referencing, but the panics are happening **with** some of the `v0.109.0` upgrades.

atoulme:
No, just anticipating that this might land after we land the upgrade to 0.109.0.

dmitryax (approved):

PR #5363, opened 2024-09-13
Continuous service discovery

dmitryax:
This change introduces a new way of how the discovery mechanism works. Instead of observing data for the first 10 seconds and creating a static `receiver_cretor` config from that, it'll be running discovery receiver watching for all supported receivers. Once a new endpoint for any supported service is available, the collector will try to fetch metrics from it. If the collector cannot get metrics from a new service right away, it'll send detailed information about what can be done to enable it to the backend, which will be reflected in the newly discovered services UI. Any services that are successfully discovered will also be reflected in the UI. The collector sends entity events as OTLP logs to deliver information about the discovered services.

The new functionality can be enabled with "splunk.continuousDiscovery" feature gate by adding `--feature-gates=splunk.continuousDiscovery` argument to the `otecol` command.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -129,6 +129,7 @@ require (
 	go.etcd.io/etcd/client/v2 v2.305.15
 	go.opentelemetry.io/collector/component/componentstatus v0.109.0
 	go.opentelemetry.io/collector/config/confighttp v0.109.0
+	go.opentelemetry.io/collector/config/configopaque v1.15.0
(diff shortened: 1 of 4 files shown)


atoulme (approved):

atoulme:
Changes look ok but lint disagrees, please take a look.

dmitryax:
> Changes look ok but lint disagrees, please take a look.

Oh right, I forgot to remove a function that was split

PR #5379, opened 2024-09-17
Support filters and pdata being emitted from monitors

atoulme:
**Description:**
This adds the ability for the monitors to emit pdata. 

A simplified cut of #4243, it no longer is applying the changes to the prometheus monitor.


internal/signalfx-agent/pkg/core/dpfilters/overridable.go
```diff
@@ -47,3 +48,26 @@ func (f *overridableDatapointFilter) Matches(dp *datapoint.Datapoint) bool {
...
+	if f.metricFilter != nil && !f.metricFilter.Matches(m.Name()) {
+		return false
+	}
+	if f.dimFilter != nil {
+		switch m.Type() {
+		case pmetric.MetricTypeGauge:
+			if m.Gauge().DataPoints().Len() == 0 {
+				return false
+			}
+			return f.dimFilter.MatchesMap(m.Gauge().DataPoints().At(0).Attributes())
```
dmitryax, line 62: I think we should filter out data points from the metric instead of dropping the whole metric if the first datapoint matches the filter. What if the second of third data point matches it instead of the first?

atoulme: Thanks I will change the logic.

pkg/receiver/smartagentreceiver/output.go
```diff
@@ -208,6 +209,41 @@ func (out *output) Copy() types.Output {
 	return &cp
 }
 
+func (out *output) SendMetrics(metrics ...pmetric.Metric) {
```
dmitryax, line 212: Is this function wired to the code? I don't see it being used... It's unclear what the scope of the PR is. If this only adds filtering capability without using it, this PR must be unrelated I guess.

atoulme: It's not used yet, this PR is a smaller take on https://github.com/signalfx/splunk-otel-collector/pull/4243 which wires this function into the prometheus monitor.

PR #5381, opened 2024-09-17
[discovery] Move continuous discovery config parts to static files

dmitryax:
We cannot assume that ${SPLUNK_ACCESS_TOKEN} and ${SPLUNK_REALM} are aways set. It's not the case for the helm chart. We should move static config part of continuous discovery from discoverer to the config files. Having it in the `agent_config.yaml` has no overhead until the continuous discovery is enabled.

### DIFF
--- a/internal/configconverter/discovery.go
+++ b/internal/configconverter/discovery.go
@@ -17,6 +17,7 @@ package configconverter
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"go.opentelemetry.io/collector/confmap"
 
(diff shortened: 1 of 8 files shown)


internal/configconverter/discovery.go
```diff
@@ -158,6 +161,25 @@ func getMetricsPipelineAndReceivers(service map[string]any) (map[string]any, []a
 	return metricsPipeline, metricsReceivers, nil
 }
 
+func setEntitiesPipelineReceivers(pipelines map[string]any, discoReceivers []any) {
+	ep, ok := pipelines["logs/entities"].(map[string]any)
+	if !ok {
+		// Entities pipeline not set, nothing to do.
```
atoulme, line 167: would a log statement make sense here?

atoulme (approved):
neat use of nop, lgtm

internal/configconverter/discovery.go
dmitryax: Yeah makes sense. But we currently don't pass a logger here. I'll address that separately

PR #5386, opened 2024-09-19
Choco upgrade should preserve collector service custom env vars

pjanotti:
**Description:**
Choco upgrade should preserve custom environment variables set by the user and should not fail if env vars contain double-quotes.

**Link to Splunk idea:**
- https://splunk.atlassian.net/browse/OTL-3041
- Follow-up to https://splunk.atlassian.net/browse/OTL-3017

**Testing:**
- Improved current test to include a "good env var" to be preserved and another "good env var" that needs to be escaped when passed in the invocation of the MSI installed.
- Manually testing the modified test via #5388

**Documentation:**
N/A

### DIFF
--- a/.github/workflows/win-package-test.yml
+++ b/.github/workflows/win-package-test.yml
@@ -379,7 +379,7 @@ jobs:
             }
             $env_vars = Get-ItemPropertyValue -Path "HKLM:\SYSTEM\CurrentControlSet\Services\splunk-otel-collector" -Name "Environment"
             $test_env_vars = @(
-              "MY_VAR_WITH_SPACES=my value",
+              "MY VAR WITH SPACES=my value",
               "GOOD_ENV_VAR_NAME=good value",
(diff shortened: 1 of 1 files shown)


.github/workflows/win-package-test.yml
```diff
@@ -378,7 +378,12 @@ jobs:
               throw "choco install failed!"
             }
             $env_vars = Get-ItemPropertyValue -Path "HKLM:\SYSTEM\CurrentControlSet\Services\splunk-otel-collector" -Name "Environment"
-            $env_vars += "MY VAR WITH SPACES=my value"
+            $test_env_vars = @(
+              "MY_VAR_WITH_SPACES=my value",
```
jeffreyc-splunk, line 382: Should we keep/add a test for var names with spaces?
```suggestion
              "MY VAR WITH SPACES=my value",
```

pjanotti: Yes, can I blame Copilot? :)

jeffreyc-splunk (approved):

PR #5387, opened 2024-09-19
Draft: add converter to update batch processor

asreehari-splunk:
when sapm exporter enables token_access_passthrough use this converter to add include_metadata with the right header key

### DIFF
--- a/internal/configconverter/update_batchproc_on_token_passthrough.go
+++ b/internal/configconverter/update_batchproc_on_token_passthrough.go
@@ -1,0 +1,52 @@
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configconverter
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/collector/confmap"
+)
+
+func UpdateBatchProcOnTokenPassthrough(_ context.Context, in *confmap.Conf) error {
+	if in == nil {
+		return nil
+	}
+
+	out := map[string]any{}
+
+	// Check for sapm receiver with include_metadata set to true
+	if in.IsSet("receivers::sapm::include_metadata") {
+		if accessTokenPassthrough, ok := in.Get("receivers::sapm::include_metadata").(bool); ok && accessTokenPassthrough {
+			// Add metadata_keys to batch processor
+			switch batchProcessor := in.Get("processors::batch").(type) {
+			case nil:
+				out["processors::batch"] = map[string]any{
+					"metadata_keys": []interface{}{"X-SF-Token"},
+				}
+			case map[string]interface{}:
+				batchProcessor["metadata_keys"] = []any{"X-SF-Token"}
+				out["processors::batch"] = batchProcessor
+			default:
+				return fmt.Errorf("unexpected type for processors::batch: %T", batchProcessor)
+			}
+		}
+	}
+
+	// Merge the modified configuration back into the original Conf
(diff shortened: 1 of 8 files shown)


internal/settings/settings_test.go
```diff
@@ -337,12 +337,18 @@ func TestSetDefaultEnvVarsSetsInterfaceFromConfigOption(t *testing.T) {
...
+			"another-gate",
+		},
+		{
+			"--feature-gates", "some-gate", "--feature-gates", "+telemetry.useOtelForInternalMetrics",
+			"--feature-gates", "another-gate",
+		},
+		{
+			"--feature-gates", "some-gate", "--feature-gates", "-telemetry.useOtelForInternalMetrics",
+			"--feature-gates", "another-gate",
+		},
```
dmitryax, line 351: Please avoid unrelated changes

internal/configconverter/update_batchproc_on_token_passthrough.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"go.opentelemetry.io/collector/confmap"
+)
+
+func UpdateBatchProcOnTokenPassthrough(_ context.Context, in *confmap.Conf) error {
+	if in == nil {
+		return nil
+	}
+
+	out := map[string]interface{}{}
```
dmitryax, line 43: nit: same for other places
```suggestion
	out := map[string]any{}
```

samiura (commented):
:LGTM

dmitryax (commented):
Please update the default config file `agent_config.yaml` as well

internal/configconverter/update_batchproc_on_token_passthrough.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// Check for sapm receiver with access_token_passthrough
+	if in.IsSet("receivers::sapm::access_token_passthrough") && in.Get("receivers::sapm::access_token_passthrough").(bool) {
```
pjanotti, line 46: I'm assuming that if there is a typo on the value of `access_token_passthrough`, e.g.: `truue`, this code is never reached, is that correct?

Can we guarantee that the type assertion never fails here?

internal/configconverter/update_batchproc_on_token_passthrough_test.go
```diff
@@ -0,0 +1,68 @@
+// Copyright The OpenTelemetry Authors
```
pjanotti, line 1: Double license header - I wish addlicense caught those...

pjanotti (commented):
nit & personal preference: instead of `_w_` and `_wo_` on the test files I would rather use `_with_` and `_no_`, up to you.

internal/configconverter/update_batchproc_on_token_passthrough.go
asreehari-splunk: thats right. it won't execute this code. let me think about that

internal/configconverter/update_batchproc_on_token_passthrough.go
```diff
@@ -0,0 +1,52 @@
...
+
+func UpdateBatchProcOnTokenPassthrough(_ context.Context, in *confmap.Conf) error {
+	if in == nil {
+		return nil
+	}
+
+	out := map[string]any{}
+
+	// Check for sapm receiver with access_token_passthrough
+	if in.IsSet("receivers::sapm::access_token_passthrough") {
```
dmitryax, line 32:
```suggestion
	if in.IsSet("receivers::sapm::include_metadata") {
```

internal/configconverter/update_batchproc_on_token_passthrough.go
```diff
@@ -0,0 +1,52 @@
...
+func UpdateBatchProcOnTokenPassthrough(_ context.Context, in *confmap.Conf) error {
+	if in == nil {
+		return nil
+	}
+
+	out := map[string]any{}
+
+	// Check for sapm receiver with include_metadata set to true
+	if in.IsSet("receivers::sapm::include_metadata") {
+		if _, ok := in.Get("receivers::sapm::include_metadata").(bool); ok {
```
dmitryax, line 33: `ok` represents whether the value casts to booled. So it's going to be true even for `include_metadata: true`. That wasn't your intention, right? Please fix and add a test case

asreehari-splunk: makes sense. I'll update it

asreehari-splunk:
closing

PR #5388, opened 2024-09-19
DNM: Test the test from fix choco upgrade with custom env vars

pjanotti:
I want to be double-safe and test the test for this, related PR #5386


pjanotti:
Ok, the double-quote issue is making the upgrade test not to finish because the MSI tries to display a dialog box and there is no one to click on it... let me modify the code to test the env vars themselves

pjanotti:
Equivalent to the test above but with the correct test code https://github.com/signalfx/splunk-otel-collector/actions/runs/10947292400/job/30396219992#step:5:189

pjanotti:
ok, tests are working as intended

PR #5406, opened 2024-09-25
PLAYGROUND -- DO NOT MERGE -- 110 upgrade

hughesjj:
My tracking branch so we can blow away the other one but apply attempted fixes


internal/receiver/signalfxgatewayprometheusremotewritereceiver/metadata.yaml
```diff
@@ -1,10 +1,9 @@
 type: "signalfxgatewayprometheusremotewrite"
-github_project: signalfx/splunk-otel-collector
```
dmitryax, line 2: Why is this removed?

PR #5451, opened 2024-10-05
test for discovery logs in test

atoulme:
This simplifies our tests by moving our assertions in the log, use golden for expected logs, and remove all log support from testutils.


tests/receivers/discovery/host_observer_test.go
```diff
@@ -17,18 +17,92 @@
...
+	require.NoError(t, err)
+
+	foundFailure := false
+	foundSuccess := false
+
+	assert.Eventually(t, func() bool {
+		if tc.OTLPReceiverSink.LogRecordCount() == 0 {
+			if err == nil {
+				err = fmt.Errorf("no logs received")
+			}
```
dmitryax, line 54: Why are these lines needed?

tests/receivers/discovery/host_observer_test.go
```diff
@@ -17,18 +17,92 @@
...
+	tc := testutils.NewTestcase(t)
+	defer tc.PrintLogsOnFailure()
+	defer tc.ShutdownOTLPReceiverSink()
+
+	_, shutdown := tc.SplunkOtelCollector("host_observer_simple_prometheus_config.yaml")
+	defer shutdown()
+
+	failure, err := golden.ReadLogs(filepath.Join("testdata", "expected", "host_observer.yaml"))
+	require.NoError(t, err)
+	success, err := golden.ReadLogs(filepath.Join("testdata", "expected", "host_observer-connected.yaml"))
```
dmitryax, line 44: Why can't we keep them in the same file? I believe they are supposed to be sent in one batch.

tests/receivers/discovery/host_observer_test.go
atoulme, line 54:
```suggestion
```

tests/receivers/discovery/host_observer_test.go
atoulme: they are not sent in one batch, no. I think we don't use the batch processor as well.

tests/receivers/discovery/host_observer_test.go
atoulme: Sorry, old code remainder. Removed.

tests/receivers/discovery/host_observer_test.go
atoulme: here, Dmitrii defines one batch as one payload of data under pdata.Logs. That means one plog.Logs, containing one plog.ResourceLogs, containing one plog.ScopeLogs, containing two plog.LogRecords.

PR #5452, opened 2024-10-05
Refactor metric tests

atoulme:
**Description:**
Building on #5451, this refactors tests to use pdata checks to test the metrics.


tests/receivers/discovery/host_observer_test.go
```diff
@@ -48,6 +49,9 @@ func TestDiscoveryReceiverWithHostObserverAndSimplePrometheusReceiverProvideStat
 
 	assert.Eventually(t, func() bool {
 		if tc.OTLPReceiverSink.LogRecordCount() == 0 {
+			if err == nil {
+				err = fmt.Errorf("no logs received")
+			}
```
atoulme, line 54:
```suggestion
```
No that can go sorry.

tests/receivers/discovery/host_observer_test.go
```diff
@@ -17,6 +17,7 @@
 package tests
 
 import (
+	"fmt"
```
atoulme, line 20:
```suggestion
```

PR #5473, opened 2024-10-08
Properly use run_state token

aknarts:
**Description:** 
Chef does not explicitly merge node.default and node.run_state data.

The complicated bit here is the hec_token https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/chef/attributes/default.rb#L21

I cannot use run_state in attributes as it is empty at that point.
My solution is to have this line after I set the token but not sure if that needs to be specified somewhere
```
node.default['splunk_otel_collector']['splunk_hec_token'] = "#{node['splunk_otel_collector']['splunk_access_token']}"
```

**Testing:** 

Tested on our internal node, Linux only

**Documentation:** 
This option was already documented but not actually working

### DIFF
--- a/deployments/chef/recipes/collector_win_config_options.rb
+++ b/deployments/chef/recipes/collector_win_config_options.rb
@@ -2,7 +2,7 @@
 # Recipe:: collector_win_config_options
 
 collector_env_vars = [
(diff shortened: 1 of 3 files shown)


atoulme (approved):
LGTM.

PR #5483, opened 2024-10-11
Move packaging to its own top level folder

atoulme:
Move packaging outside, making internal dedicated to go code.


atoulme:
linkChecker is expected to fail on the link, since the link moves to point to the expected location of the resource in main branch after the merge.

PR #5505, opened 2024-10-15
[chore] add a smoke test for the TA

hughesjj:
it's building via make locally need to add more to the test (thinking about using yq to inject some exporter config to a file and mounting such to the docker compose)

### DIFF
--- a/.github/workflows/splunk-ta-otel.yml
+++ b/.github/workflows/splunk-ta-otel.yml
@@ -1,0 +1,62 @@
+name: splunk-ta-otel-build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
+      - '.github/workflows/splunk-ta-otel.yml'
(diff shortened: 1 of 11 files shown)


atoulme (approved):

.github/workflows/splunk-ta-otel.yml
```diff
@@ -0,0 +1,65 @@
...
+      - 'pkg/**'
+      - 'tests/**'
+      - '!**.md'
+
+concurrency:
+  group: splunk-technical-addon-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  runs-on: "ubuntu-20.04"
```
atoulme, line 22: The workflow is not valid. .github/workflows/splunk-ta-otel.yml (Line: 22, Col: 3): Unexpected value 'runs-on'

PR #5545, opened 2024-10-24
Use pdata for prometheus monitor

atoulme:


internal/signalfx-agent/pkg/monitors/prometheusexporter/conversion.go
```diff
@@ -96,27 +138,51 @@ func makeHistogramDatapoints(name string, ms []*dto.Metric) []*datapoint.Datapoi
...
-			dps = append(dps, sfxclient.Cumulative(name+"_count", dims, int64(h.GetSampleCount()))) //nolint:gosec
+			metric := pmetric.NewMetric()
+			metric.SetName(name + "_count")
+			sum := metric.SetEmptySum()
+			sum.SetIsMonotonic(true)
+			sum.SetAggregationTemporality(pmetric.AggregationTemporalityCumulative)
+			dp := sum.DataPoints().AppendEmpty()
+			dp.SetIntValue(int64(h.GetSampleCount())) //nolint:gosec // disable G115
+			_ = dp.Attributes().FromRaw(dims)
+			dps = append(dps, metric)
```
dmitryax, line 149: So, we are creating one metric with one data point for each Prometheus time series. Is that correct? I don't see any combining logic down the road...

dmitryax: Looks like the existing `sfxDatapointsToPDataMetrics` implementation does pretty much the same though...

dmitryax:
Do we have any tests to confirm that the output hasn't changed?

dmitryax (approved):
LGTM assuming we have tests covering the output

internal/signalfx-agent/pkg/monitors/prometheusexporter/conversion.go
atoulme: Yes, this is a simple follow through to make sure we get the same behavior for now.

atoulme:
Yes, we have multiple tests that check with golden files that the prometheus metrics match. Both for simple gauges and counters and histograms.

PR #5546, opened 2024-10-24
use pdata for sql data

atoulme:


pjanotti:
@atoulme - reopen if needed, closing for the time being.

PR #5548, opened 2024-10-25
[chore] Update docker dependancies

dmitryax:
https://github.com/hashicorp/go-secure-stdlib/pull/136 has been resolved and released as part of v0.4.1. Blocking docker dependencies from the upgrade is no longer needed.


atoulme (approved):

atoulme:
We need this for 0.112.0, thanks! However it looks like that breaks the docker monitor.

atoulme:
This is almost ready to merge, one lint issue with a deprecated API. Given that this blocks updating the collector, I'd like to merge this, and open a follow up to fix the lint issue in parallel to updating dependencies. Please let me know if that works?

PR #5553, opened 2024-10-25
Revert "Update jmx-metric-gatherer to latest (#5534)"

atoulme:
This reverts commit 5580753a03ec42e85fac58cc254c0df48a34c6d2.

The 0.112.0 PR indicates the version of the jmx-metrics-gatherer we merged in is still not supported, so reverting back to a known stable version.


dmitryax (approved):

dmitryax:
I don't like our conditional test jobs. I think we need to enable all the tests unconditionally. It'll allow us to set make more of them required

PR #5672, opened 2024-12-06
Deprecate SPLUNK_TRACE_URL env var and don't use it in default configs

dmitryax:
Deprecate `SPLUNK_TRACE_URL` environment variable and replace with `${SPLUNK_INGEST_URL}/v2/trace`in the default configs. `SPLUNK_TRACE_URL` is still automatically set in the binary computed from on `SPLUNK_REALM` or `SPLUNK_INGEST_URL` environment variables to not break existing configurations. However, it is recommended to
  update the configurations to use `${SPLUNK_INGEST_URL}/v2/trace` instead.

Also, packaging and mass deployment solutions were update to not provide the option to set the Trace URL since it's not being used anymore in the default configs.


asreehari-splunk (approved):
LGTM

PR #5705, opened 2024-12-13
Add the filestats receiver

pjanotti:
Adding the `filestats` receiver from contrib.


tests/receivers/filestats/testdata/all_expected_metrics.yaml
```diff
@@ -0,0 +1,61 @@
...
+              dataPoints:
+                - asInt: "3"
+                  startTimeUnixNano: "1000000"
+                  timeUnixNano: "2000000"
+            name: file.size
+            unit: b
+        scope:
+          name: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/filestatsreceiver
+          version: latest
+  - resource:
```
dmitryax, line 32: Looks like an extra redundant resource after this

dmitryax (approved):

tests/receivers/filestats/filestats_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"testing"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/pdatatest/pmetrictest"
+
+	"github.com/signalfx/splunk-otel-collector/tests/testutils"
+)
+
+func TestFileStatsIntegration(t *testing.T) {
```
dmitryax, line 27: There is a bit more sophisticated test upstream https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/filestatsreceiver/integration_test.go. Why do we need to bring it here?

pjanotti: Doh! Removing it

pjanotti: I think there may be another one or two that we can drop, I will look at those.

dmitryax (approved):

PR #5706, opened 2024-12-13
Expose internal metrics at default localhost:8888 address

dmitryax:
Expose internal metrics at default `localhost:8888` address instead of `${SPLUNK_LISTEN_INTERFACE}:8888`

This can be changed by changing default values in `service::telemetry::metrics::section`:
```yaml
service:
  telemetry:
    metrics:
      readers:
        - pull:
            exporter:
              prometheus:
                host: localhost
                port: 8888
```
This also removes a warning about deprecated `service::telemetry::metrics::address`.


cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -71,7 +71,7 @@ receivers:
       - job_name: 'otel-collector'
         scrape_interval: 10s
         static_configs:
-        - targets: ["${SPLUNK_LISTEN_INTERFACE}:8888"]
+        - targets: ["0.0.0.0:8888"]
```
jinja2, line 74:
```suggestion
        - targets: ["localhost:8888"]
```
I know 0.0.0.0 will route to localhost but using localhost explicitly kind of conveys that the listener is probably only bound to it. It's just a minor thing, so feel free to merge as-is.

jinja2 (approved):

cmd/otelcol/config/collector/agent_config.yaml
dmitryax: Yeah, I was thinking about it as well. Two reasons I decided against it:
1. It'd require updating all other configs for consistency (it's a lot of them)
2. I added a comment on how to expose the metrics under a different endpoint if needed. `0.0.0.0` would work with any interface they chose. Otherwise, they would need to change this part as well.

dmitryax: I'm actually not 100% sure about point 2, targeting `0.0.0.0` may not work with specific IP addresses, only with `localhost` and `0.0.0.0`, which is no different from targeting `localhost`. If that's the case, we can update all the configs from targeting `0.0.0.0` to `localhost` separately.

jinja2: I am mostly certain 0.0.0.0 as destination address only works if server is listening on localhost

PR #5735, opened 2024-12-19
[chore] Running lint on Windows

pjanotti:
Nothing serious here, but, indication that there is no CI running pkg/extension/smartagentextension tests on Windows. Anyway, just detected issues on test code.

### DIFF
--- a/tests/internal/discoverytest/discoverytest.go
+++ b/tests/internal/discoverytest/discoverytest.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !windows
+
 package discoverytest
 
 import (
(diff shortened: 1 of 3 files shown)


tests/internal/discoverytest/discoverytest.go
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !windows
```
dmitryax, line 15: Why can't we run it on Windows?

dmitryax (approved):

tests/internal/discoverytest/discoverytest.go
dmitryax: Nevermind, I see why

pjanotti: `syscall.Stat_t` is not available on Windows

PR #5759, opened 2025-01-06
Update go version to 1.22.10

atoulme:
Update the version of Go used for testing and packaging.


crobert-1 (approved):
LGTM.

Note: It looks like `1.22.10` is correctly interpreted whether there are quotes around it or not, it's not changed to `1.22.1` without quotes. It would be nice to have a uniform value for setting `GO_VERSION` in actions. This is a nit, just something I'm pointing out as a bit annoying.

PR #5764, opened 2025-01-07
Fix expectations of JMX test

atoulme:


crobert-1:
Can you provide more context behind why the existing test is incorrect? It looks like some units changed, more metrics are being sent now, and potentially other changes have happened upstream.

I'm clarifying to ensure these changes are expected and documented, not something that's going to break something for users.

atoulme:
The JMX receiver itself updated, and the test fails because of the jmx upgrade of the jmx-metrics-gatherer jar: https://github.com/signalfx/splunk-otel-collector/pull/5729

We unfortunately don't run all integration tests on it or we would have caught it then.

The jmx-metrics-gatherer jar is cut by the opentelemetry-java-project, and they have this item in their release notes:
https://github.com/open-telemetry/opentelemetry-java-contrib/releases/tag/v1.42.0

https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1591

It looked mostly that we updated units, and nothing else. What new metrics are we adding? That might be a compounding effect of missing updates for this jar for a version or two. I will review the diff.

jinja2:
We do run these tests on upgrades to jar version. The issue seems to be the order in which it was merged. This jar version is introduced in 0.116 of the collector and we merged the jar before jmxreceiver was updated to 0.116. The jar update MR had the[ tests failing](https://github.com/signalfx/splunk-otel-collector/actions/runs/12396586075/job/34605625005?pr=5729#step:16:663) due to the version not being supported in the receiver yet.

jinja2 (approved):

crobert-1:
Thanks for clarifying @atoulme, appreciate the information and clarification, LGTM 👍

PR #5772, opened 2025-01-10
Network interface for discovery config server

atoulme:
**Description:**
Make it possible to access config server over a different network interface so we can make assertions with docker services.

### DIFF
--- a/internal/configconverter/config_server.go
+++ b/internal/configconverter/config_server.go
@@ -37,12 +37,13 @@ import (
 )
 
 const (
-	configServerEnabledEnvVar   = "SPLUNK_DEBUG_CONFIG_SERVER"
-	configServerPortEnvVar      = "SPLUNK_DEBUG_CONFIG_SERVER_PORT"
-	defaultConfigServerPort     = "55554"
-	defaultConfigServerEndpoint = "localhost:" + defaultConfigServerPort
-	effectivePath               = "/debug/configz/effective"
-	initialPath                 = "/debug/configz/initial"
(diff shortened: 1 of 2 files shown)


internal/configconverter/config_server.go
```diff
@@ -37,12 +37,13 @@ import (
...
 const (
-	configServerEnabledEnvVar   = "SPLUNK_DEBUG_CONFIG_SERVER"
-	configServerPortEnvVar      = "SPLUNK_DEBUG_CONFIG_SERVER_PORT"
-	defaultConfigServerPort     = "55554"
-	defaultConfigServerEndpoint = "localhost:" + defaultConfigServerPort
-	effectivePath               = "/debug/configz/effective"
-	initialPath                 = "/debug/configz/initial"
+	configServerEnabledEnvVar           = "SPLUNK_DEBUG_CONFIG_SERVER"
+	configServerPortEnvVar              = "SPLUNK_DEBUG_CONFIG_SERVER_PORT"
+	configServerNetworkInterfaceEnvVar  = "SPLUNK_DEBUG_CONFIG_SERVER_NETWORK_INTERFACE"
```
jinja2, line 42: Should we update the readme/docs to mention this new env var option? I see we have a blurb in our readme about the config server https://github.com/signalfx/splunk-otel-collector/blob/main/README.md?plain=1#L157-L160. We should make a changelog entry also i think

atoulme: Right. This deserves more attention than I gave it. I only wanted this for testing purposes.

atoulme:
Closing, I got the test to pass by running it in CI and updating the expectations.

PR #5779, opened 2025-01-10
move more tests to use golden

atoulme:
Move more tests away from resource_metrics to use golden instead.


crobert-1 (approved):
Functionally looks good once CI is passing 👍  _Looks like at least the imports are not grouped properly._

atoulme:
More to it, looks like. Need to revisit. I think the number of metrics is different between my machine and CI.

PR #5780, opened 2025-01-11
Add envoy metrics discovery bundle

atoulme:
**Description:**
Adds a new discovery bundle targeting envoy proxy metrics.


tests/receivers/envoy/testdata/otlp_exporter.yaml
```diff
@@ -0,0 +1,28 @@
+#receivers:
```
pjanotti, line 1: If we change the template this can become out of date, any benefit in keeping it here?

pjanotti (approved):

tests/receivers/envoy/testdata/otlp_exporter.yaml
atoulme: I'll remove.

atoulme:
I tested this locally. It worked well enough on my machine.

cmd/otelcol/config/collector/config.d.linux/receivers/envoy.discovery.yaml
```diff
@@ -0,0 +1,39 @@
...
+# in discovery mode (--discovery) and is provided for reference or customization.   #
+# Please note that any changes made to this file will need to be reconciled during  #
+# upgrades of the Collector.                                                        #
+#####################################################################################
+# prometheus:
+#   enabled: true
+#   rule:
+#     docker_observer: type == "container" and any([name, image, command], {# matches "(?i)envoy"}) and not (command matches "splunk.discovery")
+#     host_observer: type == "hostport" and command matches "(?i)envoy" and not (command matches "splunk.discovery")
+#     k8s_observer: type == "port" and pod.name matches "(?i)envoy"
```
jinja2, line 14: Is the k8s rule intentionally matching just pod name instead of the broader rule used by docker_observer? Istio sidecars for e.g. have pod name like `istio-proxy`, the image name might have envoy in it. But maybe we should only match standalone envoy to startwith

atoulme: Right, you had a note about that. I think we can update the rule to add istio-proxy but then I really need a test showing this works for an istio mesh.

atoulme: I am tempted to make a separate integration test with a kind cluster, we deploy everything and we make sure it works. I might get into that next and tweak the rule based on that.

jinja2: I think we should merge this one as-is. And have a different item to officially support istio. It is not just the pod name; [default prom port used by istio](https://istio.io/latest/docs/ops/integrations/prometheus/#option-1-metrics-merging) is also not same as the one here. Don't know if it is supported in deiscover mode, but it would be better  getting the port/path from the prometheus.io annotations for istio (like we have in our helm chart).

jinja2 (approved):

jinja2:
Discover mode config test [TestDockerObserver](https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/discoverymode/docker_observer_discovery_test.go#L38) is failing

```  
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -49,3 +49,25 @@
        	            	    (string) (len=26) "receiver_creator/discovery": (map[string]interface {}) (len=2) {
        	            	-    (string) (len=9) "receivers": (map[string]interface {}) (len=2) {
        	            	+    (string) (len=9) "receivers": (map[string]interface {}) (len=3) {
        	            	+     (string) (len=10) "prometheus": (map[string]interface {}) (len=3) {
        	            	+      (string) (len=6) "config": (map[string]interface {}) (len=1) {
        	            	+       (string) (len=6) "config": (map[string]interface {}) (len=1) {
        	            	+        (string) (len=14) "scrape_configs": ([]interface {}) (len=1) {
        	            	+         (map[string]interface {}) (len=5) {
        	            	+          (string) (len=8) "job_name": (string) (len=5) "envoy",
        	            	+          (string) (len=22) "metric_relabel_configs": ([]interface {}) (len=1) {
        	            	+           <max depth reached>
        	            	+          },
        	            	+          (string) (len=12) "metrics_path": (string) (len=17) "/stats/prometheus",
        	            	+          (string) (len=15) "scrape_interval": (string) (len=3) "10s",
        	            	+          (string) (len=14) "static_configs": ([]interface {}) (len=1) {
        	            	+           <max depth reached>
        	            	+          }
        	            	+         }
        	            	+        }
        	            	+       }
        	            	+      },
        	            	+      (string) (len=19) "resource_attributes": (map[string]interface {}) {
        	            	+      },
        	            	+      (string) (len=4) "rule": (string) (len=121) "type == \"container\" and any([name, image, command], {# matches \"(?i)envoy\"}) and not (command matches \"splunk.discovery\")"
        	            	+     },
        	            	      (string) (len=17) "prometheus_simple": (map[string]interface {}) (len=3) {
        	Test:       	TestDockerObserver
 ```

PR #5796, opened 2025-01-13
do not use resource_metrics for configproviders tests

atoulme:


tests/general/configproviders/env_test.go
```diff
@@ -41,14 +45,38 @@ service:
...
+
+	assert.EventuallyWithT(t, func(tt *assert.CollectT) {
+		for i := 0; i < len(tc.OTLPReceiverSink.AllMetrics()); i++ {
+			m := tc.OTLPReceiverSink.AllMetrics()[i]
+			for j := 0; j < m.ResourceMetrics().Len(); j++ {
+				rm := m.ResourceMetrics().At(j)
+				for k := 0; k < rm.ScopeMetrics().Len(); k++ {
+					sm := rm.ScopeMetrics().At(k)
+					for l := 0; l < sm.Metrics().Len(); l++ {
+						delete(missingMetrics, sm.Metrics().At(l).Name())
```
jinja2, line 71: I am a little confused. Can you clarify how deleting/checking for missingMetric is helpful for this test case?

atoulme: that's what resourceMetrics was doing ; over the life of the ingest, check which metrics are returned and make sure they match.

atoulme: it's a straight replacement - it just happens we have one metric to check on, but the same structure is used for other tests. We can refactor this out later to make it a call to a testutils method if we like.

jinja2: Ah, I thought with the memory metric specifically named missing metric we were doing something different.

PR #5805, opened 2025-01-15
remove unused filter processor

### LINKED ISSUE #5702: Filter processor in fargate config is not used in any pipeline
The fargate config has a filter processor, meant to filter our specific metrics. This is also mentioned in the splunk docs: https://docs.splunk.com/observability/en/gdi/opentelemetry/deployments/deployments-fargate.html#get-started

yet this filter processor is not added to any pipeline. 

https://github.com/signalfx/splunk-otel-collector/blob/4a7e59c9fca4a7072796b9e3f652a171c0452620/cmd/otelcol/config/collector/fargate_config.yaml#L79

atoulme:
**Description:**
Fixes #5702

### DIFF
--- a/cmd/otelcol/config/collector/fargate_config.yaml
+++ b/cmd/otelcol/config/collector/fargate_config.yaml
@@ -1,10 +1,5 @@
 # This collector config file is designed for use within an ECS task.
 # The collector should run in a sidecar container within an ECS task.
-config_sources:
-  env:
-    defaults:
-      METRICS_TO_EXCLUDE: []
-
 extensions:
   health_check:
     endpoint: 0.0.0.0:13133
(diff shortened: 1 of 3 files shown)


crobert-1 (approved):
Pardon my ignorance, but is the Fargate deployment an artifact released separately from the regular release cycle, or is it simply a configured collector? I realize this is functionally a no-op but users may be thrown off by the change to documentation without some kind of changelog entry. (If there's no separate release for Fargate, we can add the changelog notes now in the regular release notes CHANGELOG.)

atoulme:
It's shipped in our Dockerfile and the instructions don't work and are misleading right now, hence the original bug report. I think removing this removes confusion.

PR #5807, opened 2025-01-15
Move oracledb discovery test to checking metrics

atoulme:
**Description:**
Rather than use resource_metrics, use a check for metrics presence. A separate test checks all metrics in the same folder.


tests/receivers/oracledb/bundled_test.go
```diff
@@ -33,29 +35,85 @@ func TestOracledbDockerObserver(t *testing.T) {
...
+				"--set", "splunk.discovery.receivers.oracledb.config.password='${ORACLE_PASSWORD}'",
+				"--set", "splunk.discovery.receivers.oracledb.config.service=XE",
+				"--set", `splunk.discovery.extensions.k8s_observer.enabled=false`,
+				"--set", `splunk.discovery.extensions.host_observer.enabled=false`,
+				"--set", fmt.Sprintf("splunk.discovery.extensions.docker_observer.config.endpoint=tcp://%s", dockerSocket.ContainerEndpoint),
+			)
+		})
+	defer shutdown()
+
+	metricNames := []string{
```
crobert-1, line 65: Is there a reason for this format vs. a golden metrics file?

atoulme: I can't run the test locally to capture the oracledb golden file because the oracledb container refuses to work on an ARM architecture in a weird way. I ended up copying what was in the resource_metrics file. The other test checks that all metrics are reported, so we still have coverage.

crobert-1 (approved):

PR #5810, opened 2025-01-17
[Tanzu Tile] Add host.name resource attribute

crobert-1:

The `host.name` resource attribute should come from the Tanzu environment, not the environment in which the collector itself is running.

**Note:**
No changelog here because this won't have any impact until a new Tanzu Tile release is created.

Tested on TAS 10.0 and TAS 6.0. Working as expected. The `host.name` is populated as the IP address of the BOSH director.

### DIFF
--- a/deployments/cloudfoundry/bosh/jobs/splunk-otel-collector/templates/otel-collector-config.yaml.erb
+++ b/deployments/cloudfoundry/bosh/jobs/splunk-otel-collector/templates/otel-collector-config.yaml.erb
@@ -23,8 +23,11 @@ receivers:
         <% end %>
 
 processors:
-  resourcedetection:
-    detectors: [system]
+  resource:
+    attributes:
+    - key: host.name
+      value: <%= p('cloudfoundry.director.hostname') %>
+      action: upsert
 
 exporters:
   signalfx:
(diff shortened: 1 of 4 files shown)


jinja2 (approved):

deployments/cloudfoundry/bosh/jobs/splunk-otel-collector/templates/otel-collector-config.yaml.erb
```diff
@@ -23,8 +23,11 @@ receivers:
         <% end %>
 
 processors:
-  resourcedetection:
```
pjanotti, line 26: Can we somehow test this file so it doesn't get stale like it did here?

pjanotti (approved):

deployments/cloudfoundry/bosh/jobs/splunk-otel-collector/templates/otel-collector-config.yaml.erb
crobert-1: There isn't a quick way to test this automatically. Running a BOSH deployment requires a lot of set up and dependency installs, so it'd be some amount of effort. This file doesn't get changed much and is tested directly with each Tanzu Tile release.

The change to the config here is a bug fix, it didn't actually go stale.

PR #5813, opened 2025-01-18
Address apachespark test flakiness

atoulme:
**Description:** 
Do not exit running jobs on spark. It is possible the container may exit before the collector collects metrics

### DIFF
--- a/docker/apachespark/long_running.py
+++ b/docker/apachespark/long_running.py
@@ -26,12 +26,12 @@
 #
 
 import time
+import signal
 import threading
 import queue as Queue
 from typing import Any, Callable, List, Tuple
 
 from pyspark import SparkConf, SparkContext
-
 
 def delayed(seconds: int) -> Callable[[Any], Any]:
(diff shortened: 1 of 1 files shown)


docker/apachespark/long_running.py
```diff
@@ -76,7 +76,15 @@ def run() -> List[Tuple[int, int]]:
...
+    global interrupted
+    interrupted = True
 
 if __name__ == "__main__":
-    for i in range(100):
+    signal.signal(signal.SIGINT, signal_handler)
+    global interrupted
+    interrupted = False
+    while True:
         main()
```
pjanotti, line 88: Does main does some backoff? Anyway, LGTM.

pjanotti (approved):

docker/apachespark/long_running.py
atoulme: as I understand it, this script submits 100 times a job to spark with a random sleep in them. This change runs the test constantly until the container receives a kill signal.

PR #5828, opened 2025-01-21
[chore] Remove duplicate receiver entry

crobert-1:

There were two identical entries in the table for the Simple Prometheus receiver, so I've removed one.


docs/components.md
```diff
@@ -54,7 +54,6 @@ The distribution offers support for the following components.
 | [otlp](https://github.com/open-telemetry/opentelemetry-collector/tree/main/receiver/otlpreceiver)                                                                  | [stable]         |
 | [postgresql](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/postgresqlreceiver)                                              | [beta]           |
 | [prometheus](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver)                                              | [beta]           |
-| [prometheus_simple](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/simpleprometheusreceiver)                                 | [beta]           |
```
crobert-1, line 57: If anyone has strong preferences the other way I'm happy to remove the `simpleprometheus` entry instead.

dmitryax: We are using the component names as they are applied in the config for others. So I think we should remove `simpleprometheus` instead

crobert-1: Makes sense, I've swapped.

dmitryax (approved):

PR #5832, opened 2025-01-22
Stop supporting deprecated syntax for config source expansion

dmitryax:
The expansion of bare config sources is no longer supported.

Strings like `$ENV` or `$include:/path/to/file.yaml` will no longer be expanded. Instead, use the `${env:ENV}` or `${include:/path/to/file.yaml}` syntax. Only two symbols are allowed after `$`: `{` and `$`. The collector will log an error and fail to start if it encounters a bare config source expansion.

This syntax has been deprecated since [version 0.105.0](https://github.com/signalfx/splunk-otel-collector/blob/main/CHANGELOG.md#deprecations--6).

The reason for this change is aligning with OpenTelemetry Collector upstream, where this syntax has been removed in [0.104.0 version](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CHANGELOG.md#v1110v01040). Long term plan is to do not have [the custom config sources](https://github.com/signalfx/splunk-otel-collector/tree/main/internal/configsource) in this repository, instead consolidate the functionality with [the upstream confmap providers](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/provider).

### DIFF
--- a/internal/configsource/source.go
+++ b/internal/configsource/source.go
@@ -17,10 +17,8 @@ package configsource
 import (
 	"context"
 	"fmt"
-	"log"
 	"net/url"
 	"strings"
-	"sync"
 
 	"github.com/knadh/koanf/maps"
 	"github.com/spf13/cast"
@@ -128,20 +126,9 @@ func BuildConfigSources(ctx context.Context, configSourcesSettings map[string]Se
 // 2. Wait for an update on "watcher" func.
 // 3. Close the confmap.Retrieved instance;
 //
-// The current syntax to reference a config source in a YAML is provisional. Currently
-// single-line:
-//
-//	param_to_be_retrieved: $<cfgSrcName>:<selector>[?<params_url_query_format>]
-//
-// bracketed single-line:
+// The current syntax to reference a config source:
 //
 //	param_to_be_retrieved: ${<cfgSrcName>:<selector>[?<params_url_query_format>]}
-//
-// and multi-line are supported:
-//
-//	param_to_be_retrieved: |
-//	  $<cfgSrcName>: <selector>
-//	  [<params_multi_line_YAML>]
 //
 // The <cfgSrcName> is a name string used to identify the config source instance to be used
 // to retrieve the value.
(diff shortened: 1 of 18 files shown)


internal/configsource/source.go
```diff
@@ -151,22 +138,11 @@ func BuildConfigSources(ctx context.Context, configSourcesSettings map[string]Se
...
-// example in a YAML file:
+// <params_url_query_format> uses the same syntax as URL query parameters. Hypothetical example in a YAML file:
 //
 // component:
 //
-//	config_field: |
-//	  $yamltemplate: /etc/log_template.yaml
-//	  logs_path: /var/logs/
-//	  timeout: 10s
+//	config_field: {$file:/etc/secret.bin?binary=true}
```
jinja2, line 145:
```suggestion
//	config_field: ${file:/etc/secret.bin?binary=true}
```

internal/configsource/source.go
```diff
@@ -198,36 +174,23 @@ func BuildConfigSources(ctx context.Context, configSourcesSettings map[string]Se
...
-// for their expansion it is necessary to have a way to distinguish between them. For the non-bracketed
-// syntax the code will peek at the first character other than alpha-numeric and '_' after the '$'. If
-// that character is a ':' it will treat it as a config source and as environment variable otherwise.
-// For example:
-//
-//	component:
-//	  field_0: $PATH:/etc/logs # Injects the data from a config sourced named "PATH" using the selector "/etc/logs".
-//	  field_1: $PATH/etc/logs  # Expands the environment variable "PATH" and adds the suffix "/etc/logs" to it.
+//	  # variable DATA_PATH.
+//	  text_from_file: ${file:${DATA_PATH}/text.txt}
```
jinja2, line 178:
```suggestion
//	  text_from_file: ${file:${env:DATA_PATH}/text.txt}
```
do we need to explicitly add `env` in such examples?

internal/configsource/source.go
dmitryax: Thanks. Applied

jinja2 (approved):

dmitryax:
3 unrelated tests are still failing. Submitted separate PRs:
- https://github.com/signalfx/splunk-otel-collector/pull/5835
- https://github.com/signalfx/splunk-otel-collector/pull/5838

PR #5849, opened 2025-01-29
Remove ubuntu 16 from tests

dmitryax:
Ubuntu 16 is only getting the extended paid security support from Cannonical these days. The tests are failing most of the time due to unreliable package endpoints for Ubuntu 16. We don't remove the support but remove want to make sure the CI is stable. So the tests for Ubuntu 16 are removed.


deployments/ansible/roles/collector/meta/main.yml
```diff
@@ -21,7 +21,6 @@ galaxy_info:
       versions:
         - bionic
         - focal
-        - xenial
```
jinja2, line 24: I think this is removing 16.0 as supported platform version from the actual ansible role. Do we want to do this?

dmitryax: Good catch! I've reverted this change

jinja2 (approved):

PR #5850, opened 2025-01-30
try to match against metric received by the sink

atoulme:
Attempt to make testing catch any intermediate metrics that we may have received instead of the latest metrics in the sink.

### DIFF
--- a/tests/receivers/envoy/bundled_k8s_test.go
+++ b/tests/receivers/envoy/bundled_k8s_test.go
@@ -18,7 +18,6 @@ package tests
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -131,10 +130,10 @@ func TestEnvoyK8sObserver(t *testing.T) {
 			assert.Fail(tt, "No metrics collected")
 			return
 		}
-		var errs error
+		var err error
 		newIndex := len(sink.AllMetrics())
 		for i := index; i < newIndex; i++ {
-			err := pmetrictest.CompareMetrics(expected, m,
+			err = pmetrictest.CompareMetrics(expected, sink.AllMetrics()[i],
 				pmetrictest.IgnoreResourceAttributeValue("service.instance.id"),
 				pmetrictest.IgnoreResourceAttributeValue("net.host.port"),
 				pmetrictest.IgnoreResourceAttributeValue("net.host.name"),
@@ -162,9 +161,8 @@ func TestEnvoyK8sObserver(t *testing.T) {
 			if err == nil {
 				return
 			}
-			errs = errors.Join(errs, err)
 		}
 		index = newIndex
-		assert.NoError(tt, errs)
+		assert.NoError(tt, err)
 	}, 120*time.Second, 1*time.Second)
 }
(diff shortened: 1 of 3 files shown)


tests/testutils/golden.go
```diff
@@ -128,8 +129,15 @@ func RunMetricsCollectionTest(t *testing.T, configFile string, expectedFilePath
...
-			opts.compareMetricsOptions...)
-		assert.NoError(tt, err)
+		var errs error
+		for _, m := range sink.AllMetrics() {
+			err := pmetrictest.CompareMetrics(expected, m,
+				opts.compareMetricsOptions...)
+			if err == nil {
+				return
+			}
+			errs = errors.Join(errs, err)
```
dmitryax, line 139: I have a couple of concerns here:
1. Joined errors will be very hard to read.
2. We are scanning all of them every second which can increase time to run the tests

I'd suggest something like this:

```
        var checkedMetrics int
	assert.EventuallyWithT(t, func(tt *assert.CollectT) {
		if len(sink.AllMetrics()) == 0 {
			assert.Fail(tt, "No metrics collected")
			return
		}
		if len(sink.AllMetrics()) > checkedMetrics {
			checkedMetrics++
		}
		assert.NoError(tt, pmetrictest.CompareMetrics(expected, sink.AllMetrics()[checkedMetrics-1], opts.compareMetricsOptions...))
	}, 30*time.Second, 200*time.Millisecond)
```

I assume `assert.EventuallyWithT` cannot run concurrently here

atoulme: yes I am a bit lazy and we can incur a performance cost. Your method would work but we'd need to read all the new metrics, not just the new one, as multiple batches can be received. I can rework this code to accommodate.

dmitryax: The method I posted read all the metrics

tests/receivers/envoy/bundled_k8s_test.go
```diff
@@ -124,39 +125,46 @@ func TestEnvoyK8sObserver(t *testing.T) {
...
+				pmetrictest.IgnoreScopeMetricsOrder(),
+				pmetrictest.IgnoreScopeVersion(),
+				pmetrictest.IgnoreResourceMetricsOrder(),
+				pmetrictest.IgnoreMetricsOrder(),
+				pmetrictest.IgnoreMetricValues(),
+			)
+			if err == nil {
+				return
+			}
+			errs = errors.Join(errs, err)
```
dmitryax, line 165: I'm afraid multiple errors aggregated within a second can be hard to read. But we'll see

dmitryax (approved):

tests/receivers/envoy/bundled_k8s_test.go
dmitryax: Actually, not sure what the value of joining here, if we only join the errors from metrics received in the last second. I'd suggest throwing the last one. At least it'll be readable

tests/testutils/golden.go
dmitryax: On the other hand, this can slow down the validation by the provided frequency. So, you're right, it's better to read all of them as soon as they arrive

tests/receivers/envoy/bundled_k8s_test.go
atoulme: no problem, changed.

PR #5893, opened 2025-02-13
[Databricks] Add support for Databricks deployments

crobert-1:

This adds support for an init script in Databricks environments that will deploy the collector on every node of a cluster.

The added README should provide necessary context and deployment requirements.

### DIFF
--- a/deployments/databricks/deploy_collector.sh
+++ b/deployments/databricks/deploy_collector.sh
@@ -15,7 +15,6 @@
 #!/bin/bash
 
 set -euo pipefail
-set -x
 
 # This script is used to deploy the Splunk distribution of the OpenTelemetry Collector
 # on the current node of a Databricks cluster. Through UI configuration the script will
@@ -30,8 +29,8 @@ set -x
 #   Directions for creating a PAT: https://docs.databricks.com/en/dev-tools/auth/pat.html
 
 # Optional Variables:
-# - OTEL_VERSION: Version of the Splunk OpenTelemetry Collector to deploy as a part of this release.
-#   Default: "latest". Valid version must be >=0.120.0.
+# - SPLUNK_OTEL_VERSION: Version of the Splunk OpenTelemetry Collector to deploy as a part of this release.
+#   Default: "latest". Valid version must be >=0.119.0.
 # - SCRIPT_DIR: Installation path for the Collector and its config
 # - SPLUNK_REALM: Splunk o11y realm to send data to. Default: us0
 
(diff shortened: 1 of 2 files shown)


deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,170 @@
...
+  echo "environment variable 'SPLUNK_ACCESS_TOKEN' must be set, exiting."
+  exit 1
+fi
+
+if [ -z "${SDATABRICKS_ACCESS_TOKEN}" ]; then
+  echo "environment variable 'DATABRICKS_ACCESS_TOKEN' must be set, exiting."
+fi
+
+
+OTEL_VERSION=${OTEL_VERSION:-latest}
```
crobert-1, line 57: Happy to discuss variable names here, I'm sure they could be made more clear and obvious.

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,170 @@
...
+  echo "environment variable 'DATABRICKS_CLUSTER_HOSTNAME' must be set, exiting."
+  exit 1
+fi
+
+if [ -z "${SPLUNK_ACCESS_TOKEN}" ]; then
+  echo "environment variable 'SPLUNK_ACCESS_TOKEN' must be set, exiting."
+  exit 1
+fi
+
+if [ -z "${SDATABRICKS_ACCESS_TOKEN}" ]; then
```
jinja2, line 52:
```suggestion
if [ -z "${DATABRICKS_ACCESS_TOKEN}" ]; then
```

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,170 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/bin/bash
+
+set -euo pipefail
+set -x
```
jinja2, line 18: I think this option will print the access tokens set to vars in the script. Customers might ask for this to be removed.

deployments/databricks/deploy_collector.sh
jinja2: I see a few different variations in use `SPLUNK_OTEL_VERSION` (TA), `SPLUNK_OTEL_COLLECTOR_VERSION` (option in puppet role) and also `OTEL_VERSION` (cloudfoundry). I think prefixing with SPLUNK makes sense cause we might push patch versions which does not match with a version on upstream. Wdyt?

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,170 @@
...
+  echo "environment variable 'DATABRICKS_ACCESS_TOKEN' must be set, exiting."
+fi
+
+
+OTEL_VERSION=${OTEL_VERSION:-latest}
+OS="linux_amd64"
+OTEL_BINARY_NAME="splunk_otel_collector"
+OTEL_DOWNLOAD_BASE_URL="https://github.com/signalfx/splunk-otel-collector/releases"
+OTEL_API_URL="https://api.github.com/repos/signalfx/splunk-otel-collector/releases/latest"
+SCRIPT_DIR=${SCRIPT_DIR:-/tmp/collector_download}
```
jinja2, line 62:
```suggestion
SCRIPT_DIR="${SCRIPT_DIR:-/tmp/collector_download}"
```

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,170 @@
...
+        OTEL_VERSION=$(curl --silent "$OTEL_API_URL" |    # Get latest Collector release from GitHub api
+            grep '"tag_name":' |                          # Get tag name line
+            sed -E 's/.*"([^"]+)".*/\1/')                 # Pluck latest release version
+        if [ -z "$OTEL_VERSION" ]; then
+            echo "Failed to get tag_name for latest release from $OTEL_VERSION/latest" >&2
+            exit 1
+        fi
+fi
+
+OTEL_BINARY_DOWNLOAD_URL=${OTEL_DOWNLOAD_BASE_URL}/download/${OTEL_VERSION}/otelcol_${OS}
```
jinja2, line 79:
```suggestion
OTEL_BINARY_DOWNLOAD_URL="${OTEL_DOWNLOAD_BASE_URL}/download/${OTEL_VERSION}/otelcol_${OS}"
```

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,170 @@
...
+
+# Download Splunk's distribution of the OpenTelemetry Collector
+mkdir -p "$SCRIPT_DIR"
+wget -nv -O "$OTEL_BINARY_FILE" $OTEL_BINARY_DOWNLOAD_URL
```
jinja2, line 83:
```suggestion
wget -nv -O "$OTEL_BINARY_FILE" "$OTEL_BINARY_DOWNLOAD_URL" || { echo "Failed to download $OTEL_BINARY_DOWNLOAD_URL"; exit 1; }
```

Should we use only one of either wget or curl (curl is used to get some info earlier in the script)? Not a big deal especially if we are certain that a databricks host will have both.

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,170 @@
...
+
+[Install]
+WantedBy=multi-user.target
+"
+
+echo "$collector_service" > $SERVICE_FILE
+chmod 755 $SERVICE_FILE
+
+# The collector is run as a service on the current node
+systemctl start $OTEL_BINARY_NAME
```
jinja2, line 168:
```suggestion
systemctl enable "$OTEL_BINARY_NAME"
systemctl start $OTEL_BINARY_NAME
```
To get systemctl to start the collector on reboots. I don't know if this is meaningful for databricks especially if the init script mechanism will always run this script, so please ignore if this doesn't make sense.

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,170 @@
...
+if [ $DB_IS_DRIVER = "TRUE" ]; then
+        OPTIONAL_SPARK_RECEIVER=", apachespark"
+else
+        OPTIONAL_SPARK_RECEIVER=""
+fi
+
+collector_config="
+extensions:
+  bearertokenauth:
+    token: "$DATABRICKS_ACCESS_TOKEN"
```
jinja2, line 99:
```suggestion
    token: ${env:DATABRICKS_ACCESS_TOKEN}
```
Should we use env config provider instead? It would help keep the config file free of sensitive tokens. Also, I think we might have to escape the $ in my suggestion.

deployments/databricks/deploy_collector.sh
crobert-1: Good call. I've updated all variables that are Splunk-specific (release version, URLs, binary file names, etc.) to include the `SPLUNK_` prefix.

deployments/databricks/deploy_collector.sh
crobert-1: I agree with your concern here, but after some research, the collector is run by systemd which doesn't have access to environment variables. The collector fails to start because the environment variable can't be found.

The alternative is to add the access token as an [environment variable to the service](https://serverfault.com/questions/413397/how-to-set-environment-variable-in-systemd-service), but I think that ends up with a similar problem where the service script now contains the plaintext access token.

Any other ideas are welcome

deployments/databricks/deploy_collector.sh
crobert-1: I did some testing, the init script is run every time a node starts up, including spinning up for the first time and restarts. Based on this, I don't think we explicitly need to enable it unless you think it's more safe to do so.

deployments/databricks/deploy_collector.sh
crobert-1: I've moved to use `curl` for both commands. Both `wget` and `curl` should be installed, but consistency can't hurt.

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,188 @@
...
+  INSTALLED_SPLUNK_OTEL_VERSION=($($SPLUNK_OTEL_BINARY_FILE --version))
+  INSTALLED_SPLUNK_OTEL_VERSION=${INSTALLED_SPLUNK_OTEL_VERSION[2]}
+fi
+
+if [ ! -f $SPLUNK_OTEL_BINARY_FILE ] || [ $INSTALLED_SPLUNK_OTEL_VERSION != $SPLUNK_OTEL_VERSION ]; then
+  # If the binary is already installed it's the wrong version and it needs to be removed before downloading again
+  # to avoid "Text file busy" errors
+  if [ -f $SPLUNK_OTEL_BINARY_FILE ]; then
+    rm "$SPLUNK_OTEL_BINARY_FILE"
+  fi
```
crobert-1, line 95: After some more testing, I had to add this section to ensure the new version of the collector would be properly installed over an existing different collector version.

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,188 @@
...
+User=root
+ExecStart=$SPLUNK_OTEL_BINARY_FILE --config $CONFIG_FILE
+
+[Install]
+WantedBy=multi-user.target
+"
+
+echo "$collector_service" > $SERVICE_FILE
+chmod 755 $SERVICE_FILE
+
```
jinja2, line 184: I think we need to reload systemctl after adding new service files.
```suggestion
systemctl daemon-reload
```

deployments/databricks/deploy_collector.sh
```diff
@@ -0,0 +1,188 @@
...
+  pipelines:
+    metrics:
+      receivers: [hostmetrics$OPTIONAL_SPARK_RECEIVER]
+      processors: [batch, resourcedetection, resource]
+      exporters: [signalfx]
+"
+
+echo "$collector_config" > "$CONFIG_FILE"
+
+collector_service="
```
jinja2, line 166: Did you get a chance to try installing the collector with a package manager on databricks? I am just wondering if these nodes are fixed os type, then maybe we could install with a package manager. Looking at the service file [here](https://github.com/signalfx/splunk-otel-collector/blob/main/packaging/fpm/splunk-otel-collector.service), we'd get stuff like dedicated user/group and a better experience in general.

atoulme (approved):

deployments/databricks/deploy_collector.sh
crobert-1: https://github.com/signalfx/splunk-otel-collector/pull/5997

PR #5913, opened 2025-02-17
[chore] Add an integration test to validate internal metrics

dmitryax:
The test is needed for the 0.119.0 upgrade to validate that the output of the internal metrics isn't changed. There was a new unexpected metric added.

### DIFF
--- a/tests/general/internal_metrics_test.go
+++ b/tests/general/internal_metrics_test.go
@@ -1,0 +1,55 @@
+// Copyright Splunk, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


tests/general/internal_metrics_test.go
```diff
@@ -0,0 +1,55 @@
...
+				"scrape_duration_seconds",
+				"scrape_samples_post_metric_relabeling",
+				"scrape_samples_scraped",
+				"scrape_series_added",
+				"up",
+			),
+			pmetrictest.IgnoreResourceAttributeValue("service_instance_id"),
+			pmetrictest.IgnoreResourceAttributeValue("service_version"),
+			pmetrictest.IgnoreMetricAttributeValue("service_instance_id"),
+			pmetrictest.IgnoreMetricAttributeValue("service_version"),
```
crobert-1, line 52: Entirely unrelated to this PR, but do you know why these attributes are on both the resource and data point?

crobert-1 (approved):

tests/general/internal_metrics_test.go
dmitryax: It's a migration phase. We used to be able to set it on data point attributes (open sensus instrumentation). Now, it's partially migrated to the otel instrumentation where we can set the resource attributes. The next release will be only setting resource attributes.

crobert-1: Thanks, makes sense

PR #5914, opened 2025-02-17
[chore] Pin govuln check to latest released version

crobert-1:

`compile` is failing on `main` due to the `govulncheck@master` package requiring `>= go 1.23`. Since we're still on `go 1.22` for testing, we can pin this dependency to the latest release instead. This should keep it updated enough to save us from manually updating it, while also preserving us from breaking changes in `master` unexpectedly between releases.

**Testing:**
```
$ go install golang.org/x/vuln/cmd/govulncheck@latest
$ govulncheck --version
Go: go1.22.10
Scanner: govulncheck@v1.1.4
```


jinja2 (approved):

crobert-1:
_Failing Windows test is flaky and unrelated to this change._

PR #5917, opened 2025-02-17
[chore] Update IIS test to retry GET requests until successful

crobert-1:

This test is failing pretty consistently on either the returned status of the GET request is `500`, or the target machine actively refuses the request. This is to test to see if retrying the check after some time passes allows the server to be up and running and successfully process the command.

### DIFF
--- a/tests/zeroconfig/windows/windows_iis_test.go
+++ b/tests/zeroconfig/windows/windows_iis_test.go
@@ -17,7 +17,6 @@
 package zeroconfig
 
 import (
-	"fmt"
 	"net/http"
 	"os/exec"
 	"path"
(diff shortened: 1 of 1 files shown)


dmitryax:
Thanks for picking this up!

tests/zeroconfig/windows/windows_iis_test.go
```diff
@@ -99,23 +100,40 @@ func requireNoErrorExecCommand(t *testing.T, name string, arg ...string) {
...
+		return 0, err
+	}
+
 	defer resp.Body.Close()
-	require.Equal(t, http.StatusOK, resp.StatusCode)
+	return resp.StatusCode, err
 }
 
 func testExpectedTracesForHTTPGetRequest(t *testing.T, otlp *testutils.OTLPReceiverSink, url string, expectedTracesFileName string) {
-	requireHTTPGetRequestSuccess(t, url)
```
dmitryax, line 113: We could just move this under `assert.Eventually(t, func() bool {`. I believe it's going to be the same result. No need for extra code

dmitryax: Oh I missed that it's `assert.Eventually`. If we change it to `assert.EventuallyWithT` then this call can be moved as is.

```
		if err != nil {
			t.Log(err)
			return false
		}
```
would be `assert.NoError(t, err)` or `require.NoError(t, err)`

crobert-1: I don't think I understood exactly what you're suggesting here, but I've pushed my best guess. Let me know how it differs (if so) from your suggestion.

dmitryax: You've got it right. The only difference is that the `requireHTTPGetRequestSuccess` function is removed, but that's alright

dmitryax (approved):
Let's see if it helps

crobert-1:
_Closing in favor of https://github.com/signalfx/splunk-otel-collector/pull/5919_

PR #5918, opened 2025-02-17
[chore] Update default metric_relabel_configs for internal metrics

dmitryax:
The gRPC and HTTP metrics are now only sent at the detailed level. So, no need to exclude them anymore. However, there is another redundant metric emitted by default in 0.119.0 after the transition to OTel SDK, `promhttp_metric_handler_errors`. This change updates the default configs and the appropriate converter to exclude this metric.


crobert-1 (approved):
The logic change generally make sense to me, should be good once tests are passing.

PR #5919, opened 2025-02-18
[chore] Update IIS integration tests

dmitryax:


dmitryax:
@crobert-1 this is another try. Still failing. I think we need to look into that, see why it's failing...

crobert-1 (approved):
Agreed that this does not resolve the issue, but since it gets us a little further in the test before failing I think it would be worth merging.

PR #5931, opened 2025-02-21
[deployment/puppet] Fix unknown variable

crobert-1:

Puppet tests are failing every time on main, this attempts to resolve.

### DIFF
--- a/deployments/puppet/lib/facter/win_programdata.rb
+++ b/deployments/puppet/lib/facter/win_programdata.rb
@@ -3,6 +3,6 @@
 Facter.add(:win_programdata) do
   confine :kernel => 'windows'
   setcode do
-    ENV['PROGRAMDATA']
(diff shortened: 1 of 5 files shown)


pjanotti:
@crobert-1 I'm a bit puzzled about what changed to cause this error. Do you have any idea of the root cause?

crobert-1:
> @crobert-1 I'm a bit puzzled about what changed to cause this error. Do you have any idea of the root cause?

No change in our repo happened that contributed to this failure, I believe it may be some un-pinned dependency changed or a dependency's functionality changed somehow.

pjanotti:
The new syntax seems an improvement, more in agreement with the docs, but clearly not enough to fix the issue 🤔

crobert-1:
After more investigation we noticed the environment variables are not being detected properly, meaning any changes of this PR would break when they're detected properly. This needs to be investigated further to properly get environment variables.

PR #5972, opened 2025-03-11
Prepare TA release v1.4.1

jvoravong:
- Mostly changes for the TA release.
- Added a couple improvements to improve a related make action.

### DIFF
--- a/packaging/technical-addon/packaging-scripts/release-ta-version.sh
+++ b/packaging/technical-addon/packaging-scripts/release-ta-version.sh
@@ -13,16 +13,28 @@ TAG="Splunk_TA_otel/v$TA_VERSION"
 
 git checkout -B "$BRANCH"
 git add "$SOURCE_DIR/Makefile" "$APP_CONFIG"
-git commit -m "Updates TA to splunk-otel-collector v$SPLUNK_OTEL_VERSION and marks TA as v$TA_VERSION" || echo "version changes already comitted"
-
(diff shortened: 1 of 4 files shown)


packaging/technical-addon/packaging-scripts/release-ta-version.sh
```diff
@@ -13,16 +13,28 @@ TAG="Splunk_TA_otel/v$TA_VERSION"
...
-
+git commit -m "Updates TA to splunk-otel-collector v$SPLUNK_OTEL_VERSION and marks TA as v$TA_VERSION" || echo "version changes already committed"
 
 echo "Tagging..."
 git tag --annotate --sign --message "Release of Splunk_TA_otel version v$TA_VERSION" "$TAG"
 git push --set-upstream "$REMOTE" "$BRANCH"
-read -i "Do you want to push the tag to remote too? [y/N]" -r -n 1 USER_INPUT
-if [ "$USER_INPUT" == "y" ]; then
+
+PROMPT="Do you want to push the tag to remote too? [y/N] "
```
jvoravong, line 22: @hughesjj-splunk only expanded for Mac bash support. I was wondering if we could actually just delete having a prompt input and just always push the tag. Should we keep this prompt?

jvoravong: Got it. Yeah we should definitely tag release. Let's leave this as is then for now.

PR #5976, opened 2025-03-13
[.NET Deployer] Make UF version used in CI customizable

pjanotti:
CI change to allow testing against different versions of Splunk UF.

### DIFF
--- a/.github/workflows/dotnet-instr-deployer-add-on.yml
+++ b/.github/workflows/dotnet-instr-deployer-add-on.yml
@@ -14,11 +14,11 @@ on:
       splunk_uf_version:
         description: 'Splunk UF version'
         required: false
-        default: '9.4.1'
+        default: '9.4.0'
       splunk_uf_build_hash:
         description: 'Splunk UF build hash'
         required: false
-        default: 'e3bdab203ac8'
+        default: '6b4ebe426ca6'
 
 env:
   GO_VERSION: 1.23.6
--- a/.github/workflows/dotnet-instr-deployer-add-on.yml
+++ b/.github/workflows/dotnet-instr-deployer-add-on.yml
@@ -4,9 +4,11 @@ on:
   push:
     paths:
       - 'packaging/dotnet-instr-deployer-add-on/**'
+      - '.github/workflows/dotnet-instr-deployer-add-on.yml'
   pull_request:
     paths:
       - 'packaging/dotnet-instr-deployer-add-on/**'
+      - '.github/workflows/dotnet-instr-deployer-add-on.yml'
   workflow_dispatch:
     inputs:
       splunk_uf_version:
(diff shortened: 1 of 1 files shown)


.github/workflows/dotnet-instr-deployer-add-on.yml
```diff
@@ -4,13 +4,27 @@ on:
...
+        default: '9.4.0'
+      splunk_uf_build_hash:
+        description: 'Splunk UF build hash'
+        required: false
+        default: '6b4ebe426ca6'
 
 env:
   GO_VERSION: 1.23.6
+  # Provide default values for non-manually triggered executions
+  splunk_uf_version: ${{ github.event.inputs.splunk_uf_version || '9.4.0' }}
```
pjanotti, line 26: Notice that this doesn't need to be in sync with the defaults for `workflow_dispatch`, it is nice if they are but not needed.

.github/workflows/dotnet-instr-deployer-add-on.yml
```diff
@@ -46,7 +60,8 @@ jobs:
 
       - name: Download the Splunk UF MSI
         shell: bash
-        run: curl https://download.splunk.com/products/universalforwarder/releases/9.4.0/windows/splunkforwarder-9.4.0-6b4ebe426ca6-windows-x64.msi -o splunkforwarder.msi
+        run: |
+          curl https://download.splunk.com/products/universalforwarder/releases/${{ env.splunk_uf_version }}/windows/splunkforwarder-${{ env.splunk_uf_version }}-${{ env.splunk_uf_build_hash }}-windows-x64.msi -o splunkforwarder.msi
```
crobert-1, line 64: Do we know that the naming scheme of the msi here will always be consistent with version-buildhash?

.github/workflows/dotnet-instr-deployer-add-on.yml
```diff
@@ -4,13 +4,27 @@ on:
   push:
     paths:
       - 'packaging/dotnet-instr-deployer-add-on/**'
+      - '.github/workflows/dotnet-instr-deployer-add-on.yml'
   pull_request:
     paths:
       - 'packaging/dotnet-instr-deployer-add-on/**'
+      - '.github/workflows/dotnet-instr-deployer-add-on.yml'
   workflow_dispatch:
+    inputs:
```
crobert-1, line 13: I assume these versions will have to be updated manually, is that correct? Is there anyway to update at some frequency?

crobert-1 (approved):

.github/workflows/dotnet-instr-deployer-add-on.yml
pjanotti: Unfortunately, I couldn't find a way to get the latest version number and hash via some "latest" or similar mechanism.

.github/workflows/dotnet-instr-deployer-add-on.yml
pjanotti: As far as I can tell yes, this schema goes back many versions and hasn't changed in years.

PR #5998, opened 2025-03-19
[chore][deployments/databricks] Add docs for deploying as standalone script

crobert-1:

If a cluster is long-running, the script may not be able to be deployed as an init script. In this case, the user will manually have to run the script on each node. This PR updates documentation for directions on how to deploy manually.

### DIFF
--- a/deployments/databricks/README.md
+++ b/deployments/databricks/README.md
@@ -18,7 +18,7 @@ serverless.
 ### Overview
 
 The Splunk distribution of the OpenTelemetry Collector can be deployed on a Databricks cluster using an
-[init script](https://docs.databricks.com/en/init-scripts/index.html) on start up, or by directly running the script on each node in a long running cluster.
+[init script](https://docs.databricks.com/en/init-scripts/index.html) on start up, or by directly running the script on each node in an existing Databricks cluster.
 The Collector will run on every node in a Databricks cluster, gathering host and Apache Spark metrics.
 
 ### Init script
--- a/deployments/databricks/README.md
+++ b/deployments/databricks/README.md
@@ -10,20 +10,24 @@
 
 The OpenTelemetry Collector can be used to observe the real-time state of a Databricks cluster to
 ensure it's operating as expected and reduce mean time to repair (MTTR) in the case of downgraded performance.
+This functionality is only relevant for the classic architecture of Databricks, it will not work for
+serverless.
 
 ## Deployment
 
 ### Overview
 
 The Splunk distribution of the OpenTelemetry Collector can be deployed on a Databricks cluster using an
-[init script](https://docs.databricks.com/en/init-scripts/index.html).
+[init script](https://docs.databricks.com/en/init-scripts/index.html) on start up, or by directly running the script on each node in a long running cluster.
 The Collector will run on every node in a Databricks cluster, gathering host and Apache Spark metrics.
+
+### Init script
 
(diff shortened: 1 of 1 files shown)


deployments/databricks/README.md
```diff
@@ -10,20 +10,24 @@
...
+This functionality is only relevant for the classic architecture of Databricks, it will not work for
+serverless.
 
 ## Deployment
 
 ### Overview
 
 The Splunk distribution of the OpenTelemetry Collector can be deployed on a Databricks cluster using an
-[init script](https://docs.databricks.com/en/init-scripts/index.html).
+[init script](https://docs.databricks.com/en/init-scripts/index.html) on start up, or by directly running the script on each node in a long running cluster.
```
jinja2, line 21: Is the initscript method only usable/configurable when a cluster is being spun up? Or can a user configure init-script for an existing cluster but with the understanding that any existing nodes in the cluster need to have this script manually run on it but any new nodes in the cluster will have it run using the initscript technique?

```suggestion
[init script](https://docs.databricks.com/en/init-scripts/index.html) on start up, or by directly running the script on each node in an existing Databricks cluster.
```
Is long-running here same as existing? or is this more about a certain version of Databricks not having the initscript capability?

crobert-1: Long running here would be the same as existing **running** cluster, I'll commit your suggestion and modify to include running.

deployments/databricks/README.md
```diff
@@ -10,20 +10,24 @@
...
+This functionality is only relevant for the classic architecture of Databricks, it will not work for
+serverless.
 
 ## Deployment
 
 ### Overview
 
 The Splunk distribution of the OpenTelemetry Collector can be deployed on a Databricks cluster using an
-[init script](https://docs.databricks.com/en/init-scripts/index.html).
+[init script](https://docs.databricks.com/en/init-scripts/index.html) on start up, or by directly running the script on each node in an existing Databricks cluster.
```
crobert-1, line 21:
```suggestion
[init script](https://docs.databricks.com/en/init-scripts/index.html) on start up, or by directly running
the script on each node in an existing running Databricks cluster.
```

deployments/databricks/README.md
crobert-1: > Is the initscript method only usable/configurable when a cluster is being spun up? Or can a user configure init-script for an existing cluster but with the understanding that any existing nodes in the cluster need to have this script manually run on it but any new nodes in the cluster will have it run using the initscript technique?

I haven't been able to test autoscale to know for sure. I'd lean toward only configurable when the cluster is being spun up, but the docs aren't 100% clear from my reading. 

Relevant docs
1. If a cluster-scoped init script returns a non-zero exit code, the cluster launch fails.
2. An init script (initialization script) is a shell script that runs during startup of each cluster node before the Apache Spark driver or executor JVM starts.
3. Whenever you change any type of init script, you must restart all clusters affected by the script.
4. To remove a script from the cluster configuration, click the trash icon at the right of the script. When you confirm the delete you will be prompted to restart the cluster.

[Source 1](https://docs.databricks.com/aws/en/init-scripts), [source 2](https://docs.databricks.com/aws/en/init-scripts/cluster-scoped)

jinja2: ah, ok. I think admins of the cluster deploying the script will probably figure out what works for them best.

jinja2 (approved):

PR #6003, opened 2025-03-21
Update go version to 1.23.7

pjanotti:


crobert-1 (commented):
I'm seeing a few more references to 1.23.6 in our tests. We should either update them or explain why they don't need to be updated:
```
gitlab-ci.yml
darwin-test.yml
integration-test.yml
otelcol-fips.yml
vuln-scans.yml
windows-test.yml
```

pjanotti:
@crobert-1 thanks for pointing out those!

PR #6021, opened 2025-03-21
Make govulncheck run more reliable

pjanotti:
Broken down the execution per packages, but, in the end that was not enough and had to change the scan scope to `package`. The main disadvantage of that is that with `symbol` the tool can better asses if the code uses a function or code pattern, with `package` scan we may have to do some follow up to understand if/how the vulnerability should be treated. This is not ideal, the alternative would be have some larger runner that could take the memory hit of the tool with the `symbol` scan. 

Although the breakdown in packages was not enough I opted to keep it because it lowers the chances of more OOMs in the future and also make easier to run it locally when investigating issues. Similar to the changes on the Makefile: in the end it was not necessary to solve the problem, but, since it was already in place it is good to have it at hand.

Example run: https://github.com/signalfx/splunk-otel-collector/actions/runs/14010129646/job/39228876676?pr=6021


.github/workflows/scripts/govulncheck-run.sh
```diff
@@ -0,0 +1,33 @@
...
+FAILED=0
+
+# Repository prefix to remove from package names
+REPO_PREFIX=$(go list -m)
+
+# Run govulncheck for each package
+for pkg in $ALL_PKG_DIRS; do
+  OUTPUT_FILE="./govulncheck/$(echo "$pkg" | sed "s|^$REPO_PREFIX/||" | tr '/' '_').sarif"
+  echo -e "\nRunning govulncheck for package $pkg"
+  if ! govulncheck ${GOVULN_OPTS:-} "$pkg" > "$OUTPUT_FILE"; then
```
crobert-1, line 20: Nit: Can we style this `if` with brackets to make the style in this file consistent?

.github/workflows/scripts/govulncheck-run.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Initialize failure flag
+FAILED=0
+
+# Repository prefix to remove from package names
+REPO_PREFIX=$(go list -m)
+
+# Run govulncheck for each package
+for pkg in $ALL_PKG_DIRS; do
+  OUTPUT_FILE="./govulncheck/$(echo "$pkg" | sed "s|^$REPO_PREFIX/||" | tr '/' '_').sarif"
```
crobert-1, line 18: Can you add a comment explaining what `sed` and `tr` are doing here?

.github/workflows/scripts/govulncheck-run.sh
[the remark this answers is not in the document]
pjanotti: Yes, before #6024 was merged you could see some of the vulnerabilities, with the proper filters you can see them at https://github.com/signalfx/splunk-otel-collector/security/code-scanning?query=pr%3A6021+

Notice that the govulnchek run itself is green, it generated and uploaded the sarif file, only the code scanning report is going to show the vulnerabilities. This behavior is not being changed on this PR. If we want the run to be like the others and fail we should do a run without generating the sarif file.

.github/workflows/scripts/govulncheck-run.sh
pjanotti: f242a1a43fa30102455de2803c4cd86b62db39e6

.github/workflows/scripts/govulncheck-run.sh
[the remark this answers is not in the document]
crobert-1: Do we have any way of setting up notifications for security issues? I'd be concerned about missing them if we don't have alerts set up.

.github/workflows/scripts/govulncheck-run.sh
pjanotti: I'm not bash expert, but, I think this is the preferred style in this case, see https://google.github.io/styleguide/shellguide.html#checking-return-values.  That said that style guide has some other recommendations that the script is not following...

My understanding here is that since there is no use of conditionals the code shouldn't use square brackets. In order to use brackets here the code would have to either add $() or store the result in a variable to compare the result of the command.

.github/workflows/scripts/govulncheck-run.sh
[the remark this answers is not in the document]
pjanotti: yeah, I was thinking the same, but, that is a separate issue. This PR ensures that the runs don't fail because OOM. If notification doesn't work we can get rid of the sarif run and do a normal run, then it would fail and output reasonable messages. I actually don't know why we opted to use the sarif upload. I'm opting to separate that consideration from this PR.

pjanotti:
@crobert-1 just to be clear, we can really get rid of all of these by just setting the scan mode to be `package` instead of `symbol` which this PR had to do anyway to get reliable runs. Targeting the packages seem to be more about the future if the code base grows something that we should avoid if possible.

PR #6033, opened 2025-03-26
Fail govulncheck if vulnerabilities are found

pjanotti:
For consistency with other vulnerabilities scan the workflow should fail if vulnerabilites are detected.

Example of other vulnerability scans failing:
- https://github.com/signalfx/splunk-otel-collector/actions/runs/14010129646/job/39228865795
- https://github.com/signalfx/splunk-otel-collector/actions/runs/14010129646/job/39228866006

### DIFF
--- a/.github/workflows/scripts/govulncheck-run.sh
+++ b/.github/workflows/scripts/govulncheck-run.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -euo pipefail
+set -eo pipefail
 
 mkdir -p ./govulncheck 2>/dev/null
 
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/govulncheck-run.sh
```diff
@@ -13,23 +13,34 @@ FAILED=0
 # Repository prefix to remove from package names
 REPO_PREFIX=$(go list -m)
 
+if [[ "$GOVULN_OPTS" =~ .*--format[[:space:]]+([a-z]+).* ]]; then
+  FORMAT=${BASH_REMATCH[1]}
```
crobert-1, line 17: Can you add a comment explaining what's going on here?

PR #6044, opened 2025-03-27
[tiny][chore] Upgrade hashicorp/vault-plugin-auth-gcp to v0.20.1

pjanotti:
Supersedes #6015 that is incorrectly updating the toolchain


crobert-1 (approved):
_Approval pending passing tests_

PR #6071, opened 2025-03-31
[chore][codecov] Add coverage to auto-instrumentation workflow

crobert-1:


crobert-1:
This ends up only adding coverage to files under `packaging/tests` (test files), so I don't think this is helpful at this time.

PR #5974, opened 2025-03-12
[chore][deprecation] Remove testing for Oracle Linux 7

crobert-1:
**Description:** <Describe what has changed.>

Premier support has ended for Oracle Linux 7, and it's considered end of life (EOL) as of December 31, 2024. This removes testing for the platform.

Premier support has been ended, Oracle Linux Extended Support is available through June 2028.

[Source](https://blogs.oracle.com/linux/post/extend-your-os-upgrade-timeline-oracle-linux-7s-extended-support-with-expanded-coverage)

PR #6073, opened 2025-03-31
[chore][CI] Remove EOL distros from test matrices

pjanotti:
Both OpenSUSE 12 (leap 42) and CentOS 7 reached EOL. Removing them from the test matrices.


crobert-1 (commented):
We may want to consider adding a breaking changelog to notify users. (We added one the last time we [dropped testing for an EOL distro](https://github.com/signalfx/splunk-otel-collector/pull/5974))

pjanotti:
`puppet-rake-spec` failure is a known issue.

jinja2 (commented):
It looks like we have a copy of the readme content on the [puppetforge](https://forge.puppet.com/modules/signalfx/splunk_otel_collector/readme) page for our module. Should we be releasing a new version to push latest readme for the same? I am not sure how the docs are synced.

pjanotti:
I *think* the readme for all mass deployment tools is only updated when we make a new release for those (which are not tied to the collector release). That said it is time to make a new release for those.

PR #6075, opened 2025-03-31
[config/gateway] Update GW config and make it token passthrough ready

dloucasfx:
Update out of the box gw configs and make them passthrough ready

Goal:
-----
To keep things consistent we follow the sfx exporter passthrough logic.
If the token coming from the sender is not set or set to empty, use the exporter defined token.

Use Case 1:
-----------
Scenario: The headers_setter extension is defined in the exporter but include_metadata is false
Expected result: Use the token defined in the exporter , NOOP

Issue:
- The extension sets X-SF-Token with value from default_value; "" if default_value is not set

Solution:
- Set the default_value with the same value as the exporter

Use Case 2:
-----------
Scenario: The headers_setter extension is defined in the exporter and include_metadata is true
Criteria: Sender set X-SF-Token to empty string
Expected result: Use the token defined in the exporter

Issue:
- The extension does NOT use the default_value when the from_context (X-SF-TOKEN) is found. 
So we end up using empty string.

Solution:
- No way to use the default_value  (We agreed that this is a unlikely use case and if user explicitly set the token to empty string, then it's most likely intended )

Use Case 3:
-----------
Scenario: The headers_setter extension is defined in the exporter and include_metadata is true
Criteria: Sender does NOT send the X-SF-Token header
Expected result: Use the token defined in the exporter

Issue:
- The extension sets `X-SF-Token` with value from default_value; "" if default_value is not set.

Solution:
- Set the default_value with the same value as the exporter


dmitryax (approved):
LGTM. This also needs to be applied to the agent config (can be a separate PR)

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -2,6 +2,11 @@
 # data to Splunk products.
 
 extensions:
+  headers_setter:
+    headers:
+      - action: upsert
```
jinja2, line 7: Should we set `default_value` for this? I am just wondering what happens if the incoming req does not have the token header and does setting default to `"${SPLUNK_ACCESS_TOKEN}"` here help.

cmd/otelcol/fips/config/gateway_config.yaml
```diff
@@ -27,8 +32,12 @@ receivers:
     protocols:
       grpc:
         endpoint: "${SPLUNK_LISTEN_INTERFACE}:4317"
+        # Whether to preserve incoming access token and use instead of exporter token
```
jinja2, line 35:
```suggestion
        # Uncomment below config to preserve incoming access token and use it instead of the token value set in exporter config
```
just a tiny change to the wording to make it more clear for a user

cmd/otelcol/config/collector/gateway_config.yaml
dloucasfx: @jinja2 
This is a great use case you brought up! and correct, if the sender failed to set a value for the `X-SF-TOKEN` header, or there is a proxy in between that might be stripping down the headers, then the exporter on the GW will end up using an empty string as a value and fails to propagate the data to ingest.

But thinking about it little bit more, not setting a token value is the same as setting the wrong or expired token and user should be made aware of it (ie: from the export errors in the logs), otherwise it will "silently" work using the GW's default token and most likely it is not the intended goal.

dmitryax: Oh, I was under impression that it doesn't touch the header if cannot get anything from the context. This would be the most common use case with `include_metadata` disabled.

cmd/otelcol/fips/config/gateway_config.yaml
dloucasfx: Thanks! Agree! I used existing comment, updating all files...

cmd/otelcol/config/collector/gateway_config.yaml
dloucasfx: @dmitryax right, if the header does not exist, then yes, it won't touch it, but it's possible to send the header with an empty value. Which I think it would be an oversight from the user

dmitryax: > Which I think it would be an oversight from the user

I agree. I don't think we should address that

dloucasfx: Also note, the `default_value` is only used when the `X-SF-TOKEN` is not found in the `from_context`, but when it's found, regardless whether the value is empty string or not, it will not be used 
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/headerssetterextension/internal/source/context.go#L24-L26

dloucasfx: Found multiple issues with keeping `headers_setter` in the exporter:

1- Expanding on what Jina brought up, even when `include_metadata` is not used by the receiver, ie: the user does not want to use token passthrough, the exporter will end up using empty string for token.
When the request comes in, the headers are not kept (`include_metadata` is not set) the logic [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/headerssetterextension/internal/source/context.go#L24-L26) will end overwriting the GW token with the default value, which is an empty string

2 - two additional observations when `include_metadata` is set to true, ie: user intends to use token passthrough:
a- If the sender set the `X-SF-TOKEN` but not the value: non-default value would not work and empty string value will be used 
b- If the sender didn't add the header `X-SF-TOKEN` , non-default would work and the gw token will NOT be used. If non-default is not set, empty string will be used


I can't find a way to get the cover the intended behavior , if we set non-default value, then 2b would be a problem and what I have mentioned in earlier comments about lack of errors in case of misconfig.
If we don't set the default value, then we run in 1 and 2a

dmitryax: > b- If the sender didn't add the header X-SF-TOKEN , non-default would work and the gw token will NOT be used. If non-default is not set, empty string will be used

I'm not sure I understand this one. If user set `include_metadata` but doesn't set the `X-SF-TOKEN` we should use the default token as a fallback

dmitryax: So far it seems like we should set `default_value` to the configured and that should give as the expected behavior, unless I'm missing something

dloucasfx: Updated config based on the below:

Goal:
-----
To keep things consistent we follow the sfx exporter passthrough logic.
If the token coming from the sender is not set or set to empty, use the exporter defined token.

Use Case 1:
-----------
Scenario: The headers_setter extension is defined in the exporter but include_metadata is false
Expected result: Use the token defined in the exporter , NOOP

Issue:
- The extension sets X-SF-Token with value from default_value; "" if default_value is not set

Solution:
- Set the default_value with the same value as the exporter

Use Case 2:
-----------
Scenario: The headers_setter extension is defined in the exporter and include_metadata is true
Criteria: Sender set X-SF-Token to empty string
Expected result: Use the token defined in the exporter

Issue:
- The extension does NOT use the default_value when the from_context (X-SF-TOKEN) is found. 
So we end up using empty string.

Solution:
- No way to use the default_value (We agreed that this is a unlikely use case and if user explicitly set the token to empty string, then it's most likely intended )

Use Case 3:
-----------
Scenario: The headers_setter extension is defined in the exporter and include_metadata is true
Criteria: Sender does NOT send the X-SF-Token header
Expected result: Use the token defined in the exporter

Issue:
- The extension sets `X-SF-Token` with value from default_value; "" if default_value is not set.

Solution:
- Set the default_value with the same value as the exporter

jinja2 (approved):

jinja2:
Error in test seems to be from [this env var](https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/win-package-test.yml#L464) being passed 
```
Error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):

error decoding 'extensions': error reading configuration for "headers_setter": decoding failed due to the following error(s):

'headers[0].default_value' expected type 'string', got unconvertible type 'int', value: '12345'
2025/04/01 21:01:11 main.go:92: application run finished with error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):
```

PR #6078, opened 2025-04-01

srv-gh-o11y-gdi:


packaging/technical-addon/Splunk_TA_otel/configs/ta-agent-config.yaml
```diff
@@ -151,9 +151,6 @@ exporters:
     verbosity: detailed
 
 service:
-  telemetry:
```
jinja2, line 154: Should this be replaced with something like below so we respect the value set for SPLUNK_LISTEN_INTERFACE variable like the config being removed does?
```
  telemetry:
    metrics:
      readers:
        - pull:
            exporter:
              prometheus:
                host: "${env:SPLUNK_LISTEN_INTERFACE}"
```

dmitryax: I'm not sure why do we need to use `SPLUNK_LISTEN_INTERFACE` here. We don't have it in any other default config. It's localhost:8888 by default and it's not supposed to by accessed from the outside, the collector should be the only one calling that endpoint. I agree, potentially this might be a breaking change for someone who's scraping collector metrics from the outside, but I'd assume the chance of running into that is negligible for TA.

jinja2: I was referring to the usage of SPLUNK_LISTEN_INTERFACE in the target of the prometheus/internal receiver. I think we should just update the receiver to target `localhost` so things are consistent.

jinja2: TA config seems to be the only one using SPLUNK_LISTEN_INTERFACE for internal metrics. I think we'll be okay if we merge it as-is (will this even publish a new TA?)

dmitryax: Sounds good. We can take this as a follow-up before the TA release

dmitryax (approved):

jinja2 (approved):

PR #6081, opened 2025-04-02
[chore][CI] Remove or justify remaining qemu usages

pjanotti:
Remove or justify remaining qemu usages.


pjanotti:
Puppet test failure is expected

.github/workflows/integration-test.yml
```diff
@@ -208,6 +209,7 @@ jobs:
         with:
           name: docker-otelcol-${{ matrix.ARCH }}
           path: ./docker-otelcol/${{ matrix.ARCH }}
+      # Not all images used in the docker-compose have arm64 support, for those it is necessary to use qemu
```
crobert-1, line 212: Nit: It would be helpful to be clear about which images don't have arm64 support. This would make it easier in the future to know when we're able to remove qemu.

crobert-1 (approved):

.github/workflows/integration-test.yml
pjanotti: a4a5e5c

PR #6117, opened 2025-04-09
[chore][codecov] Add all packages to coverage results

crobert-1:

There are multiple go modules in this repository which should be included in codecov results. To be included, they must explicitly be listed in the value passed to go's `coverpkg` argument.


Makefile.Common
```diff
@@ -42,7 +42,10 @@ IMPI=impi
 BUILD_TYPE?=release
 
 # COVER_PKGS is the list of packages to include in the test coverage
-COVER_PKGS := $(shell go list ./... | tr "\n" ",")
+EXCLUDE_MODS=-not -path "./examples/*" -not -path "./tests/*"
```
crobert-1, line 45:
```suggestion
EXCLUDE_MODS=-not -path "./examples/*"
```
Although coverage for `./tests` isn't super helpful, it will be helpful to show if there's test helper code that isn't actually used (and can be deleted).

dmitryax (approved):

PR #6120, opened 2025-04-10
[chore] Fix Puppet CI: Lint and Rake spec

pjanotti:
`rake spec` started to fail a few weeks back and `puppet-lint` more recently. The later was clearly caused by the old puppet 6 (already reached EOL) being used in the container image used to run lint and rake tests not being available anymore.

Dropping puppet 6 from our test matrix and updating the container image and applying needed lint and rake fixes we can get both jobs passing again. Notes:

* The `metadata.json` is used by rake and not the test matrix of actual tests, the later seems to be what is reflected on the README.md
* A few lint checks had too many hits and are being skipped for the time being - creating a backlog item to track that


deployments/puppet/Dockerfile
```diff
@@ -1,24 +1,25 @@
...
+RUN dpkg --force-overwrite -i puppet-tools-release-bookworm.deb
+RUN apt update
+RUN apt install -y pdk
+RUN rm -f *.deb
 
 WORKDIR /etc/puppetlabs/code/modules/splunk_otel_collector
 COPY ./ ./
-RUN gem install bundler -v 2.4.22 && bundle install
+RUN pdk update
+RUN gem install bundler -v 2.4.13 && bundle install
```
crobert-1, line 22: Is there a reason we have to downgrade the bundler version?

deployments/puppet/spec/classes/init_spec.rb
```diff
@@ -9,7 +9,15 @@
...
   on_supported_os.each do |os, facts|
-    let(:facts) { facts }
+    # When running rake spec it checks for absolute paths and on Windows paths
+    # are built from facts. However, the needed facts are not set since rake
+    # spec is running on a non-windows system.
+    let(:facts) {{
+      'win_temp' =>'C:\\Windows\\Temp',
+      'win_programfiles' => 'C:\\Program Files',
+      'win_programdata' => 'C:\\ProgramData',
+      'win_systemdrive' => 'C:'}.merge( facts )
```
crobert-1, line 19: I'm concerned that these will break in a customer deployment. Is the added comment still relevant even in non-CI/CD environments? (Or is this logic even used in the deployment itself?)

_I'm not very familiar with this functionality, so just want to make sure._

deployments/puppet/Dockerfile
pjanotti: Not required, but if we keep the higher version it gives the following warning when building the image

```
but your lockfile was generated with 2.4.13. Installing Bundler 2.4.13 and restarting using that version. 
```

I didn't look into regenerating the lock file

deployments/puppet/spec/classes/init_spec.rb
pjanotti: This is the `spec` for the rake test. On actual deployments the facts are created as expected, i.e.: on Windows those facts are going to be present. Perhaps I need to update the comment to make that clear 🤔

deployments/puppet/manifests/collector_win_config_options.pp
```diff
@@ -27,5 +27,5 @@
             {}
         }
 
-    $collector_env_vars = stdlib::merge($base_env_vars, $gomemlimit, $listen_interface)
-}
\ No newline at end of file
+    $collector_env_vars = $base_env_vars + $gomemlimit + $listen_interface
```
pjanotti, line 30: Per documentation this is the same as stdlib::merge, see https://forge.puppet.com/modules/puppetlabs/stdlib/reference#stdlibmerge

deployments/puppet/spec/classes/init_spec.rb
pjanotti: 3a21cf9f

crobert-1 (approved):
🎉

PR #6129, opened 2025-04-14
Add GH release for the Splunk OTel .NET Deployer TA

pjanotti:
Following the pattern for the publication of mass deployment releases: when the TA version is updated on a specific file it triggers the creation of a tag that in turn triggers GitLab to create a signed build. For now this only opens pre-release and drafts on GitHub, publication on `splunkbase` will be addressed separately.


.gitlab-ci.yml
```diff
@@ -1838,3 +1838,77 @@ chef-release:
...
+  dependencies:
+    - dotnet-instrumentation-deployer-builder
+  script:
+    - pushd out/distribution/ && shasum -a 256 * > checksums.txt && popd
+    - |
+      set -e
+      if [[ "${CI_COMMIT_TAG:-}" =~ ^dotnet-instrumentation-deployer-ta-v([0-9]+\.[0-9]+\.[0-9]+[^ ]*) ]]; then
+        VERSION=${BASH_REMATCH[1]}
+        VERSION_CHANGELOG="$(awk -v version="$VERSION" '/^## / { if (p) { exit }; if ($2 == version) { p=1; next } } p && NF' CHANGELOG.md)"
+        SPLUNK_OTEL_DOTNET_VERSION=$(grep -oP 'version = \K.*' assets/windows_x86_64/bin/Splunk.OTel.DotNet.psm1 | tr -d '"')
```
pjanotti, line 1908: This is a bit fragile but since it is created as draft we can always fix before publishing it.

PR #6132, opened 2025-04-16
Adds a template to generate modular input and use it in a golang runner

hughesjj:
We've been talking about doing this for a while.  This expands on [Paulo's work](https://github.com/signalfx/splunk-otel-collector/tree/main/packaging/dotnet-instr-deployer-add-on/internal/modularinput) getting an XML reader for modular input and go binary for running the addon instead of the .cmd and .sh files we've historically used.

One of the most common frustrations we've seen is the duplicative logic between windows and linux that was necessary when using .cmd and .sh.  As we're now supporting three addons, and looking to potentially introduce flavors, this isn't tenable in the future.  It currently requires duplicative testing between windows and linux.  Given we can only currently test windows using splunk internal provisioning, this also makes development super annoying since we can only test linux with splunk-docker.

Writing the addon in golang allows us to reuse almost all of the same logic between windows and linux, allows us to use gotest on some of that logic, and given the significantly increased shared logic, we can be more confident that windows will work when running docker tests.

### DIFF
--- a/.github/workflows/splunk-ta-otel.yml
+++ b/.github/workflows/splunk-ta-otel.yml
@@ -73,6 +73,11 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
           cache-dependency-path: '**/go.sum'
 
+      - name: install tools
+        run: |
+          set -o pipefail
+          make install-tools
+
       - name: Test addon generator
         run: |
           set -o pipefail


packaging/technical-addon/internal/packaging/targz.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func PackageAddon(sourceDir, outputFile string) error {
```
pjanotti, line 11: I think this is not being used right now... anyway, why do you think this needs to be implemented directly? Portability?

pjanotti (commented):
I think it is looking good, will do a more detailed review once it is out of draft. It should have a CI workflow running the tests.

PR #6143, opened 2025-04-18

srv-gh-o11y-gdi:


jinja2:
Added a label so this isn't merged. We don't want to revert the cherrypicked commit and if I close it, scheduled workflow will reopen the same in a few hours.

PR #6165, opened 2025-04-22
[ps-installation-script] better error message in case of failed instrumentation installer download

lachmatt:
**Description:**
Produce a better message in case of failed attempt to download .NET instrumentation installer.
Follows the same pattern as other download attempts in the script (e.g. [1](https://github.com/signalfx/splunk-otel-collector/blob/bcb797f9dab0a85b0e54229e7580723e793ac43d/packaging/installer/install.ps1#L282-L288)).
Additionally, sets the security protocol (which might be needed on a legacy systems) similarly to other downloads in the script ([1](https://github.com/signalfx/splunk-otel-collector/blob/bcb797f9dab0a85b0e54229e7580723e793ac43d/packaging/installer/install.ps1#L280),[2](https://github.com/signalfx/splunk-otel-collector/blob/bcb797f9dab0a85b0e54229e7580723e793ac43d/packaging/installer/install.ps1#L252)).


packaging/installer/install.ps1
```diff
@@ -456,7 +456,17 @@ if ($with_dotnet_instrumentation) {
         $download = "https://github.com/signalfx/splunk-otel-dotnet/releases/latest/download/$module_name"
         $dotnet_autoinstr_path = Join-Path $tempdir $module_name
         echo "Downloading .NET Instrumentation installer ..."
-        Invoke-WebRequest -Uri $download -OutFile $dotnet_autoinstr_path -UseBasicParsing
+        try {
+            [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```
RassK, line 460: should include tls13?

lachmatt: I considered that but decided it'd be preferable to:
- set it consistently in the script (as similar is configured in multiplace places already)
- use enum values instead of magic numbers e.g.
 `[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12 -bor [Net.SecurityProtocolType]::Tls13`

RassK: Makes sense to separate it to a new issue then, if there are multiple places. The idea is not to require older security protocols but instead use newest as possible and fallback if not possible.

[Net.SecurityProtocolType]::Tls13 is available in .NET FX 4.8+. I think PowerShell 5.1 what we require is still at 4.5. So the enum is not available there.

RassK (approved):

PR #6166, opened 2025-04-23

srv-gh-o11y-gdi:


crobert-1:
_This is downgrading the otel-js version, I don't think we want this._

PR #6178, opened 2025-04-30

srv-gh-o11y-gdi:


crobert-1:
This is downgrading the dependency because of how https://github.com/signalfx/splunk-otel-js is tagging versions as latest. It's updating major versions in parallel, and the `v2` version happens to be tagged as latest currently.

crobert-1:
`latest` is currently set to `v2.16.1` as `v3.1.2` was released a day before.

PR #6184, opened 2025-04-30
update rules to not use deprecated syntax and allow for manual pushes

hughesjj:
1. [only/except is deprecated](https://docs.gitlab.com/ci/yaml/#only--except)
2. Allows for manual runs of technical addon branches if you manually protect them


pjanotti (approved):
I'm not familiar with the GitLab rules feature, but, the implied logic LGTM

PR #6189, opened 2025-05-02
[chore] Mark github.com/signalfx/signalfx-agent module as local

dmitryax:
Going forward we need to update all the usages of github.com/signalfx/signalfx-agent to github.com/signalfx/splunk-otel-collector/internal/signalfx-agent


crobert-1 (approved):
The thought process makes sense to me, LGTM once lint and the rest of CI is passing

PR #6201, opened 2025-05-05

srv-gh-o11y-gdi:


dmitryax (changes requested):
We don't want to downgrade

PR #6202, opened 2025-05-06
Add metrics needed for process overview panel

technimad-splunk:
**Description:** 
The splunk observability ootb dashboards contain a panel to show detailed process metrics. This PR adds the config to enable collection of the metrics needed for this dashboard panel to work.
Collection of process metrics is disabled by default, this is still the case. This PR adds commented config, this makes enabling the process collection more convenient, as all needed config is already provided.


cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -53,6 +53,11 @@ receivers:
       processes:
       # System processes metrics, disabled by default
       # process:
+      #   metrics:
+      #     process.cpu.utilization:
```
pjanotti, line 78: Could you please add a comment indicating that these 2 metrics are not enabled by default by should be enable if one wants to use the process panel?

atoulme:
Sorry for the long wait. Those metrics are to be enabled specifically by customers for a subset of processes and should not be on by default. We might need to link to docs here to make it more explicit to customers why and how they should go about this.

technimad-splunk:
The current docs link for this would be: https://help.splunk.com/en/splunk-observability-cloud/manage-data/available-data-sources/supported-integrations-in-splunk-observability-cloud/applications-hosts-and-servers/host-metrics-receiver#process-0

pjanotti:
Closing this PR, if appropriate, please, follow up in internally.

PR #6203, opened 2025-05-07
Rework TA makefiles

hughesjj:
This is some clean up, also adds target for new makefiles

### DIFF
--- a/Makefile.Common
+++ b/Makefile.Common
@@ -170,10 +170,12 @@ govulncheck:
 		exit 1; \
 	fi
 
+# CD's into SRC_ROOT so that this may be called/included from other Makefiles
 .PHONY: gogen
 gogen:
 	cd $(SRC_ROOT) && go generate ./...
 
+# CD's into SRC_ROOT so that this may be called/included from other Makefiles
 .PHONY: bundle.d
 bundle.d:
 	cd $(SRC_ROOT) && go install github.com/signalfx/splunk-otel-collector/internal/confmapprovider/discovery/bundle/cmd/discoverybundler


Makefile.Common
```diff
@@ -169,3 +169,13 @@ govulncheck:
...
+
+.PHONY: gogen
+gogen:
+	cd $(SRC_ROOT) && go generate ./...
+
+.PHONY: bundle.d
+bundle.d:
+	cd $(SRC_ROOT) && go install github.com/signalfx/splunk-otel-collector/internal/confmapprovider/discovery/bundle/cmd/discoverybundler
+	cd $(SRC_ROOT) && go generate -tags bootstrap.bundle.d ./...
+	cd $(SRC_ROOT) && go generate -tags bundle.d ./...
```
crobert-1, line 181: Typically we use `Makefile.Common` for targets that are relevant for every directory that has a reference to it. The top-level `Makefile` is used for repo-wide targets.

Is there a reason to have these targets in `Makefile.Common` when they're not directory specific, and just needed for the repo as a whole? (If it's in `Makefile` we can remove the `cd` commands and simply run the requested operation.)

packaging/technical-addon/Makefile
```diff
@@ -1,12 +1,19 @@
 # Used for building
-SOURCE_DIR=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
-BUILD_DIR=$(realpath $(SOURCE_DIR)/../../)/build
-COMMON_MAKEFILE:=$(abspath $(SOURCE_DIR)/../../Makefile.Common)
-
+THIS_MAKEFILE_DIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
```
crobert-1, line 2: There's a lot of path logic here, is there some reason we can't just use relative paths and `include` from there? (In other words, is there some reason we can't just do `include ../../Makefile.Common` and `include ../../instrumentation/Makefile`?)

packaging/technical-addon/Makefile
```diff
@@ -31,89 +40,130 @@ MODINPUT_CONFIG_GENERATOR := $(TOOLS_DIR)/modinput_config_generator
...
+	go build -o $(MODINPUT_CONFIG_GENERATOR) ./cmd/modinput_config_generator
 
 .PHONY: test-ta-build-tools
 test-ta-build-tools: ta-build-tools
-	cd $(SOURCE_DIR) && go build -o $(TOOLS_DIR)/modinput_config_generator ./cmd/modinput_config_generator
-	$(MODINPUT_CONFIG_GENERATOR) -source-dir=$(SOURCE_DIR)/cmd/modinput_config_generator/internal/testdata -schema-name=Sample_Addon --build-dir=$(BUILD_DIR) || exit 1; \
+	$(MODINPUT_CONFIG_GENERATOR) -source-dir=$(ADDONS_SOURCE_DIR)/cmd/modinput_config_generator/internal/testdata -schema-name=Sample_Addon --build-dir=$(BUILD_DIR) || exit 1; \
 	mkdir -p $(BUILD_DIR)/Sample_Addon/$(PLATFORM)_$(SPLUNK_ARCH)/bin; \
-	cd $(SOURCE_DIR) && go build $(GO_BUILD_FLAGS) -o $(BUILD_DIR)/Sample_Addon/$(PLATFORM)_$(SPLUNK_ARCH)/bin/Sample_Addon ./cmd/modinput_config_generator/internal/testdata/pkg/sample_addon/runner || exit 1; \
+	go build $(GO_BUILD_FLAGS) -o $(BUILD_DIR)/Sample_Addon/$(PLATFORM)_$(SPLUNK_ARCH)/bin/Sample_Addon $(ADDONS_SOURCE_DIR)/cmd/modinput_config_generator/internal/testdata/pkg/sample_addon/runner || exit 1; \
```
crobert-1, line 49: Probably unrelated to this PR, feel free to drop: When I see a target that starts `test-`, I assume it's simply for testing. Is there a reason this target has to always build first? Also, why is this building manually rather than using a `build-` target of some kind?

packaging/technical-addon/Makefile
```diff
@@ -31,89 +40,130 @@ MODINPUT_CONFIG_GENERATOR := $(TOOLS_DIR)/modinput_config_generator
...
+			mkdir -p $(BUILD_DIR)/$$schema/$${platform}_$(SPLUNK_ARCH)/bin; \
+			GOOS=$$platform GOARCH=$$arch go build $(GO_BUILD_FLAGS) -o $(BUILD_DIR)/$$schema/$${platform}_$(SPLUNK_ARCH)/bin/$$schema $(ADDONS_SOURCE_DIR)/pkg/$$schema_lower/runner || exit 1; \
+		done \
+	done
+
+.PHONY: generate-technical-addon-linux-autoinstrumentation
+generate-technical-addon-linux-autoinstrumentation:
+	cd $(AUTOINSTRUMENTATION_DIR) && $(MAKE) dist
+	mkdir -p $(BUILD_DIR)
+	PLATFORM="all" \
```
crobert-1, line 84: I would have assumed the `PLATFORM` here would be set to `linux` or something along those lines since this target is for `linux-`. Why is it set to `all`?

packaging/technical-addon/Makefile
```diff
@@ -31,89 +40,130 @@ MODINPUT_CONFIG_GENERATOR := $(TOOLS_DIR)/modinput_config_generator
...
+	BUILD_DIR="$(BUILD_DIR)" \
+	SOURCE_DIR="$(ADDONS_SOURCE_DIR)" \
+	AUTOINSTRUMENTATION_DIR="$(AUTOINSTRUMENTATION_DIR)" \
```
crobert-1, line 87: I'm seeing a common pattern in this Makefile is to have commands setting variables at the end. What is the value and purpose of this? Is `AUTOINSTRUMENTATION_DIR` going to be used after this for something else?

crobert-1 (commented):
Mostly questions as I'm not very familiar with TA logic and what's being changed here 👍

Makefile.Common
crobert-1: I'm not very good with best practice Makefile layouts myself, I just generally stick with whatever's being done upstream in `contrib`. The general technique there is sub directories have their own `Makefile` which import the top level `Makefile.Common`. The top-level `Makefile` also imports `Makefile.Common`, and then adds repo-wide make targets. (That may not be best here, just adding context for my original comment.)

> The difficulty with the targets in the technical addon is that many (most?) targets require building the bundle (for discovery mode) and the collector (I mean, what doesn't?), and we're also pulling some stuff from the autoinstrumentation makefile now.

I think my ignorance is related to this comment. To make sure I'm following,"many targets" is referring to targets found in which Makefile (or Makefiles)? You're referencing `packaging/technical-addon/Makefile`?

crobert-1: Okay, I'll leave this up to you then. At this point, I don't think the `cd` is needed here as it's not being called from any sub-directory yet. It sounds like you've got more work going on that's building on top of this, and maybe this is helpful for local development. It's technically a no-op for running it at the top-level directory anyways, so it's fair to leave. Just confusing without the context.

packaging/technical-addon/Makefile
crobert-1: That's fair. It would be helpful to me to either stub out the "build a test environment" to its own make target and call it here, or add a comment above this target to clarify.

packaging/technical-addon/Makefile
```diff
@@ -31,89 +29,116 @@ MODINPUT_CONFIG_GENERATOR := $(TOOLS_DIR)/modinput_config_generator
...
+	@echo "Generating runner configs..."
+	@for schema in $(MODULAR_INPUT_SCHEMAS); do \
+		schema_lower=$$(echo $$schema | tr '[:upper:]' '[:lower:]'); \
+		echo "Generating config and TA scaffold for $$schema..."; \
+		mkdir -p $(BUILD_DIR)/$$schema/; \
+		$(MODINPUT_CONFIG_GENERATOR) -source-dir=$(ADDONS_SOURCE_DIR) -schema-name=$$schema --build-dir=$(BUILD_DIR) || exit 1; \
+		ls -lAh $(BUILD_DIR); \
+	done
+
+# Build all runner binaries.. This may require some more build deps
```
crobert-1, line 53:
```suggestion
# Build all runner binaries. This may require some more build deps
```
Is the second sentence meant to be a TODO, or warning, or something else?

jvoravong (approved):

PR #6256, opened 2025-05-27
Clean up some of the socket proxy code to simplify code

atoulme:
fly by PR as I was rereading this code and found it could be tidied up a bit.


jvoravong:
The failing tests look legitimate and related. Please look into them.
> Error: receivers/mysql/bundled_test.go:35:2: declared and not used: dockerSocketProxy
> Error: receivers/mysql/bundled_test.go:56:98: undefined: dockerSocket

PR #6258, opened 2025-05-28

srv-gh-o11y-gdi:


jvoravong:
The lint check is failing because we use deprecated testcontainer methods. The functions we previously used to create testcontainer networks with static names (e.g. NetworkRequest and CreateNetwork) are no longer supported in the latest testcontainers-go versions. The latest API (network.New) no longer allows setting a fixed network name, network names are now auto-generated. Because of this, we'll need to refactor how we set up multiple containers to use the same network.

go.mod
```diff
@@ -762,3 +768,7 @@ replace github.com/pires/go-proxyproto v1.0.0 => github.com/peteski22/go-proxypr
 
 // github.com/veraison/go-cose v1.2.0 doesn't exists but required by the latest github.com/Microsoft/hcsshim
 replace github.com/veraison/go-cose v1.2.0 => github.com/veraison/go-cose v1.1.1
+
+// This is only needed for splunk-otel-collector 0.127.0 and can be removed in 0.128.0
+// See: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40103/files#diff-2bb8e77bc0a07447b4df98ee53c775b57b842fe5d8a649c52506406aeb18c1c8
+replace github.com/prometheus/common v0.64.0 => github.com/prometheus/common v0.62.0
```
jvoravong, line 774: My IDE was auto-resolving github.com/prometheus/common to v0.64.0, which causes compile errors. This appears to be a known upstream issue that’s fixed in a later version. I’ve added a temporary replace directive for 0.127.0 to pin the correct version, this should be removed before the 0.128.0 release.

jvoravong:
Lint is now failing for several instances of errors like these.

> go.opentelemetry.io/collector/semconv/v1.22.0" is deprecated: [v0.126.0] This package has been deprecated, use go.opentelemetry.io/otel/semconv instead. (staticcheck)

pjanotti (approved):

PR #6259, opened 2025-05-28
[discovery] Enable more metrics by default for SQL Server receiver

crobert-1:

This is to match out of the box content.


jvoravong (approved):
LGTM. The lychee link check failures are fixed in main and can be ignored in this PR.

PR #6271, opened 2025-05-30
[deployment/chef] Fix custom variables handling on Windows

crobert-1:

This fixes how Chef handles custom variables for the `splunk-otel-collector` on Windows deployments. The MSI doesn't support custom variables, it supports a set of known variables. To properly set custom variables we need to use the registry.

Also, update version of the collector being used in testing.

### DIFF
--- a/deployments/chef/metadata.rb
+++ b/deployments/chef/metadata.rb
@@ -3,7 +3,7 @@ maintainer 'Splunk, Inc.'
 maintainer_email 'signalfx-support@splunk.com'
 license 'Apache-2.0'
 description 'Install/Configure the Splunk OpenTelemetry Collector'
-version '0.16.0'
+version '0.18.0'
 chef_version '>= 16.0'
 
 supports 'amazon'
--- a/deployments/chef/metadata.rb
+++ b/deployments/chef/metadata.rb
@@ -3,7 +3,7 @@ maintainer 'Splunk, Inc.'
 maintainer_email 'signalfx-support@splunk.com'
 license 'Apache-2.0'
 description 'Install/Configure the Splunk OpenTelemetry Collector'
-version '0.18.0'
+version '0.17.0'
 chef_version '>= 16.0'
 
 supports 'amazon'
--- a/deployments/chef/recipes/default.rb
+++ b/deployments/chef/recipes/default.rb
@@ -14,7 +14,7 @@ if platform_family?('windows')
 
   # Older MSI versions can't properly setup the collector configuration
   # in this case, we need to use the registry to set the environment variables.
-  # Testing also showed custom configuration variables also require this.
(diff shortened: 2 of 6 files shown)


deployments/chef/metadata.rb
```diff
@@ -3,7 +3,7 @@
 maintainer_email 'signalfx-support@splunk.com'
 license 'Apache-2.0'
 description 'Install/Configure the Splunk OpenTelemetry Collector'
-version '0.16.0'
+version '0.18.0'
```
crobert-1: Changelog is a bit off here. There's a [changelog entry for `0.17.0`](https://github.com/signalfx/splunk-otel-collector/blob/main/deployments/chef/CHANGELOG.md#chef-v0170), but there was never an update in the release version, and no release tag for it. I'm updating to `0.18.0` to make it clear there's a new release which includes `0.17.0`, but it's in addition to `0.17.0`.

If anyone feels strongly, happy to address this differently. I could just add this PR to `0.17.0` and create the release and it won't make any difference to the end user.

deployments/chef/recipes/default.rb
```diff
@@ -12,9 +12,10 @@
   include_recipe 'splunk_otel_collector::collector_win_config_options'
   include_recipe 'splunk_otel_collector::collector_win_install'
 
-  # Older MSI versions can't properly setup the collector configuration
+  # Older MSI versions can't properly setup the collector configuration,
```
crobert-1:
```suggestion
  # Older MSI versions can't properly setup the collector configuration
```

deployments/chef/recipes/default.rb
```diff
@@ -14,7 +14,8 @@
 
   # Older MSI versions can't properly setup the collector configuration
   # in this case, we need to use the registry to set the environment variables.
-  unless node['splunk_otel_collector']['collector_msi_is_configurable']
+  # Testing also showed custom configuration variables also require this.
```
crobert-1:
```suggestion
  # Custom variables also require using the registry, as the MSI doesn't support it.
```

deployments/chef/metadata.rb
jvoravong: Recommendation:
- If a 0.17.0 release already exists, proceed with creating a 0.18.0 release.
- If a 0.17.0 release does not exist, you have two options:
  - Create a 0.17.0 release that includes all new changes since the last release. The new changes in this PR would also go into 0.17.0.
  - Create both 0.17.0 and 0.18.0 releases, ensuring that each release contains the appropriate changes unique to its version.

crobert-1: Yeah, good points all around. I think I'll just move this PR's changes into `0.17.0`. This will make sure there isn't a dead reference to `0.17.0` as there is currently 👍

PR #6284, opened 2025-06-02
[chore][ansible] Update linux tests to use locally built artifact

crobert-1:

Changes have been broken down into individually described pieces below.

**GitHub Workflow**

Add build steps for building the linux packages locally. The built artifacts are then downloaded by the ansible linux test job under the runner's `/tmp` directory. As a result of needing to build packages and upload/download them, I changed the workflow's default path to be the repo's head, and the ansible test steps have the working directory as `deployments/ansible`.

**Ansible functionality changes**

When `install_local_artifact` is enabled, the `splunk-otel-collector` image is installed from `/tmp` on the controller machine. Functionality should remain unchanged when this variable is false.

**Ansible test changes** 

Updated all existing molecule tests to use `install_local_artifact` which means that the package built from source is being used in testing, rather than a remote released version. Added a new test that uses a set version of the `splunk-otel-collector` to make sure that remote downloading and installation still works as expected on each platform.


deployments/ansible/molecule/default_install_remote_version/Dockerfile.j2
crobert-1, line 1: Note: Everything under `default_install_remote_version` was a direct copy from `default`. The only changes are as follows: 
1. Remove the setting of `install_local_artifact`
2. Set `splunk_otel_collector_version`

deployments/ansible/molecule/with_instrumentation/converge.yml
```diff
@@ -3,10 +3,11 @@
   hosts: all
   become: yes
   vars:
+    install_local_artifact: true
     splunk_access_token: fake-token
     splunk_realm: fake-realm
     install_splunk_otel_auto_instrumentation: true
-    splunk_otel_auto_instrumentation_version: 0.126.0
+    splunk_otel_auto_instrumentation_version: latest
```
crobert-1, line 10: Keeping up to date with the `latest` is what I believe is the best option available, as this artifact is not published by the `splunk-otel-collector` repo.

.github/workflows/ansible.yml
crobert-1, line 1: The added build steps in this workflow are copied from the [linux_package_test workflow](https://github.com/signalfx/splunk-otel-collector/blob/ce02a29c81230676f2ed4e539c0bcb84dfe1392e/.github/workflows/linux-package-test.yml).

deployments/ansible/molecule/custom_vars/verify.yml
crobert-1, line 1: I removed the version check as the local artifact version would need to be parsed from filenames and then set as ansible variables to compare. This seems more complicated than it's worth, but I can the required logic and check if anyone feels strongly.

deployments/ansible/roles/collector/defaults/main.yml
```diff
@@ -101,3 +101,7 @@ splunk_dotnet_auto_instrumentation_enable_profiler: false
 splunk_dotnet_auto_instrumentation_enable_profiler_memory: false
 splunk_dotnet_auto_instrumentation_iisreset: true
 splunk_dotnet_auto_instrumentation_additional_options: {}
+
+# Variables to be used only for testing
```
crobert-1, line 105: Since `install_local_artifact` is only for testing I didn't add any information on usage and functionality in the `README`. I'd say with test variables I'd prefer to not document them publicly.

_Willing to hear from others if there's any disagreement here though._

jvoravong: In Ansible, parameters like install_local_artifact={{ some_path }} are a common pattern for deploying in environments with limited remote access. To avoid confusion or blocking that use case later, I’d suggest renaming this to something like local_testing_enabled and reserving install_local_artifact for future local path-based support.

deployments/ansible/molecule/with_instrumentation/converge.yml
jvoravong: This seems fine since splunk_otel_auto_instrumentation_version updates infrequently. At worst, a new splunk_otel_auto_instrumentation_version  release could occasionally cause a flaky Asnbile idempotency test failure, but that should be rare and low impact.

jvoravong (approved):
A couple small comments, overall LGTM and approving. I’ve used Ansible before but not with this project, so it maybe let this sit for other reviews as well.

deployments/ansible/roles/collector/defaults/main.yml
crobert-1: Thanks for the extra information, I wasn't aware of that naming pattern. I'll rename it 👍

PR #6265, opened 2025-05-29
[chore] Upgrade release version used in mass deployment testing

crobert-1:
**Description:** <Describe what has changed.>

Many mass deployment tests are using old versions of the Splunk distribution of the OpenTelemetry Collector. This updates the versions being used to latest.

PR #6288, opened 2025-06-04
[chore][ansible] Upgrade version tested for Windows

crobert-1:


crobert-1:
`Ansible / Linux Test` will pass once https://github.com/signalfx/splunk-otel-collector/pull/6265 is merged and this is rebased on top.

.github/workflows/ansible.yml
```diff
@@ -187,13 +187,21 @@ jobs:
...
+        run: |
+          mkdir /tmp/log
+          molecule --debug -v --base-config ./molecule/config/windows.yml test -s ${{ matrix.scenario }} -p ${{ matrix.distro }}
         env:
           PY_COLORS: '1'
           ANSIBLE_FORCE_COLOR: '1'
           MOLECULE_VAGRANT_BOX_URL: /tmp/vagrant.box
           MOLECULE_VAGRANT_BOX_VERSION: " "  # version must not be set when using a local box
 
+      - name: Uploading log files
```
crobert-1: These logs are coming from the specific step of installing the MSI currently. We could potentially just upload logs on molecule test failure as that would save space, but if the failure messages aren't very clear it's helpful to have successful logs as well for comparing.

The log files aren't very big (~1.5MB each), so I think we should keep them for now.

.github/workflows/ansible.yml
```diff
@@ -187,13 +187,22 @@ jobs:
...
+        if: success() || failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: ansible-molecule-windows-${{ matrix.distro }}-${{matrix.ansible}}-${{ matrix.scenario }}
+          path: /tmp/log/*
```
jvoravong:
```suggestion
          path: /tmp/log/*
          retention-days: 5
```

.github/workflows/ansible.yml
crobert-1: After testing, I've found that keeping the log files and adding the async behavior has not provided any value. 

Async was added because if the ansible command for installing the MSI didn't complete before the github job timed out, the job didn't complete, so no logs were kept. However, even after adding the ansible async functionality, if async is what triggers the ansible command to stop, the logs are once again dropped from the failed install. The were never any async logs either at the path specified.

In summary, we never actually got logs from a failed installation. If we only get successful logs, we may as well drop them. _I'll remove from this PR._

deployments/ansible/molecule/custom_vars/windows-converge.yml
```diff
@@ -8,9 +8,9 @@
     splunk_api_url: https://fake-splunk-api.com
     splunk_hec_url: https://fake-splunk-hec.com
     splunk_hec_token: fake-hec-token
-    splunk_otel_collector_version: 0.48.0
+    splunk_otel_collector_version: 0.126.0
     splunk_otel_collector_config: '{{ansible_env.ProgramData}}\Splunk\OpenTelemetry Collector\custom_config.yml'
-    splunk_otel_collector_config_source: ./custom_collector_config_deprecated.yml
+    splunk_otel_collector_config_source: ./custom_collector_config.yml
```
crobert-1, line 13: This was causing a generic failure on Windows, as the deprecated config was invalid. The failure message on Windows was the service timed out starting up.

deployments/ansible/roles/collector/tasks/otel_win_install.yml
```diff
@@ -61,12 +61,91 @@
   set_fact:
     msi_arguments: "{{ msi_unfiltered_arguments | reject('eq', '') | list }}"
 
+- set_fact:
```
crobert-1, line 64: In testing, if variables without spaces have double quotes, MSI installation fails. This logic then separates arguments into two lists, one that has spaces, one without. The args without spaces are left unchanged. Arguments with spaces have the value surrounded by double quotes. After this, the lists are then combined again into the `msi_arguments` fact.

deployments/ansible/roles/collector/tasks/otel_win_install.yml
```diff
@@ -77,21 +156,18 @@
 - name: Merge custom config into the default config
   ansible.builtin.import_tasks: config_override.yml
   when:
-    - splunk_collector_msi_is_configurable is undefined
```
crobert-1, line 80: MSI installation cannot directly set custom variables. This means that the old functionality for setting service registry keys must still be used for custom variables, even if the MSI is configurable.

deployments/ansible/roles/collector/tasks/otel_win_reg.yml
```diff
@@ -1,17 +1,57 @@
...
+
 - name: Set Splunk OpenTelemetry Collector registry value
+  when:
+    - splunk_otel_collector_options_list is defined
   ansible.windows.win_regedit:
     path: "{{ splunk_otel_collector_service_registry_key }}"
     state: present
     name: Environment
-    data: "{{ splunk_otel_collector_options_list | sort }}"
+    data: "{{ splunk_otel_collector_options_list | sort | unique }}"
```
crobert-1, line 43: For the sake of [Ansible's idempotency check](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_intro.html#desired-state-and-idempotency), we need to make sure that entries in the registry are unique (which is why the `| unique` statement is added). Without this check, the custom variables are simply added again, resulting in duplicates.

deployments/ansible/roles/collector/tasks/otel_win_reg.yml
```diff
@@ -1,17 +1,57 @@
...
+  when:
+    - splunk_collector_msi_is_configurable is defined
+    - splunk_otel_collector_options_list is defined
+  set_fact:
+    splunk_otel_collector_options_list:
+      "{{ registry_contents.properties.Environment.value + splunk_otel_collector_options_list }}"
+
 - name: Set Splunk OpenTelemetry Collector registry value
+  when:
+    - splunk_otel_collector_options_list is defined
```
crobert-1, line 38: This variable is not set when the MSI is configurable and there aren't any custom variables, so there's no work to do in that case.

deployments/ansible/CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
   configure the command line arguments for the Splunk OpenTelemetry Collector
   service.
 
+### 🧰 Bug fixes 🧰
+
+- Fix bug causing MSI installation failure when specified variables had spaces in their values (relevant when `splunk_otel_collector_version` >= `0.98.0`). ([#6288](https://github.com/signalfx/splunk-otel-collector/pull/6288))
```
pjanotti:
```suggestion
- Fix bug causing Windows deployment failure when specified variables had spaces in their values (relevant when `splunk_otel_collector_version` >= `0.98.0`). ([#6288](https://github.com/signalfx/splunk-otel-collector/pull/6288))
```

deployments/ansible/CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
...
+- Fix bug causing custom variables to not be set for MSI installation (relevant when `splunk_otel_collector_version` >= `0.98.0`). ([#6288](https://github.com/signalfx/splunk-otel-collector/pull/6288))
```
pjanotti:
```suggestion
- Fix bug causing custom variables to not be set on Windows deployment (relevant when `splunk_otel_collector_version` >= `0.98.0`). ([#6288](https://github.com/signalfx/splunk-otel-collector/pull/6288))
```

PR #6303, opened 2025-06-09
[ansible] Fix incorrect environmental variable naming

crobert-1:

Without the `register` and `failed_when` checks, the test passes with output saying that lines would have been added if values don't match. The goal is to test to see if an env var is set, by matching lines in a file. If the env var is set to the expected value, no lines would be added. So rather than failing the test, there was just a log message saying a line would be added. This change makes it so tests fail if the env var isn't set to the expected value.

This is already being done in other tests, `with_instrumentation` [for example.](https://github.com/signalfx/splunk-otel-collector/blob/be1bce1d7544b25a2f51e5411d5c076af7a195bb/deployments/ansible/molecule/with_instrumentation/verify.yml#L85)

Edit: After adding verification, `custom_vars` was failing. This was happening because in https://github.com/signalfx/splunk-otel-collector/pull/6270 the variable was named as `OTELCOL_CONFIG`, but the test was checking for `OTELCOL_OPTIONS`. I think `OTELCOL_OPTIONS` was more clear so I'm updating the environment variable, rather than the test.

### DIFF
--- a/deployments/ansible/galaxy.yml
+++ b/deployments/ansible/galaxy.yml
@@ -2,7 +2,7 @@
 namespace: signalfx
 name: splunk_otel_collector
 description: Ansible collection for Splunk OpenTelemetry Collector
-version: 0.32.0
+version: 0.31.0
 readme: README.md
 authors:
   - Splunk Inc.
--- a/deployments/ansible/galaxy.yml
+++ b/deployments/ansible/galaxy.yml
@@ -2,7 +2,7 @@
 namespace: signalfx
 name: splunk_otel_collector
 description: Ansible collection for Splunk OpenTelemetry Collector
-version: 0.31.0
+version: 0.32.0
 readme: README.md
 authors:
   - Splunk Inc.
--- a/deployments/ansible/roles/collector/templates/splunk-otel-collector.conf.j2
+++ b/deployments/ansible/roles/collector/templates/splunk-otel-collector.conf.j2
@@ -1,4 +1,4 @@
-OTELCOL_CONFIG={{ splunk_otel_collector_command_line_args }}
+OTELCOL_OPTIONS={{ splunk_otel_collector_command_line_args }}
 SPLUNK_CONFIG={{ splunk_otel_collector_config }}
 SPLUNK_ACCESS_TOKEN={{ splunk_access_token }}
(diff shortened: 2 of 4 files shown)


deployments/ansible/galaxy.yml
```diff
@@ -2,7 +2,7 @@
 namespace: signalfx
 name: splunk_otel_collector
 description: Ansible collection for Splunk OpenTelemetry Collector
-version: 0.31.0
+version: 0.32.0
```
crobert-1:
```suggestion
version: 0.31.0
```
There are other bug fixes for `0.32.0` included in https://github.com/signalfx/splunk-otel-collector/pull/6288, I'll wait until both of these PRs are merged and then create a release.

deployments/ansible/roles/collector/templates/splunk-otel-collector.conf.j2
```diff
@@ -1,4 +1,4 @@
-OTELCOL_CONFIG={{ splunk_otel_collector_command_line_args }}
+OTELCOL_OPTIONS={{ splunk_otel_collector_command_line_args }}
```
jinja2, line 1: Just to confirm, `OTELCOL_OPTIONS` is the env var expected by the installer, right? Since I see only one place where this is being updated, I assume that this env var is consumed somewhere when starting the otelcol bin in this form (maybe in ExecStart).

pjanotti: Yes, this is correct.

crobert-1: @pjanotti Can you share how (or where) this is being used? I only see it being used in the test at this point.

jinja2 (approved):

deployments/ansible/roles/collector/templates/splunk-otel-collector.conf.j2
pjanotti: The collector packages use `OTELCOL_OPTIONS` at the `ExecStart` configured via package deployment https://github.com/signalfx/splunk-otel-collector/blob/be1bce1d7544b25a2f51e5411d5c076af7a195bb/packaging/fpm/splunk-otel-collector.service#L7

PR #6304, opened 2025-06-11
Add version change support to Windows install script

pjanotti:
**Description:**
Change the Windows install script to support changing the collector version, without requiring manual uninstall.

**Testing:**
Added a new test to cover the main scenario for this change

### DIFF
--- a/.github/workflows/win-package-test.yml
+++ b/.github/workflows/win-package-test.yml
@@ -287,6 +287,37 @@ jobs:
         if: ${{ always() && matrix.WITH_FLUENTD == 'true' }}
         run: Get-Content -Path "${env:SYSTEMDRIVE}\opt\td-agent\td-agent.log"
 
+  windows-script-upgrade-test:
+    strategy:
+      matrix:
+        OS: [ "windows-2022" ]
+    runs-on: otel-windows # Tests with uninstall require a more powerful runner
+    needs: [msi-test, windows-install-script-test]
+    timeout-minutes: 45
+    steps:
+      - name: Check out the codebase.
+        uses: actions/checkout@v4
+
+      - name: Downloading msi build
+        uses: actions/download-artifact@v4
+        with:
+          name: msi-build-${{ matrix.OS }}
+          path: ./dist
+
+      - name: Set the MSI_COLLECTOR_PATH and INSTALL_SCRIPT_PATH environment variable
+        run: |
+          $ErrorActionPreference = 'Stop'
+          $msi_path = Resolve-Path .\dist\splunk-otel-collector*.msi
+          Test-Path $msi_path
+          "MSI_COLLECTOR_PATH=$msi_path" | Out-File -FilePath $env:GITHUB_ENV -Append
+          $ps1_path = Resolve-Path .\packaging\installer\install.ps1
+          Test-Path $ps1_path
+          "INSTALL_SCRIPT_PATH=$ps1_path" | Out-File -FilePath $env:GITHUB_ENV -Append
+
+      - name: Run the script upgrade tests
+        run: |
+          go test -timeout 15m -v github.com/signalfx/splunk-otel-collector/tests/windows-install-script
+
   choco-build:
     runs-on: ${{ matrix.OS }}
     strategy:
--- a/.github/workflows/win-package-test.yml
+++ b/.github/workflows/win-package-test.yml
@@ -290,7 +290,7 @@ jobs:
   windows-script-upgrade-test:
     strategy:
       matrix:
-        OS: [ "windows-2022" ]
+        OS: [ "windows-2022", "windows-2025" ]
     runs-on: otel-windows # Tests with uninstall require a more powerful runner
     needs: [msi-test, windows-install-script-test]
     timeout-minutes: 45
@@ -321,7 +321,7 @@ jobs:
       - name: Run the script upgrade tests
         shell: powershell # Use PowerShell to run the script since running the go test from pwsh.exe hides the Get-ExecutionPolicy cmdlet.
         run: |
-          go test -timeout 15m -v github.com/signalfx/splunk-otel-collector/tests/windows-install-script
+          go test -v github.com/signalfx/splunk-otel-collector/tests/windows-install-script
 
   choco-build:
     runs-on: ${{ matrix.OS }}
--- a/.github/workflows/win-package-test.yml
+++ b/.github/workflows/win-package-test.yml
@@ -303,6 +303,10 @@ jobs:
         with:
           name: msi-build-${{ matrix.OS }}
           path: ./dist
+      - name: Ensure required ports in the dynamic range are available
+        run: |
+          $ErrorActionPreference = 'Continue'
+          & ${{ github.workspace }}\.github\workflows\scripts\win-required-ports.ps1
 
       - name: Set the MSI_COLLECTOR_PATH and INSTALL_SCRIPT_PATH environment variable
         run: |
@@ -315,6 +319,7 @@ jobs:
           "INSTALL_SCRIPT_PATH=$ps1_path" | Out-File -FilePath $env:GITHUB_ENV -Append
 
       - name: Run the script upgrade tests
+        shell: powershell # Use PowerShell to run the script since running the go test from pwsh.exe hides the Get-ExecutionPolicy cmdlet.
         run: |
           go test -timeout 15m -v github.com/signalfx/splunk-otel-collector/tests/windows-install-script
 
--- a/packaging/installer/install.ps1
+++ b/packaging/installer/install.ps1
@@ -124,6 +124,10 @@
     If specified, the -mode parameter will be ignored.
     .EXAMPLE
     .\install.ps1 -config_path "C:\SOME_FOLDER\my_config.yaml"
+.PARAMETER preserve_prev_default_config
+   (OPTIONAL) Preserve the default configuration files, located at `$Env:ProgramData\Splunk\OpenTelemetry Collector`, of previous version when upgrading the collector. By default it is $false since version changes can include breaking configuration changes.
+   .EXAMPLE
+    .\install.ps1 -preserve_prev_default_config $true
 #>
 
 param (
@@ -145,6 +149,7 @@ param (
     [string]$msi_path = "",
     [string]$msi_public_properties = "",
     [string]$config_path = "",
+    [bool]$preserve_prev_default_config = $false,
     [string]$collector_msi_url = "",
     [string]$fluentd_msi_url = "",
     [string]$dotnet_psm1_path = "",
@@ -370,8 +375,27 @@ function download_collector_package([string]$collector_version=$collector_versio
 }
 
 # check registry for the agent msi package
-function msi_installed([string]$name) {
-    return (Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Where { $_.DisplayName -eq $name }) -ne $null
+function is_msi_installed([string]$name) {
+    return $null -ne (Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Where { $_.DisplayName -eq $name })
+}
+
+function get_msi_installation_sids([string]$name) {
+    $sids = [string[]]@()
+
+    $uninstallEntry = Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* -ErrorAction SilentlyContinue | 
+        Where-Object { $_.DisplayName -eq $name }
+    if ($uninstallEntry) {
+        $userInstalls = Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\*\Products\*\InstallProperties' -ErrorAction SilentlyContinue |
+            Where-Object { $_.DisplayName -eq $name }
+        foreach ($user in $userInstalls) {
+            if ($user.PSPath -match 'UserData\\(?<SID>S-1-[0-9\-]+)') {
+                $sid = $Matches['SID']
+                $sids += , @($sid)
+            }
+        }
+    }
+
+    return $sids
 }
 
 function update_registry([string]$path, [string]$name, [string]$value) {
@@ -413,6 +437,22 @@ function install_msi([string]$path) {
     Write-Host "- Done"
 }
 
+function uninstall_msi([string]$product_name) {
+    Write-Host "Uninstalling $product_name ..."
+    $uninstall_entry = Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* -ErrorAction SilentlyContinue | 
+        Where-Object { $_.DisplayName -eq $product_name } | Select-Object -First 1
+    if (-not $uninstall_entry) {
+        throw "Failed to find the uninstall entry for $product_name"
+    }
+    Write-Host "/X `"$($uninstall_entry.PSChildName)`" /qn /norestart"
+    $proc = (Start-Process msiexec.exe -Wait -PassThru -ArgumentList "/X `"$($uninstall_entry.PSChildName)`" /qn /norestart")
+    if ($proc.ExitCode -ne 0) {
+        Write-Warning "The uninstall attempt failed with error code $($proc.ExitCode)."
+        Exit $proc.ExitCode
+    }
+    Write-Host "- Done"
+}
+
 $ErrorActionPreference = 'Stop'; # stop on all errors
 
 # check administrator status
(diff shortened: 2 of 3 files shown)


.github/workflows/win-package-test.yml
```diff
@@ -287,6 +287,42 @@ jobs:
...
+          Test-Path $msi_path
+          "MSI_COLLECTOR_PATH=$msi_path" | Out-File -FilePath $env:GITHUB_ENV -Append
+          $ps1_path = Resolve-Path .\packaging\installer\install.ps1
+          Test-Path $ps1_path
+          "INSTALL_SCRIPT_PATH=$ps1_path" | Out-File -FilePath $env:GITHUB_ENV -Append
+
+      - name: Run the script upgrade tests
+        shell: powershell # Use PowerShell to run the script since running the go test from pwsh.exe hides the Get-ExecutionPolicy cmdlet.
+        run: |
+          go test -timeout 15m -v github.com/signalfx/splunk-otel-collector/tests/windows-install-script
```
crobert-1: Why do we need a github job timeout and a timeout on the go test itself?

packaging/installer/install.ps1
```diff
@@ -370,8 +375,27 @@ function download_collector_package([string]$collector_version=$collector_versio
 }
 
 # check registry for the agent msi package
-function msi_installed([string]$name) {
-    return (Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Where { $_.DisplayName -eq $name }) -ne $null
+function is_msi_installed([string]$name) {
+    return $null -ne (Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Where { $_.DisplayName -eq $name })
```
crobert-1: Is there any way to try to use central variables for paths instead of plain strings? (It would be good to move to central variables for any being reused, if possible)

packaging/installer/install.ps1
```diff
@@ -370,8 +375,27 @@ function download_collector_package([string]$collector_version=$collector_versio
...
+function get_msi_installation_sids([string]$name) {
+    $sids = [string[]]@()
+
+    $uninstallEntry = Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* -ErrorAction SilentlyContinue | 
+        Where-Object { $_.DisplayName -eq $name }
+    if ($uninstallEntry) {
+        $userInstalls = Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\*\Products\*\InstallProperties' -ErrorAction SilentlyContinue |
+            Where-Object { $_.DisplayName -eq $name }
+        foreach ($user in $userInstalls) {
+            if ($user.PSPath -match 'UserData\\(?<SID>S-1-[0-9\-]+)') {
```
crobert-1, line 395: Can you explain what this check means, and what exactly the goal of this method is? (For context, I don't know what "sids" means here)

packaging/installer/install.ps1
```diff
@@ -413,6 +437,22 @@ function install_msi([string]$path) {
     Write-Host "- Done"
 }
 
+function uninstall_msi([string]$product_name) {
+    Write-Host "Uninstalling $product_name ..."
+    $uninstall_entry = Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* -ErrorAction SilentlyContinue | 
+        Where-Object { $_.DisplayName -eq $product_name } | Select-Object -First 1
+    if (-not $uninstall_entry) {
+        throw "Failed to find the uninstall entry for $product_name"
```
crobert-1: Can we make this error message more clear for this specific case? It's a bit too generic and makes me think that the uninstall command itself was unsuccessful, not that there was no key found for the given name.

Minor clarification, if the given `product_name` isn't found, doesn't that mean that there's no work to do so we could even consider it successful, rather than throwing an error?

packaging/installer/install.ps1
```diff
@@ -413,6 +437,22 @@ function install_msi([string]$path) {
...
+    }
+    Write-Host "/X `"$($uninstall_entry.PSChildName)`" /qn /norestart"
```
crobert-1:
```suggestion
    Write-Host "msiexec /X `"$($uninstall_entry.PSChildName)`" /qn /norestart"
```
Nit: Just thinking this may make it a bit more clear in the log output that we're running `msiexec`

packaging/installer/install.ps1
```diff
@@ -413,6 +437,22 @@ function install_msi([string]$path) {
...
+    $proc = (Start-Process msiexec.exe -Wait -PassThru -ArgumentList "/X `"$($uninstall_entry.PSChildName)`" /qn /norestart")
```
crobert-1, line 451: Does casing on arguments matter? Sometimes I see uppercase and sometimes lowercase, depending on the source. (In this case I'm looking specifically at `/X`.)

packaging/installer/install.ps1
```diff
@@ -427,12 +467,47 @@ if (!(check_if_admin)) {
...
+    if ($collector_sids.Count -eq 0) {
+        $skip_uninstall = $true
+        Write-Warning "The $service_name service exists but it is not on the Windows installation database."
+    }
+    else {
+        if ($collector_sids.Count -gt 1) {
+            $sids_list = $collector_sids -join ", "
+            throw "The Splunk OpenTelemetry Collector is already installed for multiple users (SIDs: $sids_list). Uninstalling the collector and remove remaining users installations."
+        }
+        else {
```
crobert-1: The `if` throws, so can we remove the `else` here? It might make the logic flow a bit easier to follow.

packaging/installer/install.ps1
```diff
@@ -427,12 +467,47 @@ if (!(check_if_admin)) {
...
+            if ("S-1-5-18" -ne $collector_sids[0]) {
```
crobert-1: What's the significance of this specific sid in this context? (I'm only seeing generic information from Windows documentation)

packaging/installer/install.ps1
```diff
@@ -427,12 +467,47 @@ if (!(check_if_admin)) {
...
 
-if (msi_installed -name "Splunk OpenTelemetry Collector") {
-    throw "The Splunk OpenTelemetry Collector is already installed. Remove or uninstall the Collector and rerun this script."
-}
-
 if (Get-Service -Name $service_name -ErrorAction SilentlyContinue) {
-    throw "The $service_name service is already installed. Remove or uninstall the Collector and rerun this script."
+    Write-Host "The $service_name service is already installed. Checking installation for automatic update."
+
+    $skip_uninstall = $false
```
crobert-1: Nit: Can we change the name of the variable to be positive? The later check would then be "if uninstall..." rather than "if not skip uninstall".

packaging/installer/install.ps1
```diff
@@ -427,12 +467,47 @@ if (!(check_if_admin)) {
...
+    $collector_sids = get_msi_installation_sids -name "Splunk OpenTelemetry Collector"
```
crobert-1: We have the plaintext string `Splunk OpenTelemetry Collector` quite a few places in this file, can we use a central variable for this?

packaging/installer/install.ps1
```diff
@@ -370,8 +375,27 @@ function download_collector_package([string]$collector_version=$collector_versio
 }
 
 # check registry for the agent msi package
-function msi_installed([string]$name) {
-    return (Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Where { $_.DisplayName -eq $name }) -ne $null
+function is_msi_installed([string]$name) {
+    return $null -ne (Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Where { $_.DisplayName -eq $name })
+}
+
+function get_msi_installation_sids([string]$name) {
```
crobert-1: `name` is a little generic, can we get more specific? Or, alternatively, do we even need a passed in name here or can we just use the preset collector MSI name?

tests/windows-install-script/windows_install_script_test.go
```diff
@@ -0,0 +1,171 @@
...
+		"-ExecutionPolicy", "Bypass",
+		"-File", getFilePathFromEnvVar(t, "INSTALL_SCRIPT_PATH"),
+		"-access_token", "fake-token",
+	}
+
+	if version != "" {
+		args = append(args, "-collector_version", version)
+	} else if msiPath != "" {
+		args = append(args, "-msi_path", msiPath)
+	}
```
crobert-1: Since we already confirmed one of these is set (and only one), we could either make this an `if/else` or add an `else` at the end of this that fails the test as it's an invalid case.

I bring this up because the way it's written implies that there's a case where neither are hit, and it's valid to continue in that case.

tests/windows-install-script/windows_install_script_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	output, err := cmd.CombinedOutput()
+	t.Logf("Install output: %s", string(output))
+	require.NoError(t, err, "Failed to install collector (version:%q msiPath:%q)", version, msiPath)
+}
+
+func verifyServiceExists(t *testing.T, scm *mgr.Mgr) {
+	service, err := scm.OpenService(serviceName)
+	require.NoError(t, err)
+	defer service.Close()
```
crobert-1: Do we need defer here? I guess it doesn't hurt in case this method ever gets changed, but at this point it's just unnecessary from what I can tell.

tests/windows-install-script/windows_install_script_test.go
crobert-1, line 1: Can we update this file for consistency with how backslashes are handled? Sometimes we're using backticks, sometimes double-quotes with escaped backslashes. I realize it probably won't be 100% consistent one way or the other, but any movement in that direction would be great.

tests/windows-install-script/windows_install_script_test.go
```diff
@@ -0,0 +1,171 @@
...
+	// Check for pending file rename operations
+	pendingFileRenameKey, err := registry.OpenKey(
+		registry.LOCAL_MACHINE, `SYSTEM\CurrentControlSet\Control\Session Manager`, registry.QUERY_VALUE)
+	require.NoError(t, err)
+	defer pendingFileRenameKey.Close()
+	pendingFileRenameEntries, _, err := pendingFileRenameKey.GetStringsValue("PendingFileRenameOperations")
+	if err != nil {
+		if err != registry.ErrNotExist {
+			require.NoError(t, err, "Failed to read PendingFileRenameOperations")
+		}
```
crobert-1: Could we just make this block a `require.ErrorIs` or something along those lines?

tests/windows-install-script/windows_install_script_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	for _, fileName := range pendingFileRenameEntries {
+		if strings.Contains(fileName, "plunk") {
```
crobert-1: Why `plunk`?

tests/windows-install-script/windows_install_script_test.go
```diff
@@ -0,0 +1,171 @@
...
+	oldCollectorVersion = "0.94.0"
+	// Service name
+	serviceName = "splunk-otel-collector"
+	// Service display name
+	serviceDisplayName = "Splunk OpenTelemetry Collector"
+)
+
+func TestUpgradeFromNonMachineWideVersion(t *testing.T) {
+	os.Setenv("VERIFY_ACCESS_TOKEN", "false")
+	defer os.Unsetenv("VERIFY_ACCESS_TOKEN")
```
crobert-1: As long as this isn't run in parallel (which I don't think it is), we could just use [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv) here instead.

crobert-1 (commented):
Lots of questions and clarifications, just trying to learn more about Windows here. 👍

tests/windows-install-script/windows_install_script_test.go
pjanotti: I'm going to keep the earlier failure for completeness and add an `else` that should never be reached here.

tests/windows-install-script/windows_install_script_test.go
pjanotti: We could let the finalization close the handle, but it seems better to make it deterministic.

tests/windows-install-script/windows_install_script_test.go
pjanotti: Same as the `t.SetEnv`: I keep forgetting that these (newish?) APIs exist...

tests/windows-install-script/windows_install_script_test.go
crobert-1: I'm not sure I follow your comment. `defer` just means it's called after the current method returns. There's currently no functional difference between having defer here and not having it, it just doesn't serve a purpose at the end of the method. (Unless I'm missing something here)

packaging/installer/install.ps1
pjanotti: I will have to introduce you to constants in Powershell, you are not going to like it :), but makes sense.

pjanotti: There are other "constants" that are not actually constants, the whole script could benefit from some basic linting... I will add to the KTLO backlog

packaging/installer/install.ps1
pjanotti: ops, left over "debugging"

packaging/installer/install.ps1
pjanotti: No, msiexec is not case-sensitive, I guess I went with upper-case here because it is written in uppercase on the registry. I don't feel strong about it, I will keep as it is, let me know if you prefer otherwise.

packaging/installer/install.ps1
pjanotti: >  if the given product_name isn't found, doesn't that mean that there's no work to do so we could even consider it successful, rather than throwing an error?

The check that is necessary to uninstall was made before calling this function, so in practice it should never happen - we need a uninstall race, however, I wanted to cover the possibility here and make clean that the entry was not there anymore.

As a more general approach it seems reasonable to have some "enum" to indicate the exact reason and let the caller handle it, that said as a script this function is already exiting the script in case of failure, like the install one. Let's handle that in the KTLO to lint/improve this script.

.github/workflows/win-package-test.yml
pjanotti: Forgot that the default is already 10 minutes - I was assuming that it was like 60 seconds.

tests/windows-install-script/windows_install_script_test.go
pjanotti: Oh, sorry, missed what you meant initially, yeah no need for `defer` here/

tests/windows-install-script/windows_install_script_test.go
pjanotti: Missed this one on first round.

packaging/installer/install.ps1
```diff
@@ -370,8 +377,29 @@ function download_collector_package([string]$collector_version=$collector_versio
 }
 
 # check registry for the agent msi package
-function msi_installed([string]$name) {
-    return (Get-ItemProperty HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\* | Where { $_.DisplayName -eq $name }) -ne $null
+function is_msi_installed([string]$product_name) {
+    return $null -ne (Get-ItemProperty $UninstallWildcardRegPath | Where { $_.DisplayName -eq $product_name })
+}
+
+function get_msi_installation_sids([string]$product_name) {
```
crobert-1, line 384: Just to confirm I'm following: SIDs here is referring to [security identifiers](https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers). We are getting the SIDs to ensure that whatever user is executing the command has proper permissions for uninstalling and installing the collector on the target system, is that the idea?

packaging/installer/install.ps1
crobert-1: No preference on my end, just thought it was interesting and unexpected. 👍

packaging/installer/install.ps1
```diff
@@ -427,12 +470,47 @@ if (!(check_if_admin)) {
...
+    $uninstall_collector = $true
+    $collector_sids = get_msi_installation_sids -product_name $CollectorServiceDisplayName
+    if ($collector_sids.Count -eq 0) {
+        $uninstall_collector = $false
+        Write-Warning "The $service_name service exists but it is not on the Windows installation database."
+    }
+    else {
+        if ($collector_sids.Count -gt 1) {
+            $sids_list = $collector_sids -join ", "
+            throw "The Splunk OpenTelemetry Collector is already installed for multiple users (SIDs: $sids_list). Uninstalling the collector and remove remaining users installations."
```
crobert-1: The message here makes it seem like the script is proceeding with the uninstallation process, specifically `Uninstalling the collector...` Is this meant to be telling the user to do this, or something else?

packaging/installer/install.ps1
pjanotti: Yes, if it was installed with a different user we need that user to do uninstallation. Note that this an issue until `v0.94.0` of the MSI, after that all installations are machine-wide and will always be under the LocalSystem SID.

packaging/installer/install.ps1
pjanotti: I will update the message. This should be a rare situation, but it is definitely a possibility with versions prior to `v0.94.0`

crobert-1 (approved):

PR #6306, opened 2025-06-11
[chore][CICD] Move cross-compile into reusable workflow

crobert-1:

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -15,27 +15,10 @@ env:
   GO_VERSION: 1.23.8
 
 jobs:
-  setup-environment:
-    name: setup-environment
-    runs-on: ubuntu-24.04
-    steps:
-      - name: Check out the codebase.
-        uses: actions/checkout@v4
-
-      - name: Set up Go
-        uses: actions/setup-go@v5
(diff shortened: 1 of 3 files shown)


.github/workflows/setup-environment.yml
```diff
@@ -0,0 +1,25 @@
+name: setup-environment
+
+on:
+  workflow_call:
+
+env:
+  GO_VERSION: 1.23.8
```
crobert-1, line 7: It'd be nice to define `GO_VERSION` somewhere centrally in our repo and have all workflows simply refer to a single source of truth, so any input on how to accomplish that in the cleanest way would be appreciated.

jinja2: Couple of options which might work -
1. Have a central .env file and update jobs to read from this.
A file with content `GO_VERSION=1.23.8`. We load the env in each job with a step like
```
- name: Load Go version
  run: echo "GO_VERSION=$(grep GO_VERSION .github/envs/go.env | cut -d '=' -f2)" >> $GITHUB_ENV
```

2. Make it a github action variable in the repo Settings. I guess we can then access it with `${{ vars.GO_VERSION }} ` in the workflows.

PR #6308, opened 2025-06-12
[TA] Linux Java autoinstrumentation TA

hughesjj:
Adds Linux java autoinstrumentation functionality for TA.  This is a separate TA from the `Splunk_TA_otel` we typically vend.

Will need to coordinate with docs team on release of TA w.r.t configuring inputs.conf for this (in addition to changelog)

### DIFF
--- a/packaging/technical-addon/internal/modularinput/spec.go
+++ b/packaging/technical-addon/internal/modularinput/spec.go
@@ -40,6 +40,7 @@ type GenericModularInput struct {
 	ModularInputs map[string]*ModInput
 	SchemaName    string
 }
+
 type TemplateData struct {
 	ModularInputs map[string]ModInputConfig `yaml:"modular-inputs"`
 	SchemaName    string                    `yaml:"modular-input-schema-name"`


packaging/technical-addon/internal/testaddon/splunk.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package testcommon
+package testaddon
```
hughesjj, line 15: Perhaps I should move this code under `tests/testutils` at the root of this repo instead of under `packaging/technical-addon`?

packaging/technical-addon/internal/modularinput/spec.go
```diff
@@ -40,7 +40,6 @@ type GenericModularInput struct {
 	ModularInputs map[string]*ModInput
 	SchemaName    string
 }
-
```
pjanotti: Keep the white line. Are we running lint on these files? If not better to put in the backlog

packaging/technical-addon/Makefile
```diff
@@ -24,32 +27,42 @@ MODULAR_INPUT_SCHEMAS=Splunk_TA_otel_linux_autoinstrumentation
 TOOLS_DIR := $(BUILD_DIR)/.tools
 MODINPUT_CONFIG_GENERATOR := $(TOOLS_DIR)/modinput_config_generator
 
+# Overrides
+# Packaging addons and starting splunk takes some time...
+GOTEST_OPT= -v -timeout 16m  -parallel 1 --tags=$(GO_BUILD_TAGS)
```
pjanotti: `-parallel 1` enforces no tests running concurrently, isn't better to leave the tests to decide it via `t.Parallel` and related?

packaging/technical-addon/Makefile
```diff
@@ -24,32 +27,42 @@ MODULAR_INPUT_SCHEMAS=Splunk_TA_otel_linux_autoinstrumentation
...
+
 # Make targets
 .PHONY: ta-build-tools
 ta-build-tools:
 	@echo "Building tools..."
 	@mkdir -p $(TOOLS_DIR)
-	go build -o $(MODINPUT_CONFIG_GENERATOR) ./cmd/modinput_config_generator
+	cd $(ADDONS_SOURCE_DIR) && go build -o $(MODINPUT_CONFIG_GENERATOR) ./cmd/modinput_config_generator
```
pjanotti, line 39: Q.: is it fine to `cd` if this target is used by others? I mean suppose that some target uses this one but then uses a different one that assumes other location, is it the `cd` scoped to each target?

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/assets/default/app.conf.tmpl
```diff
@@ -0,0 +1,28 @@
+######################################################
+#
+# {{ .SchemaName }}
+#
+# Copyright (C) 2005-2020 Splunk Inc. All Rights Reserved.
```
pjanotti: Update? Or even remove the copyright notice?

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/internal/testdata/happypath/expected/expected-zeroconfig.txt
```diff
@@ -0,0 +1,5 @@
+JAVA_TOOL_OPTIONS=-javaagent:/foo/bar
+OTEL_RESOURCE_ATTRIBUTES=splunk.zc.method=splunk-otel-auto-instrumentation-v2.15.0,asdasd
```
pjanotti, line 2: Q.: should the version float?

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+	if *validateFlag {
+		if err != nil {
+			return -1
+		}
+		return 0
+	}
+	if err != nil {
+		log.Errorf("Error parsing modinput: %+v", err)
+		panic(err)
+	}
```
pjanotti: nit: 
```suggestion

	err = mip.ProcessXML(xmlInput)
	if *validateFlag {
		if err != nil {
			return -1
		}
		return 0
	}
	if err != nil {
		log.Errorf("Error parsing modinput: %+v", err)
		panic(err)
	}
```

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+	schemeFlag := flag.Bool("scheme", false, "Print the scheme and exit")
+	validateFlag := flag.Bool("validate-arguments", false, "Validate the arguments and exit")
+	flag.Parse()
+	if *schemeFlag {
+		// TODO we've never actually implemented this, let's figure it out in the future
+		return 0
+	}
+	defaultLogFilepath, err := modularinput.GetDefaultLogFilePath(config.SchemaName)
+	if err != nil {
+		panic(err)
```
pjanotti, line 49: Is the panic really helpful?

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		log.Errorf("error running splunk linux autoinstrumentation addon: %+v", err)
+		panic(err)
+	}
+	// TODO set up traps, but not urgent given this is a run-once style script
+	return 0
+}
+
+func Run(modInputs *SplunkTAOtelLinuxAutoinstrumentationModularInputs) error {
+	lowerModInput := strings.ToLower(modInputs.Remove.Value)
```
pjanotti: nit:
```suggestion
	removeValue := strings.ToLower(modInputs.Remove.Value)
```

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+	if "false" == lowerModInput {
+		return Instrument(modInputs)
+	}
+	if "true" == lowerModInput {
+		return DeInstrument(modInputs)
+	}
+
+	return fmt.Errorf("unknown value for 'remove' modular input, expected (true|false) given %v", strings.ToLower(modInputs.Remove.Value))
```
pjanotti: nit:
```suggestion
	return fmt.Errorf("unknown value for 'remove' modular input, expected (true|false) given %q", modInputs.Remove.Value)
```

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func DeInstrument(modInputs *SplunkTAOtelLinuxAutoinstrumentationModularInputs) error {
+	if err := RemoveJavaInstrumentation(modInputs); err != nil {
+		return err
+	}
+	if err := RemovePreloadInstrumentation(modInputs); err != nil {
```
pjanotti: nit: to keep the typical install/uninstall preload should be removed prior to the java instrumentation.

nit: typically the install should come before of the source file

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+func AutoinstrumentLdPreload(modInputs *SplunkTAOtelLinuxAutoinstrumentationModularInputs) error {
+	if "true" != strings.ToLower(modInputs.AutoinstrumentationEnabled.Value) {
+		log.Println("Autoinstrumentation of /etc/ld.so.preload is disabled in inputs.conf, not configuring")
+		return nil
+	}
+	found, err := grepFile(modInputs.AutoinstrumentationPath.Value, modInputs.AutoinstrumentationPreloadPath.Value)
+	if err != nil {
+		return err
+	}
+	if !found {
```
pjanotti: nit: revert the if condition and return early if `found` with a return nil inside the if branch.

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+		if strings.ToLower(modInputs.Backup.Value) != "false" {
+			if err = backupFile(modInputs.AutoinstrumentationPreloadPath.Value); err != nil {
+				return err
+			}
+		}
+		f, err2 := os.OpenFile(modInputs.AutoinstrumentationPreloadPath.Value, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
```
pjanotti:
```suggestion
		f, errFileOp := os.OpenFile(modInputs.AutoinstrumentationPreloadPath.Value, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
```

or even `errOpenFile` and `errWriteFile` below

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+		if err2 != nil {
+			return fmt.Errorf("error opening %s: %w", modInputs.AutoinstrumentationPreloadPath.Value, err2)
+		}
+		defer f.Close()
+		if _, err2 = f.WriteString(modInputs.AutoinstrumentationPath.Value + "\n"); err2 != nil {
+			return fmt.Errorf("error writing to %s: %w", modInputs.AutoinstrumentationPreloadPath.Value, err2)
+		}
+		log.Printf("Successfully autoinstrumented preload at %v with %v\n", modInputs.AutoinstrumentationPreloadPath.Value, modInputs.AutoinstrumentationPath.Value)
```
pjanotti: nit: consistency of the formatting for the path, some places is %s others %v

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+	found, err := grepFile(modInputs.Remove.Value, modInputs.AutoinstrumentationPreloadPath.Value)
+	if err != nil {
+		return err
+	}
+	if found {
+		content, err := os.ReadFile(modInputs.AutoinstrumentationPreloadPath.Value)
+		if err != nil {
+			return err
+		}
+		if strings.ToLower(modInputs.Backup.Value) != "false" {
```
pjanotti: I'm seeing the pattern for booleans, what will be the cost of making the type conversion when reading the modinput file? Perhaps something for a follow up.

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	return nil
+}
+
+func RemovePreloadInstrumentation(modInputs *SplunkTAOtelLinuxAutoinstrumentationModularInputs) error {
+	found, err := grepFile(modInputs.Remove.Value, modInputs.AutoinstrumentationPreloadPath.Value)
+	if err != nil {
+		return err
+	}
+	if found {
```
pjanotti, line 154: nit: here the reverse should be `if !found` returning after the log message

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+		newContent := strings.ReplaceAll(string(content), modInputs.AutoinstrumentationPreloadPath.Value, "")
+		return os.WriteFile(modInputs.AutoinstrumentationPreloadPath.Value, []byte(newContent), 0644) // #nosec G306
+
+	}
+	log.Printf("Autoinstrumentation preload (%s) does not exist or does not contain configured autoinstrumentation (%s)", modInputs.AutoinstrumentationPreloadPath.Value, modInputs.AutoinstrumentationPath.Value)
+	return nil
+}
+
+func backupFile(currPath string) error {
+	if _, err := os.Stat(currPath); errors.Is(err, os.ErrNotExist) {
```
pjanotti: nit: for file operations the more resilient pattern is to try to do the actual operation and identify the error in case of failure. The rational being that the result of the test here can change before the actual operation. That said in practice the chances of that happening are almost none in this specific case.

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+		if err != nil {
+			return err
+		}
+		if strings.ToLower(modInputs.Backup.Value) != "false" {
+			if err = backupFile(modInputs.AutoinstrumentationPreloadPath.Value); err != nil {
+				return err
+			}
+		}
+		newContent := strings.ReplaceAll(string(content), modInputs.AutoinstrumentationPreloadPath.Value, "")
+		return os.WriteFile(modInputs.AutoinstrumentationPreloadPath.Value, []byte(newContent), 0644) // #nosec G306
```
pjanotti: I'm not sure if it is worth a comment on the code, but why the `nosec` here?

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+	log.Printf("Autoinstrumentation preload (%s) does not exist or does not contain configured autoinstrumentation (%s)", modInputs.AutoinstrumentationPreloadPath.Value, modInputs.AutoinstrumentationPath.Value)
+	return nil
+}
+
+func backupFile(currPath string) error {
+	if _, err := os.Stat(currPath); errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+	dir := filepath.Dir(currPath)
+	ogFilename := filepath.Base(currPath)
```
pjanotti: `ogFilename`? `og` what is meant by that?

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
```diff
@@ -0,0 +1,201 @@
...
+	newPathName := filepath.Join(dir, fmt.Sprintf("%s.%v", ogFilename, time.Now().UnixNano()))
+	if _, err := fileutils.CopyFile(currPath, newPathName); err != nil {
+		return err
+	}
+	return nil
+}
+
+func grepFile(search string, filepath string) (bool, error) {
```
pjanotti: nit: `grep` for me implies regex capabilities, `fileContains` seems a better name. It seems worth to document that it return false,nil if file doesn't exist.

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/javazeroconfig.go
```diff
@@ -0,0 +1,151 @@
...
+`
+
+// TemplateData hughesjj@
+// I've gone back and forth on whether to just wrap modInputs and add custom as needed
+// As of now, only resourceattributes differ from what's default in modInputs
+// I can't just add a transformer, given it would differ from the nodejs case
+// Since goland does not support autocompletion in the template, I've decided
+// to just duplicate them all for now, to decouple from the customer interface
+// as defined in inputs.conf
+type TemplateData struct {
```
pjanotti: nit: does the type and functions need to be public?

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/javazeroconfig.go
```diff
@@ -0,0 +1,151 @@
...
+		EnableMetrics:          modInputs.MetricsEnabled.Value,
+		ServiceName:            modInputs.OtelServiceName.Value,
+		OtlpEndpoint:           modInputs.OtelExporterOtlpEndpoint.Value,
+		OtlpEndpointProtocol:   modInputs.OtelExporterOtlpProtocol.Value,
+		MetricsExporter:        modInputs.OtelMetricsExporter.Value,
+		LogsExporter:           modInputs.OtelLogsExporter.Value,
+	}
+
+	if err = os.MkdirAll(filepath.Dir(modInputs.JavaZeroconfigPath.Value), 0755); err != nil {
+		err = fmt.Errorf("error creating java zeroconfig path, could not make parent directories: %v", err)
```
pjanotti:
```suggestion
		err = fmt.Errorf("error creating java zeroconfig path, could not make parent directories: %w", err)
```

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/javazeroconfig.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	filePath, err := os.Create(modInputs.JavaZeroconfigPath.Value)
+	if err != nil && !errors.Is(err, os.ErrExist) {
+		return err
+	}
+	if err = tmpl.Execute(filePath, templateData); err != nil {
+		return fmt.Errorf("failed to execute template: %w %v", err, templateData)
+	}
+	log.Printf("Successfully generated java autoinstrumentation config at %s \n", filePath.Name())
```
pjanotti: nits: besides the suggestion notice also the formatting for the path (consistency mentioned somewhere else)

```suggestion
	log.Printf("Successfully generated java autoinstrumentation config at %s\n", filePath.Name())
```

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/javazeroconfig.go
```diff
@@ -0,0 +1,151 @@
...
+		return err
+	}
+	return nil
+}
+
+func RemoveJavaInstrumentation(modInputs *SplunkTAOtelLinuxAutoinstrumentationModularInputs) error {
+	if strings.ToLower(modInputs.Backup.Value) != "false" {
+		if err := backupFile(modInputs.JavaZeroconfigPath.Value); err != nil && !errors.Is(err, os.ErrNotExist) {
+			log.Fatalf("error backing up java auto instrumentation configuration, refusing to remove (specify backup=false in inputs.conf if backup not needed): %v", err)
+			return err
```
pjanotti: Is this required by the compiler? I mean `log.Fatalf` will terminate the program and the caller won't receive the error.

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/javazeroconfig.go
```diff
@@ -0,0 +1,151 @@
...
+	if "" != modInputs.DeploymentEnvironment.Value {
+		resourceAttributes = fmt.Sprintf("%s,deployment.environment=%s", resourceAttributes, modInputs.DeploymentEnvironment.Value)
+	}
+	if "" != modInputs.ResourceAttributes.Value {
+		resourceAttributes = fmt.Sprintf("%s,%s", resourceAttributes, modInputs.ResourceAttributes.Value)
+	}
+
+	tmpl, err := template.New("JavaZeroConfig").Parse(configTemplate)
+	if err != nil {
+		log.Fatalf("error generating zeroconfig file at %s from template: %#v", modInputs.JavaZeroconfigPath.Value, err)
```
pjanotti: Why here `%#v` and other calls to log `%v` for errors?

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
pjanotti: or if we don't care about specif return value in case of error
```suggestion

	err = mip.ProcessXML(xmlInput)
	if *validateFlag {
		return err != nil
	}
	if err != nil {
		log.Errorf("Error parsing modinput: %+v", err)
		panic(err)
	}
```

packaging/technical-addon/Makefile
hughesjj: ADDONS_SOURCE_DIR is exclusive to this directory and thus this target

I wound up taking this Makefile out of being referenced anywhere else because it was getting... circular..

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/internal/testdata/happypath/expected/expected-zeroconfig.txt
hughesjj: I could add it to the "update" build target, fair point.

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
hughesjj: Gosec complains when we have anything but `0600` for new file permissions, but historically we've written it with `0644` (implicit with `022` as default `umask`)
https://github.com/securego/gosec#usage

To my understanding, and consistent with other `/etc/ld.so` files on my local machine, non-root users (root is the recommended user creating this file) need to be able to read the file too, but admittedly I haven't found definitive documentation on this

See https://man7.org/linux/man-pages/man8/ld.so.8.html#FILES for preload man pages

packaging/technical-addon/Makefile
hughesjj: Thank you!  I got rid of host network and this should no longer be needed

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/assets/default/app.conf.tmpl
hughesjj: I will update, yes

As this is copy pasta I'll need to figure out where this came from and update it there too, but that's a separate issue

packaging/technical-addon/Makefile
hughesjj: so, TL;dr: CAn't be used by others anymore, I should take this out, perhaps a bad rebase.  As-is it's redundant.

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/javazeroconfig.go
hughesjj: removed the log.fatalf in favor of log.error

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
hughesjj: I believe any formatting issues are resolved -- I've run `make fmt` on the directory and `make lint`, fwiw

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
hughesjj: not sure what you mean by 
 > nit: typically the install should come before of the source file

but otherwise addressed

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
hughesjj: just reused `err` instead

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
hughesjj: We could also add in a helper function, but agreed it'd be easier to read to just add typing information to the modinput config generator and add a bit more to the parsing

Also agreed that should be a followup review, as we get a clearer picture of how to best handle our implementation of modular inputs

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
hughesjj: original.. I was trying to do the golang standard as I understand it to shorten or abbreviate words, but honestly I prefer the java way

perhaps `origFilename` ?

packaging/technical-addon/pkg/splunk_ta_otel_linux_autoinstrumentation/runner/main.go
pjanotti: ops, I meant that reading the source code I would expect to first read the code for installing and then the code for uninstalling. I guess I meant 

> nit: typically the install should come before __on__ the source file

packaging/technical-addon/internal/testaddon/splunk.go
pjanotti: I think it is fine at the current location.

pjanotti (approved):

PR #6309, opened 2025-06-12
[chore][ansible] Run Windows tests against current build

crobert-1:

Essentially the same effort as https://github.com/signalfx/splunk-otel-collector/pull/6284. The functionality of installing the MSI has been split into two files, one for fetching the MSI from the public releases URL, one for fetching the MSI from the locally built artifacts. The split logic is to copy the MSI to ansible's target machine at a set path. Functionality can proceed the same way for both installations from there.


.github/workflows/ansible.yml
crobert-1, line 1: `agent-bundle-windows`, `msi-custom-actions`, and `msi-build` are all directly copied from https://github.com/signalfx/splunk-otel-collector/blob/main/.github/workflows/win-package-test.yml

.github/workflows/ansible.yml
```diff
@@ -257,6 +359,11 @@ jobs:
       - name: Check out the codebase.
         uses: actions/checkout@v4
 
+      - uses: actions/download-artifact@v4
+        with:
+          name: msi-build
```
crobert-1, line 373: For simplicity in Ansible logic, use a single MSI with a preset name, rather than having to pass around variables and parse versions.

deployments/ansible/roles/collector/tasks/win_download_remote_msi.yml
crobert-1, line 1: This file is simply tasks moved from `otel_win_install.yml` for cleaning up the organization of functionality.

deployments/ansible/molecule/custom_vars/windows-verify.yml
```diff
@@ -45,23 +45,6 @@
       assert:
         that: not service_status.changed
 
-    - name: Download splunk-otel-collector 0.126.0 MSI
```
crobert-1, line 48: Removed version check for the same reason as the [linux change](https://github.com/signalfx/splunk-otel-collector/pull/6284#discussion_r2127399483). We can revisit if anyone has a strong preference here.

deployments/ansible/roles/collector/tasks/otel_win_install.yml
```diff
@@ -1,42 +1,18 @@
 ---
 
-- name: Get splunk-otel-collector latest version
```
crobert-1, line 3: Deleted lines from this file are simply being moved to the `win_download_remote_msi,yml` task in the same directory.

PR #6311, opened 2025-06-13
Add entites for Agent - GW setup

dloucasfx:


cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -159,6 +159,8 @@ exporters:
   # Entities (applicable only if discovery mode is enabled)
   otlphttp/entities:
     logs_endpoint: "${SPLUNK_INGEST_URL}/v3/event"
+    # Use instead when sending to gateway
+    # endpoint: http://${SPLUNK_GATEWAY_URL}:4318
```
crobert-1:
```suggestion
    # Use instead when sending to gateway
    # endpoint: ${SPLUNK_GATEWAY_URL}:4318
```
Is `http://` required here? Same question for other usages. I'm wondering why the value provided isn't consistent with usage of other environment variables.

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -142,6 +142,22 @@ exporters:
     token: "${SPLUNK_ACCESS_TOKEN}"
     endpoint: "${SPLUNK_INGEST_URL}/v1/log"
     log_data_enabled: false
+  # To send entities (applicable only if discovery mode is enabled)
+  otlphttp/entities:
```
crobert-1, line 146: If this is only relevant when discovery is enabled, is there some way to tie this to discovery mode? (The pipeline is only enabled when discovery mode is enabled?)

cmd/otelcol/config/collector/gateway_config.yaml
```diff
@@ -142,6 +142,22 @@ exporters:
...
+    auth:
+      authenticator: headers_setter
+
+connectors:
+  # Routing connector to separate entity events from regular logs
+  routing/logs:
+    default_pipelines: [logs]
+    table:
+      - context: log
+        condition: instrumentation_scope.attributes["otel.entity.event_as_log"] == true
```
crobert-1, line 159: Where does `otel.entity.event_as_log` come from? Is there some way we could use a central variable or constant here? Not required as it might be a bit messy, just concerned about the possibility of this changing in the future.

crobert-1 (commented):
I don't have much context for this change, please let me know if I'm misunderstanding anything here.

Have you been able to test the routing connector functionality added here?

pjanotti (approved):

crobert-1:
I haven't seen that `linux-package-test/docker-otelcol-verify` failure before:
```
exec /usr/lib/splunk-otel-collector/agent-bundle/bin/python: no such file or directory
```
Please check to make sure it's not caused by this PR, we can address later if it's unrelated.

cmd/otelcol/config/collector/agent_config.yaml
dloucasfx: yes, http is needed. By default OTLP receiver on GW is not configured for TLS.

cmd/otelcol/config/collector/gateway_config.yaml
dloucasfx: This is on the GW side and discovery might not be enabled there but enabled at the sender

cmd/otelcol/config/collector/gateway_config.yaml
dloucasfx: This is coming from the sender (agent where discovery is enabled).
We can use env variable, but I don't think it's worth it

dloucasfx:
> I don't have much context for this change, please let me know if I'm misunderstanding anything here.
> 
> Have you been able to test the routing connector functionality added here?

yes, this was validated manually

cmd/otelcol/config/collector/agent_config.yaml
dloucasfx: @crobert-1 
I guess you are assuming that http scheme is included in the env var `SPLUNK_GATEWAY_URL` ?
I have seen a mix , where the `SPLUNK_GATEWAY_URL` is the IP address of the GW. 
If you look at this, we are explicitly defining the scheme https://github.com/signalfx/splunk-otel-collector/blame/main/cmd/otelcol/config/collector/agent_config.yaml#L155-L156 the only other place where we are not, is using grpc . 

 ~`SPLUNK_GATEWAY_URL` is not set in the conf, I wonder if they left the scheme out in case some servers running on the GW are configured to support https and other don't~
 Based on the above, we don't expect the scheme to be part of `SPLUNK_GATEWAY_URL` .
 I guess we need to clearly document this

dmitryax: We don't use the `SPLUNK_GATEWAY_URL` env var in the helm chart. Instead we dynamically configure the exporters to point to the gateway service. See https://github.com/signalfx/splunk-otel-collector-chart/blob/6d10b38a5d28e326d52f0a16dc22475a7f50a7b4/helm-charts/splunk-otel-collector/templates/config/_otel-agent.tpl#L852. User doesn't have to do anything other than setting `gateway.enabled` to `true`

dmitryax: Given that you added the splitting connector in the gateway, we can just use existing `otlp` exporter instead

dloucasfx: @dmitryax 
Yes, makes sense.
PR is updated, please review

crobert-1 (approved):

PR #6329, opened 2025-06-16
[chore][CICD] Reuse more GitHub workflow jobs

crobert-1:

Builds on https://github.com/signalfx/splunk-otel-collector/pull/6306. 

Notes: 
1. Some of the Windows build and packaging workflows are only being used in one place in this PR, but it will be utilized by https://github.com/signalfx/splunk-otel-collector/pull/6309
2. All input arguments to reusable workflows have been moved to be all capitalized.
3. Deleted `darwin-test.yml` as functionality is now covered by `build-and-test.yml`. If we want to keep `darwin-test.yml` for some reason, let me know.


.github/actions/setup-environment/action.yml
crobert-1, line 1: This PR moves `setup-environment` from a [reusable workflow to a composite action](https://docs.github.com/en/actions/sharing-automations/avoiding-duplication), as the steps should be run in the same job as the dependent functionality.

.github/actions/setup-environment/action.yml
```diff
@@ -0,0 +1,24 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Check out the codebase.
+      uses: actions/checkout@v4
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: ${{ inputs.GO_VERSION }}
```
pjanotti: Is it worth to have `GO_VERSION` as an input? Couldn't we just define it as constant here?

.github/workflows/agent-bundle-windows.yml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  agent-bundle-windows:
+    runs-on: ${{ inputs.OS }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/cache@v4
+        with:
+          path: ${{ inputs.PIP_CACHE_DIR }}
+          key: agent-bundle-windows-pip-${{ hashFiles('packaging/bundle/collectd-plugins.yaml', 'packaging/bundle/scripts/requirements.txt') }}
```
pjanotti, line 25: Let's track this in a separate KTLO issue after this one is merged, but the second file doesn't exist.

.github/actions/setup-environment/action.yml
```diff
@@ -0,0 +1,24 @@
...
+inputs:
+  GO_VERSION:
+    description: 'Version of golang to setup'
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Check out the codebase.
+      uses: actions/checkout@v4
```
pjanotti: Won't be the caller doing the checkout? Or can we remove the checkout from the caller?

.github/workflows/agent-bundle-linux.yml
```diff
@@ -0,0 +1,33 @@
...
+    inputs:
+      ARCH:
+        required: true
+        type: string
+env:
+  GO_VERSION: 1.23.10
+
+jobs:
+  agent-bundle-linux:
+    runs-on: ${{ fromJSON('["ubuntu-24.04", "otel-arm64"]')[inputs.ARCH == 'arm64'] }}
```
pjanotti, line 14: not related to the main goal of the PR but can we run this on ubuntu arm instead of our runner?

.github/workflows/msi-build.yml
```diff
@@ -0,0 +1,55 @@
...
+  GO_VERSION: 1.23.10
+
+jobs:
+  msi-build:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Check out the codebase.
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
pjanotti: Can we use this in other places also?

.github/actions/setup-environment/action.yml
crobert-1: Good point. I'll remove from here, my understanding is the caller should have it.

.github/workflows/agent-bundle-linux.yml
crobert-1: Good point. I'll update here, but if anything breaks as a result I'll wait to make the change in a separate PR. 👍

.github/workflows/msi-build.yml
crobert-1: Are you referring to the workflow as a whole, `msi-build`, or something else? From what I can tell I've updated all references to use this workflow as a whole (just updating the Ansible workflow now), but maybe I'm misunderstanding your question

.github/actions/setup-environment/action.yml
crobert-1: Good question, I'm fine either way. The end goal either way is to end up using a central variable, as [discussed previously](https://github.com/signalfx/splunk-otel-collector/pull/6306#discussion_r2141237537). With that in mind though, it'd be simpler to move to fewer variables now, so I'll move to just defining it in the action here, rather than passing it in.

crobert-1: Did some research, it looks like composite actions don't have a way to specify an environment variable in the same way workflows and regular actions have. The options were to either use a plaintext string, or use the calling workflow's environment variable. I went with the calling workflow's environment variable for now.

.github/workflows/otelcol-fips.yml
```diff
@@ -27,7 +27,7 @@ env:
 
 jobs:
   otelcol-fips:
-    runs-on: ${{ fromJSON('["ubuntu-24.04", "otel-arm64"]')[matrix.GOARCH == 'arm64'] }}
+    runs-on: ${{ fromJSON('["ubuntu-24.04", "ubuntu-24.04-arm"]')[matrix.GOARCH == 'arm64'] }}
```
dmitryax: Why do we move to another image here?

dmitryax:
This is nice 👍

.github/workflows/otelcol-fips.yml
crobert-1: @pjanotti [suggested it above](https://github.com/signalfx/splunk-otel-collector/pull/6329#discussion_r2155326373). I'm not very familiar with the nuances of the otel arm runner, but my thought in changing as suggested was that it will be an easier maintenance burden going forward to use the GitHub runner.

I can make this change in a follow up PR, or we can discuss further if that's preferred.

.github/workflows/agent-bundle-linux.yml
dmitryax: @pjanotti are you sure there is nothing on top in our image? I guess we should confirm that with @atoulme who created them. Also, I'd say it should be in a separate PR

dmitryax (approved):
LGTM apart from unrelated switch to another image

.github/workflows/otelcol-fips.yml
crobert-1: I'll revert this change and we can revisit at a later date.

.github/workflows/msi-build.yml
pjanotti: I misread I thought that it was using fetch-depth 1, that in theory should be "cheaper", but, I guess that here we need all the repo history and tags.

pjanotti: If there is a specific reason to not use the default here, i.e.: `fetch-depth: 1`, it will be good to have a comment

pjanotti (approved):

.github/workflows/msi-build.yml
crobert-1: I'm going to remove this in favor of default (as long as tests pass). This was directly copied without thinking about reasoning of fetch depth, I don't see a reason to go against default here, I don't see a need for all history since this is just building the most recent commit.

crobert-1: _Made the same change for build-packages workflow as I don't believe it's necessary there either._

PR #6333, opened 2025-06-17
[chore][chef] Run tests against locally built artifact

crobert-1:

Same idea as https://github.com/signalfx/splunk-otel-collector/pull/6284, updating Chef tests to run against locally built artifacts.

- Introduced a new variable `local_artifact_testing_enabled` that will install the collector from a local artifact when set to true.
- Added workflow build functionality is all copied from other workflows. Once https://github.com/signalfx/splunk-otel-collector/pull/6329 is merged we can reuse the reusable workflows.

### DIFF
--- a/.github/workflows/chef-test.yml
+++ b/.github/workflows/chef-test.yml
@@ -47,17 +47,17 @@ jobs:
     needs: [ setup-environment, chef-lint-spec-test ]
     strategy:
       matrix:
-        SYS_BINARIES: [ "binaries-linux_amd64", "binaries-linux_arm64", "binaries-windows_amd64" ]
+        SYS_BINARIES: [ "binaries-linux_amd64", "binaries-windows_amd64" ]
     uses: ./.github/workflows/compile.yml
     with:
       sys_binary: ${{ matrix.SYS_BINARIES }}
 
   agent-bundle-linux:
     needs: [chef-lint-spec-test]
-    runs-on: ${{ fromJSON('["ubuntu-24.04", "otel-arm64"]')[matrix.ARCH == 'arm64'] }}
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
-        ARCH: [ "amd64", "arm64" ]
+        ARCH: [ "amd64" ]
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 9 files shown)


deployments/chef/recipes/collector_win_install.rb
```diff
@@ -13,11 +13,20 @@
                       node['splunk_otel_collector']['collector_version']
                     end
 
-remote_file 'Download msi' do
-  path "#{ENV['TEMP']}/splunk-otel-collector-#{collector_version}-amd64.msi"
-  source "#{node['splunk_otel_collector']['windows_repo_url']}/splunk-otel-collector-#{collector_version}-amd64.msi"
-  action :create
-  only_if { !::File.exist?(node['splunk_otel_collector']['collector_version_file']) || (::File.readlines(node['splunk_otel_collector']['collector_version_file']).first.strip != collector_version) }
+remote_destination_path = "#{ENV['TEMP']}/splunk-otel-collector-#{collector_version}-amd64.msi"
```
crobert-1, line 16: `collector_version` may not actually match the local artifact's version, but for simplicity of the remaining logic I figure there isn't much harm in keeping the name the same for testing.

.github/workflows/chef-test.yml
```diff
@@ -110,14 +327,34 @@ jobs:
...
+      # It's simpler to unarchive and move the built artifacts manually to the
+      # cookbook files directory. The files/ dir under the Chef deployment is a special
+      # directory in Chef, used for copying files from the cookbook to the host running
+      # Chef.
+      # Cookbook files reference: https://docs.chef.io/files/
+      - name: Extract artifacts
+        run: |
+          mkdir files
+          Get-ChildItem /tmp -Recurse -Include 'splunk-otel-collector*.msi' | Select FullName -OutVariable msiPath
+          echo $msiPath.FullName
```
crobert-1:
```suggestion
```

.github/workflows/chef-test.yml
```diff
@@ -80,20 +277,40 @@ jobs:
...
+      # Chef doesn't have an easy way to get names of files from unarchived packages
+      # It's simpler to manually unarchive and move the built artifacts to the
+      # cookbook files directory here. The files/ dir under the Chef deployment is a special
+      # directory in Chef, used for copying files from the cookbook to the host running
+      # Chef.
+      # Cookbook files reference: https://docs.chef.io/files/
+      - name: Extract artifacts
+        run: |
+          mkdir -p files
+          deb_path=$(find /tmp/deb-amd64-package/splunk-otel-collector*amd64.deb)
```
crobert-1, line 304: For all three paths, `deb_path`, `rpm_path`, and `msiPath`, we're relying on the directories to only have a single file that matches the regex. This might fail in the future if built artifacts change names, or end up having more than one matching artifact.

dmitryax (approved):

PR #6354, opened 2025-06-23
[chore] Update urllib3 (Python)

pjanotti:
Updating urllib3 (Python)


.github/workflows/scripts/gendependabot.sh
```diff
@@ -51,7 +51,7 @@ get_pip_ignores() {
 
     ignore:
       - dependency-name: "urllib3"
-        versions: [">=2"]
+        versions: [">=3"]
```
crobert-1, line 54: Why are other dep requirements saying `urllib<3` but this is `>=3`?

pjanotti: This is to ignore updates for 3 and above

PR #6358, opened 2025-06-24
Update OTel deps to v0.128.0

pjanotti:
Updating OTel deps to v0.128.0


tests/receivers/jmx/cassandra/jmx_cassandra_discovery_test.go
```diff
@@ -66,8 +69,13 @@ func jmxCassandraAutoDiscoveryHelper(t *testing.T, ctx context.Context, configFi
 	factory := otlpreceiver.NewFactory()
 	port := 16745
 	c := factory.CreateDefaultConfig().(*otlpreceiver.Config)
-	c.GRPC.NetAddr.Endpoint = fmt.Sprintf("localhost:%d", port)
-	endpoint := c.GRPC.NetAddr.Endpoint
+	endpoint := fmt.Sprintf("localhost:%d", port)
+	c.GRPC = configoptional.Some(configgrpc.ServerConfig{
+		NetAddr: confignet.AddrConfig{
+			Endpoint:  endpoint,
```
crobert-1, line 75:
```suggestion
	c.GRPC = configoptional.Some(configgrpc.ServerConfig{
		NetAddr: confignet.AddrConfig{
			Endpoint:  fmt.Sprintf("localhost:%d", port),
```
Nit, feel free to drop: Other tests just set endpoint directly, we can do the same here for consistency.

pjanotti: I assumed that `endpoint` was used somewhere else but didn't look, let me check...

pjanotti: It is passed down as an environment variable, that's why it is saved in the local `endpoint`.

PR #6360, opened 2025-06-24
[chore][CICD] Fix broken Ansible workflow

crobert-1:

I believe this was broken by https://github.com/signalfx/splunk-otel-collector/pull/6329 ([failure example](https://github.com/signalfx/splunk-otel-collector/actions/runs/15834169378)), since the workflow wasn't formatted properly it didn't show up in the PR as failed, you have to go to actions to see the faiilure.

1. Compile a windows AMD64 image
2. Can only have one of: `runs-on` or `uses`
3. Download properly named msi package

### DIFF
--- a/.github/workflows/ansible.yml
+++ b/.github/workflows/ansible.yml
@@ -35,7 +35,7 @@ jobs:
   cross-compile:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


jvoravong (approved):
LGTM.

You're right. If the syntax is invalid, the workflow won’t run at all giving the impression of a false positive. We should consider adding a [linter check](https://github.com/rhysd/actionlint) or something similar in a separate PR.

PR #6363, opened 2025-06-25
[chore][CICD] Add GitHub action linter

crobert-1:

The goal is to lint GitHub actions to ensure there are no syntax errors. As seen in #6360, if a workflow has a syntax error the failure is not shown in the PR introducing the error, causing workflows to break silently until after the PR is merged, causing more noise on `main`. As suggested: https://github.com/signalfx/splunk-otel-collector/pull/6360#pullrequestreview-2954822075

**Testing:**

I broke the Ansible workflow in the same way it had broken on `main`, and the `actionlint` [job failed](https://github.com/signalfx/splunk-otel-collector/actions/runs/15888090505/job/44804872103?pr=6363) with the following message:
```
.github/workflows/ansible.yml:36:5: when a reusable workflow is called with "uses", "runs-on" is not available. only following keys are allowed: "name", "uses", "with", "secrets", "needs", "if", and "permissions" in job "cross-compile" [syntax-check]
   |
36 |     runs-on: ubuntu-24.04
   |     ^~~~~~~~
```

### DIFF
--- a/.github/actionlint.yml
+++ b/.github/actionlint.yml
@@ -7,47 +7,65 @@ self-hosted-runner:
 
 # Path-specific configurations.
 paths:
-  # All Windows Paths should ignore errors related to PowerShell formatting.
-  # Linter shellcheck is not supported for PowerShell
-  .github/workflows/**/*win*.{yml,yaml}:
-    ignore:
-      # Ignore the specific error from shellcheck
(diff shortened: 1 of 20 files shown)


pjanotti (approved):
It does like double-quotes around things that can contain spaces, etc 😄 👍🏼

PR #6368, opened 2025-06-26
[gitlab release] remove TA from release condition

hughesjj:
TA releases should be run in a separate release pipeline

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -439,7 +439,6 @@ agent-bundle-windows:
 .ta-trigger:
   rules:
     - <<: *main-condition
-    - <<: *release-condition
     - <<: *no-schedules-condition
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -439,7 +439,6 @@ agent-bundle-windows:
 .ta-trigger:
   rules:
     - <<: *main-condition
-    - <<: *release-condition
```
pjanotti, line 442: Shouldn't this be a ta-release-condition instead? This way you can add a label to trigger the TA release only with its only label as it is done for mass deployments.

hughesjj: I plan to handle that in one of the tickets we're going over in grooming today ^^

pjanotti (approved):

PR #6370, opened 2025-06-26
Test custom cmd line set via Windows install script

pjanotti:
Adding a test to validate the Windows install script support of customization of the command-line used to launch the collector service. This can be used, for instance, to enable discovery when installing the collector via the install script.

### DIFF
--- a/.github/workflows/scripts/win-test-services.ps1
+++ b/.github/workflows/scripts/win-test-services.ps1
@@ -6,7 +6,8 @@ param (
     [string]$with_fluentd = "true",
     [string]$with_msi_uninstall_comments = "",
     [string]$api_url = "https://api.${realm}.signalfx.com",
-    [string]$ingest_url = "https://ingest.${realm}.signalfx.com"
+    [string]$ingest_url = "https://ingest.${realm}.signalfx.com",
+    [string]$with_svc_args = ""
 )
 
 $ErrorActionPreference = 'Stop'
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/win-test-services.ps1
```diff
@@ -96,3 +97,18 @@ if ($installed_version -gt [Version]"0.97.0.0") {
 If (!(Test-Path -Path "${Env:ProgramData}\Splunk\OpenTelemetry Collector\*_config.yaml")) {
     throw "No config files found in ${Env:ProgramData}\Splunk\OpenTelemetry Collector these files are expected after the install"
 }
+
+$svc_commandline = ""
+try {
+    $svc_commandline = Get-ItemPropertyValue -Path "HKLM:\SYSTEM\CurrentControlSet\Services\splunk-otel-collector" -Name "ImagePath"
+} catch {
+    throw "Failed to retrieve the service command line from the registry."
```
crobert-1, line 105: Just a clarification: Would this be worth failing the test even if `with_svc_args` is empty? I'm seeing that previously this test would pass in this case, but now it would fail.

I'm asking because we could potentially just put this new try-catch inside the `if ($with_svc_args...` statement, but I'm not sure if there's value in having it checked every time or not.

.github/workflows/scripts/win-test-services.ps1
```diff
@@ -96,3 +97,18 @@ if ($installed_version -gt [Version]"0.97.0.0") {
...
+}
+
+if ($with_svc_args -ne "") {
+    if (-not $svc_commandline.EndsWith($with_svc_args)) {
+        throw "Service command line does not match the expected arguments. Found: '$svc_commandline', Expected to end with: '$with_svc_args'"
+    } else {
+        Write-Host "Service command line matches the expected arguments."
```
crobert-1, line 112: I'm not seeing this log message anywhere in this PR's test results. Is that expected?

pjanotti: The workflow/job name here is a bit misleading... you should look at the `windows-installer-script-test` workflow and then you will find this string in the logs, e.g.: https://github.com/signalfx/splunk-otel-collector/actions/runs/15914978072/job/44890752134?pr=6370#step:4:3448

Perhaps, this is something to be improved, since the same test script is also used (with different arguments) in the `win-package-test` workflow in the `windows-install-script-test` jobs... the names are overlapping...

.github/workflows/scripts/win-test-services.ps1
pjanotti: The `ImagePath` must be present on the registry with the name of the executable, you can consider it redundant with other tests here, e.g.: service started, but it won't cause new failures. So yes, even if the `with_svc_args` is empty the installation failed if `ImagePath` is not present (of course the code is not checking the contents).

crobert-1 (approved):

PR #6383, opened 2025-06-30
change choco to release with github artifacts and dotnet deployer addon

hughesjj:
Currently, a failure in choco release will prevent github or the dotnet deployer from releasing, even though they don't reference or depend on the choco release


hughesjj:
ugh.. okay, actually the .nupkg part of the code needs to exist for the sha256 checksums in the github release, apparently?

PR #6416, opened 2025-07-09
[chef][windows] Support option to set service CLI args

pjanotti:
Analog to implementation done for Linux via #6290.

### DIFF
--- a/deployments/chef/kitchen.windows.yml
+++ b/deployments/chef/kitchen.windows.yml
@@ -61,7 +61,7 @@ suites:
         collector_additional_env_vars:
           MY_CUSTOM_VAR1: value1
           MY_CUSTOM_VAR2: value2
-        collector_command_line_args: --discovery
+        collector_command_line_args: --discovery --set=processors.batch.timeout=10s
         with_fluentd: true
     verifier:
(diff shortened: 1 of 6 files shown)


deployments/chef/kitchen.windows.yml
```diff
@@ -57,10 +57,11 @@ suites:
         splunk_memory_total_mib: "256"
         splunk_hec_token: fake-hec-token
         splunk_listen_interface: "0.0.0.0"
-        collector_version: 0.126.0
+        collector_version: 0.127.0
         collector_additional_env_vars:
           MY_CUSTOM_VAR1: value1
           MY_CUSTOM_VAR2: value2
+        collector_command_line_args: --discovery
```
jinja2: I assume chef can handle multiple space-separated args here ootb but just wondering if we should update the test case

pjanotti: [48c40a1](https://github.com/signalfx/splunk-otel-collector/pull/6416/commits/48c40a1d8d0782dda8e72900a6955afaea2c164f)

jinja2 (approved):

PR #6417, opened 2025-07-10
[chore] Adds release-nupkg to the dependencies of the github-release job

hughesjj:
Also renames `nupkg-release` to `release-nupkg`

I believe this is why we haven't seen nupkg be released since making github release independent of choco

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1564,7 +1564,7 @@ release-installers:
   before_script:
     - cp -f dist/signed/install.ps1 packaging/installer/install.ps1
 
-nupkg-release:
+release-nupkg:
   extends: .trigger-filter
   stage: release
   dependencies:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -1780,6 +1780,7 @@ github-release:
     - sign-exe
     - sign-osx
     - release-debs
+    - release-nupkg
```
pjanotti, line 1783: This is the real fix 😄

.gitlab-ci.yml
```diff
@@ -1585,13 +1585,13 @@ nupkg-release:
     - Test-Path .\dist\signed\splunk-otel-collector.${version}.nupkg
   artifacts:
     paths:
-      - dist/signed/*.nupkg
+      - .\dist\signed\*.nupkg
```
pjanotti, line 1588: This should be immaterial, anyway, it won't cause any problems if running on Windows.

PR #6419, opened 2025-07-10
[receiver/discovery] Reduce amount of attributes sent the entities

dmitryax:
We don't need all the attributes from the discovered endpoints. Reducing the list to what is being used


internal/receiver/discoveryreceiver/endpoint_tracker.go
```diff
@@ -368,31 +356,50 @@ func extractIdentifyingAttrs(from pcommon.Map, to pcommon.Map) {
 	}
 }
 
-func deduceServiceName(attrs pcommon.Map) string {
-	if val, ok := attrs.Get(string(conventions.ServiceNameKey)); ok {
-		return val.AsString()
+func extractServiceName(endpointType observer.EndpointType, endpointEnv observer.EndpointEnv) string {
+	labels, labelsFound := endpointEnv["labels"].(map[string]string)
+	podEnv, podEnvFound := endpointEnv["pod"].(observer.EndpointEnv)
```
jinja2, line 361: Is my understanding correct that we'll have this `pod` map available with labels, etc for `Port` type observer?

internal/receiver/discoveryreceiver/endpoint_tracker.go
```diff
@@ -368,31 +356,50 @@ func extractIdentifyingAttrs(from pcommon.Map, to pcommon.Map) {
...
-		if val, ok := labels.Map().Get("app.kubernetes.io/name"); ok {
-			return val.AsString()
+	if !labelsFound && podEnvFound {
+		// If the endpoint is a pod, we can extract labels from the pod env.
+		labels, labelsFound = podEnv["labels"].(map[string]string)
+	}
+
+	// First, try to extract the service name from labels.
+	if labelsFound {
+		if val, ok := labels["app.kubernetes.io/name"]; ok {
```
jinja2, line 372: should we use the more specific labels["app.kubernetes.io/instance"] here and fallback to name?

internal/receiver/discoveryreceiver/endpoint_tracker.go
dmitryax: Right

internal/receiver/discoveryreceiver/endpoint_tracker.go
dmitryax: I reproduced existing behavior. I'll probably apply your suggestion as a separate PR as an explicit improvement

jinja2 (approved):

hughesjj:
Mind adding a preliminary message to the `#unreleased` part of the changelog?

PR #6439, opened 2025-07-17
[chore] forgot to update test for 1.4.3 release

hughesjj:
happy path test fails on last line which is checking this.  Test fails indicating the desired (`v0.128.0`) app was detected in the addon instead of the "erroneously expected" `v0.122.0`

This should never happen again given the update we're about to make to the script


dmitryax (approved):
Please update the PR/commit title

hughesjj:
failure in integration-test-docker is unrelated and honest we should remove that from the Github Actions for this code path

PR #6465, opened 2025-07-21
[DRAFT] Strip debug symbols from executable

pjanotti:
Cheap way to evaluate the effect on the MSI size


pjanotti:
This was just to get a build and measure the effects, no need to merge or any other action.

PR #6468, opened 2025-07-22
Revert "do not fetch the complete history (#6443)"

dmitryax:


jinja2:
The chef-test errors look unrelated to these changes, so shouldn't be blocking. Going to try to see if we can switch to newer version of ruby+debian in a different [PR](https://github.com/signalfx/splunk-otel-collector/pull/6470)

PR #6472, opened 2025-07-22
[DRAFT] Change image used to build Linux agent-bundle

pjanotti:
Avoid depending on Ubuntu for agent bundle files

### DIFF
--- a/packaging/bundle/Dockerfile
+++ b/packaging/bundle/Dockerfile
@@ -23,13 +23,14 @@ ARG DOCKER_REPO=docker.io
 
 
 ######## Base image for subsequent stages ########
-FROM ${DOCKER_REPO}/ubuntu:24.04 AS base
+FROM ${DOCKER_REPO}/python:24.04 AS base
 
 ARG ARCH
+ARG PYTHON_VERSION
+
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN sed -i -e '/^deb-src/d' /etc/apt/sources.list && \
-    apt-get update
+RUN apt-get update
 
 
 ######## Java monitor dependencies and monitor jar compilation ########


packaging/bundle/Dockerfile
```diff
@@ -23,13 +23,14 @@ ARG DOCKER_REPO=docker.io
 
 
 ######## Base image for subsequent stages ########
-FROM ${DOCKER_REPO}/ubuntu:24.04 AS base
+FROM ${DOCKER_REPO}/python AS base
```
dmitryax: Should we pin the image version to avoid unexpected changes and potentially failures?

pjanotti: Although the version is actually pinned via the same pattern as the other usage of the `python` image in this file, i.e.: leveraging the `PYTHON_VERSION` environment variable, it is more readable if it is explicit together with the image,

pjanotti: Slight mistake on my part: it is used below but not as an image, anyway pinning the version via https://github.com/signalfx/splunk-otel-collector/pull/6472/commits/34d592ed4e976281914fc8a788bf1ff04c3062ba

dmitryax (approved):

packaging/bundle/Dockerfile
```diff
@@ -23,13 +23,13 @@ ARG DOCKER_REPO=docker.io
 
 
 ######## Base image for subsequent stages ########
-FROM ${DOCKER_REPO}/ubuntu:24.04 AS base
+FROM ${DOCKER_REPO}/python:${PYTHON_VERSION} AS base
```
jinja2, line 26: Do you know what is the python image based on? given the cmds are same it should be ubuntu or debian, but am wondering which one are using now?

jinja2: Looked up the image, it is debian bookworm.

I see there is a patch version for python available, [3.13.5](https://hub.docker.com/layers/library/python/3.13.5/images/sha256-4ddc936bb22ccb7af9355ed648625db58ca896dd5f40d5fcdbe53b0d7b33ebf5) with 5 fewer high-level vuln than our current version. Wdyt about updating the PYTHON_VERSION?

pjanotti: Let give me it a spin

pjanotti:
Converting to draft until I can investigate the `integration-test-docker` failures.

PR #6487, opened 2025-07-25
[DRAFT] Remove (already deprecated) smartagent/collectd/mysql

pjanotti:
Removing the already deprecated `smartagent/collectd/mysql`


cmd/otelcol/config/collector/config.d.linux/properties.discovery.yaml.example
```diff
@@ -17,7 +17,3 @@ splunk.discovery:
       enabled: true
     postgresql:
       enabled: true
-    smartagent/collectd/mysql:
-      enabled: false
-    smartagent/postgresql:
```
pjanotti, line 22: opportunistic removal: we should not be using deprecated things on examples.

PR #6495, opened 2025-07-26
[chore] Simplify discovery bundler more

dmitryax:
Simplify, remove redundant stuff and consolidate the source for the generated discovery properties files under new metadata directory inside the builder command tree

This is need in order to make it possible to separate static parts like status from the fields that should be available in the discovery config interface


internal/confmapprovider/discovery/README.md
```diff
@@ -141,7 +141,7 @@ I. Receivers
 * `kafkametrics` ([Linux and Windows](./bundle/bundle.d/receivers/kafkametrics.discovery.yaml))
 * `mongodb` ([Linux and Windows](./bundle/bundle.d/receivers/mongodb.discovery.yaml))
 * `mysql` ([Linux and Windows](./bundle/bundle.d/receivers/mysql.discovery.yaml))
-* `nginx` ([Linux and Windows](./bundle/bundle.d/receivers/nginx.discovery.yaml))
+* `nginx` ([Linux](./bundle/bundle.d/receivers/nginx.discovery.yaml))
```
dmitryax, line 144: This is to replicate the existing state in the docs. I don't know why don't have nginx in the windows bundle. If that's a mistake we could restore it separately.

dloucasfx: @dmitryax 
We used to use the Smart Agent nginx collectd monitor, which was a linux only monitor.
But since we moved to start using OTEL nginx, I think for completion we should update the windows bundle to include nginx. 
It can be part of this PR or separate PR.

dloucasfx (changes requested):

internal/confmapprovider/discovery/README.md
dloucasfx: in addition, if we are no longer using receivers/extensions that are Linux only, can we simplify and remove the windows bundle and consolidate all systems in a single bundle?

dmitryax: I'm good with that. Once we have an example of a component that doesn't work for one of the platforms, we could bring this logic back. However, I would make this in a follow-up PR with a proper changelog entry

PR #6503, opened 2025-07-29
[DRAFT][salt] Add option to set service CLI args

pjanotti:
Adds a specific option to set the command-line arguments used to launch the collector service on installed via Salt.

The test for this new config setting is piggy-backing on the test for custom environment variables due to the similarities and to avoid yet another test for some CI that is already relatively long to run.

### DIFF
--- a/deployments/salt/splunk-otel-collector/collector_config.sls
+++ b/deployments/salt/splunk-otel-collector/collector_config.sls
@@ -26,11 +26,14 @@
 
 {% set splunk_listen_interface = salt['pillar.get']('splunk-otel-collector:splunk_listen_interface', '') %}
 
+{% set splunk_otel_collector_command_line_args = salt['pillar.get']('splunk-otel-collector:splunk_otel_collector_command_line_args', '') %}
+
 {% set collector_additional_env_vars = salt['pillar.get']('splunk-otel-collector:collector_additional_env_vars', {}) %}
 
 /etc/otel/collector/splunk-otel-collector.conf:
   file.managed:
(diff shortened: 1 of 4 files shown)


deployments/salt/splunk-otel-collector/collector_config.sls
```diff
@@ -26,11 +26,14 @@
...
 {% set splunk_listen_interface = salt['pillar.get']('splunk-otel-collector:splunk_listen_interface', '') %}
 
+{% set splunk_otel_collector_command_line_args = salt['pillar.get']('splunk-otel-collector:splunk_otel_collector_command_line_args', '') %}
+
 {% set collector_additional_env_vars = salt['pillar.get']('splunk-otel-collector:collector_additional_env_vars', {}) %}
 
 /etc/otel/collector/splunk-otel-collector.conf:
   file.managed:
     - contents: |
+        OTELCOL_OPTIONS={{ splunk_otel_collector_command_line_args }}
```
dloucasfx, line 36: The PR looks good! Approved!
Just a quick question as I am not sure if it's an issue, but do we need to sanitize the input especially that we are blindly setting it as env variable value?

dloucasfx (approved):

deployments/salt/splunk-otel-collector/collector_config.sls
pjanotti: The scope of the env var is only the collector service, the onus is on the user to validate that they have valid command-line parameters, if not the collector service will fail at start time.

PR #6509, opened 2025-07-30
[discovery/receivers] enable rabbitmq default metrics

tgtanyagoel:
This is to match out of the box content.


dmitryax:
Here is the source file now https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/discoverybundler/metadata/receivers/rabbitmq.yaml. Please update that one and run `make bundle.d`

dmitryax (approved):

PR #6524, opened 2025-08-01
[chore] only release when tagged

hughesjj:
Only cuts a release of the TA collector when there's a tagged version

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -488,8 +488,11 @@ agent-bundle-windows:
       when: never
     - <<: *main-condition
     - <<: *no-schedules-condition
-    - &collector-ta-release-condition
+    - &collector-ta-release-precondition
       if: $CI_COMMIT_BRANCH =~ /^release\/technical-addon\/v[0-9]+\.[0-9]+\.[0-9]+$/
+      when: always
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -488,9 +488,12 @@ agent-bundle-windows:
       when: never
     - <<: *main-condition
     - <<: *no-schedules-condition
-    - &collector-ta-release-condition
+    - &collector-ta-release-precondition
```
pjanotti, line 491: shouldn't the branch condition be fully deleted?

hughesjj: I'd argue not.  Only steps which actually release to github should require the tag imo.

We want a way to actually run these tests without needing a tag to do it.  Otherwise we're just going to be pinging Dmitry all the time to delete things when they go wrong, imo.

With this PR, this is saying

"auto run tests + packaging when it's a merge into main, when it's development for the TA, or when we're trying to release the TA", which makes sense imo.

That said, I could see an argument for making the "development branch" a manual action instead of "always"

pjanotti: ah, got it, makes sense - it is fine to keep it.

PR #6526, opened 2025-08-02
Migrate splunk ta otel collector to golang

hughesjj:
This migrates the addon `Splunk_TA_otel` to a golang version, which should be easier to maintain long term.

The idea is we could get rid of maintaining two scripts (the .cmd and .sh), and instead vend binaries.  This also makes signal passing much easier to the Otel TA.

Even better would be to modify `cmd/otelcol` to vend an additional binary that waits for modinput-like xml on startup, but we decided against this when talking about such a ~year ago

### DIFF
--- a/packaging/technical-addon/pkg/splunk_ta_otel_collector/runner/main.go
+++ b/packaging/technical-addon/pkg/splunk_ta_otel_collector/runner/main.go
@@ -70,7 +70,7 @@ func run() int {
 	}
 
 	if err = Run(mip); err != nil {
-		log.Errorln(fmt.Errorf("error running splunk linux autoinstrumentation addon: %w", err))
+		log.Errorln(fmt.Errorf("error running splunk collector addon: %w", err))
 		return -1
 	}
 	return 0


pjanotti:
> Even better would be to modify cmd/otelcol to vend an additional binary that waits for modinput-like xml on startup, but we decided against this when talking about such a ~year ago

I think a dedicated cmd as the best option in general, not to mix with the normal otelcol but a separate cmd. Anyway, there will be challenges because there are is more than a very thin wrapper under cmd/otelcol

pjanotti (approved):

hughesjj:
> > Even better would be to modify cmd/otelcol to vend an additional binary that waits for modinput-like xml on startup, but we decided against this when talking about such a ~year ago
> 
> I think a dedicated cmd as the best option in general, not to mix with the normal otelcol but a separate cmd. Anyway, there will be challenges because there are is more than a very thin wrapper under cmd/otelcol

Yeah, I completely agree... the complexity was the main reason I didn't go with this approach

pjanotti:
We can always re-open if needed, but it is stale at this point.

PR #6536, opened 2025-08-05
[splunk_ta_otel] Workaround ungraceful shutdown of splunkd on Windows

pjanotti:
When `splunkd` dies ungracefully on Windows it doesn't terminate the 
`cmd` process created to run `Splunk_TA_otel.cmd`, the parent process on which `Splunk_TA_otel_utils.ps1` is waiting. As a workaround this change listens on the grand parent process instead (`splunkd.exe`). This does not improve on other issues related to shutdown of the modular input and the collector, but ensures that the collector process is terminated if the parent `splunkd` process terminates (gracefully or not).

### DIFF
--- a/packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
+++ b/packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
@@ -15,9 +15,12 @@ function isOtelProcessRunning($processName)
 	$parent = (Get-WmiObject win32_process | ? processid -eq $PID).parentprocessid
 	$Global:parentPid = $parent
 	otelLogWrite "INFO Parent process id: $parent"
+	$grandParent = (Get-WmiObject win32_process | ? processid -eq $parentPid).parentprocessid
+	$Global:grandParentPid = $grandParent
+	otelLogWrite "INFO GrandParent process id: $Global:grandParentPid"
 	$child = (Get-WmiObject win32_process | ? parentprocessid -eq $parent | ? processname -eq $processName).processid
 	$Global:otelPid = $child
-	otelLogWrite "INFO Collector process id: $child"
+	otelLogWrite "INFO Collector process id: $otelPid"
 	$otelProcess = Get-Process -Id $child
 	if($otelProcess)
 	{
@@ -55,7 +58,7 @@ otelLogWrite "INFO Splunk_TA_otelutils.ps1 started"
 start-sleep -s 3
 if (isOtelProcessRunning($otelProcessName)) {
 	otelLogWrite "INFO Otel agent running"
-	waitForExit($parentPid)
+	waitForExit($grandParentPid)
 	forceStopOtelProcess($otelPid)
 	CheckOtelProcessStop($otelPid)
 } else {


packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
```diff
@@ -1,63 +1,85 @@
...
 
 function otelLogWrite
 {
    Param ([string]$logstring)
    $timestamp = Get-Date -Format "yyyy-MM-dd HH:mm:ss.fff K"
-   Add-content $splunkTAPSLogFile -value "$timestamp $logstring"
+
+   # Write-Warning is much less verbose than Write-Error and the other Write-* may end up
+   # not being captured by the parent process (expected to be the cmd that launched the script)
+   Write-Warning "$timestamp Splunk_TA_otel_utils.ps1: $logstring"
```
dloucasfx: I know we talked about it, please remind me why we are not using `Write-Information` instead?

packaging/technical-addon/Splunk_TA_otel/default/inputs.conf
```diff
@@ -11,7 +11,7 @@ sourcetype = Splunk_TA_otel
 [Splunk_TA_otel://Splunk_TA_otel]
 disabled=false
 start_by_shell=false
-interval = 60
+interval = 0
```
crobert-1: What is the purpose of `interval`? (I'm less asking about why it's changed here, more asking what it's used for )

Also, I see a number of examples ([README](https://github.com/signalfx/splunk-otel-collector/blob/7870d6ef9b9b5a4762862e99c61ad2f8024917d7/packaging/technical-addon/cmd/modinput_config_generator/README.md#L22) and [default for testing and examples](https://github.com/signalfx/splunk-otel-collector/blob/7870d6ef9b9b5a4762862e99c61ad2f8024917d7/packaging/technical-addon/cmd/modinput_config_generator/internal/testdata/pkg/sample_addon/assets/default/inputs.conf.tmpl#L4)) where the value is still being set to `60`. This isn't going to break, it just means the tests and examples no longer match default behavior. It doesn't have to block this PR, but it'd be good if they're kept in sync unless there's a good reason not to.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel.cmd
```diff
@@ -190,11 +190,13 @@ if "%discovery_properties_value%" == "" (
 call :extract_bundle
 
 set "command_line=%splunk_TA_otel_app_directory%%splunk_otel_process_name%"
+call :splunk_TA_otel_log_msg "INFO" "Terminating any orphaned instances of the collector (if present)..."
+start /B "" /I /WAIT powershell -ExecutionPolicy ByPass "& { Get-WmiObject Win32_Process | ? ProcessName -eq '%splunk_otel_process_name%' | ? CommandLine -like '*%command_line%*' | ForEach-Object { taskkill.exe /f /pid $_.ProcessId } }" >> "%splunk_TA_otel_log_file%"
```
crobert-1, line 194: Is there any testing to ensure this works as expected? I assume you're able to manually test at least?

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
```diff
@@ -1,63 +1,85 @@
...
+}
+
+function getParentProcessId($id)
+{
+	return (Get-WmiObject Win32_Process | Where-Object ProcessId -eq $id).ParentProcessId
+}
+
+function getParentAndGrandParent($id)
```
crobert-1: Do we need this method, or could we just change it to `getGrandParentPID`? I prefer to not have the multiple return values, and the parent PID functionality is already accounted for by `getParentProcessId`.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
```diff
@@ -1,63 +1,85 @@
...
    Param ([string]$logstring)
    $timestamp = Get-Date -Format "yyyy-MM-dd HH:mm:ss.fff K"
-   Add-content $splunkTAPSLogFile -value "$timestamp $logstring"
+
+   # Write-Warning is much less verbose than Write-Error and the other Write-* may end up
+   # not being captured by the parent process (expected to be the cmd that launched the script)
+   Write-Warning "$timestamp Splunk_TA_otel_utils.ps1: $logstring"
+}
+
+function getParentProcessId($id)
```
crobert-1: Nit: This file currently has the following method names:
```
getParentProcessId
getParentAndGrandParent
getOtelProcessPid
forceStopProcess
```

It would be really good to be consistent with naming around process ID. I think it's good to clarify the type of the return value, so I definitely prefer including the suffix `Id` over not including it. For anything `Pid`, I guess we'd need to clarify whether it's parent ID or process ID.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
```diff
@@ -1,63 +1,85 @@
...
+{
+	return (Get-WmiObject Win32_Process | Where-Object ProcessId -eq $id).ParentProcessId
+}
+
+function getParentAndGrandParent($id)
+{
+	try {
+		$parentId = getParentProcessId($id)
+		otelLogWrite "INFO Parent process id: $parentId"
```
crobert-1: Are these useful for every run? Should we make them debug if they're not always useful?

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
```diff
@@ -1,63 +1,85 @@
...
+		$grandParentId = getParentProcessId($parentId)
+		otelLogWrite "INFO GrandParent process id: $grandParentId"
```
crobert-1: I see a comment below stating the value of directly using `Write-Error`, but what's the purpose of `otelLogWrite`?

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
```diff
@@ -1,63 +1,85 @@
...
-
-function CheckOtelProcessStop($processId)
-{
-	$processInfo = Get-WmiObject win32_process | ? processid -eq $processId
-	if($processInfo) {
-		otelLogWrite "ERROR Otel agent didn't stop"
+otelLogWrite "INFO Splunk_TA_otelutils.ps1 started"
+try {
+	otelLogWrite "INFO sleeping ..."
+	start-sleep -s 3
```
crobert-1: I know this is just moved and not changed with this PR, but maybe as a future enhancement we should be doing some kind of polling with a deadline, rather than a single sleep 3 seconds and check here.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
```diff
@@ -1,63 +1,85 @@
...
 	} else {
-		otelLogWrite "INFO Otel agent stopped"
-	}
+		otelLogWrite "INFO Otel agent running"
+		$parentJob      = Start-Job { Wait-Process -Id $using:parentId } -Name "cmd job"
+		$grandParentJob = Start-Job { Wait-Process -Id $using:grandParentId } -Name "splunkd job"
+		$otelJob        = Start-Job { Wait-Process -Id $using:otelPid } -Name "Otel agent job"
+		otelLogWrite "INFO waiting on termination of any one of splunkd, cmd, and Otel agent"
+		$finishedJob = Wait-Job -Any -Job $parentJob,$grandParentJob,$otelJob
+		otelLogWrite "INFO at least one of the jobs finished:`n`t'$($finishedJob.Name)': finished with $($finishedJob.JobStateInfo)"
```
crobert-1: Maybe include in this log message that the collector process is stopped as a result of this. (Please feel free to ignore if it'll be obvious in the logs, especially if the `waiting on termination` and `at least one of the jobs...` logs are going to be next to each other in the log file.)

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
crobert-1: Does the comment above the line you're asking about explain why? It seems like it explains why, but maybe there's something different you're asking about.

pjanotti: Yes, Curtis is correct: `Write-Information`, specifically, ends up not being captured by any log. My guess is that it uses `stdout` and the `cmd` call is only capturing `stderr`.

packaging/technical-addon/Splunk_TA_otel/default/inputs.conf
pjanotti: This is supposed to be the time that `splunkd` waits to relaunch the script when it terminates, see https://help.splunk.com/en/splunk-enterprise/developing-views-and-apps-for-splunk-web/10.0/modular-inputs/modular-inputs-configuration#ariaid-title4

That said I'm questioning if it is behaving as documented due to some issues that I'm encountering in manual tests.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel.cmd
pjanotti: Manual test, we need to add tests for the type of issue that is being addressed in this PR.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
pjanotti: I was trying to keep the change smaller and initially this was a refactor of renaming a method that did more than what its name implied. However, I see your point: perhaps better to go for the full refactor to make things easier to read.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
pjanotti: We need a better method to log stuff from this script, right now, there is no way to make these DEBUG. Anyway, this should be very low volume since on normal operation there aren't many starts of the script.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
pjanotti: Right now, just adding the timestamp.

dloucasfx (approved):

packaging/technical-addon/Splunk_TA_otel/default/inputs.conf
pjanotti: It does behave as documented, but splunkd and process it launches, modular inputs, and processes launched by modular inputs on Windows have a somewhat subtle interactions. The key is as long as splunkd interprets that the modular input exited it restarts taking into account the interval as documented per link above.

dloucasfx (approved):

dloucasfx:
@pjanotti you just need to add the changelog

pjanotti:
@dloucasfx nice that the check for change log entries is in place! Thanks!

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
```diff
@@ -1,63 +1,88 @@
...
+	Param ([string]$msg)
+	$timestamp = Get-Date -Format "yyyy-MM-dd HH:mm:ss.fff K"
+
+	# Write-Warning is much less verbose than Write-Error and the other Write-* may end up
+	# not being captured by the parent process (expected to be the cmd that launched the script)
+	Write-Warning "${scriptName}: $timestamp $msg"
 }
 
-function isOtelProcessRunning($processName)
+function Get-Win32_ProcessByNameAndParentId
```
crobert-1, line 15: It might be helpful to add a comment explaining this method's functionality, especially regarding the fact that in the case of a timeout it returns null (it doesn't error or log).

pjanotti:
Per offline conversation omitting a changelog.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
```diff
@@ -1,63 +1,88 @@
...
+	Write-Log "INFO Otel agent running (PID $($otelProcess.ProcessId))"
+	$parentJob      = Start-Job { Wait-Process -Id $using:parentProcess.ProcessId }      -Name "cmd job"
+	$grandParentJob = Start-Job { Wait-Process -Id $using:grandParentProcess.ProcessId } -Name "splunkd job"
+	$otelJob        = Start-Job { Wait-Process -Id $using:otelProcess.ProcessId }        -Name "Otel agent job"
+
+	Write-Log "INFO waiting on termination of splunkd, cmd, or the Otel agent"
+	$finishedJob = Wait-Job -Any -Job $parentJob,$grandParentJob,$otelJob
+
+	Write-Log "INFO at least one of the jobs finished:`n`t'$($finishedJob.Name)': finished with $($finishedJob.JobStateInfo)"
+	Stop-ProcessById $otelProcess.ProcessId
```
crobert-1, line 79: Out of complete ignorance: Why don't we stop the splunkd job?

pjanotti: Stopping a job object in this case would cause the code waiting on the process to end, not the process that we are waiting on - so it doesn't have the desired effect.

Given your Q I notice that, perhaps, it will be good form to call `Remove-Job` on the job objects. However, In this specific case it is not a problem because they will be terminated when powershell exits.

packaging/technical-addon/Splunk_TA_otel/windows_x86_64/bin/Splunk_TA_otel_utils.ps1
pjanotti: Sorry, missed this comment before merging - will add once we have to change something else. Notice that the invocation of the method itself is kind of self documenting given the name of the function and the use of named parameters for timeout and sleep interval.

crobert-1: No worries. The actual functionality makes sense from the name, my main goal was just to document returns values in special cases. Not a blocker by any means 👍

PR #6584, opened 2025-08-12
[chore] fix TA orca test config

dloucasfx:
**Description:**
This test is supposed to validate SA agent bundle/collectd.
- Renamed the config
- Used original config and removed mysql and added collectd/cpu 

As I have mentioned, ORCA test is no longer needed, so we will take it out in the future, but this is good enough to get the release out


packaging/technical-addon/packaging-scripts/cicd-tests/agent-bundle/ta-agent-collectd-config.yaml
dmitryax, line 1: Why do we need the whole config with irrelevant elements? I'll be much harder to maintain than just 1 receiver and exporter.

packaging/technical-addon/packaging-scripts/cicd-tests/agent-bundle/orca-collectd-test.sh
```diff
@@ -45,6 +45,7 @@ while [ $ATTEMPT -le $MAX_ATTEMPTS ]; do
     # Copy logs from container
     scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i ~/.orca/id_rsa -r -P $SSH_PORT "splunk@$ip_addr:/opt/splunk/var/log/splunk/" "$TEST_FOLDER"
     if safe_grep_log "Starting otel agent" "$TEST_FOLDER/splunk/Splunk_TA_otel.log" && \
+        safe_grep_log "collectd\/cpu" "$TEST_FOLDER/splunk/otel.log" && \
```
dmitryax, line 48: Can't the fix be just grepping for `collectd/apache` and keeping the existing config?

dloucasfx: The string for `collectd/apache` is not in the logs, I will have to look for something else, however, this is not the issue as I wanted to test and make sure the collectd "python" like `collectd/apache`, generates the `config dir` correctly and does not fail (hence this PR). 
On the other hand, I added `collectd/cpu` as it's a built in collectd plugin .
One last note, based on internal discussion, collect python plugins are planned to be removed very soon

packaging/technical-addon/packaging-scripts/cicd-tests/agent-bundle/ta-agent-collectd-config.yaml
dloucasfx: As I mentioned in the description, I just copied the original file being used in the test, prior to the regression.
```
This test is supposed to validate SA agent bundle/collectd.

Renamed the config
Used original config and removed mysql and added collectd/cpu
As I have mentioned, ORCA test is no longer needed, so we will take it out in the future, but this is good enough to get the release out
```

I really didn't want to spend too much time on this as we should take the whole orca test out soon, I just want this to pass so we can get the release out

dloucasfx: @dmitryax simplified the config

PR #6594, opened 2025-08-13
[chore] Add nolint:gosec pragma for local helper command

dmitryax:


crobert-1 (approved):
Thanks for including the explanation with the nolint, very helpful 👍

PR #6595, opened 2025-08-13
Exposing configs via expvarz

pjanotti:
**Description:**
Expose the initial and effective configs via the `expvar` middleware of the `zpages` extension. This is the first step to eventually remove the custom endpoint currently used to expose the collector configuration for support cases.

This change:

- Adds a `configconverter` to expose the configs to `expvar`.
- Updates the support scripts to collect the JSON exposed via the `zpages`.
- Enables the `expvar` endpoint for all configurations with `zpages` (as the current config debug endpoint it is also enabled by default).

See `.chloggen/expose-config-via-zpages-expvar.yaml` for further information.

Some useful commands to get nicely formatted YAML from the zpages:

On Windows Powershell 5.1:
```powershell
(Invoke-WebRequest http://localhost:55679/debug/expvarz).Content | ConvertFrom-Json | Select-Object -ExpandProperty splunk.config.effective
```

On Bash (with `curl` and `jq`):
```bash
curl http://localhost:55679/debug/expvarz --silent | jq -r '.["splunk.config.initial"]'
```

**Testing:**
Manual validation with the collector and unit tests for the `expvar` `configconverter`.


internal/configconverter/expvar.go
```diff
@@ -0,0 +1,97 @@
...
+	effectiveMutex sync.RWMutex
+}
+
+// GetExpvarConverter returns the singleton instance of expvarConverter that publishes
+// two entries to the `expvar` JSON map:
+//   - "splunk.config.effective": a map with the effective configuration being used by the collector.
+//   - "splunk.config.initial": a map with the initial configuration being used by the collector.
+func GetExpvarConverter() *ExpvarConverter {
+	expvarConverterOnce.Do(func() {
+		expvarPublishOnce.Do(func() {
```
dmitryax: Why do we need `expvarPublishOnce` if it's already wrapped in `expvarConverterOnce`?

internal/configconverter/expvar.go
```diff
@@ -0,0 +1,97 @@
...
+			expvar.Publish("splunk.config.effective", expvar.Func(func() any {
+				instance := GetExpvarConverter()
+				instance.effectiveMutex.RLock()
```
dmitryax: why do we need the mutexes if it's controlled by `expvarConverterOnce.Do` which cannot be called more than once?

internal/configconverter/expvar.go
```diff
@@ -0,0 +1,97 @@
...
+			expvar.Publish("splunk.config.initial", expvar.Func(func() any {
+				instance := GetExpvarConverter()
+				instance.initialMutex.RLock()
+				defer instance.initialMutex.RUnlock()
+				configYAML, _ := yaml.Marshal(instance.initial)
+				return string(configYAML)
+			}))
+		})
+
+		expvarConverterInstance = &ExpvarConverter{
```
dmitryax, line 49: Can we just move this thing to the beginning of `expvarConverterOnce.Do(func() {` and avoid calling `instance := GetExpvarConverter()` from the Publish methods?

internal/configconverter/expvar.go
pjanotti: This is actually a callback that is going to be called whenever the zpages with the expvar page is loaded. In practice we don't expect concurrency here but since the updates to the config could be triggered concurrently I wanted to protect against the (small) risky of torn/corrupted configuration.

internal/configconverter/expvar.go
pjanotti: The comment on the declaration of `expvarPublishOnce` explains that, but it will be better to move it here to make it clear the purpose. We could make it with `init()` but I wanted to only do the expvar publication if the instance was going to be created.

dmitryax: But this line is already inside `expvarConverterOnce.Do(func() {})` so it's already guaranteed to be called only once, right?

pjanotti: Not in the tests, the instance is recreated with different settings for test purposes.

dmitryax: What tests fail if we remove `expvarPublishOnce`? Typically we should write tests to cover production code, not the other way around.

pjanotti: You are right, it probably made more sense in some earlier iteration, but right now I can remove it by changing a test or two. Let me see how that looks like.

pjanotti: Removed the atypical code and changed only a low value test, better this way.

dmitryax (approved):

PR #6621, opened 2025-08-19
Add the `docker_stats` receiver

pjanotti:
Adds the `docker_stats` receiver to the Splunk distribution of the OpenTelemetry collector.


.chloggen/add-dockerstatsreceiver.yaml
```diff
@@ -0,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: dockerstatsreceiver
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Added the `docker_stats` receiver to the Splunk OpenTelemetry Collector."
```
crobert-1:
```suggestion
note: "Added the `docker_stats` receiver to the Splunk distribution of the OpenTelemetry Collector."
```
Nit: Wording here can be pretty important around naming of the distribution. From what I remember it's very important to include the word distribution in some form. If this is too long of a description you can reduce it to something like "Added the `docker_stats` receiver to the Splunk distro."

PR #6625, opened 2025-08-19
[tests/general] Add test for default logs configuration

crobert-1:

The goal is to test our default recommendation for a [logs configuration file](https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/logs_config_linux.yaml). This PR adds a test that runs the collector as a process that has a filelog receiver set to read from the syslog, then use a `splunk_hec` exporter. A `splunk_hec` receiver sink has been added to collect the exported logs for testing, similar to the OTLP receiver sink.

**Testing:**
[Passing test](https://github.com/signalfx/splunk-otel-collector/actions/runs/17082464640/job/48439851638?pr=6625):
```
=== RUN   TestDefaultLogConfig
--- PASS: TestDefaultLogConfig (2.01s)
```

### DIFF
--- a/tests/general/default_config_linux_test.go
+++ b/tests/general/default_config_linux_test.go
@@ -93,7 +93,7 @@ func TestDefaultLogConfig(t *testing.T) {
 				}
 			}
 		}
-		require.Greater(c, tc.HECReceiverSink.LogRecordCount(), 0)
+		require.Positive(c, tc.HECReceiverSink.LogRecordCount())
 		require.True(c, foundSyslog)
 	}, 20*time.Second, 500*time.Millisecond)
 


cmd/otelcol/config/collector/logs_config_linux.yaml
```diff
@@ -695,7 +698,8 @@ extensions:
   file_storage/filelogs:
     # Location where checkpoint files are stored. This directory must exist and
     # must be writable by the collector..
-    directory: /var/lib/otelcol/filelogs
+    directory: "${SPLUNK_FILE_STORAGE_EXTENSION_PATH}"
```
crobert-1: This change was required as the collector does not have permission to create the original default directory in tests. I thought moving to an environment variable was a cleaner solution (as customers may also run into the same issue) than attempting a workaround in testing.

tests/testutils/testcase.go
```diff
@@ -141,13 +162,16 @@ func (t *Testcase) newCollector(initial Collector, configFilename string, builde
...
 		"OTLP_ENDPOINT":  t.OTLPEndpointForCollector,
+		"SPLUNK_HEC_URL": t.HECEndpointForCollector,
 		"SPLUNK_TEST_ID": t.ID,
 	}
 
 	if configFilename != "" {
-		collector = collector.WithConfigPath(
-			path.Join(".", "testdata", configFilename),
-		)
+		if !filepath.IsAbs(configFilename) {
```
crobert-1, line 190: We specifically want to test the default configuration file, not a copy. Checking for absolute path enables this functionality. Feel free to let me know if you have any suggestions for how better to handle this. Up until this point all tests have used a test config under a local `./testdata` directory. These tests are not impacted by this change. Ideas are welcome if anyone has a better way of solving this.

cmd/otelcol/config/collector/logs_config_linux.yaml
```diff
@@ -8,6 +8,9 @@
 # - SPLUNK_INGEST_URL: The Splunk ingest URL, e.g. https://ingest.us0.signalfx.com
 # - SPLUNK_LISTEN_INTERFACE: The network interface the agent receivers listen on.
 # - SPLUNK_MEMORY_LIMIT_MIB: 90% of memory allocated
+# - SPLUNK_FILE_STORAGE_EXTENSION_PATH: Path to directory that will be used to contain file offsets
+#   used by the filelog receivers. This allows the receivers to pick up where they left off in case of
+#   a collector restart. This directory must be writable by the collector. e.g. /var/lib/otelcol/filelogs
```
crobert-1:
```suggestion
#   used by the filelog receivers ( e.g. `/var/lib/otelcol/filelogs`). This allows the receivers to pick up
#   where they left off in case of a collector restart. This directory must be writable by the collector.
```

cmd/otelcol/config/collector/logs_config_linux.yaml
```diff
@@ -683,19 +686,18 @@ exporters:
...
   # Checkpoints allow the receiver to pick up where it left off in the case of a
   # collector restart.
   file_storage/filelogs:
-    # Location where checkpoint files are stored. This directory must exist and
-    # must be writable by the collector..
-    directory: /var/lib/otelcol/filelogs
+    # Location where checkpoint files are stored. This directory must be writable
+    # by the collector.
+    directory: "${SPLUNK_FILE_STORAGE_EXTENSION_PATH}"
+    create_directory: true
```
crobert-1: I chose to enable `create_directory` by default as it helps new installation startup proceed more smoothly. When running in a new test or production environment without this option being enabled, users have to manually create the directory or the collector will fail on startup.

.chloggen/default_logs_test.yaml
```diff
@@ -0,0 +1,20 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
jinja2: I was under the impression that logs_config_linux.yaml file is just for reference. Is this used by an installer or something?

crobert-1: It's just a reference at this point. I was thinking of marking it as breaking in case anyone is directly using it in their environment, but if that's unnecessary we can mark it as an enhancement.

dmitryax: I think we should set `SPLUNK_FILE_STORAGE_EXTENSION_PATH` by default. That will make this PR an `enhancement`. That should be probably another PR?

crobert-1: Good idea @dmitryax, I've opened https://github.com/signalfx/splunk-otel-collector/pull/6656 to handle moving to an environment variable here, and setting it by default.

crobert-1: #6656 has been merged, this PR no longer requires a changelog entry

tests/testutils/hec_receiver_sink.go
crobert-1, line 1: `hec_receiver_sink.go` and `hec_receiver_sink_test.go` are heavily lifted from the neighboring files `otlp_receiver_sink.go` and `otlp_receiver_sink_test.go`.

PR #6627, opened 2025-08-19
Prepare 0.132.0 changelog

dmitryax:


atoulme (approved):

dloucasfx (approved):
LGTM, one thing through as I am not sure if it will be an issue as a pull will redirect to an issue if it does not exists, but when I worked on the previous changelog, I have noticed that majority of the links are to issues instead of PRs, so I wonder if we should default to issue instead of PR

dmitryax:
@dloucasfx I've added a script that prepares the changelog and figure outs correct links that don't cause redirects https://github.com/signalfx/splunk-otel-collector/pull/6628

PR #6631, opened 2025-08-21
[chore] delete unused code

atoulme:


internal/signalfx-agent/pkg/monitors/internalmetrics/monitor.go
```diff
@@ -32,10 +31,9 @@ type Config struct {
 // Monitor for collecting internal metrics from the simple server that dumps
 // them.
 type Monitor struct {
-	Output    types.Output
-	AgentMeta *meta.AgentMeta
```
atoulme, line 36: I could not find a place where this field was ever set.

internal/signalfx-agent/pkg/core/dpfilters/set.go
```diff
@@ -9,9 +9,6 @@ import (
 // for a datapoint to be matched.
 type FilterSet struct {
 	ExcludeFilters []DatapointFilter
-	// IncludeFilters are optional and serve as a top-priority list of matchers
-	// that will cause a datapoint to always be sent
-	IncludeFilters []DatapointFilter
```
atoulme, line 14: Those filters were never set.

dmitryax: Aren't they exposed in the smartagent receivers config interface?

atoulme: No, they were not afaict

PR #6634, opened 2025-08-21
[chore] Remove routingprocessor

pjanotti:
This component was deleted from upstream. Removing it from our distribution.


crobert-1 (commented):
`internal/components/components_test.go` also has a reference that needs removed.

crobert-1 (approved):
Maybe when we bring over contrib's changelog we make the prefix `(Splunk, Contrib) ...` to show it's in both (if we remember)

dmitryax (approved):

PR #6674, opened 2025-08-27
[DRAFT][chore] Smoking test TA on GH runner

pjanotti:

### DIFF
--- a/.github/workflows/splunk-ta-otel.yml
+++ b/.github/workflows/splunk-ta-otel.yml
@@ -13,6 +13,16 @@ on:
       - 'pkg/**'
       - 'tests/**'
       - '!**.md'
+  workflow_dispatch:
+    inputs:
+      splunk_uf_version:
+        description: 'Splunk UF version'
+        required: false
+        default: '9.4.0'
+      splunk_uf_build_hash:
+        description: 'Splunk UF build hash'
+        required: false
+        default: '6b4ebe426ca6'
 
 concurrency:
   group: splunk-technical-addon-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/splunk-ta-otel.yml
```diff
@@ -147,6 +161,76 @@ jobs:
           name: Splunk_TA_otel
           path: ${{github.workspace}}/build/out/distribution/Splunk_TA_otel.tgz
 
+  smoke-test-windows:
+    runs-on: windows-latest
```
crobert-1: Maybe just personal preference, but it may help in future debugging if we pin this to a specific Windows version. We've had trouble at times debugging wondering if it was something functionally that broke, or just an environment change, so it may be helpful to reduce running environment changes. We already have to bulk update runner versions, so I assume it won't add too much maintenance burden.

jinja2 (approved):

.github/workflows/splunk-ta-otel.yml
```diff
@@ -147,6 +161,76 @@ jobs:
...
+          }
+
+          if ($found) {
+              Write-Host "Collector was launched after $elapsed seconds"
+          } else {
+              Write-Host "Timeout reached. Collector was not launched after $elapsed seconds"
+              exit 1
+          }
+
+      - name: Check Status
```
crobert-1:
```suggestion
      - name: Log status after test
```
Nit: Feel free to reword, but I would have assumed from the name `Check` that it's checking something and could cause the test to fail if the checks fail.

.github/workflows/splunk-ta-otel.yml
```diff
@@ -147,6 +161,76 @@ jobs:
...
+        if: always()
+        run: |
+          cd "${Env:ProgramFiles}\SplunkUniversalForwarder\var\log\splunk"
+          Start-Sleep 30
```
crobert-1: What's the purpose of the sleep call here?

.github/workflows/splunk-ta-otel.yml
```diff
@@ -147,6 +161,76 @@ jobs:
...
+          dir
+          Write-Host "`n======= otel.log contents ======="
+          Get-Content otel.log -ErrorAction SilentlyContinue
+          Write-Host "`n======= Splunk_TA_otel.log contents ======="
```
crobert-1, line 228: I thought about potentially only logging on failure, but I think it'll be helpful when something fails to have successful logs as a reference to see what changed or went wrong on the failed run 👍

.github/workflows/splunk-ta-otel.yml
pjanotti: Not needed anymore removing, removing.

crobert-1 (approved):

PR #6677, opened 2025-08-28
[chore] Remove github.com/jaegertracing/jaeger dependency

dmitryax:
Remove github.com/jaegertracing/jaeger dependency to avoid getting CVEs associated with it. Use the slim github.com/jaegertracing/jaeger-idl module for the model instead.


jinja2 (approved):

crobert-1 (approved):
🎉

PR #6680, opened 2025-08-29
[chore][prepare-changelog] Properly handle changelog entries with multiple issues

crobert-1:

When preparing the changelog notes for v0.133.0 I found the prepare changelog script couldn't properly handle changelog entries with more than one issue. The `grep` and `sed` operations were only looking for `(#NUMBER)`, filtering out valid values such as `(#NUMBER1, #NUMBER2)` and `(#NUMBER1, #NUMBER2, #NUMBER3)`. This fixes the operations to be able to properly link issues when there are more than one.

### DIFF
--- a/.github/workflows/scripts/prepare-changelog.sh
+++ b/.github/workflows/scripts/prepare-changelog.sh
@@ -50,7 +50,7 @@ convert_pr_issue_links() {
     local result="$content"
     
     # Find all (#number) patterns
-    local numbers=$(echo "$content" | grep -oE '\(#[0-9]+((, #[0-9]+)?)*\)' | grep -oE '[0-9]+' | sort -u)
+    local numbers=$(echo "$content" | grep -oE '\(#[0-9]+((, ?#[0-9]+)?)*\)' | grep -oE '[0-9]+' | sort -u)
 
     local total=$(echo "$numbers" | wc -w)
 
@@ -58,7 +58,9 @@ convert_pr_issue_links() {
         local type=$(check_pr_or_issue "$repo_url" "$number")
 
         # Replace #number with proper markdown link
-        result=$(echo "$result" | sed "s|#$number|[#$number]($repo_url/$type/$number)|g")
+        # \([^0-9]\) - This is a capture group to ensure sed ONLY matches the given number. This avoids matching
+        # numbers that are part of larger numbers (e.g., #123 should not match #1234).
+        result=$(echo "$result" | sed "s|#$number\([^0-9]\)|[#$number]($repo_url/$type/$number)\1|g")
     done
 
     echo "$result"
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts/prepare-changelog.sh
```diff
@@ -50,17 +50,17 @@ convert_pr_issue_links() {
     local result="$content"
     
     # Find all (#number) patterns
-    local numbers=$(echo "$content" | grep -oE '\(#[0-9]+\)' | grep -oE '[0-9]+' | sort -u)
-    
+    local numbers=$(echo "$content" | grep -oE '\(#[0-9]+((, #[0-9]+)?)*\)' | grep -oE '[0-9]+' | sort -u)
```
crobert-1: The change here is to capture optional groups of `, #number` after the first required group. The regex captures as many as it can find, I tested 1 issue up to 4 to ensure it worked as expected.

.github/workflows/scripts/prepare-changelog.sh
```diff
@@ -50,17 +50,17 @@ convert_pr_issue_links() {
...
-    
+
     for number in $numbers; do
         local type=$(check_pr_or_issue "$repo_url" "$number")
-        
-        # Replace (#number) with proper markdown link
-        result=$(echo "$result" | sed "s|(#$number)|([#$number]($repo_url/$type/$number))|g")
+
+        # Replace #number with proper markdown link
+        result=$(echo "$result" | sed "s|#$number|[#$number]($repo_url/$type/$number)|g")
```
crobert-1: `grep` is already filtering down possible matches to at least start the match with `(#number`, so I figure this isn't too dangerous of a regex modification on sed.

Also, for a given repository's changelog notes, the assumption is already that `#number` is a local repo's issue or PR, so even if there's a false positive (an issue or PR not surrounded by parentheses) it will still be the correct local repo reference.

crobert-1 (commented):
_My editor automatically removes trailing whitespace in lines on save, that's where all of the whitespace changes are coming from_

crobert-1:
> Minor suggestion: update the template to show that how multiple issues should be typed, I mainly worried about the space.

Good catch Paulo. I'll test some more regex to see if I can get it to work with spaces being optional, otherwise I'll add a note that they're required in the template.

dmitryax (approved):

crobert-1:
I found that the space between issues (in the case of plural) is added [automatically every time by chloggen](https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/09eb2ffc724d7471fb8979acca243227aced0841/chloggen/internal/chlog/summary.tmpl#L4). However, I don't want to depend on that so I've made the space optional in the grep regex (added a `?` after it).

Also, I found that the original `sed` search and replace that I had made was too general, allowing for incorrect functionality if one number was a substring of another. I've fixed the sed command for this case.

PR #6718, opened 2025-09-05
[fips] Use native FIPS support as per go 1.24

atoulme:


cmd/otelcol/main_fipsonly.go
```diff
@@ -13,8 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build linux && boringcrypto && cgo && (amd64 || arm64)
+//go:build requirefips
 
-package main
+//go:debug fips140=only
```
jinja2:
```suggestion
//go:debug fips140=on
```
`only` is a stricter experimental mode where even just using an algo like sha-1 to generate a non-crypto use hash will error. I think we should start with the std `on` mode. I am referring to this blog for the difference in 2 modes https://go.dev/blog/fips140#a-native-developer-experience

PR #6748, opened 2025-09-11
[otel-ta] Move control of log file from inputs.conf to collector YAML

pjanotti:
Removed the configuration to select log file from the `inputs.conf` and instead use the collector YAML configuration file to control telemetry and logs from the collector. The new default is to `error` level and writing to `stderr`, this output is automatically captured by `splunkd` and logged on `$SPLUNK_HOME/var/log/splunk/splunkd.log`. Note: if the TA scripts are used with the Splunk OTel Collector prior to `v0.135.0` logs do not follow the prefix convention understood by `splunkd` any log messages are labeled as `ERROR` in the `splunkd.log` file, e.g.:
```terminal
09-11-2025 18:03:17.960 +0000 ERROR ExecProcessor [1737 ExecProcessor] - message from "/opt/splunk/etc/apps/Splunk_TA_otel/linux_x86_64/bin/Splunk_TA_otel.sh" 2025/09/11 18:03:17 settings.go:524: Set soft memory limit set to 460 MiB
09-11-2025 18:03:17.960 +0000 ERROR ExecProcessor [1737 ExecProcessor] - message from "/opt/splunk/etc/apps/Splunk_TA_otel/linux_x86_64/bin/Splunk_TA_otel.sh" 2025/09/11 18:03:17 settings.go:373: Set garbage collection target percentage (GOGC) to 400
```

The level issue above was fixed via #6749 

Users that still want the log file should use the `service.telemetry.logs` configuration in the YAML file to control how the collector launched by `Splunk_TA_otel` is going to log. If they want to keep the same behavior prior to this change they can add the following to the `service` configuration in the YAML configuration file:
```terminal
  telemetry:
    logs:
      level: info
      output_paths: [ "${env:SPLUNK_HOME}/var/log/splunk/otel.log" ]
```

Since there no more a default log file the configuration for Splunk to collect the old default log file was also removed from the default `inputs.conf`.

PR #6749, opened 2025-09-11
[otel-ta] Change internal/settings log format for better Splunk TA reporting

pjanotti:
Keep the current log formatting for `internal/settings/`, but when the environment var `SPLUNK_HOME` is defined assume that it is running as TA and log in a format that is understood by `splunkd`, see https://docs.splunk.com/Documentation/Splunk/9.4.2/AdvancedDev/ModInputsLog. This avoids the log messages from `internal/settings` showing up as `ERROR` on `splunkd.log`

`SPLUNK_HOME` is not defined for the machine after the installation of Splunk, however, it is defined when launching any TA so it seems a safe enough check.

Related to #6748

### DIFF
--- a/internal/settings/settings.go
+++ b/internal/settings/settings.go
@@ -617,7 +617,7 @@ func confirmRequiredEnvVarsForDefaultConfigs(paths []string) error {
 			requiredEnvVars := []string{RealmEnvVar, TokenEnvVar}
 			for _, v := range requiredEnvVars {
 				if len(os.Getenv(v)) == 0 {
-					logWarn("Usage: %s=12345 %s=us0 %s", TokenEnvVar, RealmEnvVar, os.Args[0])
+					logError("Usage: %s=12345 %s=us0 %s", TokenEnvVar, RealmEnvVar, os.Args[0])
 					return fmt.Errorf("ERROR: Missing required environment variable %s with default config path %s", v, path)
 				}
 			}
@@ -714,6 +714,13 @@ func logInfo(format string, v ...any) {
 func logWarn(format string, v ...any) {
 	if isSplunkHomeDefined() {
 		format = "WARN " + format
+	}
+	log.Printf(format, v...)
+}
+
+func logError(format string, v ...any) {
+	if isSplunkHomeDefined() {
+		format = "ERROR " + format
 	}
 	log.Printf(format, v...)
 }


.chloggen/fix-internal-settings-log-level-for-otel-ta.yaml
crobert-1, line 1: The title of this PR contains `[chore]`, does it need a changelog?

.chloggen/fix-internal-settings-log-level-for-otel-ta.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'ent
```
crobert-1:
```suggestion
change_type: 'enhancement'
```

internal/settings/settings.go
```diff
@@ -616,7 +617,7 @@ func confirmRequiredEnvVarsForDefaultConfigs(paths []string) error {
 			requiredEnvVars := []string{RealmEnvVar, TokenEnvVar}
 			for _, v := range requiredEnvVars {
 				if len(os.Getenv(v)) == 0 {
-					log.Printf("Usage: %s=12345 %s=us0 %s", TokenEnvVar, RealmEnvVar, os.Args[0])
+					logWarn("Usage: %s=12345 %s=us0 %s", TokenEnvVar, RealmEnvVar, os.Args[0])
```
crobert-1: Should we log this as an error instead of `WARN`?

.chloggen/fix-internal-settings-log-level-for-otel-ta.yaml
pjanotti: I don't think so because in principle no one will be seeing this new behavior, the changelog itself will be when we release the TA with a collector with this change.

crobert-1: Is there any way to ensure this changelog will only appear in the TA release changelog? Or do we just try to remember to remove it from changelog notes on other releases?

pjanotti: I'm getting things confused, sorry. This one I will remove the `[chore]` and mark this as enhancement on the changelog for the collector. On the change for the TA, there won't be a chloggen file since we are doing those manually at the moment. Thanks for noticing this.

.chloggen/fix-internal-settings-log-level-for-otel-ta.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
```
crobert-1, line 2:
```suggestion
change_type: bug_fix
```
Do you think `bug_fix` would be more accurate?

crobert-1 (approved):

.chloggen/fix-internal-settings-log-level-for-otel-ta.yaml
pjanotti: I think it is not a bug per se and users won't ever see this on the TA since when we release the TA logging to splunkd.log the collector will be already fixed.

PR #6777, opened 2025-09-17
[chore][config server test] Address flakiness

crobert-1:

This change is to ensure the test is not impacted by previous tests or changes on the host machine by waiting until the config server isn't responding. This should help narrow down potential causes of flakiness.

### DIFF
--- a/internal/configconverter/config_server_test.go
+++ b/internal/configconverter/config_server_test.go
@@ -41,7 +41,7 @@ func TestConfigServer_RequireEnvVar(t *testing.T) {
 	require.EventuallyWithT(t, func(c *assert.CollectT) {
 		assert.True(c, isPortAvailable(defaultConfigServerPort))
 		_, err := client.Get(url)
-		assert.Error(t, err)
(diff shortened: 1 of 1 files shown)


internal/configconverter/config_server_test.go
```diff
@@ -34,6 +35,15 @@ import (
 
 func TestConfigServer_RequireEnvVar(t *testing.T) {
 	waitForRequiredPort(t, defaultConfigServerPort)
+	url := fmt.Sprintf("http://%s/debug/configz/initial", defaultConfigServerEndpoint)
+	client := &http.Client{}
+
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.True(c, isPortAvailable(defaultConfigServerPort))
```
crobert-1, line 42: This is probably unnecessary, as the check at the beginning of the method, `waitForRequiredPort` is a `required.EventuallyWithT` call to this method waiting for it to be available. With the flakiness we've been seeing though, I assume it can't hurt.

PR #6778, opened 2025-09-17
Add godebug installer

atoulme:


packaging/installer/install.sh
```diff
@@ -1562,6 +1569,7 @@ parse_args_and_install() {
   echo "Ingest Endpoint: $ingest_url"
   echo "API Endpoint: $api_url"
   echo "HEC Endpoint: $hec_url"
+  echo "GODEBUG: godebug"
```
crobert-1:
```suggestion
  echo "GODEBUG: $godebug"
```

crobert-1 (commented):
Is there any way to add tests for this?

.chloggen/add_godebug_installer.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: installer
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: expose setting GODEBUG as part of bash and powershell install scripts
```
pjanotti:
```suggestion
note: Expose setting `GODEBUG` as an option of the install scripts.
```

atoulme:
> Is there any way to add tests for this?

yes likely. I think we have tests that take the installer script and run it and check the outcome. I can follow up I guess.

packaging/installer/install.ps1
```diff
@@ -655,6 +660,7 @@ $collector_env_vars = @{
     "SPLUNK_BUNDLE_DIR"       = "$bundle_dir";
     "SPLUNK_CONFIG"           = "$config_path";
     "SPLUNK_HEC_TOKEN"        = "$hec_token";
+    "GODEBUG"                 = "$godebug";
```
pjanotti: don't add it here by default: do like `$network_interface` below, check if it is not empty then add it to this collection.

PR #6820, opened 2025-10-02
[chore] Make msi-build GH Job more reliable

pjanotti:
Make GH `msi-build` job more reliable by removing the usage of a Linux Docker image with Wine. Now, both GitHub and GitLab CI build the MSI on Windows using version 3.14 RTM of the Wix Toolset.

The call to the script building the MSI was removed from the Makefile to avoid needing to setup go just to target any of the Makefile targets - I'll add a backlog item so we can review its usage and optimize it.

This change was validated on GL via a pipeline run, see [here](https://cd.splunkdev.com/o11y-gdi/splunk-otel-collector-releaser/-/pipelines/30316293) some of the changes in this branch are not in this PR because they were made only to facilitate manual run of the GL pipeline.


.github/workflows/reusable-msi-build.yml
```diff
@@ -12,23 +12,23 @@ env:
 
 jobs:
   msi-build:
-    runs-on: ubuntu-24.04
+    runs-on: ${{ inputs.OS }}
```
pjanotti, line 15: In principle this could be a single OS but since the callers of this reusable workflow are already using different OS versions I used it as a way to run on different hosts. Anyway, this seems something that we could potentially remove, but, that should be a separate PR since it touches other workflow fils.

packaging/msi/msi-builder/build-launcher.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
pjanotti, line 1: This script is a rename of `packaging/msi/msi-builder/docker-entrypoint.sh` and changing the Docker based files with the addition of the parameters `REPO_DIR` and `WORK_DIR`.

packaging/msi/msi-builder/build.sh
```diff
@@ -154,7 +150,7 @@ parse_args_and_build() {
     files_dir="${build_dir}/msi"
     msi_name="splunk-otel-collector-${version}-amd64.msi"
 
-    if [ -z "$skip_build_dir_removal" && -d "$build_dir" ]; then
+    if [ -z "$skip_build_dir_removal" ] && [ -d "$build_dir" ]; then
```
pjanotti, line 153: Shellcheck suggestion

packaging/msi/msi-builder/build.sh
```diff
@@ -184,16 +180,16 @@ parse_args_and_build() {
     cd ${WORK_DIR}
 
     configFilesWsx="${build_dir}/configfiles.wsx"
-    heat dir "$files_dir" -srd -sreg -gg -template fragment -cg ConfigFiles -dr INSTALLDIR -out "${configFilesWsx//\//\\}"
```
pjanotti, line 187: This was something need on Wine, not needed anymore when running natively on Windows Git bash (it understands `/` as path separator)

.gitlab-ci.yml
```diff
@@ -889,16 +889,22 @@ build-msi:
     - msi-custom-actions-package
   before_script:
     # build the MSI with the signed exe
-    - mkdir -p bin
-    - cp -f dist/signed/otelcol_windows_amd64.exe bin/otelcol_windows_amd64.exe
-  id_tokens:  # http://go/gitlab-17
-    CI_JOB_JWT:
-      aud: $CICD_VAULT_ADDR
+    - New-Item -Path bin -ItemType Directory -Force
```
pjanotti, line 892: Switched everything to Powershell, the default shell on our Windows runners.

packaging/msi/msi-builder/build.sh
```diff
@@ -154,7 +150,7 @@ parse_args_and_build() {
     files_dir="${build_dir}/msi"
     msi_name="splunk-otel-collector-${version}-amd64.msi"
 
-    if [ -z "$skip_build_dir_removal" && -d "$build_dir" ]; then
+    if [ -z "$skip_build_dir_removal" ] && [ -d "$build_dir" ]; then
         rm -rf "$build_dir"
     else
         echo "Skipping build directory removal"
```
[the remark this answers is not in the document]
pjanotti: The else branch can be hit if the directory doesn't exist...

.github/workflows/reusable-msi-build.yml
```diff
@@ -12,23 +12,23 @@ env:
...
     steps:
       - name: Check out the codebase.
         uses: actions/checkout@v5
 
+      - name: Uninstall default WiX
+        run: choco uninstall wixtoolset
+
+      - name: Install WiX 3.14.0
+        run: choco install wixtoolset --version 3.14.0 --allow-downgrade --force
```
jinja2, line 24: Is it ok to use an old version? Iirc, you were discussing moving to wix 6 recently

jinja2 (approved):

.github/workflows/reusable-msi-build.yml
pjanotti: Yes, we need to be on 3.14, given syntax and tool changes upgrading to the latest is a separate work item. Notice that we can't use 3.14.1 (the latest 3.14 released): it has a bug that breaks the installation when it requires elevation, however, 3.14.0 has a security issue so we must enforce that `RemoveFolderEx` is not used. Both conditions are enforced on the script.

crobert-1 (approved):
Just took a quick look at the reusable GitHub job, looks good to me. 👍

PR #6823, opened 2025-10-03
[chore] fix update-nodejs-agent step

jinja2:

fix update-nodejs-agent step in gitlab


.gitlab-ci.yml
```diff
@@ -327,6 +327,9 @@ update-nodejs-agent:
   stage: update-deps
   needs: []
   dependencies: []
+  before_script:
+    - !reference [.go-cache, before_script]
+    - apt-get update && apt-get install -y jq
```
crobert-1:
```suggestion
    - apt update && apt install -y jq
```
Should we be using `apt` instead of `apt-get`? (Not sure if it makes much of a difference here, but we're using `apt` in the workflows already, not `apt-get`)

PR #6965, opened 2025-11-23
[Do not merge] Example showing how to sample profiles

atoulme:
This example is showing how to deploy the collector against a Java app instrumented by the Java agent, sending profiles to the collector. The example concentrates on the ability of the collector to sample profiles by using a random uuid associated with each log representing a profile.


crobert-1:
@atoulme Is there an end goal with this PR? Is there something to do here, or can we close?

PR #7263, opened 2026-02-25
[chore] push docker images to docker hub

atoulme:
Add logic to push docker images to both docker hub and quay.


.gitlab-ci.yml
```diff
@@ -1048,7 +1204,7 @@ build-push-linux-image:
           echo "Creating and pushing ${IMAGE_NAME}:${latest_tag} manifest"
           docker buildx imagetools create --tag ${IMAGE_NAME}:${latest_tag} ${IMAGE_NAME}:${arch_tag}
         fi
-        echo "${IMAGE_NAME}:${arch_tag}" > tags_to_sign_${arch}
+        echo "${IMAGE_NAME}:${arch_tag}" >> tags_to_sign_${arch}
```
pjanotti, line 881: Any specific reason for this change or just good form?

atoulme: since we now have 2 jobs in parallel, I got spooked that file can be overridden. I think the name of the file might even need to be unique per parallel run

.gitlab-ci.yml
```diff
@@ -1010,20 +1011,24 @@ build-push-linux-image:
   id_tokens:  # http://go/gitlab-17
     CI_JOB_JWT:
       aud: $CICD_VAULT_ADDR
+  parallel:
+    matrix:
+      - BASE_TAG: [ "quay.io/signalfx/splunk-otel-collector", "splunk/otel-collector" ]
   script:
     - *docker-reader-role
     - docker login -u $CIRCLECI_QUAY_USERNAME -p $CIRCLECI_QUAY_PASSWORD quay.io
+    - - docker login -u splunk -p $DOCKER_HUB_TOKEN
```
jinja2, line 845:
```suggestion
    - docker login -u splunk -p $DOCKER_HUB_TOKEN
```

also, should the username be hidden? or is the org name/repo namespace also username for dockerhub?

.gitlab-ci.yml
pjanotti: The file is uploaded as an artifact so you need to add `${BASE_TAG}` to the file name and ensure that wherever it is used it is properly finding the expected files. I mean it already handles `${arch}` you need to ensure that it also handles the addition of `${BASE_TAG}` to the name.

.gitlab-ci.yml
atoulme: we can hide it, I'll do that

PR #7412, opened 2026-04-08
OTL-4045 Update collector configs to use observability.splunkcloud.com URLs

akhileshsingh85:
Replace signalfx.com endpoints with observability.splunkcloud.com across:
- Collector YAML configs (cmd/otelcol/config and fips/config)
- internal/settings: default URL construction from SPLUNK_REALM
- internal/configconverter testdata, confmapprovider tests
- Technical addon YAML configs and general config tests

**Description:** <Describe what has changed.>

**Link to Splunk idea:** <Link to Splunk idea, see https://ideas.splunk.com>

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

PR #7413, opened 2026-04-08
[OTL-4045] installer update to use observability.splunkcloud.com urls

akhileshsingh85:
Description:
Migrated all default API and ingest endpoint URLs from signalfx.com to .observability.splunkcloud.com across the repository. The
following URL patterns were updated:

    api.{realm}.signalfx.com → api.{realm}.observability.splunkcloud.com
    ingest.{realm}.signalfx.com → ingest.{realm}.observability.splunkcloud.com
    Runtime defaults (internal/settings/settings.go) — API, ingest, trace, and HEC URL construction from realm now
    targets .observability.splunkcloud.com
    Installer scripts (install.sh, install.ps1, chocolateyinstall.ps1) — updated default URL templates and help text
    Deployment configs — Ansible, Chef, Puppet, Salt, Nomad default URL parameters and templates
    Collector configs — agent_config.yaml, gateway_config.yaml, fargate_config.yaml, ecs_ec2_config.yaml, etc. (both standard and FIPS variants)
    Tests and test fixtures — Go, Python, shell test expected values updated accordingly
    Documentation — All README examples updated

Testing: Existing function test should work fine with new endpoint.


packaging/choco/splunk-otel-collector/tools/chocolateyinstall.ps1
```diff
@@ -77,10 +77,20 @@ if ($access_token) {
 set_env_var_value_from_package_params $env_vars $pp "SPLUNK_REALM" "us0"
 $realm = $env_vars["SPLUNK_REALM"] # Cache the realm since it is used to build various default values.
 
-set_env_var_value_from_package_params $env_vars $pp "SPLUNK_INGEST_URL"         "https://ingest.$realm.signalfx.com"
-set_env_var_value_from_package_params $env_vars $pp "SPLUNK_API_URL"            "https://api.$realm.signalfx.com"
+# Migrate old signalfx.com URLs to observability.splunkcloud.com
+foreach ($url_var in @("SPLUNK_INGEST_URL", "SPLUNK_API_URL", "SPLUNK_HEC_URL")) {
```
akhileshsingh85: re-evaluating the need for this change. It was added to handle the upgrade test. When first install was done with older version. The old version sets the env with older domain. During the upgrade, that environment variable need to be updated to new domain. This logic handles that case.
Scenario to validate: Will this effect if user explicitly sets those environment?

.chloggen/otl-4045-replace-signalfx-urls.yaml
```diff
@@ -0,0 +1,47 @@
...
+  Windows (`install.ps1`):
+  ```
+  -api_url https://api.<realm>.signalfx.com
+  -ingest_url https://ingest.<realm>.signalfx.com
+  -hec_url https://ingest.<realm>.signalfx.com/v1/log
+  ```
+
+  If you already set `SPLUNK_API_URL`, `SPLUNK_INGEST_URL`, or `SPLUNK_HEC_URL` explicitly in your
+  deployment, no change is required — those values override the installer defaults. The legacy
+  `*.signalfx.com` endpoints remain available during the transition period.
```
dmitryax: We shouldn't put such big text in the changelog. Let's put it in [the upgrade guidelines](https://github.com/signalfx/splunk-otel-collector?tab=readme-ov-file#upgrade-guidelines) and reference here

packaging/choco/splunk-otel-collector/tools/chocolateyinstall.ps1
dmitryax: I don't think we should replace the values for users automatically. We should add deprecation messages instead, but that can be done in a separate PR

packaging/technical-addon/Splunk_TA_otel/default/inputs.conf
```diff
@@ -0,0 +1,19 @@
...
+index = _internal
+sourcetype = Splunk_TA_otel
+splunk_access_token_file=$SPLUNK_OTEL_TA_HOME/local/access_token
+splunk_api_url=https://api.us0.observability.splunkcloud.com
+splunk_bundle_dir=$SPLUNK_OTEL_TA_PLATFORM_HOME/bin/agent-bundle
+splunk_collectd_dir=$SPLUNK_OTEL_TA_PLATFORM_HOME/bin/agent-bundle/run/collectd
+splunk_ingest_url=https://ingest.us0.observability.splunkcloud.com
+splunk_listen_interface=localhost
+splunk_realm=us0
+splunk_config=$SPLUNK_OTEL_TA_HOME/configs/ta-agent-config.yaml
```
dmitryax: Why new files have to be added?

.github/workflows/win-package-test.yml
```diff
@@ -320,6 +320,16 @@ jobs:
...
             $env_vars = Get-ItemPropertyValue -Path "HKLM:\SYSTEM\CurrentControlSet\Services\splunk-otel-collector" -Name "Environment"
+            # Read the URLs written by the previous installation. The new installer preserves
+            # existing registry values without migration, so whatever the old version set must
+            # survive the upgrade unchanged — regardless of which domain it defaulted to.
+            $installed_env = @{}
+            foreach ($entry in $env_vars) {
+              $k, $v = $entry.Split("=", 2)
+              $installed_env[$k] = $v
+            }
+            $expected_api_url = $installed_env["SPLUNK_API_URL"]
```
akhileshsingh85: This is done to make sure that the already existing ENV, due to previous install, do not get changed due to the upgrade. 
So collect the current value and pass it on to win-test-service.ps1 script to validate the old value.

.github/workflows/win-package-test.yml
```diff
@@ -320,6 +323,16 @@ jobs:
...
+            $expected_ingest_url = $installed_env["SPLUNK_INGEST_URL"]
```
dmitryax: Do we have an upgrade test case that actually executes this path?

README.md
```diff
@@ -161,6 +161,99 @@ manually before the backward compatibility is dropped. For every configuration u
...
+- `https://ingest.<realm>.signalfx.com/v1/log` → `https://ingest.<realm>.observability.splunkcloud.com/v1/log`
+
+**Migration Notes — Proxy and Firewall Configuration:**
+If you have firewall rules or proxy allowlists scoped to `*.signalfx.com`, update them to also allow
+`*.observability.splunkcloud.com`, or the specific realm-scoped hostnames:
+- `api.<realm>.observability.splunkcloud.com`
+- `ingest.<realm>.observability.splunkcloud.com`
+
+If you already set `SPLUNK_API_URL`, `SPLUNK_INGEST_URL`, `SPLUNK_TRACE_URL`, or `SPLUNK_HEC_URL`
+explicitly in your deployment, no change is required — explicit values always override the defaults.
```
dmitryax: We should mention that there are no immediate changes required but the user should start updating those env vars

README.md
```diff
@@ -161,6 +161,99 @@ manually before the backward compatibility is dropped. For every configuration u
...
+-ingest_url https://ingest.<realm>.signalfx.com
+-hec_url https://ingest.<realm>.signalfx.com/v1/log
+```
+
+Chocolatey:
+```
+choco install splunk-otel-collector --params "'/SPLUNK_API_URL:https://api.<realm>.signalfx.com /SPLUNK_INGEST_URL:https://ingest.<realm>.signalfx.com /SPLUNK_HEC_URL:https://ingest.<realm>.signalfx.com/v1/log'"
+```
+
+#### Deployment tooling
```
dmitryax: Let's move those to corresponding changelogs

.github/workflows/win-package-test.yml
akhileshsingh85: Yes. Choco-test-> scenario upgrade will execute this path.

dmitryax: Ok but that test doesn't set `SPLUNK_API_URL` and other envs. How does it trigger this path?

pkg/modularinput/testdata/Splunk_TA_otel.inputs.xml
```diff
@@ -12,14 +12,14 @@
       <param name="interval">60</param>
       <param name="sourcetype">Splunk_TA_otel</param>
       <param name="splunk_access_token_file">$SPLUNK_OTEL_TA_HOME/local/access_token</param>
-      <param name="splunk_api_url">https://api.us0.signalfx.com</param>
+      <param name="splunk_api_url">https://api.us0.observability.splunkcloud.com</param>
```
jinja2, line 15: While this PR is only updating the test fixtures, I am wondering if we need to do anything for TA v2.

@pjanotti do we need to update `packaging/ta-v2/CHANGELOG.md`? I am not sure if we have made any v2 releases yet. 

My understanding is that TA v2 inherits the signalfx.com -> observability.splunkcloud.com change from the collector binary (PR #7412 which changed settings.go). Next TA build that picks this up will resolve to new endpoints. There are no explicit inputs for urls in the [default input](https://github.com/signalfx/splunk-otel-collector/blob/main/packaging/ta-v2/assets/default/inputs.conf) so if a user wants to set overrides, they'd have to set it via `splunk_collector_env_vars`. Should we make a note in the TA changelog about this?

.chloggen/otl-4045-replace-signalfx-urls.yaml
```diff
@@ -0,0 +1,19 @@
...
+# One or more tracking issues related to the change
+issues: [7413]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  For full migration details, proxy/firewall configuration notes, and per-tool instructions for
+  keeping legacy `*.signalfx.com` endpoints, see the
+  [0.149.0 → 0.150.0 upgrade guideline](../README.md#from-01500-to-01510) in README.md.
```
jinja2:
```suggestion
  [0.149.0 → 0.150.0 upgrade guideline](../README.md#from-01490-to-01500) in README.md.
```
I assume this is the correct anchor in README

pkg/modularinput/testdata/Splunk_TA_otel.inputs.xml
pjanotti: Good Qs.:
> do we need to update `packaging/ta-v2/CHANGELOG.md`?

No, since `packaging/ta-v2/CHANGELOG.md` was created things changed a bit and the TA v2 just uses the same changelog as the collector. I will remove this file to avoid future confusion.

A bit more detail: as you noted the TA v2 uses the same configs and code settings as the collector, so no specific TA v2 changes are needed. The [default input](https://gihub.com/signalfx/splunk-otel-collector/blob/v0.149.0/packaging/ta-v2/assets/default/inputs.conf) and the [input spec](https://github.com/signalfx/splunk-otel-collector/blob/v0.149.0/packaging/ta-v2/assets/README/inputs.conf.spec) are generated from the [collector TA scheme](https://github.com/signalfx/splunk-otel-collector/blob/v0.149.0/cmd/otelcol/ta_scheme.xml) and it doesn't carry any URL - once more they come from defaults in configs and code of the collector.

The URL in the test file is immaterial to that, it processes generic XML inputs for modular inputs, it was just a convenience that I used the same as the actual TA v2 input XML.

dmitryax (approved):

PR #7479, opened 2026-04-22
Add internal metrics to keep track of used splunk custom config source.

dloucasfx:
This PR adds new internal metric `otelcol_splunk_configsource_usage` to track the usage of the splunk custom config source .
The decision is to emit the metric for enabled splunk config source only.

In order to inject metrics in the telemetry meter provider, we have to use a custom extension as the config source has no access to `TelemetrySettings` 
This extension is loaded by default through a config map converter.


internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,197 @@
...
+
+	if configSourcesMap, ok := retrieved["config_sources"].(map[string]any); ok {
+		t.logger().Debug("TelemetryHook: config_sources found in retrieved config", zap.String("scheme", scheme))
+
+		for csType := range configSourcesMap {
+			if isCustomConfigSource(csType) {
+				t.logger().Info("TelemetryHook: custom config source detected",
+					zap.String("type", csType),
+					zap.String("scheme", scheme))
+				t.usedSources[csType] = true
```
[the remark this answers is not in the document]
dloucasfx: Not an issue as we go through this once

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,197 @@
...
+	return nil
+}
+
+// SetTelemetrySettings injects the service component.TelemetrySettings into the hook.
+// Called by the configsourcetelemetryextension on Start().
+func (t *TelemetryHook) SetTelemetrySettings(settings component.TelemetrySettings) {
+	t.mutex.Lock()
+	t.telemetrySettings = &settings
+	t.mutex.Unlock()
+
```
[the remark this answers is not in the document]
dloucasfx: Not an issue

internal/configconverter/configsourcetelemetry.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	extensions, _ := out["extensions"].(map[string]any)
+	if extensions == nil {
+		extensions = map[string]any{}
+	}
+	if _, exists := extensions[configsourcetelemetryextension.TypeStr]; !exists {
+		extensions[configsourcetelemetryextension.TypeStr] = map[string]any{}
+	}
+	out["extensions"] = extensions
```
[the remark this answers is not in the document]
dloucasfx: I will fix it

cmd/otelcol/main.go
```diff
@@ -99,14 +99,16 @@ func runFromCmdLine(args []string) {
...
 
+	telemetryHook := configsource.NewTelemetryHook()
 	confMapConverterFactories := collectorSettings.ConfMapConverterFactories()
 	dryRun := configconverter.NewDryRun(collectorSettings.IsDryRun(), confMapConverterFactories)
 	expvarConverter := configconverter.GetExpvarConverter()
 	confMapConverterFactories = append(confMapConverterFactories,
 		configconverter.ConverterFactoryFromConverter(dryRun),
-		configconverter.ConverterFactoryFromConverter(expvarConverter))
+		configconverter.ConverterFactoryFromConverter(expvarConverter),
+		configconverter.ConverterFactoryFromFunc(configconverter.InjectConfigSourceTelemetryExtension))
```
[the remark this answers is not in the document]
dloucasfx: This is expected behavior, as a matter of fact, we have test that make sure that dry run does not run

internal/configconverter/configsourcetelemetry.go
```diff
@@ -0,0 +1,92 @@
...
+		result := make([]string, 0, len(val))
+		for _, item := range val {
+			if s, ok := item.(string); ok {
+				result = append(result, s)
+			}
+		}
+		return result
+	}
+	return []string{}
+}
```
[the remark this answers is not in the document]
dloucasfx: I opted to use an existing function from discovery

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,196 @@
...
+		t.logger().Debug("TelemetryHook: config_sources found in retrieved config", zap.String("scheme", scheme))
+
+		for csType := range configSourcesMap {
+			if isCustomConfigSource(csType) {
+				t.logger().Info("TelemetryHook: custom config source detected",
+					zap.String("type", csType),
+					zap.String("scheme", scheme))
+				t.usedSources[csType] = true
+			}
+		}
```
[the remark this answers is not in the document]
dloucasfx: Fixed

.chloggen/configsourcetelemetry.yaml
crobert-1, line 1: All plaintext references to Splunk should capitalize it.

internal/extension/configsourcetelemetryextension/extension.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/extension"
+
+	"github.com/signalfx/splunk-otel-collector/internal/confmapprovider/configsource"
+)
+
+const (
+	// TypeStr is the type string for this extension
+	TypeStr = "configsource_telemetry"
```
crobert-1: Config source seems like two words, is there any reason why the component name should keep it as one? Upstream is moving to use underscores as much as possible rather than merging words, so that's my preference (unless there's a good reason not to).

internal/configconverter/configsourcetelemetry.go
crobert-1, line 1: General comment on go code being added in this PR: There are a lot of comments that are unnecessary here. I'd ask that we go through and remove all generic/boilerplate comments and only keep ones that are legitimately useful.

Not useful examples:
```
// SetGlobalHook sets the global hook instance.
```
```
// Validate implements component.Config
```

internal/extension/configsourcetelemetryextension/extension_test.go
crobert-1, line 1: For all golang test code: As much as possible, golang recommends using `t.Context` in tests rather than `context.Background`. ([Upstream actually enables a linter that enforces this](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/125706251e9cc85666d61136a18b07ef8e8a58b5/.golangci.yml#L470), allowing exceptions of course)

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,203 @@
...
+var (
+	_ Hook = (*TelemetryHook)(nil)
+
+	// globalHook instance needed by the extension
+	globalHook *TelemetryHook
+)
+
+const (
+	meterName                    = "github.com/signalfx/splunk-otel-collector/internal/confmapprovider/configsource"
+	configSourceUsageMetricName  = "otelcol_splunk_configsource_usage"
```
crobert-1: I already brought this up elsewhere, but should `configsource` be `config_source`?

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,203 @@
...
+
+// TelemetryHook tracks the usage of custom config sources and exposes them as OpenTelemetry metrics.
+type TelemetryHook struct {
+	usedSources       map[string]bool
+	telemetrySettings *component.TelemetrySettings
+	mutex             sync.RWMutex
+}
+
+// logger returns the logger from TelemetrySettings once injected, or a no-op logger before that.
+func (t *TelemetryHook) logger() *zap.Logger {
```
crobert-1, line 49: Tiny nit, feel free to drop: Anytime I see `t` in golang I assume it's testing. It might be helpful to rename the receiver argument to `tel`

internal/configconverter/configsourcetelemetry.go
```diff
@@ -0,0 +1,58 @@
...
+
+	out := in.ToStringMap()
+
+	service, serviceExtensions, err := getServiceExtensions(out)
+	if err != nil {
+		return err
+	}
+
+	extensions := map[string]any{}
+	if raw, exists := out["extensions"]; exists && raw != nil {
```
crobert-1, line 41: I'd expect any plain strings that are being re-used to be defined as constants, preferably in some central helper file. (This is relevant for `extensions` and `service`)

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,203 @@
...
+		return nil
+	}
+
+	meter := t.telemetrySettings.MeterProvider.Meter(meterName)
+
+	// Register an observable gauge that reports which config sources are in use
+	var err error
+	_, err = meter.Int64ObservableGauge(
+		configSourceUsageMetricName,
+		metric.WithDescription("Indicates whether a custom config source is in use (1 = in use)"),
```
crobert-1, line 87: Maybe also specify that 0 is not in use?

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,203 @@
...
+// OnShutdown implements Hook and clears tracked config sources.
+func (t *TelemetryHook) OnShutdown() {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+	t.usedSources = make(map[string]bool)
+}
+
+// observeConfigSourceUsage is the callback function for the observable gauge.
+// Only config sources that are actively in use are reported (value = 1).
+func (t *TelemetryHook) observeConfigSourceUsage(_ context.Context, observer metric.Int64Observer) error {
```
crobert-1, line 149: From what I can tell this method never returns an error, can we drop the return type here?

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,203 @@
...
+
+// isCustomConfigSource reports whether csType is one of the custom config sources
+// provided by this distribution.
+func isCustomConfigSource(csType string) bool {
+	_, ok := customConfigSources[csType]
+	return ok
+}
+
+// GetUsedSources returns a copy of the currently used config sources; used for unit test.
+func (t *TelemetryHook) GetUsedSources() map[string]bool {
```
crobert-1, line 180: What's the benefit of having this as its own method for testing? Also, if it's just for testing, why isn't it in the testing file itself? (Same questions for `GetTelemetrySettings`)

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,203 @@
...
+			observer.Observe(1, metric.WithAttributes(
+				attribute.String(configSourceTypeAttributeKey, csType),
+			))
+		}
+	}
+
+	return nil
+}
+
+// customConfigSources is the fixed set of config source types provided by this distribution.
```
crobert-1: I'd like to have a test asserting the size of this map (key count), that also requires the count to match how many config sources there are. I'm not sure how programmatically count config sources, but if there's some not-entirely hack way of doing it, I think it would be really useful. This is the kind of thing we'd easily miss updating, I want a test that clearly notifies us when this needs to be updated.

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,203 @@
...
+var customConfigSources = map[string]struct{}{
+	"env":       {},
```
crobert-1, line 165: Nit: we should alphabetize the list

internal/extension/configsourcetelemetryextension/README.md
crobert-1, line 1: Any reason this component doesn't have a `metadata.yaml` file? It would be useful to conform to upstream requirements, as well as auto-generate the status table. We could then also declare the telemetry metric there, giving us built in automated testing (I assume, I haven't checked how this works for extensions in particular or the functionality we're using).

internal/extension/configsourcetelemetryextension/README.md
```diff
@@ -0,0 +1,71 @@
+# Config Source Telemetry Extension
+
+| Status                   |                                     |
+|--------------------------|-------------------------------------|
+| Stability                | [alpha]                             |
+| Supported pipeline types | N/A (extension)                     |
```
crobert-1:
```suggestion
```
Don't need this line, it's not generated for upstream extensions either.

internal/extension/configsourcetelemetryextension/README.md
```diff
@@ -0,0 +1,71 @@
...
+| Distributions            | [Splunk OpenTelemetry Collector]    |
+
+This extension bridges the gap between the Splunk collector's custom config source
+provider and the OpenTelemetry service telemetry infrastructure. It is injected
+automatically into the collector's service configuration at startup — no user
+configuration is required.
+
+## Purpose
```
crobert-1, line 13: Nit: I think this gets too into the weeds, I don't think the README needs to explain the inner workings of injection and hooks. This section is "Purpose", so we should explain why we need this, and what's being done (a telemetry metric is being generated and forwarded to our backend).

Feel free to drop, I could be wrong here.

internal/extension/configsourcetelemetryextension/README.md
```diff
@@ -0,0 +1,71 @@
...
+The Splunk OpenTelemetry Collector supports custom config sources (e.g. `vault`,
+`include`, `env`, `zookeeper`, `etcd2`) that are initialised before the collector
+service — and therefore before the service's `MeterProvider` — is available.
+This extension is started by the service with a fully initialised
+`component.TelemetrySettings`. On `Start()` it injects those settings into the
+global `TelemetryHook`, enabling the hook to register the
+`otelcol_splunk_configsource_usage` observable gauge against the **service's own**
+`MeterProvider`. The metric then appears at the collector's internal Prometheus
+endpoint (`/metrics`) and is automatically scraped and forwarded to your observability
+backend by the existing `prometheus/internal` receiver pipeline.
```
crobert-1: This seems to assume all users have the `prometheus/internal` receiver setup the way our default configs have it. Should we make this description more generic so it's clear that it not scoped to **just** that pipeline?

internal/extension/configsourcetelemetryextension/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+| Metric name                         | Type  | Attributes           | Description |
+|-------------------------------------|-------|----------------------|-------------|
+| `otelcol_splunk_configsource_usage` | Gauge | `config_source_type` | Emitted with value `1` for each custom config source that is present in the collector config. No datapoint is emitted for config sources that are not in use. |
+
+### Attribute values for `config_source_type`
+
+| Value       | Config source           |
+|-------------|-------------------------|
+| `env`       | `envvarconfigsource`    |
```
crobert-1, line 53: nit: alphabetize

internal/extension/configsourcetelemetryextension/README.md
```diff
@@ -0,0 +1,71 @@
...
+| Distributions            | [Splunk OpenTelemetry Collector]    |
+
+This extension bridges the gap between the Splunk collector's custom config source
+provider and the OpenTelemetry service telemetry infrastructure. It is injected
+automatically into the collector's service configuration at startup — no user
+configuration is required.
+
+## Purpose
+
+The Splunk OpenTelemetry Collector supports custom config sources (e.g. `vault`,
```
crobert-1: nit: alphabetize the list of config sources

crobert-1 (commented):
Lots of nits, I also may have missed the point in a few places, please feel free to push back and clarify where I did.

internal/extension/configsourcetelemetryextension/extension.go
dloucasfx: No specific reason, I will change it to `config_source_telemetry`

internal/confmapprovider/configsource/telemetry.go
dloucasfx: I think it's given that 1 is the only value

internal/extension/configsourcetelemetryextension/extension.go
dloucasfx: Done

internal/confmapprovider/configsource/telemetry.go
dloucasfx: the func signature requires error `type Int64Callback func(context.Context, Int64Observer) error`

internal/configconverter/configsourcetelemetry.go
dloucasfx: Done

internal/extension/configsourcetelemetryextension/extension_test.go
dloucasfx: I guess we need to clean up lots of code in our repo :) 
I updated my newly added tests

internal/confmapprovider/configsource/telemetry.go
dloucasfx: I brought it up in standup and folks agreed on the name.
But now that I have changed the extension, I will update the metric name as well

internal/confmapprovider/configsource/telemetry.go
dloucasfx: tbh, I don't think it's an issue

internal/confmapprovider/configsource/telemetry.go
dloucasfx: Done

internal/confmapprovider/configsource/telemetry.go
dloucasfx: It's mainly encapsulation, usedSources and telemetrySettings are unexported fields and we need to expose them when test is running from a different package, example GetTelemetrySettings which is used in the extension test.

But in general, it's cleaner and safer

internal/confmapprovider/configsource/telemetry.go
dloucasfx: That's a very good point, I thought of it but then I figured it's not worth the effort as I doubt we will add new config sources.
We might be able to do it using `configSourceFactories` , but need to check how practical it is.

dloucasfx: yes, I was able to build the list from `configSourceFactories` , however, I think the static list is more robust and I added a test which is using `configSourceFactories` to build the list and compare to the static one.
See code update.
Thanks

internal/confmapprovider/configsource/telemetry.go
dloucasfx: Nit :) 
Done

internal/extension/configsourcetelemetryextension/README.md
dloucasfx: The plan is not to expose it to user, so figured I can omit this file.
Let me know if you still think otherwise

internal/extension/configsourcetelemetryextension/README.md
dloucasfx: Done

internal/extension/configsourcetelemetryextension/README.md
dloucasfx: It just describes the whole purpose of the extension, which is to making `MeterProvider` for telemetry available.
This is not an end user extension, it's a hack to inject our telemetry

internal/extension/configsourcetelemetryextension/README.md
dloucasfx: I see what you mean, updated the note. let me know what you think

internal/extension/configsourcetelemetryextension/README.md
dloucasfx: Done

internal/extension/configsourcetelemetryextension/README.md
dloucasfx: Done

internal/configconverter/configsourcetelemetry.go
dloucasfx: That's a fair point and I would have done it in different cases (only for `extensions` , `service` is used once), however, the reason you don't have to define them as constant is because those values are not defined or owned by us, they are part OTEL confmap schema . Define them as constant will make them look like they are owned by us and they might change/drift at one point.
Also not, `getServiceExtensions` which I am reusing, already encapsulates and uses the "service" and "extensions" and they are not constants

internal/confmapprovider/configsource/telemetry.go
crobert-1: Ah I see that now, thanks for pointing that out 👍

internal/extension/configsourcetelemetryextension/README.md
crobert-1: Update looks good, thanks!

internal/confmapprovider/configsource/telemetry.go
```diff
@@ -0,0 +1,196 @@
...
+
+	result := make(map[string]bool)
+	for k, v := range t.usedSources {
+		result[k] = v
+	}
+	return result
+}
+
+// GetTelemetrySettings returns the current TelemetrySettings; used for unit test.
+func (t *TelemetryHook) GetTelemetrySettings() *component.TelemetrySettings {
```
crobert-1, line 192: I'm slightly concerned about timing and flakiness here: We're returning an object that theoretically should be locked by the mutex, but after returning there's no lock enforcement. We're exposing the underlying field in what could be a dangerous way. Maybe we should return a copy instead?

_I do understand that practically this will be set once and never modified (for now, since config can't be reloaded), so the odds of anything like this happening are likely 0, but wanted to point it out in case there's a case it could cause trouble._

crobert-1 (commented):
Can we also add goleak to the new extension? It would just be adding a new file `package_test.go` and copying [an example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/generated_package_test.go) from upstream (with minor modification).

internal/confmapprovider/configsource/telemetry.go
dloucasfx: The test is very controlled and the SetTelemetrySettings is called once only at service start and has all needed locks.
Also returning a copy is not practical here as we will have to create a deep copy to handles the fields which are made of pointer and interfaces.

I am honestly not concerned about this here, but if you are, maybe I can return the specific field the test needs `t.telemetrySettings.MeterProvider` would this help?

crobert-1: Oh you know what, I was confusing this with the `usedSources` field, never mind. This is probably fine, sorry about that!

crobert-1 (approved):

internal/components/components.go
```diff
@@ -168,6 +169,7 @@ func Get() (otelcol.Factories, error) {
 		ackextension.NewFactory(),
 		basicauthextension.NewFactory(),
 		bearertokenauthextension.NewFactory(),
+		configsourcetelemetryextension.NewFactory(),
```
dmitryax, line 172: If we always include it with the config converter, there is no reason to give users any control on the extension. We can remove this registration and user facing configuration. Then we also won't need [the deduplication logic](https://github.com/signalfx/splunk-otel-collector/pull/7479/changes#diff-e99a4a072adde7223d13c239f967e301b31d9769ff97fd186043dfea6922b729R40-R49)

dloucasfx: @dmitryax can you provide pointers on how we can do this? the extension is injected into the config by InjectConfigSourceTelemetryExtension, if we don't register it the collector can't unmarshal it. 

My understanding, an extension must be present in service.extensions to be started by the service, so it must be registered.

dmitryax: Oh you're right. I missed that part

dmitryax (approved):

PR #7480, opened 2026-04-22
[chore][deployment/puppet] Update Puppet installation instructions

crobert-1:

Update installation instructions to properly point to Splunk's artifactory instance.

### DIFF
--- a/deployments/puppet/README.md
+++ b/deployments/puppet/README.md
@@ -32,13 +32,11 @@ On Windows, the collector is installed as a Windows service and its environment
 variables are set at the service scope, i.e.: they are only available to the
 collector service and not to the entire machine.
 
-## Installation and Configuration
-
-### Installation methods
+## Installation
 
 This module can be downloaded and installed from [Splunk's JFrog Artifactory instance](https://splunk.jfrog.io/ui/repos/tree/General/puppet-splunk).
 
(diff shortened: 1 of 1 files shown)


deployments/puppet/README.md
```diff
@@ -32,9 +32,45 @@ On Windows, the collector is installed as a Windows service and its environment
...
+## Installation
 
-This module can be downloaded and installed from [Puppet Forge](https://forge.puppet.com/modules/signalfx/splunk_otel_collector).
+This module can be downloaded and installed from [Splunk's JFrog Artifactory instance](https://splunk.jfrog.io/ui/repos/tree/General/puppet-splunk).
+
+### [Connect Puppet to Artifactory](https://docs.jfrog.com/artifactory/docs/puppet-repositories#connect-puppet-to-artifactory)
+
+Add the following snippet to your `puppet.conf` file:
+
+With authentication:
```
jinja2: Is this a real usecase? Like would a customer ever have a user in our artifactory?

crobert-1: Not that I'm aware of, I included it just to be safe. If it's just extra noise I can remove it.

jinja2: I think we should remove it. It could be confusing to a user

crobert-1: I've removed it 👍

jinja2 (approved):

PR #7509, opened 2026-04-29

srv-gh-o11y-gdi:


tests/receivers/smartagent/prometheus-exporter/prometheus_exporter_test.go
crobert-1, line 1: Do we know why these aren't being sent anymore? (I'm not seeing any relevant changelogs that explain this)

jkoronaAtCisco: No, I'm going to investigate it deeper.

jkoronaAtCisco: For `otelcol_exporter_sent_metric_points` [new attributes been added upstream](https://github.com/open-telemetry/opentelemetry-collector/pull/14998):  `server.address` and `url.path`, so I had to include them in test.

Tests for ` otelcol_receiver_{accepted,refused,failed}_metric_points` were mistakenly removed from the golden file during the investigation  because they initially appeared "MISSING in actual", which turned out to be a timing issue.

Now all tests are restored and passing.

crobert-1 (approved):

internal/components/components_test.go
```diff
@@ -65,11 +68,11 @@ func TestDefaultComponents(t *testing.T) {
...
 		"elasticsearch",
 		"file_log",
-		"filestats",
-		"fluentforward",
+		"file_stats",
+		"fluent_forward",
 		"googlecloudpubsub",
 		"haproxy",
-		"hostmetrics",
+		"host_metrics",
```
jinja2, line 75: We should update configs to use the new names. Feel free to resolve my comment if the plan is to update configs in a follow up PR

jkoronaAtCisco: https://github.com/signalfx/splunk-otel-collector/pull/7522

PR #7512, opened 2026-04-29
[chore] Clarify OpAMP feature gate in configs

dmitryax:
The extension appears enabled in the default configs, but it is disabled at startup unless the `splunk.opamp.enabled` feature gate is enabled. The comments make that behavior explicit for users reading or editing the configs.

### DIFF
--- a/cmd/otelcol/config/collector/agent_config.yaml
+++ b/cmd/otelcol/config/collector/agent_config.yaml
@@ -27,6 +27,8 @@ extensions:
       endpoint: "${SPLUNK_API_URL}"
       # Use instead when sending to gateway
       #endpoint: "${SPLUNK_GATEWAY_URL}"
+  # opamp/splunk_o11y is included in the default config, but startup removes it
+  # from service.extensions unless --feature-gates=+splunk.opamp.enabled is set.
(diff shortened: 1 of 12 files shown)


crobert-1 (commented):
A couple thoughts:
1. Would it be useful to simply move every comment to just reference some documentation on the feature gate? I'm wondering if that would be simpler to keep up to date moving forward than finding/replacing all of these comments when functionality changes.
2. I'm also concerned about the wording that the feature gate "is not enabled yet." This will change in the future some day as the feature gate moves to beta, but there's a chance we might miss updating these comments when that happens, and also as Copilot points out, this doesn't clearly explain that it's currently disabled by default but could be enabled by the user.

dmitryax:
@crobert-1 I updated the comment. It should address your concerns to some extend. I don't think moving it to a separate doc would make it easier given that both the behavior and docs can be updates over time, but the config might be stay as is in the user files

crobert-1 (approved):
I think this will still need updated when moving the feature gate to beta (ideally in my mind it wouldn't need updated), but it's fine for now. Thanks for adding!

PR #7516, opened 2026-05-02
[configconverter] Drops SignalFx exporter from trace pipelines

crobert-1:

- Add a new feature gate (enabled by default) that drops the `signalfx` exporter from all trace pipelines.
- Any trace pipeline that no longer has any exporters will be dropped as well.
- Log deprecation/info notices when exporter/pipeline is removed.

### DIFF
--- a/internal/configconverter/common.go
+++ b/internal/configconverter/common.go
@@ -51,7 +51,9 @@ func ConverterFactoryFromConverter(conv confmap.Converter) confmap.ConverterFact
 func getMetricsPipelineAndReceivers(pipelines map[string]any) (map[string]any, []any, error) {
 	metricsPipeline := map[string]any{}
 	if mp, ok := pipelines["metrics"]; ok && mp != nil {
-		metricsPipeline = mp.(map[string]any)
+		if metricsPipeline, ok = mp.(map[string]any); !ok {
+			return nil, nil, fmt.Errorf("metrics pipeline is of unexpected form (%T): %v", mp, mp)
+		}
 	}
 	pipelines["metrics"] = metricsPipeline
 
--- a/internal/configconverter/common.go
+++ b/internal/configconverter/common.go
@@ -99,11 +99,11 @@ func getExtensionsFromService(service map[string]any) ([]any, error) {
 }
 
 func isSignalFxExporter(exporter string) bool {
-	return strings.ToLower(exporter) == "signalfx" || strings.HasPrefix(strings.ToLower(exporter), "signalfx/")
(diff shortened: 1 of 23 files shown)


internal/configconverter/testdata/drop_signalfx_exporter_traces_pipelines/agent_config_expected.yaml
crobert-1, line 1: Ran a diff for the original config files against expected, only changes are which exporters are in the pipelines. The actual diff for each file is quite lengthy, but that's mainly the result of difference in comments and ordering.

  | File | Trace exporters before | Trace exporters after |
  |---|---|---|
  | agent_config.yaml | [otlp_http, signalfx] | [otlp_http] |
  | ecs_ec2_config.yaml | [otlp_http, signalfx] | [otlp_http] |
  | fargate_config.yaml | [otlp_http, signalfx] | [otlp_http] |
  | full_config_linux.yaml | [otlp_http, signalfx] | [otlp_http] |
  | gateway_config.yaml | [otlp_http] | [otlp_http] |
  | logs_config_linux.yaml | no traces pipeline | no traces pipeline |
  | otlp_config_linux.yaml | [otlp_http, signalfx] | [otlp_http] |
  | upstream_agent_config.yaml | [otlp_http, signalfx] | [otlp_http] |

internal/configconverter/common.go
crobert-1, line 1: Most of these changes are moving from a specific config converter's file to a central method so it's more obvious which helpers are available. Also, some helper methods were special cased to check if something exists, then if it does and it's valid, also return a sub-key. I separated some of this logic to make it easier to reuse.

internal/configconverter/drop_signalfx_trace_correlation_test.go
```diff
@@ -0,0 +1,273 @@
...
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+			require.Equal(t, original, config)
+		})
+	}
+}
+
+func TestDropSignalFxTracesExporterIfFeatureGateEnabled(t *testing.T) {
```
crobert-1, line 115: This kind of testing will end up requiring some level of up-keep, as it will fail every time default configs are updated. The alternative is to just copy the current default configs and use them as input. I slightly prefer directly referencing the current configs to ensure the config converter works with them properly, vs potentially going stale or missing a case, but it's fine with me if we prefer to test a copy.

crobert-1:
_This is no longer necessary, closing._

PR #7522, opened 2026-05-05
[chore] Update default configs to use canonical receiver names

jkoronaAtCisco:
## Summary
- Follow-up to #7509 — replaces the legacy `fluentforward` and `hostmetrics` receiver aliases with their canonical names (`fluent_forward`, `host_metrics`) across the shipped collector configs.
- Addresses reviewer feedback on #7509 (https://github.com/signalfx/splunk-otel-collector/pull/7509#pullrequestreview-4223446774).
- Updated files: agent / upstream-agent / full / ecs_ec2 configs (plus their FIPS variants), the salt template, and the `transform_metrics` example.

## Test plan
- [x] `go test ./internal/components/...`
- [x] `go test ./cmd/otelcol/...` (includes config validation)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.chloggen/update-default-config-aliases.yaml
+++ b/.chloggen/update-default-config-aliases.yaml
@@ -1,5 +1,5 @@
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: deprecation
+change_type: breaking
 
 # The name of the component, or a single word describing the area of concern, (e.g. crosslink)
 component: config
--- a/.chloggen/update-default-config-aliases.yaml
+++ b/.chloggen/update-default-config-aliases.yaml
@@ -1,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: config
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update default configs to use canonical receiver names (`fluent_forward`, `host_metrics`) instead of legacy aliases (`fluentforward`, `hostmetrics`).
+
+# One or more tracking issues related to the change
+issues: [7522]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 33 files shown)


cmd/otelcol/config/collector/agent_config.yaml
dloucasfx, line 1: Do we need to a changelog with deprecation note? My only concern is when the alias is removed, users with old naming will break and we might miss it now that we are changing the default config. 
WDYT?

crobert-1 (commented):
There are still quite a few `hostmetrics` references in tests and test configs, we should update those as well as part of this PR. Not a definitive list, but I'm seeing some in `internal/configconverter`, `tests/general`, `internal/settings`.

Also, I see some configs under `deployments` are updated (salt) but others are not updated (Databricks, nomad). I don't have too strong of a preference if they should be updated in this PR or after, but we should be consistent either way. Also, changelog notes for each deployment that gets updated would need to be included.

cmd/otelcol/config/collector/agent_config.yaml
jkoronaAtCisco: That's right. I added changelog entry,

jkoronaAtCisco:
> There are still quite a few `hostmetrics` references in tests and test configs, we should update those as well as part of this PR. Not a definitive list, but I'm seeing some in `internal/configconverter`, `tests/general`, `internal/settings`.
> 
> Also, I see some configs under `deployments` are updated (salt) but others are not updated (Databricks, nomad). I don't have too strong of a preference if they should be updated in this PR or after, but we should be consistent either way. Also, changelog notes for each deployment that gets updated would need to be included.

@crobert-1 added missing config changes and changelog entry. Thanks!

.chloggen/update-default-config-aliases.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
```
crobert-1: I think this should be considered a breaking change. If users have custom configurations that they're using with the default configs, they might be referencing these component names in their custom config, thus rendering their custom config broken.

crobert-1 (approved):

PR #7524, opened 2026-05-05
[ta-v2][chore] Generate TAs per platform

pjanotti:
In order to work better with deployment server we need the per platform packages to use different directories, respectively:

* `Splunk_TA_otel` for the multi-platform package
* `Splunk_TA_otel_linux_x86_64` for the Linux only package
* `Splunk_TA_otel_windows_x86_64` for the Windows only package

Besides the packaging there were the following changes:

1. Exposing some new env vars to the config so the actual directory name of the TA is not hardcoded
2. Publication script now publishes each package to a different Splunk base app


packaging/ta-v2/tests/package_test.go
```diff
@@ -30,13 +30,15 @@ import (
...
-	windowsBinPath   = "Splunk_TA_otel/windows_x86_64/"
+	distDir       = "../out/distribution"
+	multiOSName   = "Splunk_TA_otel"
+	linuxName     = "Splunk_TA_otel_linux_x86_64"
+	windowsName   = "Splunk_TA_otel_windows_x86_64"
+	multiOSTgz    = multiOSName + ".tgz"
+	linuxTgz      = linuxName + ".tgz"
+	windowsTgz    = windowsName + ".tgz"
+	linuxBinDir   = "/linux_x86_64/"
+	windowsBinDir = "/windows_x86_64/"
```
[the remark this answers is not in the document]
pjanotti: I don't mind the suggestion but filepath.Join doesn't behave as described above

packaging/ta-v2/appinspect.sh
```diff
@@ -20,4 +20,15 @@ echo "Building Docker image from $SCRIPT_DIR/Dockerfile.splunk-appinspect..."
...
 docker run --rm -v "$PACKAGES_PATH:/packages:ro" "$IMAGE_NAME"
+exit_code=$?
+set -e
+echo "Appinspect exited with code $exit_code. See https://dev.splunk.com/enterprise/reference/appinspect/appinspectcliref for exit codes."
+if [ "$exit_code" -eq 101 ]; then
+    echo "Appinspect found issues in the packages."
+    exit 1
+else
+    echo "Appinspect passed with no issues."
+    exit 0
```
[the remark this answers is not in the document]
pjanotti: Incorrect

packaging/ta-v2/tests/package_test.go
```diff
@@ -194,44 +203,52 @@ func TestMultiOSPackageContents(t *testing.T) {
...
 
 			mandatoryPaths := []string{
-				modularInputName + "/configs/agent_config.yaml",
-				modularInputName + "/configs/gateway_config.yaml",
-				modularInputName + "/default/app.conf",
-				modularInputName + "/default/inputs.conf",
-				modularInputName + "/README/inputs.conf.spec",
-				modularInputName + "/static/appIcon_2x.png",
-				modularInputName + "/static/appIcon.png",
+				filepath.Join(root, "configs/agent_config.yaml"),
```
crobert-1: This is a question out of total ignorance: I'm seeing quite a few `filepath.Join` calls with built-in "/" to strings here, and this test is supposed to be covering Windows. Are these `/` separators being handled correctly for Windows, and is the test here mirroring actual functionality?

If there's any noteworthy difference, is there any reason we can't just move to using separate arguments to `filepath.Join` to mark the actual file path?

packaging/ta-v2/tests/package_test.go
```diff
@@ -267,18 +282,30 @@ func TestPackageMandatoryFiles(t *testing.T) {
...
-				assert.False(t,
-					strings.HasSuffix(appConfVersion, "-linux-x86-64") || strings.HasSuffix(appConfVersion, "-windows-x86-64"),
-					"Multi-OS app.conf version %q should not have OS/arch suffix", appConfVersion)
-			case "Linux":
-				assert.True(t, strings.HasSuffix(appConfVersion, "-linux-x86-64"),
-					"Linux app.conf version %q should end with '-linux-x86-64'", appConfVersion)
-			case "Windows":
-				assert.True(t, strings.HasSuffix(appConfVersion, "-windows-x86-64"),
-					"Windows app.conf version %q should end with '-windows-x86-64'", appConfVersion)
+			require.Regexp(t, `^\d+\.\d+\.\d+(-[0-9]+-[a-z0-9]+)?$`, appConfVersion)
```
crobert-1, line 286: A comment to show expected format might be helpful (since I'm slow with regex)

packaging/ta-v2/tests/package_test.go
```diff
@@ -267,18 +282,30 @@ func TestPackageMandatoryFiles(t *testing.T) {
...
+			var appConfPackageID string
+			inPackageSection := false
+			for _, line := range strings.Split(appConfContent, "\n") {
+				trimmed := strings.TrimSpace(line)
+				if trimmed == "[package]" {
+					inPackageSection = true
+					continue
+				}
+				if inPackageSection {
+					if strings.HasPrefix(trimmed, "[") {
```
crobert-1: Maybe move `[` into a variable and call it `newSectionDelimiter` for clarity?

packaging/ta-v2/tests/package_test.go
```diff
@@ -267,18 +282,30 @@ func TestPackageMandatoryFiles(t *testing.T) {
...
+						break
+					}
+					if strings.HasPrefix(trimmed, "id") && strings.ContainsRune(trimmed, '=') {
+						parts := strings.SplitN(trimmed, "=", 2)
```
crobert-1: Maybe we use `strings.ReplaceAll` to remove all spaces and make all of these checks simpler? (I'm assuming the need for so many sub-checks is to account for optional spaces around `=`?) I guess if the ID itself can have spaces this wouldn't work.

If that doesn't work, could we use regex to match optional spaces around `=` with the `id` prefix?

The existing logic is just a bit hard to follow. I'll leave it up to you though, if you think those just make it more complicated feel free to push back here.

pkg/modularinput/ta_launcher_test.go
```diff
@@ -1507,3 +1587,89 @@ func TestHandleLaunchAsTA_ValidationMode_ValidatorReceivesItems(t *testing.T) {
...
+			name:     "multi OS path",
+			execPath: "/opt/splunk/etc/apps/Splunk_TA_otel/platform/bin/Splunk_TA_otel",
+			want:     "Splunk_TA_otel",
+		},
+		{
+			name:     "platform path",
+			execPath: "/opt/splunk/etc/apps/Splunk_TA_otel_platform/platform/bin/Splunk_TA_otel",
+			want:     "Splunk_TA_otel_platform",
+		},
+		{
```
crobert-1, line 1611: One other empty case would be the execPath `/`, we'd expect "" still (From [filepath.Dir documentation](https://pkg.go.dev/path/filepath#Dir))

.gitlab-ci.yml
```diff
@@ -1526,7 +1526,40 @@ ta-v2-splunkbase-release:
...
       if [[ "$should_release" = true ]] && [[ -z "$DRY_RUN" ]]; then
         echo "Releasing TA v2 to Splunkbase"
-        APP_ID=7125 TA_PACKAGES_PATH=packaging/ta-v2/out/distribution VERSION_TAG=${RELEASE_VERSION} ./.gitlab/publish-ta-v2-to-splunkbase.sh
+        TA_DIR=packaging/ta-v2/out/distribution
+
+        declare -A log_files pids
+        for entry in \
+            "7125:${TA_DIR}/Splunk_TA_otel.tgz" \
+            "8968:${TA_DIR}/Splunk_TA_otel_linux_x86_64.tgz" \
+            "8969:${TA_DIR}/Splunk_TA_otel_windows_x86_64.tgz"
```
crobert-1, line 1535: Where are these prefixes coming from? (eg `7125:`) It looks like these are supposed to be app IDs, where are those defined?

crobert-1 (approved):
Just some minor nits

packaging/ta-v2/tests/package_test.go
pjanotti: Yes, `/` is a valid path separator on Windows and you can combine then in the same path, something like `C:\Windows/system32` is a valid path. That said some tools/libraries may have trouble with that, but as you noticed not here in the tests. Anyway, I will change everything to `filepath.Join`

.gitlab-ci.yml
pjanotti: Yes, these are the app ids on splunkbase site

packaging/ta-v2/appinspect.sh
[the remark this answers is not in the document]
pjanotti: Check https://dev.splunk.com/enterprise/reference/appinspect/appinspectcliref

.gitlab/publish-ta-v2-to-splunkbase.sh
crobert-1, line 1: Do you think it would be valuable to add a `--dry-run` flag (or something along those lines) and add testing for this script? That way when updates are made we can make sure it's working as much as possible without actually pushing?

pjanotti: that is a good idea, what I've been using is running the script and then deleting the uploaded packages - by default they are "invisible"

PR #7538, opened 2026-05-12

srv-gh-o11y-gdi:


crobert-1:
Closing so it can be re-generated on top of dep bumps.

PR #7556, opened 2026-05-13
[all] Update deprecated aliases

crobert-1:

Aliases updated:
- `cloudfoundry` -> `cloud_foundry`
- `kafkametrics` -> `kafka_metrics`
- `kubeletstats` -> `kubelet_stats`
- `metricstransform` -> `metrics_transform`
- `tcpcheck` -> `tcp_check`


cmd/otelcol/config/collector/full_config_linux.yaml
```diff
@@ -499,7 +499,7 @@ processors:
   # Enables the metrics transform processor with default settings
   # Full configuration here: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstransformprocessor
   # NOTE: This processor settings need to be change when using this processor
-  #transforms:
+  #metrics_transform:
```
crobert-1, line 502: Since this was commented out and already broken, I don't believe this needs a changelog entry.

deployments/cloudfoundry/bosh/jobs/splunk-otel-collector/templates/otel-collector-config.yaml.erb
```diff
@@ -3,7 +3,7 @@
 <% else %>
 ---
 receivers:
-  cloudfoundry:
+  cloud_foundry:
```
crobert-1, line 6: Tanzu releases are handled separately from mainline releases so this doesn't need a changelog entry

Makefile
crobert-1, line 1: I'm a bit torn on whether to keep it as `kafkametrics` vs `kafka-metrics` here, neither are _exactly_ right, happy to discuss or update if anyone has thoughts otherwise.

.chloggen/deprecated_aliases_default_configs.yaml
```diff
@@ -0,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: config
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update deprecated receiver alias from `kubeletstats` to `kubelet_stats` in default configs
```
dmitryax, line 8: There only two changelog entries here while the PR changes more components. However, we already have those changelog items from upstream, right? SO we don't need any I guess

crobert-1: A couple things:
1. I've only included these two because the changelog entries are for updates to our default configs, which are breaking changes.
2. The remaining alias updates in this PR have no different impact than the upstream changes, so upstream changelog entries are sufficient, IMO.

jinja2 (approved):

PR #7581, opened 2026-05-20
Migrate from `libsplunk` to the OTel Injector

pjanotti:
## Summary

Replaces the custom `libsplunk.so` auto-instrumentation shim with the official [OpenTelemetry injector](https://github.com/open-telemetry/opentelemetry-injector)(`libotelinject`), removing the need to maintain a bespoke C shim in this repo.

## Key changes

- **Custom shim removed**: `instrumentation/src/main.c`, `Dockerfile`, `devel.Dockerfile`, and the per-language test fixtures under `instrumentation/tests/{dotnet,java,nodejs}/` are all deleted.
- **Build now vendors the upstream injector**: `instrumentation/Makefile` clones `open-telemetry/opentelemetry-injector` (pinned via `instrumentation/packaging/injector-release.txt`) and builds it via its own Makefile instead of compiling local C source. Vendoring is needed to allow environment variables with prefix `SPLUNK_`.
- **New CI job** `update-injector` in `.gitlab-ci.yml` (+ `.gitlab/update-injector.sh`) runs on a schedule to open a PR bumping the pinned injector release, mirroring the existing `update-dotnet-agent` job pattern.
- **Renamed artifacts**: `libsplunk` → `libotelinject` throughout `.gitlab-ci.yml` (job names, build output filenames, release assets).
- **Config layout change**: auto-instrumentation config moves from `/etc/splunk/zeroconfig/*.conf` to `/etc/opentelemetry/injector/{default_env.conf,injector.conf}`, matching the layout expected by the official injector. Config now supports both `OTEL_` and `SPLUNK_` env var prefixes (e.g. `SPLUNK_METRICS_ENABLED`, `SPLUNK_PROFILER_ENABLED`).
- **.NET on arm64**: dotnet auto-instrumentation test matrix now runs on arm64 (previously excluded), with a smaller exclude list limited to specific old-glibc distros.

Assisted-by: Claude and Codex


.chloggen/move-from-libsplunk-to-otel-injector.yaml
```diff
@@ -0,0 +1,23 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
jinja2: should this be breaking instead?

.chloggen/move-from-libsplunk-to-otel-injector.yaml
```diff
@@ -0,0 +1,23 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
crobert-1, line 2: This is listed as breaking, but it's not clear what exactly breaks here. Is there something the user needs to change in their config or deployment functionality to fix the break, or make things work as expected? If so, we need to make sure it's clear from the changelog.

.github/workflows/auto-instrumentation.yml
```diff
@@ -76,9 +62,9 @@ jobs:
           distro=$(for d in $dockerfiles; do echo -n "\"$d\","; done)
           arch="\"amd64\", \"arm64\""
           testcase="\"express\",\"tomcat\",\"dotnet\""
-          # splunk-otel-dotnet currently not supported on arm64
-          exclude='{"ARCH": "arm64", "TESTCASE": "dotnet"}'
-          matrix="{\"DISTRO\": [${distro%,}], \"ARCH\": [${arch}], \"TESTCASE\": [${testcase}], \"exclude\": [${exclude}]}"
+          excluded_distros="centos-8 debian-bullseye oraclelinux-8 rockylinux-8"
```
crobert-1, line 64: Is there some documented reason these distros are excluded?

.github/workflows/reusable-build-instrumentation-package.yml
```diff
@@ -0,0 +1,40 @@
...
+          path: instrumentation/build/opentelemetry-injector
+
+      - name: Build ${{ inputs.ARCH }} ${{ inputs.SYS_PACKAGE }} package
+        run: make -C instrumentation/ ${{ inputs.SYS_PACKAGE }}-package ARCH="${{ inputs.ARCH }}"
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: splunk-otel-auto-instrumentation-${{ inputs.ARCH }}-${{ inputs.SYS_PACKAGE }}
+          path: ./instrumentation/dist/*.${{ inputs.SYS_PACKAGE }}
```
crobert-1, line 58: What's going on with `*.` here in the path? It looks like regex, but I wouldn't expect that when specifying an upload path.

.gitlab/update-injector.sh
crobert-1, line 1: Just a note: Since this will be GitLab, please just check after this PR is merged to ensure it's working properly. (No need to follow up, just leaving as a reminder)

instrumentation/README.md
```diff
@@ -97,3 +97,52 @@ configuration of the Collector and Auto Instrumentation for supported platforms.
    object library will then be preloaded for all subsequent processes and inject the environment variables from the
    `/etc/splunk/zeroconfig` configuration files for Java and Node.js processes.
 
+## Running the `auto-instrumentation` CI Workflow Locally
```
crobert-1, line 104: This is super helpful. Maybe as a future enhancement we create a script that handles all of this, with a few input arguments? (We could file a follow up issue)

_Same comment for added testing section in `packaging/tests/README`_

instrumentation/packaging/fpm/common.sh
```diff
@@ -85,48 +85,48 @@ download_nodejs_agent() {
...
 download_dotnet_agent() {
     local tag="$1"
-    local dest="$2"
-    local pkg="splunk-opentelemetry-dotnet-linux-glibc-x64.zip"
+    local arch="$2"
+    local dest="$3"
+    local pkg
+    case "$arch" in
+        arm64) pkg="splunk-opentelemetry-dotnet-linux-glibc-arm64.zip" ;;
+        *)     pkg="splunk-opentelemetry-dotnet-linux-glibc-x64.zip" ;;
```
crobert-1, line 94: Would it work to move to reusable constant strings here with the arch being substituted?

instrumentation/packaging/fpm/common.sh
```diff
@@ -85,48 +85,48 @@ download_nodejs_agent() {
...
+    esac
     local dl_url="$DOTNET_AGENT_RELEASE_URL/download/$tag/$pkg"
 
     echo "Downloading $dl_url ..."
     curl -sfL "$dl_url" -o /tmp/$pkg
 
-    echo "Extracting $pkg to $dest ..."
-    mkdir -p "$dest"
-    unzip -d $dest /tmp/$pkg
+    echo "Extracting $pkg to $dest/glibc ..."
```
crobert-1, line 101: Why **always** the subdir glibc? I'm a bit concerned that this may impact users in unexpected ways, unless this is needed as a result of the move to upstream injector?

packaging/installer/install.sh
```diff
@@ -943,7 +886,7 @@ install() {
         exit 1
       fi
       apt-get -y update
-      apt-get -y install apt-transport-https gnupg
+      apt-get -y install apt-transport-https gnupg libcap2-bin
```
crobert-1, line 896: Do we know exactly why `libcap2-bin` is needed? (These kinds of things are frequent causes of vulns so I just want to make sure we're not unnecessarily exposing ourselves here)

packaging/tests/installer_test.py
```diff
@@ -57,30 +57,43 @@
...
-    "CORECLR_PROFILER_PATH": DOTNET_AGENT_PATH,
-    "DOTNET_ADDITIONAL_DEPS": f"{DOTNET_HOME}/AdditionalDeps",
-    "DOTNET_SHARED_STORE": f"{DOTNET_HOME}/store",
-    "DOTNET_STARTUP_HOOKS": f"{DOTNET_HOME}/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll",
-    "OTEL_DOTNET_AUTO_HOME": DOTNET_HOME,
-    "OTEL_DOTNET_AUTO_PLUGINS":
-        "Splunk.OpenTelemetry.AutoInstrumentation.Plugin,Splunk.OpenTelemetry.AutoInstrumentation",
-}
+DOTNET_AGENT_PATH = f"{DOTNET_HOME}/glibc/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so"
+DOTNET_AGENT_ARM64_PATH = f"{DOTNET_HOME}/glibc/linux-arm64/OpenTelemetry.AutoInstrumentation.Native.so"
```
crobert-1: Why not use a single variable, then use Python's [string formatting](https://docs.python.org/3/library/string.html#format-examples) to fill in the arch value in `get_dotnet_agent_path`?

packaging/tests/installer_test.py
```diff
@@ -415,39 +440,33 @@ def test_installer_with_instrumentation_default(distro, arch, method):
...
-            assert container_file_exists(container, DOTNET_AGENT_PATH)
+        assert container_file_exists(container, get_dotnet_agent_path(arch))
 
         config_attributes = rf"splunk\.zc\.method={zc_method}"
 
         if method == "preload":
-            # verify libsplunk.so was added to /etc/ld.so.preload
-            verify_config_file(container, PRELOAD_PATH, LIBSPLUNK_PATH, None)
+            # verify libotelinject.so was added to /etc/ld.so.preload
+            verify_config_file(container, PRELOAD_PATH, LIBOTELINJECT_PATH, None)
```
crobert-1, line 448: Nit, feel free to drop: It might make sense to pull all of these checks into their own method to make it slightly easier to follow. (I realize this was existing functionality though, so not a big deal.)

packaging/tests/instrumentation/instrumentation_test.py
```diff
@@ -44,30 +44,28 @@
...
 
 JAVA_AGENT_PATH = f"{LIB_DIR}/splunk-otel-javaagent.jar"
-JAVA_CONFIG_PATH = "/etc/splunk/zeroconfig/java.conf"
-CUSTOM_JAVA_CONFIG_PATH = TESTS_DIR / "instrumentation" / "libsplunk-java-test.conf"
-
 NODE_AGENT_PATH = f"{LIB_DIR}/splunk-otel-js.tgz"
-NODE_CONFIG_PATH = "/etc/splunk/zeroconfig/node.conf"
-CUSTOM_NODE_CONFIG_PATH = TESTS_DIR / "instrumentation" / "libsplunk-node-test.conf"
+DOTNET_AGENT_PATH = f"{LIB_DIR}/splunk-otel-dotnet/glibc/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so"
+DOTNET_AGENT_ARM64_PATH = f"{LIB_DIR}/splunk-otel-dotnet/glibc/linux-arm64/OpenTelemetry.AutoInstrumentation.Native.so"
```
crobert-1: Same as other comment, would it make sense to use a single variable for the PATH, then substitute the ARCH?

crobert-1 (commented):
_I'm not very familiar with this functionality, so mostly dumb questions or nits._

packaging/installer/install.sh
```diff
@@ -93,7 +93,7 @@ preload_path="/etc/ld.so.preload"
 default_instrumentation_version="latest"
 default_obi_version="v0.6.0"
 default_deployment_environment=""
-instrumentation_so_path="/usr/lib/splunk-instrumentation/libsplunk.so"
+instrumentation_so_path="/usr/lib/splunk-instrumentation/libotelinject.so"
```
jinja2, line 97: Should we continue recognizing the old `libsplunk.so` installation here, or preferably detect the instrumentation package through `dpkg`/`rpm`? `uninstall()` only checks this new library path, so running the new installer with `--uninstall` against an existing `libsplunk.so` installation appears to remove the collector but skip `splunk-otel-auto-instrumentation`, leaving the old package and preload/configuration state behind.

An old-install/new-uninstall test would be helpful. Or if we are intentionally keeping this a breaking change, a note detailing this behavior would be good to have.

packaging/installer/install.sh
```diff
@@ -918,10 +861,10 @@ EOH
...
 DefaultEnvironment="CORECLR_PROFILER={918728DD-259F-4A6A-AC2B-B85E1B658318}"
-DefaultEnvironment="CORECLR_PROFILER_PATH=${dotnet_install_dir}/linux-x64/OpenTelemetry.AutoInstrumentation.Native.so"
-DefaultEnvironment="DOTNET_ADDITIONAL_DEPS=${dotnet_install_dir}/AdditionalDeps"
-DefaultEnvironment="DOTNET_SHARED_STORE=${dotnet_install_dir}/store"
-DefaultEnvironment="DOTNET_STARTUP_HOOKS=${dotnet_install_dir}/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll"
+DefaultEnvironment="CORECLR_PROFILER_PATH=${dotnet_install_dir}/glibc/linux-${dotnet_native_arch}/OpenTelemetry.AutoInstrumentation.Native.so"
+DefaultEnvironment="DOTNET_ADDITIONAL_DEPS=${dotnet_install_dir}/glibc/AdditionalDeps"
+DefaultEnvironment="DOTNET_SHARED_STORE=${dotnet_install_dir}/glibc/store"
+DefaultEnvironment="DOTNET_STARTUP_HOOKS=${dotnet_install_dir}/glibc/net/OpenTelemetry.AutoInstrumentation.StartupHook.dll"
 DefaultEnvironment="OTEL_DOTNET_AUTO_HOME=${dotnet_install_dir}"
```
jinja2: Does the home path need a `/glibc` too? https://github.com/open-telemetry/opentelemetry-injector/blob/v0.9.2/src/dotnet.zig#L810

packaging/tests/instrumentation/instrumentation_test.py
```diff
@@ -275,11 +296,10 @@ def test_tomcat_instrumentation(distro, arch):
             r"telemetry\.sdk\.language": r"Str\(java\)",
             r"service\.name": rf"Str\(service_name_from_java\)",
             r"deployment\.environment": rf"Str\(deployment_environment_from_java\)",
-            r"com\.splunk\.sourcetype": r"Str\(otel\.profiling\)",
         }
 
-        # overwrite the default libsplunk config with the custom one for testing
-        copy_file_into_container(container, CUSTOM_JAVA_CONFIG_PATH, JAVA_CONFIG_PATH)
+        # overwrite default env var, copy per-language env vars
+        copy_file_into_container(container, CUSTOM_JAVA_ENV_PATH, INJECTOR_DEFAULT_ENV_PATH)
```
jinja2, line 339: Could we add a precedence case here where the same variable is set in both the application/process environment and `default_env.conf`?

The current shim uses [`setenv(..., 0)`](https://github.com/signalfx/splunk-otel-collector/blob/ca321d998ff4cd5d3e0997645e7e4be8df99beb4/instrumentation/src/main.c#L97-L101), so an existing environment value wins. But [injector seems to overwrite it](https://github.com/open-telemetry/opentelemetry-injector/blob/v0.9.2/src/root.zig#L384), so the value from `default_env.conf` wins instead. For example, this test could start the application with `OTEL_SERVICE_NAME=service_name_from_app` while the fixture contains `OTEL_SERVICE_NAME=service_name_from_java`, and assert the intended result.

I’m not sure whether the new precedence is okay, but it appears to change existing behavior. Could we also document that behavior?

instrumentation/README.md
jinja2, line 51: Can we update this section? It still instructs users to preload `libsplunk.so` and configure files under `/etc/splunk/zeroconfig`. The [requirements section](https://github.com/signalfx/splunk-otel-collector/blob/main/instrumentation/README.md#prerequisitesrequirements) also says that .NET is amd64-only.

instrumentation/packaging/fpm/test.sh
```diff
@@ -32,12 +32,12 @@ if [ ! -f /etc/ld.so.preload ]; then
     echo "/etc/ld.so.preload not found" >&2
     exit 1
 fi
-if ! grep -q "$LIBSPLUNK_INSTALL_PATH" /etc/ld.so.preload; then
-    echo "$LIBSPLUNK_INSTALL_PATH not found in /etc/ld.so.preload" >&2
+if ! grep -q "$LIBOTELINJECT_INSTALL_PATH" /etc/ld.so.preload; then
+    echo "$LIBOTELINJECT_INSTALL_PATH not found in /etc/ld.so.preload" >&2
```
jinja2: Is this script still used? I couldn’t find a workflow that invokes it. It also expects package installation to add `libotelinject.so` to `/etc/ld.so.preload`, while the [package test below explicitly verifies that installation does not modify the preload file](https://github.com/signalfx/splunk-otel-collector/pull/7581/changes#diff-f70b3375fab4c7ef12dcac26401cbf5dd4295a32292f14a2d04bedf7077ba6d3R439).

Should this script be removed, or updated to match the behavior covered by the active test suite?

packaging/tests/instrumentation/instrumentation_test.py
jinja2, line 1: Could we add an old-to-new auto-instrumentation package upgrade case to the test suite? I could not find an existing test covering the transition from the released `libsplunk.so` package to the `libotelinject.so` package for deb/rpm upgrade

.chloggen/move-from-libsplunk-to-otel-injector.yaml
```diff
@@ -0,0 +1,23 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Use the official OpenTelemetry injector release instead of the custom injector shim for auto-instrumentation.
+
+# One or more tracking issues related to the change
+issues: [7581]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
jinja2, line 16: I may be missing part of the intended upgrade flow, but I noticed a few meaningful differences between the old and new configuration models. It might be useful to include an upgrade guide for existing users, rather than only describing the new paths.

In particular, could the guidance clarify:

- Whether existing installations are migrated automatically.
- How `/etc/ld.so.preload` should be changed from `libsplunk.so` to `libotelinject.so`.
- How values under `/etc/splunk/zeroconfig` should be migrated to `default_env.conf`.
- How users with different per-runtime or per-service values should configure them with the new shared file, including which values take precedence.
- Whether the old configuration files should be removed.
- For package upgrades, what follow-up is required for the new preload path and config to take effect, including whether instrumented applications need to be restarted.

I’m not fully confident that I understand all the migration implications from the diff alone, so documenting the intended upgrade path would also help confirm which changes are automatic and which are intentionally manual.

jinja2 (commented):
@pjanotti I want to confirm the intended sequencing. You mentioned updating the mass-deployment tests as follow-up after this is released (i.e. the new RPM/DEB is published and resolved by latest). Reading the implementation, I think this needs mass-deployment code changes before that promotion, not just test updates.

All four mass deployment playbooks default the package to `latest` but still emit the old layout in their released implementations. For example, Ansible playbook [writes /etc/splunk/zeroconfig files](https://github.com/signalfx/splunk-otel-collector/blob/v0.157.0/deployments/ansible_collections/signalfx/splunk_otel_collector/roles/collector/tasks/linux_auto_instrumentation.yml#L77-L129) and [preloads libsplunk.so](https://github.com/signalfx/splunk-otel-collector/blob/v0.157.0/deployments/ansible_collections/signalfx/splunk_otel_collector/roles/collector/templates/ld_so_preload.j2).

Once the 0.158.0 becomes latest, fresh runs of these released versions won't work with the new auto-instrumentation.

Am I misunderstanding the plan to rollout mass deployment updates here?

.chloggen/move-from-libsplunk-to-otel-injector.yaml
pjanotti: Configuration files are now following the OTel injector standard, @jinja2 also had very good questions in this regard. I think I need to add a README.md to the migration 🤔

.chloggen/move-from-libsplunk-to-otel-injector.yaml
pjanotti: These are very good points, I think it deserves a dedicated file, probably too much information to the changelog.

.github/workflows/auto-instrumentation.yml
pjanotti: Good Q: the issue is that the GLIB on these is too old for the .NET auto-instrumentation, see @pjanotti
[Remove dotnet tests for arm distros with too old GLIB](https://github.com/signalfx/splunk-otel-collector/pull/7581/commits/f252ab248e5542b3d72b0979a3b11d6df37d878d)

I ended up losing the specific error message, but it is clear that it requires an update of GLIB on the image but this is an issue for the .NET instrumentation.

.github/workflows/reusable-build-instrumentation-package.yml
pjanotti: This is a file glob, there are names by arch (amd64 vs arm64) and the extension by package system, e.g.: deb vs rpm. In other words, the artifacts from the build step depends both on `SYS_PACKAGE` and `ARCH`:

`make -C instrumentation/ ${{ inputs.SYS_PACKAGE }}-package ARCH="${{ inputs.ARCH }}"`

.gitlab/update-injector.sh
pjanotti: The DRY_RUN should make that easy in this case. Pending resolution until I have a chance to run it in GitLab

instrumentation/packaging/fpm/common.sh
pjanotti: I'm not sure if this is what you intended, so let me know if I misunderstood. The packages are arch dependent so is the dotnet instrumentation. The package build script is in charge of downloading the instrumentation, so it seems appropriate that it chooses the package based on the arch. Moving this higher seems to expose a detail about dotnet instrumentation too high in the chain.

Let me know if you meant something else.

instrumentation/packaging/fpm/common.sh
pjanotti: This is a requirement from the injector to support dotnet instrumentation of different OSes, see https://github.com/open-telemetry/opentelemetry-injector/blob/30f5f46177bfbb5497577d31c1dea74d1b565129/README.md?plain=1#L119-L140

We are not benefiting from that since we don't have a package for Alpine.

packaging/installer/install.sh
pjanotti: Good Q, this actually a dependency of the collector, see https://github.com/signalfx/splunk-otel-collector/actions/runs/29796692790/job/88531125411#step:9:390

```
Unpacking splunk-otel-collector (0.156.0-74) ...
dpkg: dependency problems prevent configuration of splunk-otel-collector:
 splunk-otel-collector depends on libcap2-bin; however:
  Package libcap2-bin is not installed.

dpkg: error processing package splunk-otel-collector (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 splunk-otel-collector

FAILED
```

We need a follow up to see the proper action here: document it? add to the install script? etc

packaging/tests/installer_test.py
pjanotti: This a good point, but we want to port these tests to go, so in principle not investing much in changing them, just doing enough to ensure coverage for the new library. I will track this one when migrating the tests.

packaging/tests/installer_test.py
pjanotti: 40f25891f

packaging/tests/instrumentation/instrumentation_test.py
pjanotti: 40f25891f

packaging/installer/install.sh
pjanotti: Either we have a test gap or this doesn't break the instrumentation 🤔

packaging/tests/instrumentation/instrumentation_test.py
pjanotti: Per [injector source code comment](https://github.com/open-telemetry/opentelemetry-injector/blob/30f5f46177bfbb5497577d31c1dea74d1b565129/src/config.zig#L120) there is a merge prior to setting the environment variable so this should preserve the current behavior, but I will double-check this explicitly.

packaging/installer/install.sh
pjanotti: I also would like to remove this special configuration for systemd, per https://github.com/open-telemetry/opentelemetry-injector/pull/130, but that probably is better if done outside of this PR.

instrumentation/packaging/fpm/test.sh
pjanotti: yeah, I couldn't find any usage of this script, removing it...

pjanotti: e500d8e88

instrumentation/README.md
pjanotti: f9ba89d9d

packaging/installer/install.sh
pjanotti: 943d8be2b

instrumentation/packaging/fpm/common.sh
crobert-1: Sorry, forgot this was bash. I was thinking of just having a simpler getter method (or an equivalent functionality) that could just take the arch and format the package name accordingly. Something like `pkg=fmt.Sprintf("splunk-opentelemetry-dotnet-linux-glibc-%s.zip", arch)`, but it's probably more complicated that the current form to make it work in bash.

packaging/installer/install.sh
crobert-1: I don't need anything more at this point, I just wanted to confirm it's needed. Thanks for doing the work of checking and sharing the results!

.chloggen/move-from-libsplunk-to-otel-injector.yaml
pjanotti: Now, this is now covered by the `subtext:` section.

.gitlab/update-injector.sh
pjanotti: Tested and fixed via dry run in gitlab

instrumentation/README.md
pjanotti: Created ticket to track this.

packaging/installer/install.sh
pjanotti: Fixed, the old package is now removed.

packaging/tests/instrumentation/instrumentation_test.py
pjanotti: Done.

.chloggen/move-from-libsplunk-to-otel-injector.yaml
pjanotti: These were added in various changes, including a repeat of the blurb in the mass deployments: 3d1532f9d

packaging/tests/instrumentation/instrumentation_test.py
pjanotti: 5d9da43e2

instrumentation/packaging/injector-release.txt
```diff
@@ -0,0 +1 @@
+v0.9.2
```
jinja2: Should we update to the latest https://github.com/open-telemetry/opentelemetry-injector/releases/tag/v0.11.0?

deployments/chef/templates/injector.conf.erb
```diff
@@ -0,0 +1,12 @@
+jvm_auto_instrumentation_agent_path=<%= node['splunk_otel_collector']['auto_instrumentation_java_agent_jar'] %>
+nodejs_auto_instrumentation_agent_path=/usr/lib/splunk-instrumentation/splunk-otel-js/node_modules/@splunk/otel/instrument.js
+dotnet_auto_instrumentation_agent_path_prefix=/usr/lib/splunk-instrumentation/splunk-otel-dotnet
+<% unless @with_java -%>
+auto_instrumentation_disabled=jvm
```
jinja2: [This test](https://github.com/open-telemetry/opentelemetry-injector/blob/v0.9.2/src/config.zig#L1038-L1055) seems to imply that the injector will only respect the last `auto_instrumentation_disabled` set here. Should we make it comma-separated like `auto_instrumentation_disabled=jvm,dotnet`?

deployments/chef/recipes/auto_instrumentation.rb
jinja2: I am confused if `install` work safely on an existing host. I think when `auto_instrumentation_version` is `latest`, this should use `action :upgrade`. [Upgrade](https://docs.chef.io/client/18/resources/bundled/package/#actions) is stated as the action which guarantees installation of the newest available version.

For e.g., if 0.159.0 is already installed, `action :install` can leave that version in place. But now `with_otel_injector` is set to `true` because the requested value is `latest`, so the recipe deletes the legacy config and writes libotelinject.so into preload. Version 0.159.0 contains only `libsplunk.so`, leaving the preload entry pointed at a missing file.

.chloggen/move-from-libsplunk-to-otel-injector.yaml
```diff
@@ -0,0 +1,35 @@
...
+  This is a breaking change for existing auto-instrumentation installations. Configuration files moved from
```
jinja2, line 17: Can we add a section in [UPGRADING](https://github.com/signalfx/splunk-otel-collector#upgrade-guidelines) readme? We can move some of the finer details into the upgrading section and link to it from subtext. I think we should mention mass-deployment too, something like - 
```
When using Ansible, Chef, Puppet, or Salt, do not edit the generated `default_env.conf` directly. Move the settings into the corresponding deployment-tool parameters and upgrade to an injector-compatible version of that deployment module before promoting the new auto-instrumentation package.
```

And I would also highlight that user should backup `/etc/splunk/zeroconfig/` in case they need to downgrade or something

.chloggen/move-from-libsplunk-to-otel-injector.yaml
```diff
@@ -0,0 +1,35 @@
...
+  `/etc/splunk/zeroconfig/` to `/etc/opentelemetry/injector/` (`default_env.conf` and `injector.conf`). Existing
+  values are not migrated, and the legacy files are removed during upgrade and cannot be recovered.
```
jinja2: Not blocking but I wonder if we considered preserving the legacy config in a documented backup location instead of deleting the entire directory

instrumentation/packaging/fpm/preuninstall.sh
```diff
@@ -15,11 +15,11 @@
 # limitations under the License.
 
 PRELOAD_PATH="/etc/ld.so.preload"
-LIBSPLUNK_PATH="/usr/lib/splunk-instrumentation/libsplunk.so"
+LIBOTELINJECT_PATH="/usr/lib/splunk-instrumentation/libotelinject.so"
 
-if [ -f "$PRELOAD_PATH" ] && grep -q "$LIBSPLUNK_PATH" "$PRELOAD_PATH"; then
-    echo "Removing $LIBSPLUNK_PATH from $PRELOAD_PATH"
-    sed -i -e "s|$LIBSPLUNK_PATH||" "$PRELOAD_PATH"
+if [ -f "$PRELOAD_PATH" ] && grep -q "$LIBOTELINJECT_PATH" "$PRELOAD_PATH"; then
```
jinja2: I was looking at the Debian [docs](https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#summary-of-ways-maintainer-scripts-are-called), and I think there may be a pre-existing upgrade bug here, although I am not sure why it has not been reported before.

This script is passed to FPM using `--before-remove`. FPM installs it as the Debian `prerm` or RPM `%preun` hook so that preload path is removed when the package is uninstalled. But my concern is that the old package's hook also runs during a normal RPM/DEB upgrade. This script removes the preload entry without distinguishing an upgrade from an uninstall, and the incoming package does not appear to restore it.

instrumentation/packaging/injector-release.txt
pjanotti: yes, thanks f07e377d7

deployments/chef/templates/injector.conf.erb
pjanotti: Excellent catch! All 4 mass deployments were treating this wrongly, fixing via 498b4ab33

deployments/chef/recipes/auto_instrumentation.rb
pjanotti: This was subtle issue, but per my reading you are corrected, fixing via 828658cea

instrumentation/packaging/fpm/preuninstall.sh
pjanotti: Removing the old hook via 468760757

.chloggen/move-from-libsplunk-to-otel-injector.yaml
pjanotti: I didn't gave it much thought because the files shouldn't be part of the new package, so uninstall would need an explicit option to delete them. However, in retrospect this seems like a good idea giving that the configuration is preserved if somebody misses the release notes and these files are small anyway. Added via b352edf24, notice the side effect for mass deployments noted via d4a27a320

.chloggen/move-from-libsplunk-to-otel-injector.yaml
pjanotti: 4adda10fd

instrumentation/packaging/fpm/preuninstall.sh
pjanotti: 9e96973d5

pjanotti:
CI failures are unrelated

jinja2 (approved):

jinja2 (approved):
LGTM!

PR #7578, opened 2026-05-19
[DRAFT] POC for packaging OpAMP Supervisor to manage collector with launcher

kelseyma:
**Description:** 
This is a reference PR for the migration to using a launcher to enable direct or supervisor mode. Direct mode will continue to run the collector as is, while supervisor mode will start OpAMP Supervisor to manage the collector.

**Breakdown of PRs:** 
- [x] 1. Launcher core (#7589): add launcher and unit tests, but no package entrypoint changes.
- [x] 2. Windows launcher lifecycle/logging (#7613): add windows launcher support.
- [x] 3. Gated manual build targets (#7771): add gated make targets for building the binaries, but keep artifact packaging unchanged 
    - [x] a) Launcher improvements (#7769): process config flags and add flexibility for handling multiple configs
- [x] 4. Packaging
    - [x] a) add launcher/supervisor into binary compile builds: #7889
    - [x] b) add binaries into linux packaging and install script
    - [x] c) add binaries into windows packaging and install script
    - [x] d) any changes needed for mass deployments
- [x] 5. Final prod prep: add additional e2e tests, etc

**Testing:** Verified the following on Ubuntu/Windows
- Linux brownfield upgrade from current Splunk Otel Collector, default direct mode.
- Linux brownfield switch to supervisor by env var only, then rollback to direct mode. 
- Linux brownfield switch from direct w/ opamp feature gate to supervisor mode.
- Linux greenfield install with supervisor enabled.
- Windows brownfield MSI upgrade from current Splunk Otel Collector, default direct mode.
- Windows brownfield registry-only switch to supervisor, then rollback to direct mode. 
- Windows brownfield switch from direct w/ opamp feature gate to supervisor mode.
- Windows greenfield MSI install with supervisor enabled.

PR #7589, opened 2026-05-26
Add OpAMP Supervisor launcher core for remote configuration support

kelseyma:
**Description:** 
This PR breaks out the launcher core from the larger OpAMP Supervisor POC (https://github.com/signalfx/splunk-otel-collector/pull/7578) so it can be reviewed independently. It adds a new `splunk-otel-collector-launcher` entrypoint and internal OpAMP Supervisor launcher config package. 

The launcher keeps direct mode as the default (keeping existing collector runtime behavior), using `SPLUNK_OPAMP_SUPERVISOR_ENABLED=true` to switch into supervisor mode. In supervisor mode it:
-  generates the initial supervisor config, preserving existing generated supervisor config on later starts
- derives or preserves OpAMP endpoint settings and enables the same capabilities supported by opamp extension plus remote config support 
- starts the collector with any args (if set) and a sanitized collector config with the splunk opamp extension disabled (as supervisor adds its own)
- ensures existing default behavior for the listen interface (agent_config.yaml => 127.0.01, gateway_config.yaml => 0.0.0.0)

This slice intentionally does not include packaging/install changes, full Windows service behavior, or config converter changes. Windows currently gets a compile-only launcher stub, followup PR will add service behavior separately.

### DIFF
--- a/cmd/splunk-otel-collector-launcher/main_windows.go
+++ b/cmd/splunk-otel-collector-launcher/main_windows.go
@@ -1,0 +1,27 @@
+// Copyright Splunk Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build windows
+
+package main
+
+import (
+	"fmt"
+
+	"github.com/signalfx/splunk-otel-collector/internal/opampsupervisor/launcher"
+)
+
+func run(_, _ []string, _ launcher.Paths) error {
+	return fmt.Errorf("windows launcher service support is not implemented")
+}
--- a/internal/opampsupervisor/launcher/config.go
+++ b/internal/opampsupervisor/launcher/config.go
@@ -1,0 +1,379 @@
+// Copyright Splunk Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package launcher
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	SupervisorEnabledEnvVar = "SPLUNK_OPAMP_SUPERVISOR_ENABLED"
+
+	CollectorConfigEnvVar = "SPLUNK_CONFIG"
+	GatewayURLEnvVar      = "SPLUNK_GATEWAY_URL"
+	IngestURLEnvVar       = "SPLUNK_INGEST_URL"
+
+	opampSplunkExtension  = "opamp/splunk_o11y"
+	opampFeatureGate      = "splunk.opamp.enabled"
+	collectorConfigEnvRef = "${" + CollectorConfigEnvVar + "}"
+)
+
+// Paths contains package installation paths and supervisor state paths.
+type Paths struct {
+	CollectorExecutable  string
+	SupervisorExecutable string
+	SupervisorConfig     string
+	StorageDirectory     string
+	UseHUPConfigReload   bool
+}
+
+// Command is the executable, arguments, and environment the launcher should run.
+type Command struct {
+	Path string
+	Args []string
+	Env  []string
+}
+
+// SupervisorConfig is the subset of the upstream opampsupervisor config this
+// launcher renders for package-managed supervisor mode.
+type SupervisorConfig struct {
+	Server       supervisorServer       `yaml:"server"`
+	Storage      supervisorStorage      `yaml:"storage"`
+	Agent        supervisorAgent        `yaml:"agent"`
+	Capabilities supervisorCapabilities `yaml:"capabilities"`
+}
+
+type supervisorServer struct {
+	TLS      any            `yaml:"tls,omitempty"`
+	Headers  map[string]any `yaml:"headers,omitempty"`
+	Endpoint string         `yaml:"endpoint"`
+}
+
+type supervisorCapabilities struct {
+	ReportsEffectiveConfig     bool `yaml:"reports_effective_config"`
+	ReportsHealth              bool `yaml:"reports_health"`
+	ReportsAvailableComponents bool `yaml:"reports_available_components"`
+	AcceptsRemoteConfig        bool `yaml:"accepts_remote_config"`
+	ReportsRemoteConfig        bool `yaml:"reports_remote_config"`
+	ReportsOwnMetrics          bool `yaml:"reports_own_metrics"`
+	ReportsHeartbeat           bool `yaml:"reports_heartbeat"`
+}
+
+type supervisorStorage struct {
+	Directory string `yaml:"directory"`
+}
+
+type supervisorAgent struct {
+	Env                map[string]string `yaml:"env,omitempty"`
+	Executable         string            `yaml:"executable"`
+	ConfigFiles        []string          `yaml:"config_files"`
+	Args               []string          `yaml:"args,omitempty"`
+	PassthroughLogs    bool              `yaml:"passthrough_logs"`
+	UseHUPConfigReload bool              `yaml:"use_hup_config_reload,omitempty"`
+	ValidateConfig     bool              `yaml:"validate_config"`
+}
+
+// SupervisorEnabled reports whether the persisted service-scoped supervisor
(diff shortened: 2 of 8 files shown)


internal/opampsupervisor/launcher/config.go
```diff
@@ -0,0 +1,379 @@
...
+	if !errors.Is(err, fs.ErrNotExist) {
+		return fmt.Errorf("stat supervisor config %q: %w", paths.SupervisorConfig, err)
+	}
+
+	agentArgs := filterSupervisorAgentArgs(args)
+
+	configPath := strings.TrimSpace(env[CollectorConfigEnvVar])
+	if configPath == "" {
+		return fmt.Errorf("%s must be set in supervisor mode", CollectorConfigEnvVar)
+	}
```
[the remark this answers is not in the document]
kelseyma: This is intended behavior. Users should have their SPLUNK_CONFIG value set as that is the scope of what can be updated via remote config, and flags will always be applied over last and not be handled/parsed for configs in this remote config scope.

cmd/splunk-otel-collector-launcher/main_windows.go
```diff
@@ -0,0 +1,27 @@
...
+package main
+
+import (
+	"fmt"
+
+	"github.com/signalfx/splunk-otel-collector/internal/opampsupervisor/launcher"
+)
+
+func run(_, _ []string, _ launcher.Paths) error {
+	return fmt.Errorf("windows launcher service support is not implemented")
```
pjanotti, line 26: not that it matters for this PR, but a Q. for future direction: for the collector it makes sense to support run as a "Windows service" and run as standalone executable, are we going to keep that same pattern for the launcher?

internal/opampsupervisor/launcher/config.go
```diff
@@ -0,0 +1,379 @@
...
+
+// PrepareSupervisor creates the supervisor directory and writes the initial
+// supervisor config only when it does not already exist. Existing supervisor
+// config is user-editable and is preserved across launcher restarts.
+func PrepareSupervisor(args []string, env map[string]string, paths Paths) error {
+	if err := prepareSupervisorStorageDir(paths); err != nil {
+		return err
+	}
+
+	_, err := os.Stat(paths.SupervisorConfig)
```
pjanotti: In practice this a corner case concern, but given nature of files it is always possible for `os.Stat` succeed and later to have a failure when actually reading the file (the file can deleted in between). The general preference is to have this case treated at the location where the file is read.

internal/opampsupervisor/launcher/config.go
```diff
@@ -0,0 +1,379 @@
...
+	if err := writeYAML(paths.SupervisorConfig, supervisorConfig, 0o600); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// loadConfigFile reads a collector config into a generic map so direct OpAMP
+// settings can be copied into the generated supervisor config when present.
+func loadConfigFile(path string) (map[string]any, error) {
```
pjanotti: nit:
```suggestion
func loadCollectorConfigFile(path string) (map[string]any, error) {
```

internal/opampsupervisor/launcher/config.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+
+	agentArgs := filterSupervisorAgentArgs(args)
+
+	configPath := strings.TrimSpace(env[CollectorConfigEnvVar])
+	if configPath == "" {
+		return fmt.Errorf("%s must be set in supervisor mode", CollectorConfigEnvVar)
+	}
+
+	config, err := loadConfigFile(configPath)
```
pjanotti: nit:
```suggestion
	collectorConfig, err := loadConfigFile(configPath)
```

internal/opampsupervisor/launcher/paths_others.go
```diff
@@ -0,0 +1,29 @@
...
+package launcher
+
+// DefaultPaths returns the installed collector, supervisor, and state
+// locations used by the package-managed service on Linux packages.
+func DefaultPaths() Paths {
+	return Paths{
+		CollectorExecutable:  "/usr/bin/otelcol",
+		SupervisorExecutable: "/usr/bin/opampsupervisor",
+		SupervisorConfig:     "/etc/otel/collector/supervisor/supervisor_config.yaml",
+		StorageDirectory:     "/etc/otel/collector/supervisor",
```
dmitryax: I know we discussed it but I forgot that we already store log checkpoints in `/var/lib/otelcol/filelogs` . So this should  probably be `/var/lib/otelcol/supervisor`

kelseyma: Sounds good, in that case did we want to move the supervisor_config.yaml back out to the `/etc/otel/collector` level or did we still want a supervisor dir with just the config in it?

dmitryax: If it's just one file, I think it can be `/etc/otel/collector/supervisor_config.yaml`

internal/opampsupervisor/launcher/config.go
```diff
@@ -0,0 +1,379 @@
...
+// other collector feature gates continue to flow to the child collector.
+func filterFeatureGateValue(value string) (string, bool) {
+	var kept []string
+	for _, item := range strings.Split(value, ",") {
+		trimmed := strings.TrimSpace(item)
+		if trimmed == "" {
+			continue
+		}
+		gate := strings.TrimLeft(trimmed, "+-")
+		if gate == opampFeatureGate {
```
dmitryax: Why do we need to remove this feature gate? Doesn't supervisor override the extension anyway? 

Even if we want to override the feature gate, mutating the `featureGates` argument is not the right way because the FG (splunk.opamp.enabled) will be turned to enabled by default soon and removing it from the argument will get completely different behavior -- always enabled.

internal/opampsupervisor/launcher/config.go
```diff
@@ -0,0 +1,379 @@
...
+	if strings.TrimSpace(env[IngestURLEnvVar]) != "" {
+		return "${" + IngestURLEnvVar + "}/v1/opamp"
+	}
+	return ""
+}
+
+// prepareSupervisorStorageDir creates the supervisor directory before config
+// generation or supervisor persistence files are used.
+func prepareSupervisorStorageDir(paths Paths) error {
+	if err := os.MkdirAll(paths.StorageDirectory, 0o755); err != nil {
```
pjanotti: Why execute rights?

internal/opampsupervisor/launcher/config.go
```diff
@@ -0,0 +1,379 @@
...
+		return fmt.Errorf("create supervisor directory %q: %w", paths.StorageDirectory, err)
+	}
+	if err := os.Chmod(paths.StorageDirectory, 0o755); err != nil {
```
pjanotti: Why a separate os.Chmod is needed? Does it mean that os.MkdirAll can succeed but doesn't set the requested rights?

internal/opampsupervisor/launcher/config.go
```diff
@@ -0,0 +1,379 @@
...
+		}
+		kept = append(kept, trimmed)
+	}
+	return strings.Join(kept, ","), len(kept) > 0
+}
+
+// writeYAML writes generated config files with their parent directories created
+// for package-managed state paths.
+func writeYAML(path string, value any, perm os.FileMode) error {
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
```
pjanotti: Similar to above: why not use the equivalent to the `perm` parameter?

internal/opampsupervisor/launcher/config.go
kelseyma: I see, the removal is to trigger the config converter so we don't have multiple opamp extensions running at once (since supervisor adds just `opamp` named extension while our read path support is adding `opamp/splunk_o11y`). But if it's going to be enabled by default soon, the launcher should instead explicitly disable it. Would the best method for that be to always ensure feature gate flag has `-splunk.opamp.enabled` or do we have another method for disabling feature gates?

internal/opampsupervisor/launcher/config.go
pjanotti: Perhaps `perm` should be a local constant to the function itself.

cmd/splunk-otel-collector-launcher/main_windows.go
kelseyma: Yes, it'll follow the same pattern as the collector (and also opamp supervisor)

internal/opampsupervisor/launcher/config.go
kelseyma: This check was mainly to exit out of the supervisor config generation if the config already exists, as we don't want to overwrite existing configs. The reading of the supervisor config will take place in the supervisor process when started by the launcher.

internal/opampsupervisor/launcher/config.go
kelseyma: was originally adding that to allow the directory to have search permissions, but actually found that opamp supervisor can handle creation of the storage dir and set to `0o700`. Will probably remove this method as it isn't really needed anymore in the launcher process

internal/opampsupervisor/launcher/config.go
kelseyma: Good callout! I think that was codex overengineering a bit with trying to make this method generic. This is only used to write a config yaml in an existing path so will simplify it to just write the file and remove this extra logic

pjanotti (approved):

internal/opampsupervisor/launcher/config.go
kelseyma: From discussion, launcher will now:
- load SPLUNK_CONFIG, remove opamp/splunk_o11y from definitions and service list, and then write the sanitized result to a generated collector config path (no changes to feature gate flag values)
- add the subdir back for /etc/otel/collector/supervisor which will have: collector_config.yaml  and supervisor_config.yaml

dmitryax (approved):

PR #7591, opened 2026-05-27
feat: logs and metrics config for Windows

weliasz:
Adds new logs_windows_config.yaml to MSI distribution
Adds new metrics_windows_config.yaml to MSI distribution

OTEL is run with following files depending on presence of specified parameter:
agent_config.yaml -> SPLUNK_ACCESS_TOKEN
logs_windows_config.yaml -> SPLUNK_PLATFORM_LOGS_INDEX
metrics_windows_config.yaml -> SPLUNK_PLATFORM_METRICS_INDEX

### DIFF
--- a/cmd/otelcol/config/collector/logs_windows_config.yaml
+++ b/cmd/otelcol/config/collector/logs_windows_config.yaml
@@ -1,0 +1,327 @@
+receivers:
+  windowseventlog:
+    channel: Application
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+    # Excludes events from providers whose metadata DLL is not accessible
+    # to the collector process (e.g. Tenable Nessus Agent).
+    # Remove entries below if you want to collect those events despite the error.
+    exclude_providers:
+      - "Tenable Nessus Agent"
+      - "XENAGENT"
+
+  windowseventlog/security:
+    channel: Security
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+    # NOTE: Splunk blacklists the following with a message-body regex condition:
+    #   EventCode 4662 where Object Type != groupPolicyContainer
+    #   EventCode 566  where Object Type != groupPolicyContainer
+    # The exclude_events filter below drops ALL 4662/566 events — it cannot
+    # replicate the groupPolicyContainer exception. Use a filterprocessor
+    # with OTTL if you need conditional filtering.
+    #
+    # exclude_events:
+    #   - event_ids: [4662, 566]
+
+  windowseventlog/system:
+    channel: System
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  windowseventlog/printservice:
+    channel: Microsoft-Windows-PrintService/Operational
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  # --- Disabled in inputs.conf -------------------------------------------
+
+  # ForwardedEvents exists but is disabled — enable on WEF collectors only
+  windowseventlog/forwardedevents:
+    channel: ForwardedEvents
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  # DC-only channels — will silently no-op on non-DC machines due to ignore_channel_errors
+  windowseventlog/dfs_replication:
+    channel: DFS Replication
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  windowseventlog/directory_service:
+    channel: Directory Service
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  windowseventlog/file_replication:
+    channel: File Replication Service
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  windowseventlog/dns_server:
+    channel: DNS Server
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  windowseventlog/key_management:
+    channel: Key Management Service
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  windowseventlog/defender:
+    channel: Microsoft-Windows-Windows Defender/Operational
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  filelog/dhcp:
+    include:
+      - 'C:\Windows\System32\DHCP\DhcpSrvLog*'
+    start_at: beginning
+    include_file_path: true
+    operators:
+      - type: move
+        from: attributes["log.file.path"]
+        to: resource["com.splunk.source"]
+      - type: add
+        field: resource["com.splunk.sourcetype"]
+        value: DhcpSrvLog
+    storage: file_storage/filelogs
+
+  filelog/windowsupdate_legacy:
+    # For Windows 8 / Server 2008R2 / Server 2012 / Server 2012R2
+    include:
+      - 'C:\Windows\WindowsUpdate.log'
+    start_at: beginning
+    include_file_path: true
+    operators:
+      - type: move
+        from: attributes["log.file.path"]
+        to: resource["com.splunk.source"]
+      - type: add
+        field: resource["com.splunk.sourcetype"]
+        value: WindowsUpdateLog
+    storage: file_storage/filelogs
+
+  filelog/netlogon:
+    include:
+      - 'C:\Windows\debug\netlogon.log'
+    start_at: beginning
+    include_file_path: true
+    operators:
+      - type: move
+        from: attributes["log.file.path"]
+        to: resource["com.splunk.source"]
+      - type: add
+        field: resource["com.splunk.sourcetype"]
+        value: MSAD:NT6:Netlogon
+    storage: file_storage/filelogs
+
+  filelog/firewall:
+    include:
+      - 'C:\Windows\System32\LogFiles\Firewall\*'
+    start_at: beginning
+    include_file_path: true
+    operators:
+      - type: move
+        from: attributes["log.file.path"]
+        to: resource["com.splunk.source"]
+      - type: add
+        field: resource["com.splunk.sourcetype"]
+        value: WindowsFirewallLog
+    storage: file_storage/filelogs
+
+  #   --- GAP: dns.log (MonitorNoHandle) ------------------------------------
+  #   [MonitorNoHandle://$WINDIR\System32\Dns\dns.log]
+  #   The live DNS debug log is held open with an exclusive lock by the DNS
+  #   service. filelogreceiver cannot reliably read exclusively locked files.
+  #   Workaround: use a scheduled task to periodically copy dns.log to a
+  #   staging path, then monitor that path with filelog.
+
+  filelog/dns_debug:
+    include:
+      - 'C:\dns_log_copy\dns.log'
+    start_at: beginning
+    include_file_path: true
+    operators:
+      - type: move
+        from: attributes["log.file.path"]
+        to: resource["com.splunk.source"]
+      - type: add
+        field: resource["com.splunk.sourcetype"]
+        value: MSAD:NT6:DNS
+    storage: file_storage/filelogs
+
+processors:
+  batch:
+  resourcedetection:
+    detectors: [ system, env ]
+    system:
+      hostname_sources: [ "os" ]
+  # WinEventLog://Application -> sourcetype=XmlWinEventLog  source=WinEventLog:Application
+  transform/application:
+    log_statements:
+      - context: log
+        statements:
+          - set(attributes["com.splunk.sourcetype"], "XmlWinEventLog")
+          - set(attributes["com.splunk.source"], "XmlWinEventLog:Application")
+  # WinEventLog://Security -> sourcetype=XmlWinEventLog  source=WinEventLog:Security
+  transform/security:
+    log_statements:
+      - context: log
+        statements:
+          - set(attributes["com.splunk.sourcetype"], "XmlWinEventLog")
+          - set(attributes["com.splunk.source"], "XmlWinEventLog:Security")
+  # WinEventLog://System -> sourcetype=XmlWinEventLog  source=WinEventLog:System
+  transform/system:
+    log_statements:
+      - context: log
+        statements:
+          - set(attributes["com.splunk.sourcetype"], "XmlWinEventLog")
+          - set(attributes["com.splunk.source"], "XmlWinEventLog:System")
+  # WinEventLog://Microsoft-Windows-PrintService/Operational
+  # No renderXml in inputs.conf -> sourcetype=WinEventLog
+  transform/printservice:
+    log_statements:
+      - context: log
+        statements:
+          - set(attributes["com.splunk.sourcetype"], "WinEventLog")
+          - set(attributes["com.splunk.source"], "WinEventLog:Microsoft-Windows-PrintService/Operational")
+  # transform/forwardedevents — uncomment if enabling ForwardedEvents receiver
+  transform/forwardedevents:
+    log_statements:
+      - context: log
+        statements:
+          - set(attributes["com.splunk.sourcetype"], "XmlWinEventLog")
+          - set(attributes["com.splunk.source"], "XmlWinEventLog:ForwardedEvents")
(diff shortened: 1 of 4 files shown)


cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
+receivers:
+  windowseventlog:
```
pjanotti, line 2: You can already use the new naming convention:

```suggestion
  windows_event_log:
```

cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
...
+    channel: Application
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+    # Excludes events from providers whose metadata DLL is not accessible
+    # to the collector process (e.g. Tenable Nessus Agent).
+    # Remove entries below if you want to collect those events despite the error.
+    exclude_providers:
+      - "Tenable Nessus Agent"
```
pjanotti, line 11: I guess that these are coming straight from the TAs shipped with Splunk, if not: are these specific to some deployment? How did you arrive at the list of exclusions?

cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
...
+      - "XENAGENT"
+
+  windowseventlog/security:
+    channel: Security
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+    # NOTE: Splunk blacklists the following with a message-body regex condition:
```
pjanotti, line 19: I wonder if we can use the [query capabilities instead](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39055), but anyway something that doesn't need to be addressed on this PR.

cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
...
+
+  windowseventlog/defender:
+    channel: Microsoft-Windows-Windows Defender/Operational
+    start_at: beginning
+    raw: true
+    ignore_channel_errors: true
+
+  filelog/dhcp:
+    include:
+      - 'C:\Windows\System32\DHCP\DhcpSrvLog*'
```
pjanotti, line 89: Instead of assuming the Windows dir here use the corresponding env var:
```suggestion
      - '${env:SystemRoot}\System32\DHCP\DhcpSrvLog*'
```

cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
...
+        to: resource["com.splunk.source"]
+      - type: add
+        field: resource["com.splunk.sourcetype"]
+        value: DhcpSrvLog
+    storage: file_storage/filelogs
+
+  filelog/windowsupdate_legacy:
+    # For Windows 8 / Server 2008R2 / Server 2012 / Server 2012R2
+    include:
+      - 'C:\Windows\WindowsUpdate.log'
```
pjanotti, line 104:
```suggestion
      - '${env:SystemRoot}\WindowsUpdate.log'
```

cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
...
+        from: attributes["log.file.path"]
+        to: resource["com.splunk.source"]
+      - type: add
+        field: resource["com.splunk.sourcetype"]
+        value: WindowsUpdateLog
+    storage: file_storage/filelogs
+
+  filelog/netlogon:
+    include:
+      - 'C:\Windows\debug\netlogon.log'
```
pjanotti, line 118:
```suggestion
      - '${env:SystemRoot}\debug\netlogon.log'
```

cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
...
+        from: attributes["log.file.path"]
+        to: resource["com.splunk.source"]
+      - type: add
+        field: resource["com.splunk.sourcetype"]
+        value: MSAD:NT6:Netlogon
+    storage: file_storage/filelogs
+
+  filelog/firewall:
+    include:
+      - 'C:\Windows\System32\LogFiles\Firewall\*'
```
pjanotti, line 132:
```suggestion
      - '${env:SystemRoot}\System32\LogFiles\Firewall\*'
```

cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
...
+  #   --- GAP: dns.log (MonitorNoHandle) ------------------------------------
+  #   [MonitorNoHandle://$WINDIR\System32\Dns\dns.log]
+  #   The live DNS debug log is held open with an exclusive lock by the DNS
+  #   service. filelogreceiver cannot reliably read exclusively locked files.
+  #   Workaround: use a scheduled task to periodically copy dns.log to a
+  #   staging path, then monitor that path with filelog.
+
+  filelog/dns_debug:
+    include:
+      - 'C:\dns_log_copy\dns.log'
```
pjanotti, line 153:
```suggestion
      - '${env:SystemDrive}\dns_log_copy\dns.log'
```

cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
...
+    source: "otel"
+    sourcetype: "otel"
+    index: "${SPLUNK_PLATFORM_LOGS_INDEX}"
+
+extensions:
+  file_storage/filelogs:
+    directory: "${ProgramData}/Splunk/OpenTelemetry Collector/FileStorage"
+    compaction:
+      on_start: true
+      directory: /tmp/
```
pjanotti, line 258:
```suggestion
      directory: ${env:TEMP}
```

cmd/otelcol/config/collector/logs_windows_config.yaml
```diff
@@ -0,0 +1,327 @@
...
+          - set(attributes["com.splunk.source"], "XmlWinEventLog:Microsoft-Windows-Windows Defender/Operational")
+
+
+exporters:
+  splunk_hec/logs:
+    endpoint: "${SPLUNK_HEC_URL}"
+    token: "${SPLUNK_HEC_TOKEN}"
+    source: "otel"
+    sourcetype: "otel"
+    index: "${SPLUNK_PLATFORM_LOGS_INDEX}"
```
pjanotti, line 251: For my enlightment: IIRC token implies an index, if that is really correct what happens if the index and token are not referencing the same Splunk index?

pjanotti (commented):
~~I'm not sure about the removal of `SPLUNK_CONFIG` from the Windows Installer. It feels more like a different install `mode` and having that decided via defining install properties feels unfriendly to users.~~

I misread the change: this is adding the Splunk platform configurations when the  SPLUNK_PLATFORM_* env vars are configured.

pjanotti:
@weliasz forgot to mention on the review: add tests for the new configurations in CI. You can use run the configurations directly or via some of the MSI tests.

cmd/otelcol/config/collector/logs_windows_config.yaml
weliasz: During HEC configuration one can select "Allowed Indexes" and "Default Index".
Configuration of these is not mandatory. If no allowed indexes are allowed HEC can store data in any index provided by the client. By default HEC stores data in Splunk default index (typically "main"). So there is no such concept as token referencing index (just available and default). Client provided index is the one used and ingest may fail if index is not allowed by HEC. Does this answer address your concern?

cmd/otelcol/config/collector/logs_windows_config.yaml
weliasz: On Windows 2022 XENAGENT was resulting in exception so I added it as exception.
"Tenable Nessus Agent" was most probably AI generated, I can remove it and test again.

pjanotti: If this is not on the base TA, please remove it.

cmd/otelcol/config/collector/logs_windows_config.yaml
pjanotti: @weliasz please open a follow-up ticket to see how we can reproduce the feature of the original TA.

packaging/msi/splunk-otel-collector.wxs
```diff
@@ -40,13 +40,21 @@
...
       <Property Id="SPLUNK_API_URL" Secure="yes" />
       <Property Id="SPLUNK_INGEST_URL" Secure="yes" />
       <Property Id="SPLUNK_HEC_URL" Secure="yes" />
       <Property Id="SPLUNK_BUNDLE_DIR" Secure="yes" />
-
-      <CustomAction Id="Set_SPLUNK_CONFIG" Property="SPLUNK_CONFIG" Value="[CommonAppDataFolder]Splunk\OpenTelemetry Collector\[SPLUNK_SETUP_COLLECTOR_MODE]_config.yaml" />
+      <Property Id="SPLUNK_PLATFORM_LOGS_INDEX" Secure="yes" />
+      <Property Id="SPLUNK_PLATFORM_METRICS_INDEX" Secure="yes" />
+
+      <CustomAction Id="Set_SPLUNK_PLATFORM_LOGS_INDEX" Property="SPLUNK_PLATFORM_LOGS_INDEX" Value="[%SPLUNK_PLATFORM_LOGS_INDEX]" />
```
pjanotti, line 50: For consistency with all other configurations on the MSI the options should be made via _properties_ not env vars.

Using `Value=[%NAME]` tries to take the value from the environment which is inconvenient in Windows since, unless if using a shell like bash, there is no syntax to inline env vars within the scope of the command like in bash - in Powershell, you can inline the properties but they will leave after the command is done. It is more convenient to use the `Value="[NAME]"` since the user can just add those to the end of the installation command, e.g.: `msiexec /i installer.msi MY_PROP=value`

Apply the same to the next custom action below.

packaging/msi/splunk-otel-collector.wxs
```diff
@@ -40,13 +40,21 @@
       <!-- SPLUNK_*_URL properties required to start the collector -->
       <Property Id="SPLUNK_REALM" Value="us0" Secure="yes" />
 
-      <Property Id="SPLUNK_CONFIG" Secure="yes" />
       <Property Id="SPLUNK_API_URL" Secure="yes" />
       <Property Id="SPLUNK_INGEST_URL" Secure="yes" />
       <Property Id="SPLUNK_HEC_URL" Secure="yes" />
       <Property Id="SPLUNK_BUNDLE_DIR" Secure="yes" />
-
-      <CustomAction Id="Set_SPLUNK_CONFIG" Property="SPLUNK_CONFIG" Value="[CommonAppDataFolder]Splunk\OpenTelemetry Collector\[SPLUNK_SETUP_COLLECTOR_MODE]_config.yaml" />
```
pjanotti, line 49: In principle you can keep this and leverage the `COLLECTOR_SVC_ARGS` to add the new properties.

packaging/msi/splunk-otel-collector.wxs
```diff
@@ -55,7 +63,14 @@
       <InstallExecuteSequence>
          <Custom Action="CheckLaunchConditions" After="LaunchConditions" />
          <Custom Action="Set_SPLUNK_HEC_TOKEN" After="LaunchConditions" Condition="NOT SPLUNK_HEC_TOKEN" />
-         <Custom Action="Set_SPLUNK_CONFIG" After="LaunchConditions" Condition="NOT SPLUNK_CONFIG" />
+         <Custom Action="Set_SPLUNK_PLATFORM_LOGS_INDEX" After="CheckLaunchConditions" Condition="NOT SPLUNK_PLATFORM_LOGS_INDEX AND %SPLUNK_PLATFORM_LOGS_INDEX" />
```
pjanotti, line 66: These are also relying on env vars, please make the switch to install properties.

packaging/msi/splunk-otel-collector.wxs
```diff
@@ -55,7 +63,14 @@
...
+         <Custom Action="Set_SPLUNK_PLATFORM_METRICS_INDEX" After="Set_SPLUNK_PLATFORM_LOGS_INDEX" Condition="NOT SPLUNK_PLATFORM_METRICS_INDEX AND %SPLUNK_PLATFORM_METRICS_INDEX" />
+         <Custom Action="AppendDefaultConfigTo_COLLECTOR_SVC_ARGS" After="Set_SPLUNK_PLATFORM_METRICS_INDEX" Condition="COLLECTOR_SVC_ARGS AND SPLUNK_ACCESS_TOKEN" />
+         <Custom Action="SetDefaultConfigTo_COLLECTOR_SVC_ARGS" After="AppendDefaultConfigTo_COLLECTOR_SVC_ARGS" Condition="NOT COLLECTOR_SVC_ARGS AND SPLUNK_ACCESS_TOKEN" />
```
pjanotti, line 69: If you keep the `SPLUNK_CONFIG` you don't need these 2 conditions here, even if removing this can be reduced to a single one.

packaging/msi/splunk-otel-collector.wxs
```diff
@@ -55,7 +63,14 @@
...
+         <Custom Action="AppendLogsConfigTo_COLLECTOR_SVC_ARGS" After="SetDefaultConfigTo_COLLECTOR_SVC_ARGS" Condition="COLLECTOR_SVC_ARGS AND SPLUNK_PLATFORM_LOGS_INDEX" />
+         <Custom Action="SetLogsConfigTo_COLLECTOR_SVC_ARGS" After="AppendLogsConfigTo_COLLECTOR_SVC_ARGS" Condition="NOT COLLECTOR_SVC_ARGS AND SPLUNK_PLATFORM_LOGS_INDEX" />
```
pjanotti, line 71: This should only check for `SPLUNK_PLATFORM_LOGS_INDEX` and be a single condition.

packaging/msi/splunk-otel-collector.wxs
```diff
@@ -55,7 +63,14 @@
...
+         <Custom Action="AppendMetricsConfigTo_COLLECTOR_SVC_ARGS" After="SetLogsConfigTo_COLLECTOR_SVC_ARGS" Condition="COLLECTOR_SVC_ARGS AND SPLUNK_PLATFORM_METRICS_INDEX" />
+         <Custom Action="SetMetricsConfigTo_COLLECTOR_SVC_ARGS" After="AppendMetricsConfigTo_COLLECTOR_SVC_ARGS" Condition="NOT COLLECTOR_SVC_ARGS AND SPLUNK_PLATFORM_METRICS_INDEX" />
```
pjanotti, line 73: Similar here.

packaging/msi/msi-builder/build.sh
```diff
@@ -24,6 +24,8 @@ PROJECT_DIR=${PROJECT_DIR:-"${REPO_DIR}"}
 MSI_SRC_DIR=${MSI_SRC_DIR:-"${REPO_DIR}/packaging/msi"}
 WXS_PATH="${MSI_SRC_DIR}/splunk-otel-collector.wxs"
 AGENT_CONFIG="${REPO_DIR}/cmd/otelcol/config/collector/agent_config.yaml"
+LOGS_CONFIG="${REPO_DIR}/cmd/otelcol/config/collector/logs_windows_config.yaml"
```
pjanotti, line 27: For consistency with https://github.com/signalfx/splunk-otel-collector/pull/7452 these should have the prefix `splunk_`

pjanotti (commented):
It is almost there, please, look at [tests/msi/msi_test.go](https://github.com/signalfx/splunk-otel-collector/blob/main/tests/msi/msi_test.go) on how to add MSI tests for this feature.

cmd/otelcol/config/collector/logs_windows_config.yaml
weliasz: https://splunk.atlassian.net/browse/ADDON-88165

PR #7613, opened 2026-06-03
Add Windows OpAMP Supervisor launcher support

kelseyma:
This PR adds the Windows launcher support based on the larger OpAMP Supervisor POC (https://github.com/signalfx/splunk-otel-collector/pull/7578). On Windows, the implementation is more complex because the launcher cannot replace itself with the child process like Unix can. Main items on the windows side:
- Starts either `otelcol.exe` in direct mode or `opampsupervisor.exe` in supervisor mode
- Supports service handler lifecycle (report statuses, handle stop/shutdown controls)
- Forwards child stdout/stderr into Windows Event Log, configured log file, or stdout/stderr
- Ensure direct mode maintains same user facing behavior for the service

1. Direct mode with valid config (with feature gate)
    a) collector logging to file => service/launcher/collector running, opamp connection seen and collector fully starts. Exits cleanly (shutdown logs seen, all processes stopped)
    b) default collector logging (aka Event log) => same as above, logs seen in Event Log instead
2. Direct mode, kill otelcol process outside of service signals => service/launcher also stopped
3. Direct mode with invalid config (conflicting fields) => service fails to start entirely
4. Supervisor mode with valid config (default collector logging) => service/launcher/supervisor/collector running, logs seen in Event Log
    a) send valid remote config (update exporters to nop and logging to file) => saw updated effective.yaml with remote config added, collector restarted and updated effective config reported to opamp server, collector logs switched to file
    b) send invalid config (conflicting fields) => supervisor validates config and validation fails so does not apply config and continues to run with existing config
    c) stop service => all exit cleanly
5. Supervisor mode, kill opampsupervisor process outside of service signals => service/launcher stop but collector remains due to possible bug in opamp extension, will need to investigate separately
6. Supervisor mode start with invalid config (conflicting fields) => service/launcher/supervisor running, collector failed and opamp reported status is unhealthy with error, stop exits cleanly


cmd/splunk-otel-collector-launcher/main_windows.go
```diff
@@ -17,11 +17,362 @@
...
+var (
+	kernel32API = windows.NewLazySystemDLL("kernel32.dll")
+
+	allocConsoleProc = kernel32API.NewProc("AllocConsole")
+	freeConsoleProc  = kernel32API.NewProc("FreeConsole")
+)
+
+// run starts as a Windows service when invoked by the Service Control Manager
+// and falls back to interactive mode for command-line runs.
+func run(args, env []string, paths launcher.Paths) error {
```
pjanotti:
```suggestion
func runAsService(args, env []string, paths launcher.Paths) error {
```

cmd/splunk-otel-collector-launcher/main_windows.go
```diff
@@ -17,11 +17,362 @@
...
+	// Allocate a console so it can later deliver CTRL_BREAK_EVENT to a child process group.
+	// Only free the console if we allocated it successfully; otherwise we may detach
+	// an existing interactive console and lose stderr output.
+	consoleAllocated := false
+	if err := allocConsole(); err != nil {
+		if !errors.Is(err, windows.ERROR_ACCESS_DENIED) {
```
pjanotti, line 55: Why the special treatment for ERROR_ACCESS_DENIED?

pjanotti: Is that the error that you get when running in interactive mode? If yes, I suggest to allocate the console only when running as service, it will simplify the code here a bit.

dmitryax (approved):
LGTM. @pjanotti can you please take a look as well?

cmd/splunk-otel-collector-launcher/main_windows.go
```diff
@@ -17,11 +17,362 @@
...
+
+// startChild launches the selected binary in its own process group. Interactive
+// runs inherit console stdio; service runs bridge child stdout/stderr to Event
+// Log through the provided event log handle.
+func startChild(cmdSpec launcher.Command, interactive bool, elog windowsEventLog) (*childProcess, error) {
+	cmd := exec.Command(cmdSpec.Path, cmdSpec.Args...)
+	cmd.Env = cmdSpec.Env
+	cmd.SysProcAttr = &syscall.SysProcAttr{CreationFlags: windows.CREATE_NEW_PROCESS_GROUP}
+	if interactive {
+		cmd.Stdin = os.Stdin
```
pjanotti: Q.: what is `os.Stdin` when running as a service? Couldn't you blindly assign it for service and interactive modes?

pjanotti (commented):
LGTM, my Qs can be addressed later, if worth at all.

cmd/splunk-otel-collector-launcher/main_windows.go
kelseyma: I pulled this from the [existing OpAMP Supervisor code](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/26a14acf8dce2a9dd12c451f7af258602b277084/cmd/opampsupervisor/main_windows.go#L26-L34) for windows start. From their comments, they preferred to always allocate the console. It could be possible for interactive mode to not have a console and thus need this allocation, but in most cases it should be present already. Not sure how much we needed to account for that edge case so kept the current location, but open to only allocating for services if we feel that is better, let me know!

cmd/splunk-otel-collector-launcher/main_windows.go
kelseyma: Good callout! On second thought, I don't think assigning stdin is needed at all. There isn't any stdin behavior supported by otelcol or opampsupervisor so will remove to simplify this

cmd/splunk-otel-collector-launcher/main_windows.go
pjanotti: IIRC you have to compile your exe to be a Windows subsystem, i.e.: for it to be a GUI, for it to not have a console or launch it as a service. We can look into more details later, perhaps it is something that we could change upstream too.

pjanotti (approved):

PR #7646, opened 2026-06-10
Add transform/limit_histogram_buckets to default agent config 

rohit-sonawane:
## Description
The SignalFx backend silently drops explicit-bounds histograms with more than 31 explicit
boundaries (32 total buckets). gRPC metrics like `grpc.server.duration` emit 40+
boundaries by default, causing them to disappear at ingest with no error surfaced.

This adds a `transform/limit_histogram_buckets` processor to the default `metrics` pipeline
in both the standard and FIPS agent configs. It uses
`merge_histogram_buckets(31, method="limit_buckets")` from `transformprocessor`, which
merges adjacent buckets until the boundary count is within the backend limit while
preserving the total count and sum. The processor runs between `memory_limiter` and `batch`.
`tests/go.mod` is bumped to `opentelemetry-collector-contrib v0.154.0`, which introduced the
`limit_buckets` method.

New integration test `TestHistogramBucketLimitReducesBuckets` sends a synthetic 40-bound
histogram through the pipeline and asserts ≤31 bounds are emitted. `TestDefaultAgentConfig`
is updated to include the new processor.


.chloggen/otl-4345-histogram-bucket-limit.yaml
```diff
@@ -0,0 +1,23 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: config
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `transform/limit_histogram_buckets` processor to default agent metrics pipeline to cap explicit histogram buckets at 31.
```
jinja2: Tmk, we support 31 bucket boundaries (32 buckets). Can we look into this? Is this just a phrasing issue here or are we keeping 31 buckets only, in which case we should verify our backend limit and increment this to allow 32 buckets.

internal/components/components_test.go
```diff
@@ -127,12 +127,13 @@ func TestDefaultComponents(t *testing.T) {
 		"wavefront",
 		"windows_event_log",
 		"windowsperfcounters",
-		"windowsservice",
+		"windows_service",
```
jinja2, line 130: There are a lot of unrelated changes in this PR, like this and go mod upgrades etc. Can you rebase on latest main?

.chloggen/otl-4345-histogram-bucket-limit.yaml
rohit-sonawane: Good catch on the phrasing — it was misleading. The 31 in merge_histogram_buckets(31, method="limit_buckets") is the number of explicit bounds (ExplicitBounds array length), not the number of buckets. So:

31 explicit bounds → 32 total buckets (31 finite intervals + the [last_bound, +Inf) overflow)
The SignalFx backend limit is 32 total buckets = 31 explicit bounds
So 31 is correct — we're allowing exactly up to 32 buckets. I've updated the config comments and changelog to say "31 explicit bounds (32 total buckets)" to make this clear.

crobert-1:
A couple general comments:

1. Please ensure that the [AGENT instructions](https://github.com/signalfx/splunk-otel-collector-chart/blob/main/AGENTS.md) are being followed. (Some of the comments appear to be heavily AI generated which is against the instructions, but my apologies if I'm mistaken.)
2. "Splunk Idea" is referring to [Splunk Ideas](https://ideas.splunk.com/), not Jira. Jira issues should not be referred to here as this is an open source repository and Jira is private/internal.
3. The PR description is not up to date with the contents of this PR, it should be updated accordingly.

_EDIT: Just realized the AGENT instructions that I'm referring to are in the chart repo, and not this one. I'll update here to match._

rohit-sonawane:
@crobert-1  , Thanks for the notes. Removed the internal Jira reference from the PR description (no public Splunk Ideas entry exists for this). Description has been updated to match the current diff and trimmed per the AGENTS guidelines.

.chloggen/otl-4345-histogram-bucket-limit.yaml
jinja2: I think it's actually buckets, not bounds in the function. The [README](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/processor/transformprocessor/v0.154.0/processor/transformprocessor/README.md#merge_histogram_buckets) says "target_value is the maximum number of buckets to keep." So merge_histogram_buckets(31, ...) yields 31 buckets / 30 bounds. Since the backend max is 32 buckets / 31 bounds, shouldn't this be 32? Can you trace it through code and confirm?

Also, it might be worth verifying the transform output against the real backend to confirm these reduced histograms are accepted correctly. This may be overkill if the transform function already has good correctness test coverage, but given the confusion about the bounds/bucket support, a test that sends a 64-bucket histogram, reduces it to 32 buckets, and confirms it shows up correctly in the backend would be good to have.

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -133,6 +133,16 @@ processors:
     detectors: [gcp, ecs, ec2, azure, system]
     override: true
 
+  # Cap histograms to 31 explicit bounds (32 total buckets) before sending to the
+  # SignalFx backend. The backend supports a maximum of 32 buckets (31 explicit
+  # bounds); histograms with more (e.g. gRPC call-duration metrics with 40+) are
+  # otherwise dropped at ingest.
+  transform/limit_histogram_buckets:
```
jinja2, line 142: Some context about our backend support - the 32-bucket limit only applies when the signalfx exporter sends histograms in OTLP format (`send_otlp_histograms: true` is set on exporter). In the default mode it explodes each bucket into separate `_bucket` mts with `le` dimensions, and there's no 32-bucket limit on that path. Adding this transform processor to default configs will also coarsen buckets for default-mode users who never had the drop problem. This will be a breaking change for users who are not* yet on OTLP histograms. Can you confirm if this in the scope of the ask? in which case we should document this

cmd/otelcol/fips/config/agent_config.yaml
```diff
@@ -223,7 +233,7 @@ service:
       #exporters: [otlp_grpc]
     metrics:
       receivers: [host_metrics, otlp]
-      processors: [memory_limiter, batch, resourcedetection]
+      processors: [memory_limiter, transform/limit_histogram_buckets, batch, resourcedetection]
```
jinja2: Could you map out which pipelines this needs to cover vs. what's intentionally out of scope? A few I noticed: the `metrics/internal` pipeline (agent and gateway) exports to signalfx without the transform and I think we have some histograms exposed from internal telemetry, and the gateway's metrics pipeline receives OTLP from sources that may not have done the limiting (e.g. instrumentation sdks sending http req/res time histograms to gateway). Let's be explicit about where histograms are reduced vs where they can still get dropped after this change.

cmd/otelcol/config/collector/agent_config.yaml
jinja2: We've decided to make this breaking change for existing decomposed histogram metrics too

rohit-sonawane: Acknowledged — changelog is change_type: breaking and the subtext explicitly calls out the default-mode (decomposed _bucket/le) impact and how to opt out.

.chloggen/otl-4345-histogram-bucket-limit.yaml
```diff
@@ -0,0 +1,35 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: config
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `transform/limit_histogram_buckets` processor to default agent metrics pipeline to prevent SignalFx backend from silently dropping histograms that exceed the 32-bucket limit.
```
jinja2:
```suggestion
note: Add `transform/limit_histogram_buckets` processor to limit explicit-bucket histograms in the default agent metrics pipeline to at most 32 buckets to prevent native OTLP histograms from being dropped by the SignalFx backend.
```

.chloggen/otl-4345-histogram-bucket-limit.yaml
```diff
@@ -0,0 +1,35 @@
...
+  configs. Requires opentelemetry-collector-contrib v0.156.0 or later.
+
+  This is a breaking change: in the default exporter mode (`send_otlp_histograms: false`),
+  histograms are exploded into `_bucket`/`le` metric time series and are not subject to the
+  32-bucket backend limit. Users on the default mode will see reduced histogram resolution
+  (fewer `le` bucket boundaries) as a result of this processor running unconditionally.
+
+  To opt out, remove the `transform/limit_histogram_buckets` processor from the `metrics`
+  pipeline processors list and delete its definition from the `processors:` block in
+  `agent_config.yaml`.
```
jinja2: I think the current entry has the right information, but it is a bit much for a changelog. I’d suggest a middle ground that explains the user-visible behavior and the impact on decomposed histograms without going too deeply into the implementation. Something like:


```suggestion
subtext: |
  Histograms above the limit are uniformly compacted by merging adjacent buckets.
  This preserves the total count and sum, but can produce fewer than 32 buckets and
  reduce histogram resolution.

  When `send_otlp_histograms` is false, the SignalFx exporter emits histogram buckets
  as `_bucket` metric time series, with the `le` dimension identifying each bucket's
  upper bound. Compaction therefore produces fewer `_bucket` series and `le` values.

  To opt out, remove `transform/limit_histogram_buckets` from the metrics pipeline.
```

cmd/otelcol/config/collector/agent_config.yaml
```diff
@@ -136,6 +136,16 @@ processors:
     detectors: [gcp, ecs, ec2, azure, system]
     override: true
 
+  # Cap histograms to 32 buckets before sending to the SignalFx backend.
+  # When send_otlp_histograms is enabled, the backend drops histograms with more
+  # than 32 buckets at ingest. Metrics such as gRPC call-duration histograms emit
+  # 40+ buckets by default and would otherwise be silently lost.
```
jinja2:
```suggestion
# Limit explicit histograms to at most 32 buckets before export through signalfx exporter.
# With `send_otlp_histograms: true`, the backend drops histograms above this limit.
# With `send_otlp_histograms: false`, this reduces the emitted `_bucket` series and `le` boundaries.
```

Same suggestion for other places this comment shows up.

cmd/otelcol/fips/config/agent_config.yaml
jinja2: @rohit-sonawane I think we should include this in gateway config too since sdk instrumented apps could send metrics to gateway directly.

tests/general/histogram_bucket_limit_test.go
```diff
@@ -0,0 +1,165 @@
...
+		"histogram_bucket_limit.yaml",
+		func(c testutils.Collector) testutils.Collector {
+			return c.WithEnv(map[string]string{
+				"COLLECTOR_OTLP_PORT": fmt.Sprintf("%d", collectorOTLPPort),
+			})
+		},
+	)
+	defer shutdown()
+
+	// Dial the collector's OTLP gRPC receiver.
```
crobert-1: It would be easier to follow this test if the GRPC logic of sending data to the receiver was in its own helper method. (Also, can you check to see if there's some generic test helper that already does this in the repo, I would have assume this already exists?)

tests/general/histogram_bucket_limit_test.go
```diff
@@ -0,0 +1,165 @@
...
+	conn, err := grpc.NewClient(collectorAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	require.NoError(t, err)
+	defer conn.Close()
+
+	client := pmetricotlp.NewGRPCClient(conn)
+	payload := pmetricotlp.NewExportRequestFromMetrics(buildHistogramMetric(inputBounds))
+
+	// Wait for the collector's OTLP receiver to become ready, then send once.
+	require.Eventually(t, func() bool {
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
crobert-1: I'm a bit confused by the context timeout being 2 seconds, but the eventually ticker is running every 500 ms. Is this purposeful, are there any unintended side effects as a result?

tests/general/histogram_bucket_limit_test.go
```diff
@@ -0,0 +1,165 @@
...
+	defer tc.PrintLogsOnFailure()
+	defer tc.ShutdownOTLPReceiverSink()
+
+	// Reserve a port for the collector's own OTLP/gRPC receiver.
+	// The test pushes histograms to this port; the collector processes them and
+	// forwards the result to tc.OTLPReceiverSink.
+	collectorOTLPPort := testutils.GetAvailablePort(t)
+
+	_, shutdown := tc.SplunkOtelCollectorProcess(
+		"histogram_bucket_limit.yaml",
```
crobert-1, line 67: Is there any way for this test to somehow use the default config itself? I'm concerned about a "drift" happening between the actual config and the test config. The actual config might break but our tests wouldn't catch it since it's just using a point in time snapshot.

tests/general/histogram_bucket_limit_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
+	"github.com/signalfx/splunk-otel-collector/tests/testutils"
+)
+
+const (
+	// inputBounds is the number of explicit bounds in the synthetic histogram
+	// sent to the collector (40 > 32 buckets, intentionally triggering the limit).
+	inputBounds = 40
```
crobert-1: This name isn't very clear for its purpose, maybe rename it to something like `invalidInputBounds` or something like that to make it more clear?

.chloggen/otl-4345-histogram-bucket-limit.yaml
crobert-1, line 1: Terminology in this PR goes between buckets and bounds/boundaries quite a bit. I think we need to try to be as consistent as possible with messaging and naming. Are buckets or boundaries more obvious to users? Which ever one is used more and more obvious should be used, and we should stick with it across READMEs, changelog notes, configs, tests, etc.

Swapping between them makes it easy for us to stumble into off by one errors without realizing.

crobert-1 (commented):
There are a lot of comments in the YAML and code introduced in this PR that are not very useful. Please review comments and only include ones that are essential in understanding the "why" of what's going on. (IMO having lots of comments just makes it harder to follow the logic itself)

tests/general/histogram_bucket_limit_test.go
rohit-sonawane: Good call. Pulled the dial + send into newOTLPMetricsClient and exportHistogram. I did look for an existing one first , we have OTLP receiver sinks in testutils but nothing that pushes to a collector, so I kept these local for now.

tests/general/histogram_bucket_limit_test.go
rohit-sonawane: Yes. Dropped the per-send timeout to 1s and the ticker to 1s so they line up , a slow dial can't drift the retry cadence anymore.

.chloggen/otl-4345-histogram-bucket-limit.yaml
rohit-sonawane: Agreed,  Standardized on "buckets" everywhere , it matches merge_histogram_buckets and the backend's "32 buckets" phrasing. The test now asserts bucket count directly instead of counting bounds.

tests/general/histogram_bucket_limit_test.go
rohit-sonawane: Renamed to oversizedInputBuckets.

tests/general/histogram_bucket_limit_test.go
rohit-sonawane: TestDefaultAgentConfig already asserts the processor and its pipeline position against the actual agent_config.yaml, so if the config changes and this test isn't updated, that one fails. This test is doing the behavioral half  proving 40 buckets actually come out ≤32 , which the snapshot test can't.

jinja2 (approved):

PR #7662, opened 2026-06-13
[chore] Add an agent skill to help prepare changelog for release

pjanotti:


pjanotti:
Using internal tools to store such type of sources.

PR #7674, opened 2026-06-16
[chore] Allow having closed-source private Go dependency

dmitryax:
Allows consuming closed-source Go packages from a private GitHub repository `github.com/signalfx/splunk-otel-collector-components`

For now we don't pull anything from the closed sourced repo. It will only be used behing a go build flag ensuring that the open-source part is the default in can be built with no extra credentials

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -1,9 +1,26 @@
 name: 'setup environment'
 description: 'setup go environment'
 
+inputs:
+  private-modules-token:
+    description: 'GitHub token with read access to private Go module repositories'
+    required: true
+
 runs:
   using: "composite"
   steps:
+    - name: Configure private Go module access
+      shell: bash
+      env:
+        PRIVATE_MODULES_TOKEN: ${{ inputs.private-modules-token }}
+      run: |
+        if [ -z "$PRIVATE_MODULES_TOKEN" ]; then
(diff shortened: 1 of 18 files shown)


rockb1017 (approved):

.github/actions/setup-environment/action.yml
```diff
@@ -1,9 +1,26 @@
 name: 'setup environment'
 description: 'setup go environment'
 
+inputs:
+  private-modules-token:
```
jinja2, line 9: I think we should separate the build mode from the token itself. Right now passing private-modules-token is doing two things at once; it decides this is a private-components build, and it provides the token.

Could we add a second explicit input to make CI intent cleaner, something like:
```
inputs:
  private-modules:
    description: "Enable private Go module access"
    required: false
    default: "false"
  private-modules-token:
    description: "GitHub token with read access to private Go module repositories"
    required: false
```

Then only require the token when private-modules == "true". I think this will also avoid making forks/dependabot fail just from calling setup (they don't inherit secret)

dmitryax: Yeah sounds reasonable

dmitryax: Added. I'm making it available on all the jobs for now. Later we can remove it where needed

jinja2 (approved):

PR #7680, opened 2026-06-17
[chore] Automate PR for release of Ansible, Chef, and Puppet

pjanotti:
This automates the creation of a PR to release Ansible, Chef, and Puppet.


.gitlab-ci.yml
crobert-1, line 1: Why does this need to be triggered from a GitLab job vs. GitHub?

.gitlab/check-mass-deployments-release.sh
crobert-1, line 1: What do you think of modifying this script to take input variables and then operate on them accordingly? I'm thinking it would be easier to add new mass deployment cases as well as test it. It could then be called by a matrix in the workflow, the matrix having an entry for each mass deployment, or just separate jobs for each.

It would also simplify things in some ways, as the mass deployment-specific logic is pretty similar between mass deployments. If we just pull out anything specific to each one into environment variables they could be passed in and make the script simpler by itself. Or alternatively, have "MASS_DEPLOYMENT_TYPE" as the input variable, then the script does an enum on it setting variables accordingly.

An environment variable for "should create PR" could also be used for testing to make sure the script works in different scenarios.

.gitlab-ci.yml
pjanotti: I'm following the same pattern that we have for "updates", I think this way we can keep the GitHub required rights for CI on the public repo to a minimum.

.gitlab/check-mass-deployments-release.sh
pjanotti: I like the generalization idea, perhaps we could separate as much as possible the specific of each deployment so each deployment has a small script of its specifics and all common parts are shared. 🤔

crobert-1: Yeah, that sounds good 👍

pjanotti: Addressed via [b928bed](https://github.com/signalfx/splunk-otel-collector/pull/7680/commits/b928bedcbd74f6bd7f6e4f3da8d28b9373b300a6)

jinja2 (approved):

PR #7691, opened 2026-06-23
Include logical CPU count for SignalFx host metrics pipelines

dmitryax:
Follow-up to open-telemetry/opentelemetry-collector-contrib#49296.

Adds a config converter so Splunk Collector preserves user-enabled `system.cpu.logical.count` from `host_metrics` receivers by adding it to relevant SignalFx exporter `include_metrics` lists when default translation rules are enabled.

### DIFF
--- a/internal/configconverter/hostmetrics_logical_cpu_count.go
+++ b/internal/configconverter/hostmetrics_logical_cpu_count.go
@@ -1,0 +1,341 @@
+// Copyright Splunk, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configconverter
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"go.opentelemetry.io/collector/confmap"
+)
+
+const (
+	hostMetricsReceiverType      = "host_metrics"
+	hostMetricsReceiverAliasType = "hostmetrics"
+	signalfxExporterType         = "signalfx"
+	metricsPipelineType          = "metrics"
+	cpuScraperName               = "cpu"
+	logicalCPUCountMetricName    = "system.cpu.logical.count"
+)
+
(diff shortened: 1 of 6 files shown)


internal/configconverter/hostmetrics_logical_cpu_count.go
crobert-1, line 1: Nit: Upstream has moved to having host metrics be two words, it'd be ideal to match that with the file names here (`hostmetrics_` to `host_metrics_`).

internal/configconverter/hostmetrics_logical_cpu_count.go
```diff
@@ -0,0 +1,341 @@
...
+	if err != nil {
+		return false, fmt.Errorf("cannot determine exporters::%s::include_metrics: %w", exporterID, err)
+	}
+	includeMetrics = append(includeMetrics, map[string]any{"metric_name": logicalCPUCountMetricName})
+	exporter["include_metrics"] = includeMetrics
+	exporters[exporterID] = exporter
+	return true, nil
+}
+
+func metricFilterListContains(parent map[string]any, key string) (bool, error) {
```
crobert-1, line 275: The name of this function should be more specific since it's always checking if `system.cpu.logical.count` is included. The name by itself makes it look like a generic check of some kind.

internal/configconverter/hostmetrics_logical_cpu_count.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	if !changed {
+		return nil
+	}
+
+	*cfgMap = *confmap.NewFromStringMap(out)
+	return nil
+}
+
+func signalfxExportersForEnabledLogicalCPUCount(
```
crobert-1, line 92: The logical flow of this converter is pretty hard to follow. It would be great if there was some way that the parsing and actual logic could somehow be separated more.

A few potential ideas:
1. Move generic methods to some kind of central reused package
2. Name methods more specifically based on use case if they stay in this file. A lot of the names don't make inherent sense on their own, like `componentIDs`, `getMap`, `nestedMap`, `metricFilterList`, etc. I'm having to go back and forth between where they're being called and their implementation to try to remember and figure out what's going on.
3. Maybe try some kind of more eager parsing into a local struct that contains only the things we care about, then operate the logical parts on that.

Just some thoughts, I'll defer to you. I'm just having a hard time understanding what the format of any given variable is at a given time, what it's purpose is, and what each of these helper methods are actually doing.

internal/configconverter/hostmetrics_logical_cpu_count_test.go
crobert-1, line 1: Ignore if I've missed something: It may be good to include a test that fails if/when upstream enables the logical CPU metric by default, since this assumes it's disabled by default. It would be good to ensure somehow that we don't miss it if/when it happens.

PR #7695, opened 2026-06-24
[chore] Bump go.mongodb.org/mongo-driver/v2 to v2.4.2

dmitryax:
Bumps go.mongodb.org/mongo-driver/v2 from v2.3.1 to v2.4.2 to address GHSA-cp6g-7hqx-qxhp.

The newer driver requires MongoDB 4.2+, so MongoDB 4.0 is removed from the discovery integration-test matrix.


pjanotti (approved):

pjanotti:
Notice that upstream updated the dependency to v2.7.0, but this change is still needed to update our tests.

PR #7700, opened 2026-06-25
feat: adjust Windows shell script to match MSI changes

weliasz:
- Remove SIGNAL_FX mandatory token validation
- Don't overwrite ENVIRONMENT registry property, keep values created by MSI update if necessary
- Pass SIGNAL_FX token into MSI to trigger for multiple configs appending


packaging/installer/install.ps1
```diff
@@ -409,18 +409,56 @@ function update_registry([string]$path, [string]$name, [string]$value) {
...
+        [string[]] $existing_env_vars = $existing_env_value
+        foreach ($entry in $existing_env_vars) {
+            $separator_index = $entry.IndexOf("=")
+            if ($separator_index -gt 0) {
+                $key = $entry.Substring(0, $separator_index)
+                $value = $entry.Substring($separator_index + 1)
+                $merged_env_vars[$key] = $value
+            }
+            else {
+                $preserved_env_entries += $entry
```
pjanotti: This is a bit too defensive: if there is no separator is not a valid env var, no? Anyway, in principle it won't break anything that wasn't already "broken". Perhaps log, as warning, that an invalid/unexpected env var was found but unless have a strong reason to preserve it I think we should drop it instead.

packaging/installer/install.ps1
```diff
@@ -695,16 +739,22 @@ if (!(Test-Path -Path "$config_path")) {
 }
 
 $collector_env_vars = @{
-    "SPLUNK_ACCESS_TOKEN"     = "$access_token";
     "SPLUNK_API_URL"          = "$api_url";
     "SPLUNK_CONFIG"           = "$config_path";
```
pjanotti: Aren't all of these covered via the MSI? Can't we remove all of them?

Notice #7701 fixing the support to `SPLUNK_CONFIG`

pjanotti (commented):
My main Q is if isn't possible to remove the merge and setting the env var from the script was a whole. It certainly will be broken for very old versions, and also perhaps the one the introduced the platform support (see #7701).

pjanotti (approved):

PR #7706, opened 2026-06-26
[chore] update release rpm checks

jinja2:

The release helper now nudges Artifactory metadata recalculation before waiting, checks whether unchanged package uploads are actually present in the live repo metadata before deciding to skip the wait, and gives metadata more time to settle before signing. For RPMs, the release job also enables synchronous metadata calculation in the `test` stage only, so we can validate that JFrog behavior before changing beta/release publishing.

This also splits metadata signature verification into deb and per-arch rpm jobs, so each check only depends on the relevant release artifacts and the GitHub release is blocked if the live metadata/signature pair does not match what the pipeline produced.

### DIFF
--- a/packaging/release/helpers/util.py
+++ b/packaging/release/helpers/util.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import gzip
 import hashlib
 import os
 import re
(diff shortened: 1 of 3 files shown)


packaging/release/helpers/util.py
```diff
@@ -136,11 +137,17 @@ def get_md5_from_artifactory(url, user, token):
     return md5
 
 
-def trigger_metadata_calculation(calculate_url, user, token, timeout=DEFAULT_TIMEOUT):
-    # Explicitly ask Artifactory to (re)calculate the repo metadata and block
-    # until it finishes (async=0).
-    print(f"Triggering synchronous metadata calculation: {calculate_url}")
-    resp = requests.post(calculate_url, auth=(user, token), timeout=timeout)
+def trigger_metadata_calculation(calculate_url, user, token, sync=False, timeout=DEFAULT_TIMEOUT):
```
pjanotti, line 140: Why default to sync if the safe default seems to be async?

pjanotti (commented):
LGTM

packaging/release/helpers/util.py
jinja2: I don't know if sync is actually available on our repo, so i added the option to try sync and enabled it for the test stage for now (we have been relying on async up until now)

PR #7708, opened 2026-06-26
Add Linux build and packaging of launcher and OpAMP Supervisor

kelseyma:
**_Moving this to draft while the launcher approach without SPLUNK_CONFIG is investigated (env var was recently removed in some cases but launcher depends on it)_** 

**Description:**
1. Added gated build targets and packaging for Linux (not releasing yet so adding behind gates for now). OpAMP Supervisor is being built locally with a helper go module using `tool` directive. Other options considered: 
- Download binaries from otel releases (security concerns with validating binaries, also didn't have all the os/arch combos)
- Build locally within only make target using `go mod download` (needs additional processing to copy over into temp dir and to remove local dir `replace` directives, can be fragile with the processing and dir handling) 
- Clone contrib repo to build from source (heavy, increases build times, seemed like it would be too much for the use case)
2. Incorporated fixes from latest OpAMP Supervisor release and for setting config apply timeout (default value was 5s, too short and would cause the supervisor to report failed remote config application even though it actually worked)

**Open questions with this build approach:**
1. Do we want this to be handled by the automatic dep updates? Should this be added into the update otel deps job?

**Testing:** Built linux packages with and without the gate
- Without the gate (default): remains the same as before, does not include any of the new supervisor/launcher binaries or dirs
- With the gate: contains launcher/supervisor items, verified deb package installs and runs with new components


agoallikmaa:
Is it acceptable that the new 1m/2m timeout defaults aren't applied if there already is a supervisor config file?

Since Windows-specific logic for `splunk-otel-collector-launcher` and `opampsupervisor` make targets was included, I verified that they do indeed build locally.

kelseyma:
> Is it acceptable that the new 1m/2m timeout defaults aren't applied if there already is a supervisor config file?

We haven't released the launcher/supervisor yet as it's still in development so I think it should be ok

agoallikmaa (approved):

PR #7709, opened 2026-06-26
[config] Generate agent configs

crobert-1:

Having one default agent config with comments specifying what should be included when the agent is sending data to just the backend vs. gateway is confusing. This change moves to have a source config file that is then used to generate both agent configs: one that is deployed with a gateway, one without.

NOTE: Still need validation and testing, this still in a POC stage

### DIFF
--- a/cmd/default-config-generator/config_templates/agent_config_source.yaml.tmpl
+++ b/cmd/default-config-generator/config_templates/agent_config_source.yaml.tmpl
@@ -1,0 +1,281 @@
+# <-- generated comment -->
+# This file is being used to generate two config files:
+#   1. agent_config.yaml: An agent config used without a gateway
+#   2. agent_to_gateway_config.yaml: An agent config used with a gateway
+#
+# How it works:
+# Specialized blocks are used to determine which comments can be dropped, or which
+# functionality is only valid for the agent to gateway or agent without gateway scenarios.
+#
+# Specialized block format:
+# <-- gateway enabled -->
+# this line should only be rendered in the agent to gateway config file
+# <-- gateway enabled end -->
+# <-- gateway disabled -->
+# this line should only be rendered in the agent to backend config file
+# <-- gateway disabled end -->
+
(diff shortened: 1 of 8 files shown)


cmd/default-config-generator/config_templates/agent_config_source.yaml.tmpl
crobert-1, line 1: This file extension will be renamed to make it clear that it's a template

cmd/default-config-generator/main.go
crobert-1, line 1: I'm planning on converting this to be a more generic top-level script that takes a CLI argument to decide which config template to work on, to make it easier to add more in the future. It could be used like `./default-config-generator --all` or with specific configs enumerated.

cmd/default-config-generator/config_templates/agent_config_source.yaml.tmpl
```diff
@@ -0,0 +1,281 @@
...
+# How it works:
+# Specialized blocks are used to determine which comments can be dropped, or which
+# functionality is only valid for the agent to gateway or agent without gateway scenarios.
+#
+# Specialized block format:
+# <-- gateway enabled -->
+# <-- gateway disabled -->
+# <-- generated comment -->
+
+# Note: Nesting blocks is not supported
```
crobert-1: I'm going to drop this rule, it was just to make parsing and validation simpler.

cmd/otelcol/config/collector/agent_to_gateway_config.yaml
crobert-1, line 1: I'm planning on dropping this file from being generated once the template formatting is finalized (before merging this PR). I'm keeping it for now to show how templating from one source file works with multiple destination files.

The actual agent to gateway config will be introduced in a follow up **with** testing included.

dmitryax:
I like the adea of just pre-generating all the configs users might need. However, why do we need to come up with a custom templating language and not use something already available, let's go templates, instead?

PR #7730, opened 2026-06-30
[chore] Improve RPM release metadata verification

jinja2:
Updates the RPM release flow to avoid signing stale Artifactory yum metadata. release-rpms now uploads the collector and auto-instrumentation RPMs for an arch as one batch, triggers synchronous yum metadata recalculation for that stage/arch, waits for both RPMs to appear in live metadata, then downloads repomd.xml for signing and uploads the returned repomd.xml.asc.

Adds stronger publish verification: metadata verification checks live repomd.xml against the published .asc using the pinned Splunk GPG fingerprint, and a new RPM install smoke job configures the public repo with repo_gpgcheck=1, runs dnf makecache, installs both RPM packages, and checks direct URL install for the auto-instrumentation RPM.

Also removes retry from the RPM release/metadata verify path to avoid replaying a partially published state, keeps the longer release timeout, and tightens quoting on the signature upload command.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,7 +15,7 @@ variables:
     description: "Set to 'true' to do a dry-run of most jobs, publishing dev images and test packages when appropriate."
   RUN_RPM_METADATA_REPAIR:
     value: "false"
-    description: "Set to 'true' on a manual main pipeline to re-sign the current release RPM repo metadata."
+    description: "Set to 'true' on a manual main web pipeline to re-sign the current release RPM repo metadata."
   GO_VERSION: 1.26.4
   WIN_2019_BASE_IMAGE: mcr.microsoft.com/windows/servercore:ltsc2019
   WIN_2022_BASE_IMAGE: mcr.microsoft.com/windows/servercore:ltsc2022
@@ -102,6 +102,8 @@ manual_release_trigger:
 
       echo "Manual release triggered for version $RELEASE_VERSION"
   rules:
+    - if: '$RUN_RPM_METADATA_REPAIR == "true"'
+      when: never
     - if: '$CI_PIPELINE_SOURCE == "web" && $RELEASE_VERSION && $CI_COMMIT_BRANCH == "main"'
       when: on_success
   allow_failure: false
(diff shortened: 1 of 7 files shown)


.gitlab-ci.yml
```diff
@@ -1266,31 +1267,40 @@ release-rpms:
...
-  retry: 2
+  timeout: 2h
   before_script:
     - *get-artifactory-stage
     - pip3 install -r packaging/release/requirements.txt
     - apt update && apt install -y curl
     - |
       set -ex
-      for path in dist/signed/*${ARCH}.rpm; do
+      shopt -s nullglob
```
pjanotti, line 1279: Doesn't the debs also need something similar to this?

.gitlab-ci.yml
```diff
@@ -1467,6 +1465,85 @@ verify-signed-metadata:
...
+        'gpgcheck=1' \
+        'repo_gpgcheck=1' \
+        'gpgkey=https://splunk.jfrog.io/splunk/otel-collector-rpm/splunk-B3CD4420.pub' \
+        'enabled=1' \
+        > /etc/yum.repos.d/splunk-otel-collector.repo
+
+      dnf clean all
+      dnf -y --disablerepo='*' --enablerepo=splunk-otel-collector makecache
+      dnf -y install "$collector_nevra" "$instrumentation_nevra"
+      rpm -q "$collector_nevra" "$instrumentation_nevra"
```
pjanotti: nit: putting these commands in a file will make easier to run stuff on dev boxes.

pjanotti: include some lines below too

.gitlab-ci.yml
```diff
@@ -1266,31 +1267,40 @@ release-rpms:
   needs:
     - job: sign-rpms
       artifacts: true
-  retry: 2
+  timeout: 2h
```
pjanotti, line 1272: this bad? any evidence that they may take this long?

pjanotti (approved):

.gitlab-ci.yml
jinja2: In recent job history, the old flow averaged about ~24 min for x86_64 and ~22 min for aarch64, with max observed times around ~39 min for x86_64 and ~44 min for aarch64. We could reduce it, but I want to pass a long timeout to the sync http call, so I'd rather start with a big enough value and then tighten it once i have a full run (the last "test" run definitely timed out with http timeout set to 45m)

.gitlab-ci.yml
jinja2: The deb repo was not changed, and we haven't seen the same issue with deb yet. So this PR has not changed anything for the deb. We should tackle deb too though after we fix rpm

.gitlab-ci.yml
jinja2: Tbh, we do have the other verify job which checks the signatures correctly but this is me being paranoid and adding the exact install cmds we ask users to run. Do you think i should remove it? or just keep it (i'll move it from here)

pjanotti: keep it

PR #7738, opened 2026-07-01
[chore] Migrate packaging/tests for installer from Python to go

pjanotti:
Migrate `packaging/tests/installer_test.py` to golang, following up on #7689


pjanotti:
I will start with a new one after merging the migration to OTel Injector

PR #7739, opened 2026-07-01
[chore] update repo signing process

jinja2:

This removes the release pipeline steps that manually send RPM/DEB repository metadata to the signing service and upload detached metadata signatures back to Artifactory. Artifactory is now expected to generate/sign repo metadata itself.
It also removes the temporary manual metadata repair jobs and restores the DEB/RPM release jobs to the normal release flow. Package signing stays unchanged.
The post-release metadata check remains, but now verifies the live Artifactory-generated signatures:

- DEB Release.gpg

- DEB InRelease

- RPM repomd.xml.asc for both x86_64 and aarch64

All are validated against the published Splunk B3CD key. The DEB/RPM release script timeout is also passed as 5400s to give Artifactory metadata calculation more room.


.gitlab-ci.yml
```diff
@@ -1296,305 +1257,65 @@ release-rpms:
...
-
-      echo "Signature OK for ${name}"
+      set -euo pipefail
+      for platform in linux/amd64 linux/arm64; do
+        echo "Verifying DEB install on ${platform} ..."
+        docker run --rm --platform "${platform}" \
+          -e STAGE="${STAGE}" \
+          -e SPLUNK_DEB_GPG_KEY_URL="https://splunk.jfrog.io/splunk/otel-collector-deb/splunk-B3CD4420.gpg" \
+          -e SPLUNK_GPG_FINGERPRINT="58C33310B7A354C1279DB6695EFA01EDB3CD4420" \
+          ubuntu:22.04 bash -s < packaging/release/verify-deb-install.sh
```
dmitryax: Why don't we keep the jobs validating signatures for now `publish-release-rpm-metadata-signature` instead of replacing it with a script that can fail in docker? If we want improve the validation, shouldn't it be out of this scope?

jinja2: The current verification is limited to just checking .asc and .gpg are signed with B3CD4420.pub . The new script runs the full install per arch and is a better smoke test I think.

> instead of replacing it with a script that can fail in docker

we will catch the exit code here if anything in the script fails.

packaging/release/release.py
```diff
@@ -18,17 +18,37 @@
...
-    if asset:
+    if len(assets) > 1:
+        print(f"Releasing the following assets to the '{args.stage}' stage:")
+        for asset in assets:
+            print(asset.path)
+
+        if not args.force:
+            resp = input("Continue? [y/N]: ")
+            if resp.lower() not in ("y", "yes"):
+                sys.exit(1)
```
dmitryax: Why do we need the interactive mode if it's supposed to run in CI?

jinja2: Only added it for consistency. The current implementation has this same check when the script is invoked without --force and single asset

dmitryax (approved):

.gitlab-ci.yml
```diff
@@ -1246,44 +1221,29 @@ release-debs:
...
+  script:
     - |
       set -ex
       for path in dist/signed/*.deb; do
         if [ ! -f "$path" ]; then
           echo "$path not found!"
           exit 1
         fi
-        python3 packaging/release/release.py --force --stage=${STAGE} --path=$path
+        python3 packaging/release/release.py --force --stage=${STAGE} --timeout=5400 --path="$path"
```
jinja2, line 1237: We have seen these steps timeout recently. Going with a large timeout for now to get a sense of timing w/o having the retry hit

.gitlab-ci.yml
```diff
@@ -1555,46 +1317,36 @@ verify-signed-metadata:
...
-          local_sig="signed/Release.asc"
+          curl -fsSL "${base}/Release" -o metadata.deb-Release
+          curl -fsSL "${base}/Release.gpg" -o metadata.deb-Release.gpg
+          curl -fsSL "${base}/InRelease" -o metadata.deb-InRelease
+
+          gpg --status-fd=1 --verify metadata.deb-Release.gpg metadata.deb-Release \
+            | grep -Eq "^\[GNUPG:\] VALIDSIG ${SPLUNK_GPG_FINGERPRINT}( |\$)" \
+            || { echo "DEB Release.gpg did not validate against pinned key ${SPLUNK_GPG_FINGERPRINT}!"; exit 1; }
+
+          gpg --status-fd=1 --verify metadata.deb-InRelease \
```
jinja2, line 1331: This change is to add InRelease check which we didn't do previously

pjanotti (approved):

PR #7741, opened 2026-07-02
Add CyberArk config source

akgrover:
**Description:** 

Adds a `cyberark` config source so operators can keep credentials out of collector config and pull them from CyberArk at config-resolution time — the equivalent of the existing `vault` config source.

Targets the **CyberArk Credential Provider (CP)** agent installed on the collector host. There's no REST endpoint; secrets are fetched by executing the provider's bundled `CLIPasswordSDK` binary, so this source is exec-based.

### Usage

```yaml
config_sources:
  cyberark:
    app_id: collector-app
    safe: DBSecrets
    object: prod-db

components:
  some_receiver:
    endpoint: ${cyberark:Address}:${cyberark:Port}
    username: ${cyberark:UserName}
    password: ${cyberark:Password}    # empty selector also returns Password
```

Each instance is pinned to a single CyberArk object (`app_id` + `safe` + `object`, optional `folder`). One CLI call retrieves the password plus pass properties (`UserName`, `Address`, `Database`, `Port`) and object metadata (`Name`, `Safe`, `Folder`); the reference selector picks the field.

### What's included
- New package `internal/configsource/cyberarkconfigsource/` (config, factory, source, CP retriever, README, testdata).
- Factory registered in `provider.go`; `cyberark` added to `customConfigSources` telemetry map.
- Unit tests (fake command runner + fake retriever, no real binary needed) and a `.chloggen` entry.

### Password rotation & collector impact

CyberArk doesn't lease credentials — an object has a stable identity and its password *value* is rotated by the CPM. There's no lease/TTL/renewal to react to. A config source resolves `${cyberark:...}` **once**, at config load, baking a literal into each component's config; the only way a rotated credential reaches a running collector is to re-resolve the config. Two modes:

- **Static (default, `auto_refresh: false`)** — fetch once, no watcher. If CyberArk rotates *after* startup, the collector keeps the value it resolved and won't notice on its own; it's refreshed only on a process restart (systemd/k8s/fatal error), which re-resolves config against the CP agent's current local cache. Best when rotation is infrequent and a restart is acceptable.
- **Auto-refresh (`auto_refresh: true` + `poll_interval`)** — a background watcher re-reads on `poll_interval` and, when values change, triggers a full collector **config reload** (same mechanism as `vault`) that re-resolves references and rebuilds affected pipelines in-process, no restart. Trade-offs: affected pipelines briefly tear down/rebuild on reload; detection latency is bounded by `poll_interval`; each poll execs `CLIPasswordSDK`; poll failures surface as a reload error rather than being swallowed.

### Design decisions

- **Selective auto-refresh.** Because a receiver can't lazily re-fetch and refresh forces a config reload, watching is **opt-in** rather than always-on — the common static case pays no goroutine/CLI cost, and operators explicitly choose the reload trade-off when rotation frequency warrants it. This also matches CyberArk's rotation (not lease) model: the correct refresh signal is "value changed," detected by re-read + compare, so there are no `Renewable`/`LeaseDuration` branches like vault has.
- **Extensible retrieval backends.** The core (`source.go`) is backend-agnostic and talks to an internal `retriever` interface (`retrieve(ctx) → map[string]any`), selected by a `retrieval_mode` field (default `cp`). Account addressing, field/selector model, caching, and the watcher are all shared, leaving a future `ccp` mode (mTLS REST to AIMWebService) as a drop-in with **no change to `source.go`**. `ccp` currently returns `errUnsupportedMode`. Conjur — a different product with dynamic secrets, TTLs, and a different API/auth model — is intentionally left as a *separate future config source* rather than a mode here.
- **Injectable exec for testability.** The CP retriever takes a `commandRunner` func (defaulted to an `os/exec` implementation), so parsing and arg-building are fully unit-tested without the real `CLIPasswordSDK` binary, which can't run in CI.
- **Strict positional parsing.** Requests a fixed, ordered `-o` field list joined by a rare delimiter (`@#@`) and parses positionally. A field-count mismatch is an error rather than risking a silent mis-map of one credential field onto another. A configurable field list is noted as a future extension.

- `go build ./...`, `gofmt`, `go vet` clean.
- `go test ./internal/configsource/cyberarkconfigsource/...` — 19/19 pass.
- `go test ./internal/confmapprovider/configsource/...` — registration + sync checks pass.
-  Real-binary path is covered by the manual smoke test

README.md file documenting the component config, data flow of the secrets, and future enhancements

### DIFF
--- a/.chloggen/add-cyberark-config-source.yaml
+++ b/.chloggen/add-cyberark-config-source.yaml
@@ -1,0 +1,20 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: config
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add a CyberArk config source (`cyberark:`) that retrieves credentials from the CyberArk Credential Provider via the local CLIPasswordSDK binary"
+
+# One or more tracking issues related to the change
+issues: []
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Each config source instance is pinned to a single CyberArk object via `app_id`, `safe`,
+  and `object`. References such as `${cyberark:Password}` and `${cyberark:UserName}` select
+  fields from the retrieved account. Set `auto_refresh: true` with a `poll_interval` to
+  trigger a collector config reload when the credential rotates.
--- a/.chloggen/add-cyberark-config-source.yaml
+++ b/.chloggen/add-cyberark-config-source.yaml
@@ -8,7 +8,7 @@ component: config
 note: "Add a CyberArk config source (`cyberark:`) that retrieves credentials from the CyberArk Credential Provider via the local CLIPasswordSDK binary"
 
 # One or more tracking issues related to the change
-issues: []
+issues: [7741]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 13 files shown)


akgrover:

akgrover:
cc @atoulme

atoulme (approved):

PR #7746, opened 2026-07-03
instrumentation: scope auto-discovery to allowed executable paths and working dirs

technimad-splunk:
Adds two optional `libsplunk` config keys so operators can limit auto-instrumentation to processes whose executable path and/or working directory fall under an allow-list of directories, instead of instrumenting every java/node/dotnet process.

- `SPLUNK_INSTRUMENTATION_ALLOWED_EXECUTABLE_PATHS` and `SPLUNK_INSTRUMENTATION_ALLOWED_WORKING_DIRS` in the per-language `/etc/splunk/zeroconfig/*.conf` files; `:`-separated absolute directories, directory-prefix match.
- Each set key is an AND constraint; an unset key imposes no constraint. Config entries are canonicalized and guarded against path traversal (non-absolute / `..` entries are ignored).
- Backwards compatible: with neither key set, all java/node/dotnet processes are still instrumented, so existing installations are unaffected.

## Test plan
- `cd instrumentation && make all && make test-java`
- New java test cases assert injection occurs for a matching allowed path and is skipped for a non-matching one, including `..` traversal that canonicalizes away from (skipped) and back onto (allowed) the executable's directory.

Assisted-by: Cursor Composer 2.5


technimad-splunk:
For reference, the now included upstream [opentelemetry injector](https://github.com/open-telemetry/opentelemetry-injector) supports environment variables achieving similar functionality as proposed in this PR:
```
OTEL_INJECTOR_INCLUDE_PATHS: a comma-separated list of glob patterns to match executable paths
OTEL_INJECTOR_EXCLUDE_PATHS: a comma-separated list of glob patterns to exclude executable paths
OTEL_INJECTOR_INCLUDE_WITH_ARGUMENTS: a comma-separated list of glob patterns to match process arguments
OTEL_INJECTOR_EXCLUDE_WITH_ARGUMENTS: a comma-separated list of glob patterns to exclude process arguments
```

PR #7767, opened 2026-07-09
Update launcher to parse config flags and include additional supervisor configs

kelseyma:
**Description:**
Launcher (entrypoint for toggling between direct collector vs supervisor mode) parses configs flags/env, removes splunk opamp extension config, and passes the files to OpAMP Supervisor. Supervisor will merge configs and handle `service.extensions` list appending, which covers the req from the new splunk platform log/metric configs. This merge will not handle `pipelines.*.receivers/exporters` appending, but is an acceptable limitation from discussions with team (better merge support is being discussed upstream, will adopt when available).
1. Launcher updates to support `--config` flags and add flexibility (only affects supervisor mode)

Before | After | Reason
-- | -- | --
SPLUNK_CONFIG is required as value used for collector config | Parse `--config` flags, fallback to SPLUNK_CONFIG if no config flags. Reject non-file providers in supervisor mode (acceptable limitation from discussions). `enableMergeAppendOption` feature gate removed as it's a no-op in supervisor mode (supervisor owns configs and does merging) | SPLUNK_CONFIG is no longer present in certain install cases (all Windows, Linux unified ingestion)
collector_config.yaml created and passed to supervisor as sanitized version of SPLUNK_CONFIG value | Loop through configs and create a managed copy if `opamp/splunk_o11y` extension is seen in config. Naming scheme: `managed_collector_<filename><suffix-if-needed>.yaml` (e.g. managed_collector_agent_config.yaml) | Adds flexibility for multiple configs and ordering of configs |
supervisor `server` config pulled from `SPLUNK_CONFIG` | Loop through configs, last complete compatible `opamp/splunk_o11y.server.http.endpoint` block used | Adds flexibility for multiple configs and ordering of configs
`supervisor_config.yaml` created as one time migration output, allow user changes to persist (similar to agent_config.yaml) | `supervisor_config.yaml` still created once and changes preserved. New file `supervisor_runtime_config.yaml` created on start so launcher can own `agent.executable`, `agent.config_files`, and `agent.args` | Allows service args/config changes to be reflected without overwriting the user changes in `supervisor_config.yaml`

2. Updates to OpAMP Supervisor config:
- Include collector resource attributes
- Set log format to console for readability (supervisor default was json, but collector default is console)
- Set config apply timeout (default value was 5s, too short and would cause the supervisor to report failed remote config application even though it actually worked)

1. Install collector (direct mode, no flags) ==> service starts, same as existing behavior
2. Update conf to enable supervisor ==> all expected supervisor files created, supervisor & collector running
4. Update to pass flags and edit `supervisor_config.yaml` ==> config persisted, flags forwarded 
5. Update to pass `--config` flag pointing to gateway config ==> `managed_collector_gateway_config.yaml` created, all flags except config flag forwarded
6. Update config flag to use unsupported provider type ==> service fails to start, error seen in logs
7. Fresh install with unified ingestion + supervisor mode ==> flags filtered (none forwarded as only had config and merge append feature gate), supervisor `config_files` list show managed agent_config, logs config, metrics config
8. Swap ordering of config files, add another feature gate ==> `config_files` list reflects updated order, feature gate flag forwarded (without merge append)

### DIFF
--- a/internal/opampsupervisor/launcher/config.go
+++ b/internal/opampsupervisor/launcher/config.go
@@ -15,12 +15,16 @@
 package launcher
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
+	"slices"
 	"strings"
+	"text/template"
 
 	"gopkg.in/yaml.v3"
 )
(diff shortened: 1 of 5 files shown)


dmitryax (approved):
LGTM

PR #7770, opened 2026-07-10
[chore] Add splunkprivate build tag to allow private components

dmitryax:
A build-flag-gated collector build that includes the closed-source components from [signalfx/splunk-otel-collector-components](https://github.com/signalfx/splunk-otel-collector-components) (`s2sexporter`, `linebreakprocessor`, `wineventlogreceiver`). Follows on from #7674.

- `internal/components/components_splunkprivate.go` `//go:build splunkprivate`; imports the three private modules and populates package-level factory slices via `init()`.
- `internal/components/components.go` declares nil-default `private*Factories` slices and merges them into the returned `otelcol.Factories` map with duplicate detection. Non-tagged builds get an empty merge (no-op).
- `Makefile` `otelcol-splunkprivate` target is a one-liner wrapping `make otelcol` with `GOFLAGS=-tags=splunkprivate`.
- `.github/workflows/build-and-test.yml` new `splunkprivate-build` job: single Ubuntu runner, builds with the tag, that's it. Intentionally not extended to cross-compile / package / vuln-scan matrices.

Also removed `private-modules: 'true'` from jobs that only exercise default-tag Go code (gofmt, generate-metrics, lint, coverage, latest-dockerd-test, reusable-compile, reusable-unit-test, lint-examples, nomad, ta-v2, integration-test, vuln-scans docker/govulncheck, windows-test). Kept for `tidy`, which scans every file regardless of tag. Dropped now-unused `secrets: inherit` from the corresponding reusable workflow callers.

### DIFF
--- a/.github/workflows/auto-instrumentation.yml
+++ b/.github/workflows/auto-instrumentation.yml
@@ -33,7 +33,6 @@ jobs:
     uses: ./.github/workflows/reusable-compile.yml
     with:
       SYS_BINARY: ${{ matrix.SYS_BINARIES }}
-    secrets: inherit
 
   build-package:
     runs-on: ${{ matrix.ARCH == 'amd64' && 'ubuntu-24.04' || 'ubuntu-24.04-arm' }}
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -39,9 +39,6 @@ jobs:
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: ./.github/actions/setup-environment
-        with:
-          private-modules: 'true'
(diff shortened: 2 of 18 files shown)


.github/workflows/auto-instrumentation.yml
```diff
@@ -33,7 +33,6 @@ jobs:
     uses: ./.github/workflows/reusable-compile.yml
     with:
       SYS_BINARY: ${{ matrix.SYS_BINARIES }}
-    secrets: inherit
```
akhileshsingh85, line 36: Will the test break if this line is not removed? There are other files which got merged in main with field "secrets: inherit". For example puppet_test.yml file

dmitryax: We are not passing the private build it to the puppet yet. I just wanted to keep it isolated for now and expand when needed

akhileshsingh85 (approved):

PR #7771, opened 2026-07-10
Add Linux and Windows building of launcher and OpAMP Supervisor

kelseyma:
**Description:**
Added gated build targets for Linux and Windows (not releasing yet so adding behind gates for now). No changes added yet for packaging, those will follow in a later PR along with signing of binaries.

OpAMP Supervisor is being built locally and added to `internal/tools` as a required tool. This should allow `update-otel-deps` to include it in the update job. Other options considered: 
- Download binaries from otel releases (security concerns with validating binaries, also didn't have all the os/arch combos)
- Build locally within only make target using `go mod download` (needs additional processing to copy over into temp dir and to remove local dir `replace` directives, can be fragile with the processing and dir handling) 
- Clone contrib repo to build from source (heavy, increases build times, seemed like it would be too much for the use case)

**Testing:** 
- Without the gate (default): remains the same as before, `bin/` does not include any of the new supervisor/launcher binaries
- With the gate: `bin/` contains launcher and supervisor binaries


Makefile
```diff
@@ -269,6 +270,24 @@ else
 	$(LINK_CMD) otelcol_$(GOOS)_$(GOARCH)$(EXTENSION) ./bin/otelcol$(EXTENSION)
 endif
 
+.PHONY: splunk-otel-collector-launcher
+splunk-otel-collector-launcher:
+	GO111MODULE=on CGO_ENABLED=$(CGO_ENABLED) go build -trimpath -o ./bin/splunk-otel-collector-launcher_$(GOOS)_$(GOARCH)$(EXTENSION) $(BUILD_INFO) ./cmd/splunk-otel-collector-launcher
```
dmitryax: otelcollauncher maybe? I'm not sure about dashes. We don't have them anywhere else

Makefile
```diff
@@ -269,6 +270,24 @@ else
...
+ifeq ($(OS), Windows_NT)
+	$(LINK_CMD) .\bin\splunk-otel-collector-launcher$(EXTENSION) .\bin\splunk-otel-collector-launcher_$(GOOS)_$(GOARCH)$(EXTENSION)
+else
+	$(LINK_CMD) splunk-otel-collector-launcher_$(GOOS)_$(GOARCH)$(EXTENSION) ./bin/splunk-otel-collector-launcher$(EXTENSION)
+endif
+
+.PHONY: opampsupervisor
+opampsupervisor:
+	cd packaging/opampsupervisor && GO111MODULE=on CGO_ENABLED=$(CGO_ENABLED) go build -trimpath -o "$(CURDIR)/bin/opampsupervisor_$(GOOS)_$(GOARCH)$(EXTENSION)" github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor
```
dmitryax: I agree we should build it ourself (and sign btw), but why do we need the local go module? We can pin the contrib go tag and fetch it on demand. The go tag (maybe env var in Makefile) must be always in sync with the contrib deps we pull for the collector build (this can be enforced by the CI)

kelseyma: @dmitryax I tried a few different methods for building locally and each had some drawbacks/issues:
1. `go install`: didn't work because opamp supervisor go.mod has [replace directives](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/cmd/opampsupervisor/go.mod#L189-L309)
2. `go mod download`/`go list` to fetch module: needs processing to remove all of the replace directives from the go.mod (first `jq` parsing to get all of the local dir replaces, then `go mod edit` to drop replaces), concern was with the processing being fragile
3. `go get`: changes to collector's go.mod, concern was with the dependencies being linked and changing what collector/supervisor should be built with
4. temp go module: similar to PR's local mod but created in a temp dir, go.mod just requires opampsupervisor and deleted after build, concern was if we wanted traceability/visibility of the local module
6. clone entire contrib: concern was with increased build times and seemed like it would be too much for the use case

Definitely open to other suggestions if you have, or if any of these alternative options are preferred! I was originally thinking of option 2, but wasn't sure if all the processing was desired. For options 2 and 4, the supervisor version can be based on opampextension dep to keep it automatically in sync with rest of contrib deps. For the current local go module, I think we would have to add it to the `update-deps` script to sync it.

Makefile
kelseyma: good point, that name is also cleaner with matching `otelcol`. thanks, updated!

packaging/fpm/common.sh
```diff
@@ -89,18 +96,49 @@ setup_files_and_permissions() {
...
+
+        cp -f "$launcher" "$buildroot/$LAUNCHER_INSTALL_PATH"
+        sudo chown root:root "$buildroot/$LAUNCHER_INSTALL_PATH"
+        sudo chmod 755 "$buildroot/$LAUNCHER_INSTALL_PATH"
+
+        cp -f "$opampsupervisor" "$buildroot/$OPAMPSUPERVISOR_INSTALL_PATH"
+        sudo chown root:root "$buildroot/$OPAMPSUPERVISOR_INSTALL_PATH"
+        sudo chmod 755 "$buildroot/$OPAMPSUPERVISOR_INSTALL_PATH"
+
+        mkdir -p "$buildroot/$SUPERVISOR_STORAGE_PARENT_INSTALL_PATH"
```
kelseyma: For `/var/lib/otelcol` dir creation, I had this here to not affect any prod/user facing files but on further thought realistically this should probably go in both `install.sh` and `postinstall.sh` files. Will remove this as the user facing changes will come later when we are ready to officially release

**edit: since there are more temp workarounds in common.sh (like overriding the ExecStart) and realizing some concerns about custom user support, leaning towards reducing this PR to just the launcher/supervisor make targets and remove fpm packaging side changes until we are ready to release**

Makefile
dmitryax: I'm surprised replaces causing problem for downstream consumers. Please look at how collector's tools [pulled in contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/47b5da2e5c81f8da0eb470a8ce94961425fa0208/Makefile.Common#L73). They also have replaces see [mdatagen for example](https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/mdatagen/go.mod). It works fine there. Let's see if we can adopt the same model

dmitryax: Problem with the approach in the PR is that go.mod has to be kept in sync

kelseyma: I considered that approach as well, but wasn't sure if we are ok with the supervisor dependencies getting mixed in with the rest of the tools. It will mean that:
1. Supervisor or the other tools can potentially be built with differing versions from upstream
2. If there are conflicting versions during upgrade, we will need to manually resolve them as `update-deps` will fail (for example, I had to set the `genproto` version as golangci-lint depended on a version of viper that used an older monolithic version of genproto vs supervisor needing a newer version. During the initial `go mod tidy`, go found identical packages in both the monolithic and new genproto deps and reported ambiguous imports.)

However, overall it does seem like adding it in tools would simplify the building and make syncing the go.mod easier because the `update-deps` job should handle most cases automatically. I updated the PR to reflect this approach, let me know if the tradeoffs are ok.

kelseyma: > From discussion: ok with approach as long as supervisor binary doesn't end up with extra unnecessary deps from other tools and larger size as a result

Ran `go version -m bin/opampsupervisor_linux_amd64` and verified that it doesn't include any extra deps from the other tools, also compared size with supervisor locally built from contrib repo `release/v0.156.x` branch (sizes were basically the same: 31.8665 MiB vs 31.8663 MiB)

dmitryax (approved):
LGTM 👍

PR #7772, opened 2026-07-10
[config generation] POC for GO templates

crobert-1:

This is just a rough draft to get an idea of what config generation may look like with golang templates

### DIFF
--- a/cmd/default-config-generator/config_templates/agent_config_source.yaml.tmpl
+++ b/cmd/default-config-generator/config_templates/agent_config_source.yaml.tmpl
@@ -199,9 +199,7 @@ service:
     metrics:
       receivers: [host_metrics, otlp]
       processors: [memory_limiter, batch, resourcedetection]
-      exporters: [signalfx]
-      # Use instead when sending to gateway
-      #exporters: [otlp_grpc/gateway]
+      exporters: [{{- .MetricsExporter.Name -}}]
     metrics/internal:
       receivers: [prometheus/internal]
       processors: [memory_limiter, batch, resourcedetection]
@@ -232,5 +230,3 @@ service:
       {{- else -}}
       {{- .OTLPGenericExporter.Name -}}
       {{- end -}}]
-      # Use instead when sending to gateway
-      #exporters: [otlp_grpc/gateway]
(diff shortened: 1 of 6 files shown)


cmd/default-config-generator/main.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+	"text/template"
+)
+
+type AgentTemplateDestination struct {
```
crobert-1, line 41: Work is still needed on the layout here, proper organization/structure is key to how easy this will be to modify and add new generated config destinations.

cmd/default-config-generator/main.go
```diff
@@ -0,0 +1,165 @@
...
+	OTLPGenericExporter       ComponentConfig
+	ProfilingExporter         ComponentConfig
+	OpAmp                     TelemetryDestination
+	SplunkAPI                 TelemetryDestination
+	SplunkIngest              TelemetryDestination
+	ConfigDestinationFilePath string
+}
+
+func main() {
+	agentConfigs := []AgentTemplateDestination{
```
crobert-1, line 55: It'd probably be cleaner to pull out struct instantiation into separate methods, but with plain strings included, it will be pretty messy either way.

cmd/default-config-generator/config_templates/agent_config_source.yaml.tmpl
```diff
@@ -0,0 +1,225 @@
...
+        - resourcedetection
+      #- resource/add_environment
+      exporters: [
+      {{- if or .LogsExporter.Name .ProfilingExporter.Name -}}
+        {{- if .LogsExporter.Name -}}{{ .LogsExporter.Name }}{{- end -}}
+        {{- if and .LogsExporter.Name .ProfilingExporter.Name -}}, {{ end -}}
+        {{- if .ProfilingExporter.Name -}}{{ .ProfilingExporter.Name }}{{- end -}}
+      {{- else -}}
+      {{- .OTLPGenericExporter.Name -}}
+      {{- end -}}]
```
crobert-1, line 225: This is an example of the main problem with this go template approach: We rely on specific logic that accounts for all situations today, but there's no clear way to accommodate future changes.

PR #7781, opened 2026-07-13
Fix RabbitMQ discovery defaults

agoallikmaa:
**Description:**
The default RabbitMQ discovery config does not work because of 2 reasons:

- Passing `host:port` format endpoint to a config field expecting an URL with schema
  - Parsed in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.155.0/receiver/rabbitmqreceiver/config.go#L48 using `url.Parse` which produces `first path segment in URL cannot contain colon` error if schema is not provided
- AQMP endpoint being passed instead of the HTTP management endpoint because the rule matched both

**Testing:**
Verified in a minimal local kind cluster setup with RabbitMQ deployed and collector discovery enabled that it encountered an error in logs without overrides, but not when using the following overrides:
```
agent:
  discovery:
    properties:
      receivers:
        rabbitmq:
          rule:
            k8s_observer: type == "port" and port == 15672 and pod.name matches "(?i)rabbitmq.*"
          config:
            default:
              endpoint: 'http://`endpoint`'
```

### DIFF
--- a/cmd/discoverybundler/metadata/receivers/rabbitmq.yaml
+++ b/cmd/discoverybundler/metadata/receivers/rabbitmq.yaml
@@ -3,12 +3,12 @@ service_type: rabbitmq
 properties_tmpl: |
   enabled: true
   rule:
-    docker_observer: type == "container" and any([name, image, command], {# matches "(?i)rabbitmq.*"}) and not (command matches "splunk.discovery")
-    host_observer: type == "hostport" and command matches "(?i)rabbitmq.*" and not (command matches "splunk.discovery")
-    k8s_observer: type == "port" and pod.name matches "(?i)rabbitmq.*"
+    docker_observer: type == "container" and port == 15672 and any([name, image, command], {# matches "(?i)rabbitmq.*"}) and not (command matches "splunk.discovery")
+    host_observer: type == "hostport" and port == 15672 and command matches "(?i)rabbitmq.*" and not (command matches "splunk.discovery")
+    k8s_observer: type == "port" and port == 15672 and pod.name matches "(?i)rabbitmq.*"
   config:
     default:
-      endpoint: '`endpoint`'
+      endpoint: 'http://`endpoint`'
       username: {{ defaultValue }}
       password: {{ defaultValue }}
       collection_interval: 10s
--- a/tests/receivers/rabbitmq/bundled_test.go
+++ b/tests/receivers/rabbitmq/bundled_test.go
@@ -1,0 +1,81 @@
+// Copyright Splunk, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build discovery_integration_rabbitmq
+
+package tests
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+
+	"github.com/signalfx/splunk-otel-collector/tests/testutils"
+)
+
+func TestRabbitMQDockerObserver(t *testing.T) {
+	testutils.SkipIfNotContainerTest(t)
+	dockerSocketProxy, err := testutils.CreateDockerSocketProxy(t)
+	require.NoError(t, err)
+
+	tc := testutils.NewTestcase(t)
+	defer tc.PrintLogsOnFailure()
+	defer tc.ShutdownOTLPReceiverSink()
+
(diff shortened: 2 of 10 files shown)


cmd/discoverybundler/metadata/receivers/rabbitmq.yaml
```diff
@@ -3,12 +3,12 @@ service_type: rabbitmq
 properties_tmpl: |
   enabled: true
   rule:
-    docker_observer: type == "container" and any([name, image, command], {# matches "(?i)rabbitmq.*"}) and not (command matches "splunk.discovery")
-    host_observer: type == "hostport" and command matches "(?i)rabbitmq.*" and not (command matches "splunk.discovery")
-    k8s_observer: type == "port" and pod.name matches "(?i)rabbitmq.*"
+    docker_observer: type == "container" and port == 15672 and any([name, image, command], {# matches "(?i)rabbitmq.*"}) and not (command matches "splunk.discovery")
```
pjanotti, line 6: Is it really necessary to hard code the default port here?

cmd/discoverybundler/metadata/receivers/rabbitmq.yaml
```diff
@@ -3,12 +3,12 @@ service_type: rabbitmq
...
+    host_observer: type == "hostport" and port == 15672 and command matches "(?i)rabbitmq.*" and not (command matches "splunk.discovery")
+    k8s_observer: type == "port" and port == 15672 and pod.name matches "(?i)rabbitmq.*"
   config:
     default:
-      endpoint: '`endpoint`'
+      endpoint: 'http://`endpoint`'
```
pjanotti, line 11: Is it possible to chose `http://` vs `https://` via some other criteria? An example for [nginx discovery](https://github.com/signalfx/splunk-otel-collector/blob/d0498bb0500c9514fa3650eb6cc64e73d2d0deae/cmd/discoverybundler/metadata/receivers/nginx.yaml#L11)

pjanotti (commented):
One big ask: is it possible to add an integration test for discovery as done for others? See https://github.com/signalfx/splunk-otel-collector/blob/d0498bb0500c9514fa3650eb6cc64e73d2d0deae/.github/workflows/integration-test.yml#L326

cmd/discoverybundler/metadata/receivers/rabbitmq.yaml
agoallikmaa: When I checked the errors it first encountered without the `http://` prefix fix, it also tried to create an observer for the 5672 port endpoint, which is AMQP and would not work:
```
error	receivercreator@v0.153.0/observerhandler.go:229	failed to start receiver	{"resource": {"service.instance.id": "9e2544e0-a77c-41f3-a926-d09c034cc1e7", "service.name": "otel-agent", "service.version": "v0.153.0", 
"splunk_autodiscovery": "true"}, "otelcol.component.id": "discovery/k8s_observer", "otelcol.component.kind": "receiver", "otelcol.signal": "logs", "receiver": "rabbitmq", "error": "invalid runtime receiver config: receivers::rabbitmq: \"endpoint\" must be 
in the form of <scheme>://<hostname>:<port>: parse \"10.244.0.2:5672\": first path segment in URL cannot contain colon"}
```
I don't really know of a better way to use rules to exclude that but not the management http endpoint.

cmd/discoverybundler/metadata/receivers/rabbitmq.yaml
agoallikmaa: In theory it might be possible to configure RabbitMQ to use TLS the official documentation suggests doing that on a different port. There does not seem to be any hardcoded defaults for that, but the documentation shows an example of using 15671 for HTTPS. But I'm not sure if that example alone is enough to consider port 15671 as a sign to try `https://` instead.

agoallikmaa:
@pjanotti I was originally planning on adding the discovery test to chart repository (the same way it has for redis currently), but it would actually work here too. Will try to add.

cmd/discoverybundler/metadata/receivers/rabbitmq.yaml
pjanotti: Sounds reasonable, we may have to tweak these defaults later, but given the available information it seems reasonable.

cmd/discoverybundler/metadata/receivers/rabbitmq.yaml
pjanotti: Sounds reasonable, we may have to tweak these defaults later but given the available information it seems reasonable.

agoallikmaa:
@pjanotti Integration test is also present now.

tests/receivers/rabbitmq/bundled_test.go
```diff
@@ -0,0 +1,81 @@
...
+	defer tc.ShutdownOTLPReceiverSink()
+
+	_, shutdown := tc.SplunkOtelCollectorContainer("otlp_exporter.yaml", func(c testutils.Collector) testutils.Collector {
+		cc := c.(*testutils.CollectorContainer)
+		cc.Container = cc.Container.WillWaitForLogs("Everything is ready")
+		return cc.WithEnv(map[string]string{
+			"SPLUNK_DISCOVERY_LOG_LEVEL": "debug",
+		}).WithArgs(
+			"--discovery",
+			"--set", `splunk.discovery.extensions.k8s_observer.enabled=false`,
```
pjanotti, line 47: not that it matters much, but this isn't necessary, right?

pjanotti (approved):

tests/receivers/rabbitmq/bundled_test.go
agoallikmaa: Not for the test to pass, just less unrelated logs during the test then.

PR #7783, opened 2026-07-13
feat: add rollingspanlatencyprocessor

rnterbush:
## Summary

- Adds `pkg/processor/rollingspanlatencyprocessor`, a new OTel Collector processor that maintains per-span-name EWMA latency baselines and appends a configurable span attribute (`latency.category` by default) with value `slow` or `very_slow` when a span's duration is statistically anomalous relative to its own rolling history.
- Wires the processor into `internal/components/components.go` and the root `go.mod`.

## Design

**EWMA baseline** — Each (resource key tuple, span name) pair maintains an exponentially weighted mean and variance. The decay factor is time-aware: `alpha = 1 - exp(-ln2 * dt / half_life)`, so the effective weight of any sample halves every `half_life` regardless of observation frequency.

**Per-span clock** — The span's own `EndTimestamp` is used as the EWMA clock, not a batch-shared wall clock. This prevents `dt=0` for all but the first span in a batch, which would collapse `alpha` to 0 and leave variance near-zero.

**Pre-update snapshot** — A span is scored against the baseline *before* the current span is folded in, preventing a single outlier from inflating its own stddev and masking its own anomaly.

**Warmup gate** — No attribute is written until `warmup_count` observations (default 30) have been accumulated, preventing false positives at startup or after eviction.

**`min_stddev` floor** — Guards against near-zero variance with very uniform traffic or slow half-lives. Defaults to 1ms (1_000_000 ns).

**Composite key** — Resource attribute values (`service.namespace`, `service.name`, `deployment.environment.name` by default) concatenated with the span name, giving independent baselines per service/environment/operation combination.

**Eviction** — Background goroutine sweeps the stats map on a configurable interval, removing entries idle longer than `idle_timeout`. A churn warning is logged when turnover exceeds `churn_warning_ratio`. A `max_baselines` cap prevents unbounded memory growth.

**OTel metrics** — `rollingspanlatency_active_baselines` (gauge) and `rollingspanlatency_dropped_keys_total` (counter).

## Test plan

- [x] Unit tests pass: `go test ./pkg/processor/rollingspanlatencyprocessor/...`
- [x] Root module builds: `go build ./internal/components/...`
- [x] Config load test covers both default and custom config variants
- [x] Processor tests cover: warmup gate, normal/slow/very_slow labeling, independent baselines per service/namespace/environment, eviction, max_baselines cap, churn counter reset, demo scenario (normal spans never mislabeled after outliers feed back into baseline)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rnterbush:

atoulme:
Please run `make tidy-all`, `make fmt` and see what `make lint` says, that'll help with review.

pkg/processor/rollingspanlatencyprocessor/internal/metadata/generated_status.go
crobert-1, line 1: Usually this file is generated by `mdatagen`, based on the contents of a top level `metadata.yaml` file under the processor directory. I'm not seeing the `metadata.yaml` file, please add one with the relevant information included.

.chloggen/rolling-span-latency-processor.yaml
```diff
@@ -0,0 +1,20 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: rollingspanlatencyprocessor
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `rollingspanlatency` processor that appends a `latency.category` span attribute (`slow` or `very_slow`) when a span's duration is statistically anomalous relative to its own rolling EWMA baseline.
```
crobert-1: There's been a recent effort upstream for all component names to be snake case, so I believe we need to update the component name to `rolling_span_latency`. (The filenames and directory are fine the way they are, but programmatic references should be updated.)

crobert-1 (commented):
Please add a component README under the processor's top-level directory. The main things to include are the processor's description/purpose, a component stability table ([reference](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a7db2f09d562681dccb2004b4606d519bfe68cd6/processor/resourcedetectionprocessor/README.md#L3), but only include information relevant to our repo here), and configuration details. The configuration should include what options are available, which ones are required/optional, and a full example of what a common configuration looks like as well.

pkg/processor/rollingspanlatencyprocessor/ewma.go
```diff
@@ -0,0 +1,88 @@
...
+//
+// where dt is the elapsed time since the last observation. This makes the
+// effective weight of any sample halve every halfLife regardless of observation
+// frequency.
+type spanStats struct {
+	lastSeen time.Time
+	mean     float64
+	variance float64
+	count    int64
+	mu       sync.Mutex
```
crobert-1: A lot of variable references are just to read values, I think we could make this much performance-wise if we used a [ RWMutex](https://pkg.go.dev/sync#RWMutex) instead.

pkg/processor/rollingspanlatencyprocessor/factory.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/processor"
+
+	"github.com/signalfx/splunk-otel-collector/pkg/processor/rollingspanlatencyprocessor/internal/metadata"
+)
+
+// NewFactory returns the OTel processor factory for rollingspanlatency.
```
crobert-1: We should try to remove boiler plate comments as much as possible. Please try to only keep comments that are really valuable, and aren't obvious.

pkg/processor/rollingspanlatencyprocessor/config.go
```diff
@@ -0,0 +1,146 @@
...
+	// exceeded by a single eviction sweep's evicted count, triggers a warning
+	// log indicating high key churn. For example, 0.5 warns when more than 50%
+	// of active baselines were replaced within one eviction interval. Default: 0.5.
+	ChurnWarningRatio float64 `mapstructure:"churn_warning_ratio"`
+
+	// MinStddev is the minimum standard deviation (in nanoseconds) used when
+	// scoring a span. When the EWMA stddev falls below this floor the floor is
+	// used instead, preventing near-zero variance from turning small OS jitter
+	// into false positives. Default: 1ms (1_000_000 ns).
+	MinStddev float64 `mapstructure:"min_stddev"`
```
crobert-1: Why not just use `time.Duration` as the type here, vs. requiring unit to be in `ns`?

rnterbush:
@crobert-1 Thanks for the review. I've addressed your feedback. Let me know if there's any other changes you'd recommend

atoulme (approved):

PR #7786, opened 2026-07-14
[chore] Remove prometheus_simple references from discovery mode

crobert-1:

Since the upstream `prometheus_simple` receiver is unmaintained, it's a good time to update documentation and tests to move to the `prometheus` receiver. Existing default discovery functionality already uses `prometheus`, so this is a documentation and test change, only, at this time.


tests/general/discoverymode/testdata/host-observer-config.d/receivers/internal-prometheus.discovery.yaml
```diff
@@ -1,13 +1,21 @@
...
+      config:
+        scrape_configs:
+          - job_name: 'prometheus'
+            scrape_interval: invalid
+            scrape_timeout: invalid
     host_observer:
-      collection_interval: 1s
-      labels:
-        label_two: ${LABEL_TWO_VALUE}
+      config:
```
crobert-1, line 12: I've removed testing here around "configuration priority" due to a few reasons:
1. `label_three`: Discovery CLI properties (`--set variable`) can't handle the format of the prometheus receiver as it would require special characters ([]) for lists (scrape configs and static configs).
2. `label_one`: koanf merge of the Prometheus receiver's default and host observer configurations result in different behavior than what happens with the Simple Prometheus receiver. There are known issues in koanf that have no path forward (see https://github.com/knadh/koanf/issues/299 and issues linked there), so this is a known issue at this time. This has no impact on our default discovery at this time. The impact is that `default` with different overlays (eg host_observer) won't work for the Prometheus receiver, when it would have with `prometheus_simple`. Users will have to define the full configuration for the Prometheus receiver for each observer (if there are differences), without being able to re-use `default`.

tests/general/discoverymode/docker_observer_discovery_test.go
```diff
@@ -72,15 +72,11 @@ func TestDockerObserver(t *testing.T) {
 				"DOCKER_DOMAIN_SOCKET":       fmt.Sprintf("tcp://%s", dockerSocketProxy.ContainerEndpoint),
 				"LABEL_ONE_VALUE":            "actual.label.one.value",
 				"LABEL_TWO_VALUE":            "actual.label.two.value",
-				"SPLUNK_DISCOVERY_RECEIVERS_prometheus_x5f_simple_CONFIG_labels_x3a__x3a_label_x5f_three": "overwritten by --set property",
```
crobert-1, line 75: This is already being tested by [template functions tests](https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/discoverybundler/templatefunctions_test.go) and [env var tests](https://github.com/signalfx/splunk-otel-collector/blob/main/internal/confmapprovider/discovery/properties/env_var_test.go) so I think it's safe to remove this test of functionality without adding a replacement at this time. (Removing because as explained in another comment, it doesn't work properly with the Prometheus receiver.)

tests/general/discoverymode/testdata/docker-observer-config.d/properties.discovery.yaml
```diff
@@ -1,6 +1,3 @@
-splunk.discovery.receivers.prometheus_simple.config.labels::label_three: file contents should be disregarded
```
crobert-1, line 1: This is also already being tested in the same places as the other templating logic and env var tests, so I don't think we need to add another test for this.

tests/general/discoverymode/testdata/docker-observer-config.d/receivers/internal-prometheus.discovery.yaml
```diff
@@ -1,15 +1,22 @@
-prometheus_simple:
-  enabled: false # overwritten by cli and env var properties
+prometheus/test:
+  enabled: true
```
crobert-1: `--set` and `SPLUNK_DISCOVERY` environment variables don't work with forward slashes in the receiver name, so it can't be tested here. Other tests (as pointed out in other comments) already exercise this functionality, so I don't think a replacement is needed.

crobert-1 (commented):
I'm leaving a review of all the things that are different using the Prometheus receiver for discovery vs. the simple Prometheus receiver.

examples/prometheus-federation/otel-collector-config.yml
```diff
@@ -1,13 +1,18 @@
...
+            scrape_interval: 10s
+            scrape_timeout: 10s
+            # the federation endpoint:
+            # Read more about it here: https://prometheus.io/docs/prometheus/latest/federation/
+            # You can query the federation with PromQL, encoded as part of the query string.
+            static_configs:
+              - targets: [prometheus:9090]
+            metrics_path: /federate
+            params:
+              match[]: '{job="counter"}'
```
jkoronaAtCisco: params field is typed as `map[string][]string`, so it should be:
```yaml
params:
    match[]: ['{job="counter"}']
```

crobert-1: Thanks @jkoronaAtCisco, updated.

tests/general/discoverymode/docker_observer_discovery_test.go
jinja2: Are these tests also exercising receiver-property application/precedence end to end?

I think we might need to keep least one receiver-scoped override in this test. It doesn’t need to modify scrape_configs; toggling `prometheus/test.enabled` should preserve coverage of the receiver path i think

internal/confmapprovider/discovery/README.md
```diff
@@ -222,8 +223,9 @@ SPLUNK_DISCOVERY_RECEIVERS_receiver_x2d_type_x2f_receiver_x2d_name_ENABLED=<true
 SPLUNK_DISCOVERY_EXTENSIONS_observer_x2d_type_x2f_observer_x2d_name_ENABLED=<true or false>
 
 # Examples
-SPLUNK_DISCOVERY_RECEIVERS_prometheus_simple_CONFIG_labels_x3a__x3a_my_label="my_username"
-SPLUNK_DISCOVERY_RECEIVERS_prometheus_simple_ENABLED=true
+SPLUNK_DISCOVERY_RECEIVERS_oracledb_ENABLED=true
+SPLUNK_DISCOVERY_RECEIVERS_oracledb_CONFIG_service="my_username"
```
jinja2: nit
```suggestion
SPLUNK_DISCOVERY_RECEIVERS_oracledb_CONFIG_service="service_name"
```

tests/general/discoverymode/docker_observer_discovery_test.go
crobert-1: I've re-added the `--set splunk.discovery.receivers.prometheus/test.enabled=true` functionality here to make sure it takes precedence over the discovery config file's value. 👍

jinja2 (approved):

PR #7790, opened 2026-07-16
fix: align install script with current MSI version

weliasz:
Adjust install script to match new Splunk platform logs ingestion.
Fixes to failing tests

### DIFF
--- a/.chloggen/refactor-windows-installation-script-to-match-new-msi-format.yaml
+++ b/.chloggen/refactor-windows-installation-script-to-match-new-msi-format.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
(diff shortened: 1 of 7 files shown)


.chloggen/refactor-windows-installation-script-to-match-new-msi-format.yaml
```diff
@@ -0,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: installer
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Refactor Windows install script to match new MSI installation format
```
jinja2, line 8: Can you please write this to be user friendly? This does not provide any info to the user about what bug is being fixed and if they are impacted. Same for the PR description, can you list the changes and why are they needed for a reviewer to have some context?

PR #7807, opened 2026-07-22
[chore] Delete obsolete TestDiscoveryReceiverWithHostObserverAndPrometheusReceiverProvideStatusLogs

jkoronaAtCisco:

Deleted skipped discovery receiver integration test for the unmaintained `prometheus_simple` receiver.


cmd/discoverybundler/metadata/receivers/prometheus_otelcol.yaml
```diff
@@ -0,0 +1,26 @@
+receiver_id: prometheus/otelcol
```
jinja2: The PR description only mentions changes to tests, but we are introducing a new bundled target with this config. I see it is disabled by default, but we'll still be shipping this new discovery target. Can you clarify the scope of the PR?

jkoronaAtCisco: The bundled discovery target is intentional — it provides a default prometheus/otelcol config for --discovery mode, so users running the collector with auto-discovery can scrape the collector's own metrics out of the box. 

However, I agree that shipping a new bundled target is beyond the stated scope of this PR. I can either remove it and use a different approach to get the `receiverMetaMap` entry needed for the test, or split it into a separate PR if the bundled target is considered valuable to keep. WDYT?

pjanotti: The default [agent](https://github.com/signalfx/splunk-otel-collector/blob/bdcc1d1a0c245a48510f3b7c18a8e90cfd013e67/cmd/otelcol/config/collector/agent_config.yaml#L208) and [gateway](https://github.com/signalfx/splunk-otel-collector/blob/bdcc1d1a0c245a48510f3b7c18a8e90cfd013e67/cmd/otelcol/config/collector/gateway_config.yaml#L183) configs already ship a "self-scrape" for the collector metrics. This change shouldn't add that to the default discovery rules.

jkoronaAtCisco: If I understand correctly, `prometheus/otelcol` has to be registered in `receiverMetaMap` because  discovery receiver's evaluators need it to emit status logs for that receiver type. Without this entry the test receives no status logs and times out — that's why the test was originally skipped after #6530 moved status and service_type out of ReceiverEntry into the  bundler-managed map.

To avoid changing bundler metadata, I added the `receiverMetaMap` entry via a hand-written `init()` in a non-generated file within the `discoveryreceiver` package. Let me know if that's ok, or if there's a better way to handle it.

jinja2: This test predates #6530 and was [skipped](https://github.com/signalfx/splunk-otel-collector/commit/5123302950ccf57e0f5ab5a034355bafa38d6981#diff-4cec24a4d0fa5e11e0c480aff1a4a265e8c23fbcaca9fdb92399a4b9a6bc5481) before `status` and `service_type` were moved out of receiver config. Since that change intentionally made status evaluation dependent on bundled metadata, adding a `receiverMetaMap` entry solely to revive this test feels like additional product behavior for a stale test.

We already cover real host-observer Prometheus collection in [TestHostObserver](https://github.com/signalfx/splunk-otel-collector/blob/main/tests/general/discoverymode/host_observer_discovery_test.go#L38), while metric and statement status evaluation should be covered in the discovery receiver tests. If preserving this exact combined flow is still valuable, could it instead become an in-package component test using a fake observable, real prometheus receiver, and test-local metadata that is restored during cleanup? Otherwise, removing the obsolete skipped test may be cleaner than introducing permanent metadata for prometheus/otelcol.

jkoronaAtCisco: Thanks for clarification @jinja2! I decided to delete that test. It looks that it's already well covered.

jinja2 (approved):

dmitryax (approved):

PR #7812, opened 2026-07-23
[smartagent/logstash] deprecate the monitor

atoulme:
**Description:**
Deprecate the logstash monitor, to be removed on or after October 2026.

### DIFF
--- a/.chloggen/deprecate_logstash.yaml
+++ b/.chloggen/deprecate_logstash.yaml
@@ -1,0 +1,17 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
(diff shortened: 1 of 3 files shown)


MovieStoreGuy (commented):
Is there an estimated version number that we could include?

atoulme:
> Is there an estimated version number that we could include?

according to https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md, we could go with 0.163.0. That said, we keep this fairly open and sometime only remove components after an additional delay.

MovieStoreGuy:
Since we don't follow CalVer, it be useful just to provide an expected version so you don't need to go digger as much and it is fine if it slips due to reasons.

atoulme:
I'll finagle something in the release notes for all of the deprecations.

crobert-1 (approved):

PR #7817, opened 2026-07-23
[smartagent/prometheusexporter] deprecate prometheusexporter 

atoulme:
**Description:** 
The prometheusexporter monitor is deprecated, as well as all the submonitors (prometheus/redis, prometheus/prometheus, prometheus/postgres, prometheus/node, prometheus/nginx-vts, prometheus/nginx-ingress, prometheus/go).
  For all those use cases, please use a dedicated receiver or use the [prometheus receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver)

### DIFF
--- a/.chloggen/deprecate_prometheusexporter.yaml
+++ b/.chloggen/deprecate_prometheusexporter.yaml
@@ -1,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: smartagent/prometheusexporter
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: The monitor is deprecated.
+
+# One or more tracking issues related to the change
+issues: [7817]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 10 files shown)


.chloggen/deprecate_prometheusexporter.yaml
```diff
@@ -0,0 +1,18 @@
...
+note: The monitor is deprecated.
+
+# One or more tracking issues related to the change
+issues: [7817]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The prometheusexporter monitor is deprecated, as well as all the submonitors (prometheus/redis, prometheus/prometheus, prometheus/postgres, prometheus/node, prometheus/nginx-vts, prometheus/nginx-ingress, prometheus/go).
```
jinja2, line 17: Aren't the k8s component-specific monitors (e.g. smartagent/kubernetest-apiserver) also wrappers around prometheusexporter monitor? Will the deprecation msg also show up in those?

atoulme: yes, it likely involves those too. I'll update the deprecation notice, and we can hold off until those are removed from the chart.

PR #7824, opened 2026-07-23
Add oracle_encoding extension for OCI metrics

bgola-signalfx:
**Description:**
Add a new `oracle_encoding` extension that unmarshals OCI (Oracle Cloud Infrastructure) Monitoring metrics published in JSONL format into OpenTelemetry metrics format.

**Link to Splunk idea:**
https://ideas.splunk.com/ideas/SFXIMMID-I-555

**Testing:**
* Added unit tests for the low level logic
* Run manual E2E test using real OCI resources & OCI Monitoring input

**Documentation:**
See [README.md](./pkg/extension/oracleencodingextension/README.md)


.chloggen/add-oracleencodingextension.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: new_component
+component: oracleencodingextension
+note: Add the `oracle_encoding` extension, which unmarshals OCI (Oracle Cloud Infrastructure) Monitoring metrics published in JSONL format into OpenTelemetry metrics.
```
crobert-1, line 3: Can you share more context for this component? What is the use case for this? How would the collector be deployed in this environment? Can you confirm that no upstream collector components fit the required functionality here?

bgola-signalfx: We have customers with deployments in Oracle Cloud Infrastructure (OCI) who would like to monitor them using Splunk Observability Cloud (O11y).

Oracle provides a way to push OCI metrics (i.e. metrics for cloud resources deployed in Oracle Cloud) to a Kafka topic. OCI uses a [custom format](https://docs.oracle.com/en-us/iaas/Content/connector-hub/create-service-connector-monitoring-source.htm#api) for metric data.

To enable the OTel Collector to export OCI metrics to OTel-compatible consumers such as the Splunk Observability Cloud, this PR adds an OCI Encoding Extension that converts the OCI metric format to OTLP. It is similar to the [Azure Encoding Extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/encoding/azureencodingextension), but supports metrics only.

Customers will deploy the Collector on a VM within their OCI compartments using a Terraform script provided by Splunk.

I believe no equivalent component exists in either the upstream or contrib repository.

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/builder.go
```diff
@@ -0,0 +1,198 @@
...
+// it into the builder's accumulated state. Records sharing the same
+// compartment, namespace, resource group and resource ID are grouped into a
+// single ResourceMetrics. Within a ResourceMetrics, records sharing the same
+// metric name and unit are merged into a single Metric.
+func (b *metricsBuilder) unmarshalRecord(jsonRecord []byte) {
+	rec, err := b.getValidRecord(jsonRecord)
+	if err != nil {
+		b.logger.Warn("Skipping invalid OCI metric record", zap.Error(err))
+		return
+	}
```
MovieStoreGuy: Is silently erroring here okay? Should this bubble out the issue so that a decision can be made?

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/builder.go
```diff
@@ -0,0 +1,198 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics // import "github.com/signalfx/splunk-otel-collector/pkg/extension/oracleencodingextension/internal/unmarshaler/metrics"
+
+import (
+	"encoding/json"
```
MovieStoreGuy: For a performance improvement, I would suggest using `github.com/goccy/go-json` instead of the native first party pacakge.

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/builder.go
```diff
@@ -0,0 +1,198 @@
...
+
+	md := pmetric.NewMetrics()
+	for _, rm := range b.allResourceMetrics {
+		rm.MoveTo(md.ResourceMetrics().AppendEmpty())
+	}
+	return md
+}
+
+func (b *metricsBuilder) getValidRecord(jsonRecord []byte) (*ociMetricRecord, error) {
+	var rec ociMetricRecord
```
MovieStoreGuy: I would honestly move the validate portion of this as part of the type to make it easier to test in future, and it means you can remove this method entirely. 

The updated call sites would look like:

```go
if err := json.NewDecoder(bytes.NewBuffer(record)).Encode(&ocMetricRecordVar); err != nil {
  return fmt.Errorf("json unmarshal: %w", err)
}
if err := ociMetricRecordVar.Validate(); err != nil {
   return fmt.Errorf("record validation: %w",err)
}
```

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/unmarshaler.go
```diff
@@ -0,0 +1,191 @@
...
+	"io"
+	"strings"
+
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	conventions "go.opentelemetry.io/otel/semconv/v1.40.0"
+	"go.uber.org/zap"
+)
+
+// oracleCloudNamespaceKey and oracleCloudResourceGroupKey are not yet part of
+// the OTel semantic conventions oracle_cloud.* registry, so they are defined here
```
MovieStoreGuy: Is there an open issue that could be linked here for this?

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/unmarshaler.go
```diff
@@ -0,0 +1,191 @@
...
+			unmarshalRecordFn(trimmed)
+		}
+		if err != nil {
+			if err != io.EOF {
+				r.logger.Warn("Failed to read JSON input", zap.Error(err))
+			}
+			return
+		}
+	}
+}
```
MovieStoreGuy: I believe you can simplify this to:

```go 
decoder := json.NewDecoder(r)
for decoder.More() { 
   if err := decoder.Decode(&var); err != nil {
      return err
   }
   // Do something with Var
}
```

pkg/extension/oracleencodingextension/extension.go
```diff
@@ -0,0 +1,50 @@
...
+	return ex.metricUnmarshaler.UnmarshalMetrics(buf)
+}
+
+func (*oracleExtension) Start(context.Context, component.Host) error {
+	return nil
+}
+
+func (*oracleExtension) Shutdown(context.Context) error {
+	return nil
+}
```
MovieStoreGuy: You can embed `component.StartFunc` and `component.ShutdownFunc` into the struct so it will inherit the required methods to avoid needing to test noop methods.

pkg/extension/oracleencodingextension/go.mod
```diff
@@ -0,0 +1,38 @@
+module github.com/signalfx/splunk-otel-collector/pkg/extension/oracleencodingextension
+
+go 1.26.5
```
MovieStoreGuy, line 3: Can you set this to be `1.26.0` to avoid dragging the min supported version of the collector up?

MovieStoreGuy (commented):
In order to make this easier to maintain with regards to own lifecycle, can I ask that you follow the examples provided by [mdatagen](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/mdatagen#metadata-generator)?

It should be at most have a metadata.yml file, call generate, wire up all the generated values.

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/unmarshaler.go
```diff
@@ -0,0 +1,191 @@
...
+// OCI Monitoring has been observed emitting this dimension key as either
+// "resourceId" or "resourceID", so the key is matched case-insensitively.
+func extractResourceID(dimensions map[string]any) (resourceID string) {
+	if len(dimensions) == 0 {
+		return ""
+	}
+
+	for k, v := range dimensions {
+		if resourceID == "" && strings.EqualFold(k, dimensionResourceID) {
+			resourceID, _ = v.(string)
```
dloucasfx, line 174: Is it possible to have both "resourceId" and "resourceID" at the same time?
If it is, then how to address this non-deterministic case?

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/builder.go
dloucasfx, line 54: In addition to above comment whether we should bubble the error vs logging it, does it make sense to add a rate limit or set it to debug?
Note, when `UnmarshalMetrics` is called by the consumer, in this example `kafka` receiver , the error is checked/handled https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f5def4cf28f02252521e14e1d1ce5205b6167568/receiver/kafkareceiver/kafka_receiver.go#L227-L233 and logged https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f5def4cf28f02252521e14e1d1ce5205b6167568/receiver/kafkareceiver/kafka_receiver.go#L368

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/builder.go
dloucasfx: I agree with @MovieStoreGuy , goccy/go-json or jsoniter are ~4.5x faster then encoding/json , I highly recommend switching packages

pkg/extension/oracleencodingextension/go.mod
dloucasfx: This should not be an issue, as the splunk distro of the collector is currently on  1.26.5

dloucasfx (changes requested):
Overall, this looks good!
I added a few comments, also for the `log.warn` vs returning an `error`, this applies to multiple places

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/builder.go
bgola-signalfx: I believe the best we can do is to log the error details (which is handled by the `getValidRecord` func) and skip a malformed record.

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/unmarshaler.go
bgola-signalfx: I believe there is no such issue today. 

Shall I create one for `oracleCloudCompartmentIDKey, oracleCloudNamespaceKey and oracleCloudResourceGroupKey`?

BTW I've also updated the code to use the `conventions.OracleCloudRealmKey` instead of a locally defined dup definition.

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/builder.go
bgola-signalfx: done

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/builder.go
bgola-signalfx: done

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/unmarshaler.go
bgola-signalfx: Done. This led to the builder's `unmarshalRecord(jsonRecord []byte)` be replaced with `addRecord(rec ociMetricRecord)`.

pkg/extension/oracleencodingextension/go.mod
bgola-signalfx: leaving as-is

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/unmarshaler.go
bgola-signalfx: It's hard to definitively say "no, it's always either one of those" but I am pretty sure that is the case. It would be (1) super weird to have a metric with those two dimensions and (2) to have those two dimensions with different values (typically OCIDs).

https://docs.oracle.com/en-us/iaas/Content/General/Concepts/identifiers.htm

pkg/extension/oracleencodingextension/internal/unmarshaler/metrics/builder.go
bgola-signalfx: Now that I read the Dani's comment I understand the idea better. I'll switch to the error bubbling.

bgola-signalfx: I've updated the code to bubble the `UnmarshalMetrics/readJSONLLines` errors up. However I chose to still log `ociMetricRecord` semantic errors so that a single invalid metric record would not result in other valid records from the same JSON lines batch to be dropped. 

I do not have a strong opinion on this so feel free to suggest other approach.

dloucasfx (changes requested):
@bgola-signalfx looks like the `"github.com/goccy/go-json"` changes in this commit https://github.com/signalfx/splunk-otel-collector/pull/7824/changes/cd59c3beea41cbd7ff5ad62030e89f2d65143ad6 were undone and reverted to `"encoding/json"` in this commit https://github.com/signalfx/splunk-otel-collector/pull/7824/changes/1f6c7e6837bea6eefd5c45e2f494715411e5c87f#diff-b1968acd50cad7288eb12fefc707e78dc8fb8d77d1312944aac4a55b74a37004 

Once this is addressed, we can merged this. 
Thanks!

dloucasfx (approved):

PR #7825, opened 2026-07-23
[chore] Add monitoring systemd timer example

pjanotti:
Add an example that shows a potential path of migration for users that leverage the exec receiver. The specific case is for customers that can use `systemd`. The example shows a legacy script publishing InfluxDB metrics running as a systemd timer and being monitored, via logs and metrics, by the Splunk OpenTelemetry collector.


pjanotti:
🤔[.github/workflows/lychee.yml / linkChecker (pull_request)](https://github.com/signalfx/splunk-otel-collector/actions/runs/30043618917/job/89329266339?pr=7825) is about a new link being added: https://developers.redhat.com/blog/2016/09/13/running-systemd-in-a-non-privileged-container

The link exists and I don't need login but I guess that from lychee in CI doesn't work...

atoulme (approved):

PR #7829, opened 2026-07-24
[chore] add splunkinputsreceiver with a feature gate, disabled by default

atoulme:
There is no changelog as this is mostly for ongoing development and experimentation.


internal/components/components.go
```diff
@@ -161,6 +163,18 @@ import (
 	"github.com/signalfx/splunk-otel-collector/pkg/receiver/smartagentreceiver"
 )
 
+const (
+	splunkInputsSupportFeatureGateID = "splunk_inputs"
```
dmitryax: Should we call it `enableTARunner` or `dotConfSupport`? I think once we have outputs.conf support, it should be behind the same FG

dmitryax (approved):

internal/components/components.go
atoulme: yes sure

internal/components/components.go
```diff
@@ -161,6 +163,18 @@ import (
...
+)
+
+var splunkinputssupport = featuregate.GlobalRegistry().MustRegister(
+	splunkInputsSupportFeatureGateID,
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription("When enabled, the collector supports the `splunk_inputs` receiver. "+
+		"When disabled (default), the `splunk_inputs` receiver is not available and the collector will crash if it tries to run it."),
+	featuregate.WithRegisterFromVersion("v0.157.0"),
```
pjanotti: 158, right?

internal/components/components.go
atoulme: done

pjanotti (commented):
I see go.sum changes but also CI is breaking on that... anyway, LGTM minus the build break...

internal/components/components.go
atoulme: sorry, is it better to have it camel case or snake case for feature gate IDs? I used `dot_conf_support` but `dotConfSupport` works too, please let me know

internal/components/components.go
atoulme: corrected, thanks

internal/components/components.go
```diff
@@ -161,6 +163,18 @@ import (
 	"github.com/signalfx/splunk-otel-collector/pkg/receiver/smartagentreceiver"
 )
 
+const (
+	dotConfSupportFeatureGateID = "dot_conf_support"
```
pjanotti: I saw the previous comment about this but I don't like this name: .conf is just a file format, somehow I think it needs to imply "Splunk Data Inputs" or something that Splunk users are used to.

atoulme: ok changed to `enableTARunner` ; note we mean for this feature gate to be for development

internal/components/components.go
dmitryax: Looks like we typically use `camelCase` for feature gates

PR #7830, opened 2026-07-24
[chore] Improve TA dev experience on Windows

pjanotti:
Fix the TA build and tests on Windows. This change only affects the dev experience on Windows.

### DIFF
--- a/packaging/ta-v2/Makefile
+++ b/packaging/ta-v2/Makefile
@@ -13,6 +13,8 @@ endif
 
 VERSION ?= $(shell git describe --match "v[0-9]*" HEAD)
 
+CP := sh -c 'cp "$$@"' cp
+
 # Set tar exclude options based on the target OS for packaging
 ifeq ($(TA_TARGET_OS),windows)
 	TAR_EXCLUDE_OPT ?= --exclude='*/linux_$(TA_ARCH)/*'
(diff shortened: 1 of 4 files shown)


packaging/ta-v2/tests/package_test.go
```diff
@@ -166,12 +167,12 @@ func TestLinuxPackageContents(t *testing.T) {
 
 	packageName := linuxName
 	// Linux package should contain linux binaries
-	linuxBinPath := filepath.Join(packageName, linuxBinDir)
+	linuxBinPath := path.Join(packageName, linuxBinDir)
```
pjanotti, line 170: `filepath.Join` generates platform specific code, but some of the tests are checking paths inside the `tgz` which uses `/` as path separator, so changing to `path.Join` instead that also always uses `/`.

packaging/ta-v2/Makefile
```diff
@@ -13,6 +13,8 @@ endif
 
 VERSION ?= $(shell git describe --match "v[0-9]*" HEAD)
 
+CP := sh -c 'cp "$$@"' cp
```
pjanotti, line 16: Portable way to do the command.

dmitryax (approved):

PR #7842, opened 2026-07-28
New connector for persistent queueing

atoulme:
**Description:**
This is a connector that uses a persistent queue. It stores data in the queue, and reads back out of it. You can set a limit to how many bytes you want to read out to throttle the data.

Performance benchmarks are present.


.chloggen/pq_processor.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: pqprocessor
```
crobert-1: Every `pg` reference should be updated to be the full "persistent queue". `pg` could mean many different things, it's not clear enough on its own.

internal/processor/pqprocessor/README.md
```diff
@@ -0,0 +1,14 @@
+# Persistent queue processor
+
```
crobert-1, line 2: We should match upstream's component header as much as possible here, specifically the stability/telemetry type table.

internal/processor/pqprocessor/README.md
```diff
@@ -0,0 +1,14 @@
+# Persistent queue processor
```
crobert-1: I'd expect some recommendation here about order/placement in the telemetry pipeline, similar to how we always recommend the batching processor to be first.

My guess would we'd always want it last in the pipeline so that the telemetry is all in its "final form" when it's sitting in the queue?

internal/processor/pqprocessor/README.md
```diff
@@ -0,0 +1,14 @@
...
+The persistent queue processor stores data into a queue. It reads from the queue to send data down to exporters.
+
+As such, this processor is asynchronous and can enforce a bandwidth limit.
+
+The queue retries indefinitely, and is strictly applying FIFO order. As such, it cannot be shared across multiple pipelines.
+
+Example:
+```yaml
+persistent_queue:
+  folder: /var/queue
```
crobert-1: I don't think `folder` is a good name here, something like `file_path`, `storage_dir`, `path` may be more obvious.

internal/processor/pqprocessor/README.md
```diff
@@ -0,0 +1,14 @@
...
+  bandwidth: 256768 # 256 Kb
```
crobert-1: `bandwidth` isn't very self-explanatory, and the description above is only "bandwidth limit". Is this some kind of maximum storage limit? What's the unit?

internal/processor/pqprocessor/README.md
```diff
@@ -0,0 +1,14 @@
+# Persistent queue processor
+
+The persistent queue processor stores data into a queue. It reads from the queue to send data down to exporters.
+
+As such, this processor is asynchronous and can enforce a bandwidth limit.
```
crobert-1, line 5: As shown by my questions below, a configuration section would be helpful. What options are available, required, what the defaults are, units, etc.

internal/processor/pqprocessor/processor.go
```diff
@@ -0,0 +1,195 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/nsqio/go-diskqueue"
```
crobert-1: Latest release: 5 years ago
Latest merged commit: 10 months

There's are also an open PR that seems to be fixing a pretty noteworthy bug, but no update in over a year on it.

😬

internal/processor/pqprocessor/processor.go
```diff
@@ -0,0 +1,195 @@
...
+	if err := os.MkdirAll(b.config.Folder, 0o755); err != nil {
+		return err
+	}
+
+	maxSize := b.config.Bandwidth
+	if maxSize == 0 {
+		maxSize = 10_000_000
+	}
+
+	// Create a new queue with segment size of 50
```
crobert-1: I don't see anything related to segment size here, or 50. What am I missing?

crobert-1 (commented):
_Just a quick read through, I didn't spend any time on the actual `processor.go` logic_

internal/processor/pqprocessor/processor.go
atoulme: old comment, removing.

.chloggen/pq_processor.yaml
atoulme: done, it's gone

internal/processor/pqprocessor/README.md
atoulme: that table is generated upstream, based on metadata.yaml. I guess we could do that if and when we donate upstream, if that sounds good.

internal/processor/pqprocessor/README.md
atoulme: it's a throughput limit. I'll add a table with the options.

internal/processor/pqprocessor/README.md
atoulme: done - would be after batch processor, late.

internal/processor/pqprocessor/processor.go
atoulme: we might want to inline it in and patch it. It needs thorough review.

atoulme: inlined the lib

atoulme:
Closing in favor of #7956

PR #7843, opened 2026-07-28
add cim mapping sourcetype transform

omrozowicz-splunk:
**Description:** Improve sourcetype assignment in Linux log collection config. Updates `splunk_logs_config_linux.yaml` to more accurately replicate Splunk TA for *nix (Splunk_TA_nix) sourcetype assignment behaviour.

**Changes**

`file_log/varlog` — dynamic filename-based sourcetype assignment

Replaces the hardcoded `linux:varlog` sourcetype with dynamic assignment based on the log filename, matching Splunk's default behavior of using the filename without extension as the sourcetype (e.g. `syslog` → `syslog`, `kern.log` → `kern`).

Uses `regex_parser` to extract the filename stem and set `com.splunk.sourcetype`.

`file_log/varlog` — exclude vim swap files

Adds exclusion patterns for vim swap files (`*.swp`, `*.swx`, `.*.swp`, `.*.swx`) to prevent binary swap files from being ingested, which was causing incorrect `httpevent` sourcetype assignment.

`file_log/bash_history` — fix sourcetype

Changes hardcoded sourcetype from `linux:bash_history` to `bash_history` to match the TA's `inputs.conf` which explicitly sets `sourcetype = bash_history`.

`transform/nix_sourcetype` — path-based sourcetype overrides

Adds a new `transform/nix_sourcetype` processor that overrides sourcetypes for well-known paths, replicating the `source::` stanzas from the Splunk Add-on for Unix and Linux (`Splunk_TA_nix`) `props.conf` and Splunk's default `props.conf` FILE MATCH CONDITIONS. Examples:
- `.../var/log/secure` → `linux_secure`
- `.../var/log/auth.log` → `syslog`
- `.../var/log/audit/audit.log` → `linux_audit`
- `.../var/log/messages` → `syslog`
- `*.conf`, `*.cfg`, `*.ini`, etc. → `config_file`
- `/etc/passwd*`, `/etc/shadow*` → `ignored_type`

Path patterns support log-rotation suffixes (e.g. `secure.1`, `secure.2`). Splunk's content-based `delayedrule` detection (>80% line matching) is not implemented; `auth.log` → `syslog` is an explicit exception.

**Link to Splunk idea:** N/A

**Testing:** Manual testing and comparison against Linux TA

**Documentation:** Updated `docs/getting-started/linux-logs.md` with a Sourcetype assignment section describing both stages of assignment and a mapping table.


dhruvanshus-crest (approved):

docs/getting-started/linux-logs.md
```diff
@@ -111,18 +111,50 @@ file_log/varlog:
...
+
+**2. Path-based overrides (`transform/nix_sourcetype` processor)**
+
+After the filename fallback, the `transform/nix_sourcetype` processor overrides the sourcetype for well-known paths to match the sourcetypes used by the [Splunk Add-on for Unix and Linux](https://splunkbase.splunk.com/app/833) and Splunk's default `props.conf` FILE MATCH CONDITIONS. Examples:
+
+| Source path                      | Sourcetype      |
+|----------------------------------|-----------------|
+| `.../var/log/secure`             | `linux_secure`  |
+| `.../var/log/auth.log`           | `syslog`        |
+| `.../var/log/audit/audit.log`    | `linux_audit`   |
```
jinja2, line 130: Will this actually match the varlog file receiver's `/var/log/*.log` re? I thought it needs `**` to include subdirs

omrozowicz-splunk: This config mirrors what is configured in scope of the Splunk TA for Unix and Linux. For sourcetyping we have much wider scope of paths than for the default monitoring itself. The props will be applied when user add such a file to the filelog receiver scope.

omrozowicz-splunk: I added a note to the docs about that as this behaviour might be misleading for users

pjanotti (approved):

jinja2 (approved):

PR #7729, opened 2026-06-30
Warn if deprecated otlp_config_linux.yaml is used

agoallikmaa:
**Description:**
Use of the otlp_config_linux.yaml bundled with Docker images is now deprecated and the file is planned to be removed from future images. To prepare for removal, when deprecated otlp_config_linux.yaml is set as the config path, emit a warning.

PR #7848, opened 2026-07-29
Warn if deprecated otlp_config_linux.yaml is used

agoallikmaa:
**Description:**
Use of the otlp_config_linux.yaml bundled with Docker images is now deprecated and the file is planned to be removed from future images. To prepare for removal, when deprecated otlp_config_linux.yaml is set as the config path, emit a warning.

Replaces #7729 to not PR from fork, for better CI coverage.

### DIFF
--- a/.chloggen/deprecate-otlp-config-linux.yaml
+++ b/.chloggen/deprecate-otlp-config-linux.yaml
@@ -1,0 +1,5 @@
+change_type: deprecation
+component: configuration
+note: Deprecate `/etc/otel/collector/otlp_config_linux.yaml` and warn users to migrate to `/etc/otel/collector/agent_config.yaml`.
+issues: [7729]
+subtext: Set `SPLUNK_LISTEN_INTERFACE=0.0.0.0` when migrating if receivers need to listen outside the container.
--- a/.chloggen/deprecate-otlp-config-linux.yaml
+++ b/.chloggen/deprecate-otlp-config-linux.yaml
@@ -1,5 +1,5 @@
 change_type: deprecation
 component: configuration
 note: Deprecate `/etc/otel/collector/otlp_config_linux.yaml` and warn users to migrate to `/etc/otel/collector/agent_config.yaml`.
-issues: [7729]
+issues: [7848]
 subtext: Set `SPLUNK_LISTEN_INTERFACE=0.0.0.0` when migrating if receivers need to listen outside the container.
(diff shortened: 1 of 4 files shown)


.chloggen/deprecate-otlp-config-linux.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: deprecation
+component: configuration
+note: Deprecate `/etc/otel/collector/otlp_config_linux.yaml` and warn users to migrate to `/etc/otel/collector/agent_config.yaml`.
```
crobert-1:
```suggestion
note: The default config `/etc/otel/collector/otlp_config_linux.yaml` is now deprecated and will be removed in a future release.
```
It would also be helpful to include the expected timeline of removal. We can add the replacement guidelines in the `subtext` (users should move to `agent_config`), but I don't think it's necessary to call out the part about "warn users to migrate".

internal/settings/settings.go
```diff
@@ -85,6 +85,8 @@ var DefaultAgentConfigWindows = func() string {
 	return filepath.Clean(path)
 }()
 
+var statConfigFile = os.Stat
```
crobert-1, line 90: Maybe add a comment here that this is used for testing (it took me a bit to figure out why we need this)

crobert-1 (commented):
Maybe also add a note at the top of the actual config file itself also to make it clear that it's deprecated.

Also, does this functionality work for deployment methods other than docker?

.chloggen/deprecate-otlp-config-linux.yaml
agoallikmaa: Applied change and added replacement guide to subtext. I'm not sure if we have an expected timeline for this though, so didn't address that.

internal/settings/settings.go
agoallikmaa: Done.

agoallikmaa:
> Maybe also add a note at the top of the actual config file itself also to make it clear that it's deprecated.

Done

> Also, does this functionality work for deployment methods other than docker?

It is not relevant outside of docker, because that file is only shipped as part of the docker image. Though the path check isn't technically restricted to use in docker, but users would have to explicitly put that file at that path for the check to trigger outside of docker.

.chloggen/deprecate-otlp-config-linux.yaml
crobert-1, line 1: Since this path is primarily related to usage on Docker, maybe add a note somewhere in here to make it clear.

crobert-1 (approved):

PR #7850, opened 2026-07-29
Enable OpAMP extension

dloucasfx:

### DIFF
--- a/.chloggen/enable_opamp.yaml
+++ b/.chloggen/enable_opamp.yaml
@@ -1,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: extension/opamp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Enable the Splunk OpAMP extension (`opamp/splunk_o11y`) by default.
+
+# One or more tracking issues related to the change
+issues: [7850]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The feature gate ID `splunk.opamp.enabled` is now `StageBeta` which no longer removes the Splunk OpAMP extension (`opamp/splunk_o11y`) by default.
+  If you do not want to use the OpAMP extension, disable the feature gate with `--feature-gates=-splunk.opamp.enabled`. 
--- a/.chloggen/enable_opamp.yaml
+++ b/.chloggen/enable_opamp.yaml
@@ -14,5 +14,5 @@ issues: [7850]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  The feature gate ID `splunk.opamp.enabled` is now `StageBeta` which no longer removes the Splunk OpAMP extension (`opamp/splunk_o11y`) by default.
+  Promotes the `splunk.opamp.enabled` feature gate to beta which no longer removes the Splunk OpAMP extension (`opamp/splunk_o11y`) by default.
   If you do not want to use the OpAMP extension, disable the feature gate with `--feature-gates=-splunk.opamp.enabled`. 
--- a/internal/configconverter/opamp.go
+++ b/internal/configconverter/opamp.go
@@ -29,9 +29,9 @@ const (
 
 var opampFeatureGate = featuregate.GlobalRegistry().MustRegister(
 	opampFeatureGateID,
-	featuregate.StageAlpha,
-	featuregate.WithRegisterDescription("When enabled, the opamp/splunk_o11y extension is active. "+
-		"When disabled (default), the opamp/splunk_o11y extension is removed from the service.extensions configuration at startup, if it is present."),
+	featuregate.StageBeta,
+	featuregate.WithRegisterDescription("When enabled (default), the opamp/splunk_o11y extension is active. "+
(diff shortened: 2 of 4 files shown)


internal/configconverter/opamp.go
```diff
@@ -29,7 +29,7 @@ const (
 
 var opampFeatureGate = featuregate.GlobalRegistry().MustRegister(
 	opampFeatureGateID,
-	featuregate.StageAlpha,
+	featuregate.StageBeta,
 	featuregate.WithRegisterDescription("When enabled, the opamp/splunk_o11y extension is active. "+
 		"When disabled (default), the opamp/splunk_o11y extension is removed from the service.extensions configuration at startup, if it is present."),
```
crobert-1: This statement (`disabled (default)`) is out of date

.chloggen/enable_opamp.yaml
```diff
@@ -0,0 +1,19 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: extensions/opamp
```
crobert-1:
```suggestion
component: extension/opamp
```
Or maybe component should just be `config`, but I'm not sure what's best here

.chloggen/enable_opamp.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Enable the Splunk OpAMP extension by default.
```
crobert-1: The note and subtext are a bit confusing, relative to what's going on here. Out of ignorance, I assumed this meant that some new extension will now be added by default to all configs. What's really going on is `opamp/splunk_o11y` will no longer be removed from configs when specified. It's slightly different, but two key differences to me:
1. It's only relevant for the specific `opamp/splunk_o11y` name, not all `opamp` extensions.
2. Nothing is being added by default, it's just no longer being removed. (Maybe the end user doesn't care about this distinction, but some config converters are automatically adding things to people's config no matter what, so I don't want users to think that's what's happening here).

internal/configconverter/opamp.go
dloucasfx: yeah, I have the fix waiting to see if any tests fails before pushing.
Thanks!

.chloggen/enable_opamp.yaml
dloucasfx: The original feature gate went through as `extension/opamp` , I will stick to this and fix the spelling

.chloggen/enable_opamp.yaml
dloucasfx: > It's only relevant for the specific opamp/splunk_o11y name, not all opamp extensions.
Correct, hence we reference to it as "Splunk OpAMP" 

> Nothing is being added by default, it's just no longer being removed. (Maybe the end user doesn't care about this distinction, but some config converters are automatically adding things to people's config no matter what, so I don't want users to think that's what's happening here).

Prior to this change, the default behavior is to remove `opamp/splunk_o11y` unless it was explicitly enabled through the feature gate, now we are keeping it by default.

I did some rewarding, let me know if this looks better

.chloggen/enable_opamp.yaml
```diff
@@ -0,0 +1,18 @@
...
+note: Enable the Splunk OpAMP extension (`opamp/splunk_o11y`) by default.
+
+# One or more tracking issues related to the change
+issues: [7850]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The feature gate ID `splunk.opamp.enabled` is now `StageBeta` which no longer removes the Splunk OpAMP extension (`opamp/splunk_o11y`) by default.
```
crobert-1:
```suggestion
  Promotes the `splunk.opamp.enabled` feature gate to beta which no longer removes the Splunk OpAMP extension (`opamp/splunk_o11y`) by default.
```
`StageBeta` is the programmatic/internal name of the stability, not formatted for end users. We could even link to [feature gate stabilities](https://github.com/open-telemetry/opentelemetry-collector/tree/main/featuregate#feature-lifecycle) if it would be helpful.

.chloggen/enable_opamp.yaml
```diff
@@ -0,0 +1,18 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Enable the Splunk OpAMP extension (`opamp/splunk_o11y`) by default.
+
+# One or more tracking issues related to the change
+issues: [7850]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
crobert-1, line 16: Maybe add another sentence that if a user's config does not include an extension named `opamp/splunk_o11y` this has no impact? I'm wondering if a user would see the statement `enabled by default` and not understand the actual scope here.

.chloggen/enable_opamp.yaml
crobert-1: Thanks, the new wording of "no longer removes" is more clear for me 👍

.chloggen/enable_opamp.yaml
dloucasfx: yeah, makes sense ! 
thanks!

.chloggen/enable_opamp.yaml
dloucasfx: I don't think this would be an issue here, I think pointing to every use case will create more confusion.
The config resolver also include some useful log messages to when the feature is enabled but the extension is missing from the config.

crobert-1 (approved):

PR #7851, opened 2026-07-29
Update launcher to support tar package dir structure

kelseyma:
- Move creation of supervisor config dir to launcher (ensures supervisor dir only created when supervisor mode is enabled vs if done via packaging)
- Update launcher to support tar dir structure where binaries aren't in service installed `/usr/bin` dir

- Local dev testing with tar package => ran launcher with supervisor mode, verified binaries/configs found and able to run

### DIFF
--- a/cmd/otelcollauncher/main_windows.go
+++ b/cmd/otelcollauncher/main_windows.go
@@ -212,10 +212,9 @@ type childResult struct {
 // childProcess wraps the immediate child selected by the launcher. The child is
 // either otelcol.exe in direct mode or opampsupervisor.exe in supervisor mode.
 type childProcess struct {
-	cmd *exec.Cmd
-	// outputClosers unblock StdoutPipe/StderrPipe forwarding during hard-stop cleanup.
-	outputClosers []io.ReadCloser
+	cmd           *exec.Cmd
 	done          <-chan childResult
+	outputClosers []io.ReadCloser
 }
 
 // startChild launches the selected binary in its own process group. Interactive
--- a/internal/opampsupervisor/launcher/config.go
+++ b/internal/opampsupervisor/launcher/config.go
@@ -194,6 +194,11 @@ func PrepareCommand(args, environ []string, paths Paths) (Command, error) {
 // supervisor start and writes the runtime supervisor config with the current
 // launcher-managed agent fields.
 func prepareSupervisor(inputs supervisorInputs, env map[string]string, paths Paths) error {
+	supervisorDir := filepath.Dir(paths.SupervisorConfig)
+	if err := os.MkdirAll(supervisorDir, 0o700); err != nil {
+		return fmt.Errorf("create supervisor config directory %q: %w", supervisorDir, err)
+	}
+
 	collectorConfigs, err := loadCollectorConfigFiles(inputs.configFiles)
 	if err != nil {
 		return err
--- a/internal/opampsupervisor/launcher/config.go
+++ b/internal/opampsupervisor/launcher/config.go
@@ -196,7 +196,7 @@ func PrepareCommand(args, environ []string, paths Paths) (Command, error) {
 func prepareSupervisor(inputs supervisorInputs, env map[string]string, paths Paths) error {
 	supervisorDir := filepath.Dir(paths.SupervisorConfig)
 	if err := os.MkdirAll(supervisorDir, 0o700); err != nil {
-		return fmt.Errorf("create supervisor config directory %q: %w", supervisorDir, err)
+		return fmt.Errorf("failed to create supervisor config directory %q: %w", supervisorDir, err)
 	}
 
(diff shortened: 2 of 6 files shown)


internal/opampsupervisor/launcher/paths_others.go
```diff
@@ -16,17 +16,36 @@
...
 package launcher
 
-// DefaultPaths returns the installed collector, supervisor, and state
-// locations used by the package-managed service on Linux packages.
+import (
+	"os"
+	"path/filepath"
+)
+
+// DefaultPaths returns collector, supervisor, and state locations for Linux
```
crobert-1, line 24: I'm pretty out of the loop here, so I apologize if this is obvious: I'm noticing the build flag is simply `!windows`, which means this may be running on mac, but the comment here is only linux. Is that on purpose?

internal/opampsupervisor/launcher/config.go
```diff
@@ -194,6 +194,11 @@ func PrepareCommand(args, environ []string, paths Paths) (Command, error) {
 // supervisor start and writes the runtime supervisor config with the current
 // launcher-managed agent fields.
 func prepareSupervisor(inputs supervisorInputs, env map[string]string, paths Paths) error {
+	supervisorDir := filepath.Dir(paths.SupervisorConfig)
+	if err := os.MkdirAll(supervisorDir, 0o700); err != nil {
+		return fmt.Errorf("create supervisor config directory %q: %w", supervisorDir, err)
```
crobert-1: Should this error message include some kind of wording to make it clear that the create command failed here?

internal/opampsupervisor/launcher/paths_others.go
```diff
@@ -16,17 +16,36 @@
...
+// packages. Tar bundles use paths relative to the launcher executable.
 func DefaultPaths() Paths {
+	launcherExecutable, err := os.Executable()
+	if err != nil {
+		launcherExecutable = "/usr/bin/otelcollauncher"
+	}
+	return pathsForExecutable(launcherExecutable)
+}
+
+func pathsForExecutable(launcherExecutable string) Paths {
```
crobert-1, line 34: Where are all of these paths being set/used originally? (I'm wondering if there's any kind of simpler way to get and set these values that would be closer to the source).

internal/opampsupervisor/launcher/paths_others.go
kelseyma: This was mainly for ease of running the unit tests and setup of files in [cmd/otelcollauncher](https://github.com/signalfx/splunk-otel-collector/tree/main/cmd/otelcollauncher). We'll only be including the launcher in the linux and windows packages for now, as I don't believe we support a full mac package yet. If it's preferred to make it clear about only linux (vs others), let me know and I can update the files/build tags in both packages!

internal/opampsupervisor/launcher/paths_others.go
kelseyma: These are coming from the hardcoded paths set in the installer/packaging [scripts](https://github.com/signalfx/splunk-otel-collector/blob/main/packaging/installer/install.sh#L61) of where the binaries and dirs are added. For example, the [service file](https://github.com/signalfx/splunk-otel-collector/blob/main/packaging/fpm/splunk-otel-collector.service#L7) hardcodes the `/usr/bin/otelcol` location as expected for ExecStart.
I didn't really see any constants declared for these in the code, so I just added them individually here as they're mainly set/referenced in install scripts.

internal/opampsupervisor/launcher/config.go
kelseyma: good callout! codex was being brief on the messages, went through and updated all to be clearer

cmd/otelcollauncher/main_windows.go
```diff
@@ -212,10 +212,9 @@ type childResult struct {
 // childProcess wraps the immediate child selected by the launcher. The child is
 // either otelcol.exe in direct mode or opampsupervisor.exe in supervisor mode.
 type childProcess struct {
-	cmd *exec.Cmd
-	// outputClosers unblock StdoutPipe/StderrPipe forwarding during hard-stop cleanup.
-	outputClosers []io.ReadCloser
+	cmd           *exec.Cmd
 	done          <-chan childResult
+	outputClosers []io.ReadCloser
```
kelseyma, line 217: just some lint fixes for main_windows and main_windows_test

internal/opampsupervisor/launcher/paths_others.go
crobert-1: I see. My preference would be to be as specific as possible with these build constraints to make supported platforms obvious, but for now it's fair to just match `cmd/otelcollauncher`. It'd be great if things just failed to start on unsupported platforms with obvious log messages about it (if it doesn't already), but not necessary for this PR.

crobert-1 (approved):

PR #7852, opened 2026-07-30
Adding the Splunk Secret Storage config provider

pjanotti:
Adds a `splunk_secret` config source for retrieving secrets from Splunk's
local secret storage, plus CI coverage and supporting infra changes.

## New component: `splunk_secret` config source

- [internal/configsource/splunksecretconfigsource/](internal/configsource/splunksecretconfigsource/):
  new config source that resolves `${splunk_secret:<realm>:<name>}` via
  splunkd's `storage/passwords` REST endpoint (`endpoint`, `session_key`,
  `app`, `user`, `insecure_skip_verify`, `timeout` config fields).

## TA launcher: expose splunkd session info

- [pkg/modularinput/ta_launcher.go](pkg/modularinput/ta_launcher.go) now sets
  `SPLUNK_SERVER_URI`/`SPLUNK_SESSION_KEY` env vars from stdin so
  `splunk_secret` works out-of-the-box as a TA modular input.

## CI: end-to-end smoke test

- [.github/workflows/ta-v2.yaml](.github/workflows/ta-v2.yaml) (Linux +
  Windows): creates a test secret via REST, patches `agent_config.yaml` to
  reference it, restarts Splunk, and asserts the resolved value appears in
  `/metrics`. Also drops the deprecated Splunk 8.2.0 matrix entry and fixes
  app-name/management-port bugs in the existing tests.

## Collateral

- [Dockerfile.windows](packaging/ta-v2/Dockerfile.windows): requires a
  `SPLUNK_ADMIN_PASSWORD` build-arg and forces `mgmtMode = tcp` so port 8089
  is reachable.
- [run-local-package-in-container.ps1](packaging/ta-v2/run-local-package-in-container.ps1):
  publishes port 8089.
- [run-local-package-in-container.sh](packaging/ta-v2/run-local-package-in-container.sh):
  honors a caller-supplied `SPLUNK_PASSWORD`.


pjanotti:
lychee and changelog errors are for the link to the new component being added.

dloucasfx (approved):

.github/workflows/ta-v2.yaml
crobert-1, line 1: Would there be any way to move more of this test functionality into a script or standalone test? I wonder if it would make it possible to do testing locally instead of 100% relying on GitHub to run it in CICD.

.github/workflows/ta-v2.yaml
```diff
@@ -228,8 +296,30 @@ jobs:
...
+          "splunk_password=$pw" | Out-File -FilePath $env:GITHUB_OUTPUT -Encoding utf8 -Append
+
       - name: Install Splunk UF
-        run: msiexec.exe /i "$PWD\splunkforwarder.msi" /qn AGREETOLICENSE=Yes
+        env:
+          SPLUNK_PASSWORD: ${{ steps.gen_password.outputs.splunk_password }}
+        run: msiexec.exe /i "$PWD\splunkforwarder.msi" /qn AGREETOLICENSE=Yes SPLUNKPASSWORD="$Env:SPLUNK_PASSWORD"
+
+      - name: Configure Splunk UF management port for TCP access
+        run: |
```
crobert-1, line 313: I'm seeing some steps with the `shell` specified as `pwsh`, some not. Is there a reason some simple way to know when it should be powershell vs not?

internal/configsource/splunksecretconfigsource/README.md
```diff
@@ -0,0 +1,85 @@
+# Splunk Secret Config Source (Alpha)
```
crobert-1, line 1: I'd prefer formatting the README closer to upstream's format, eg moving to using the header table that includes the stability there (instead of the title). Some entries are irrelevant in the upstream table.

internal/configsource/splunksecretconfigsource/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Use the `splunk_secret` config source to retrieve secrets stored in Splunk's local
+[secret storage](https://dev.splunk.com/enterprise/docs/developapps/manageknowledge/secretstorage/secretstoragerest)
+(the `storage/passwords` REST endpoint served by splunkd on the management port,
+typically `8089`) and inject them into your collector configuration.
+
+This is primarily intended for use when the collector is launched as a Splunk TA
+modular input, where the `SPLUNK_SERVER_URI` and `SPLUNK_SESSION_KEY` environment
+variables are automatically set by the TA launcher shim (see
+`pkg/modularinput/ta_launcher.go`) from the `server_uri` and `session_key` fields
```
crobert-1: It'd be better for this to be a link rather than path, especially since it's not immediately clear if it's an absolute or relative path in the repo.

internal/configsource/splunksecretconfigsource/README.md
```diff
@@ -0,0 +1,85 @@
...
+Splunk provides on stdin. `endpoint` and `session_key` default to the values of
+these environment variables, so the config source works out-of-the-box when
+running as a TA modular input without any explicit configuration.
+
+## Configuration
```
crobert-1, line 14: I'd prefer to have the list of options, specifying what's required, description of options, default values, etc, then an actual config example, rather than the config example with a bunch of comments. This would better match upstream's format, and it's to know what's required vs. optional in the documented config.

internal/configsource/splunksecretconfigsource/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+A `<realm>:<name>` pair is only guaranteed unique within the Splunk user/app
+namespace it was created in, so different users and/or apps can each store a
+credential with the same realm and name. If a selector matches credentials
+owned by more than one user/app namespace, the lookup fails with an ambiguous
+selector error; set the `user` and/or `app` config fields to the owning
+user/app to disambiguate.
+
+```yaml
+components:
```
crobert-1: I think there should be one full config example in the README to show potential usage with other components. I'm having a hard time understanding what the actual usage would be from this README.

internal/configsource/splunksecretconfigsource/config.go
crobert-1, line 1: Please review comments introduced in this PR, we should only include comments that are non-obvious. (It's hard to follow the code right now with how many comments are here, at least for me)

internal/configsource/splunksecretconfigsource/factory.go
```diff
@@ -0,0 +1,111 @@
...
+		User:           defaultUser,
+		Timeout:        defaultTimeout,
+	}
+}
+
+func (f *splunkSecretFactory) CreateConfigSource(_ context.Context, settings configsource.Settings, logger *zap.Logger) (configsource.ConfigSource, error) {
+	cfg := settings.(*Config)
+
+	if cfg.Endpoint == "" {
+		return nil, &errMissingEndpoint{errors.New("cannot connect to splunkd without an endpoint")}
```
crobert-1: I think the error messages are a bit ambiguous the way they're currently written, specifically "without an endpoint" and "without a session_key". We should clarify that the user config does not set these configuration options, and that these configuration options are required. There's a slight chance here that the user isn't able to immediately connect the dots.

internal/configsource/splunksecretconfigsource/source.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, &errRequestFailed{fmt.Errorf("failed to read response body for selector %q: %w", selector, err)}
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, &errUnexpectedStatusCode{fmt.Errorf("splunkd returned status %d for selector %q: %s", resp.StatusCode, selector, string(body))}
```
crobert-1: I'm pretty concerned about leaking credentials accidentally, can you confirm that's not a risk with this error message (and the rest in this file?)

internal/configsource/splunksecretconfigsource/source.go
```diff
@@ -0,0 +1,152 @@
...
+		owners := make([]string, 0, len(parsed.Entry))
+		for _, entry := range parsed.Entry {
+			owners = append(owners, entry.ACL.Owner+":"+entry.ACL.App)
+		}
+		return nil, &errAmbiguousSelector{fmt.Errorf("selector %q matches credentials owned by multiple user:app namespaces %v; set the User and App config fields to disambiguate", selector, owners)}
+	}
+
+	clearPassword := parsed.Entry[0].Content.ClearPassword
+	if clearPassword == "" {
+		return nil, &errMissingClearPassword{errors.New("secret has no clear_password value")}
```
crobert-1: Should we be including which secret/selector is being referred to here? (Or will it be clear to the user?)

internal/configsource/splunksecretconfigsource/source.go
```diff
@@ -0,0 +1,152 @@
...
+type splunkSecretConfigSource struct {
+	client     *http.Client
+	logger     *zap.Logger
+	endpoint   string
+	sessionKey string
+	app        string
+	user       string
+}
+
+func newConfigSource(cfg *Config, app, user string, timeout time.Duration, logger *zap.Logger) *splunkSecretConfigSource {
```
crobert-1, line 81: I'm wondering if the current split of functionality between `CreateConfigSource` and `newConfigSource` is the right way to handle this. My main concern is that both methods are performing validation and modification of data. I guess I don't fully understand why they need to be separate methods, and if they do need to be separate, why is some logic in one method vs. the other?

internal/configsource/splunksecretconfigsource/source.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// Retrieve fetches the clear text value of the secret identified by selector from splunkd's
+// storage/passwords REST endpoint. The selector must be in the form "<realm>:<name>", e.g.
+// "myrealm:myuser". The realm segment can be empty (i.e. selector starting with ":") to
+// reference a secret stored without a realm.
+func (s *splunkSecretConfigSource) Retrieve(ctx context.Context, selector string, _ *confmap.Conf, _ confmap.WatcherFunc) (*confmap.Retrieved, error) {
+	realm, name, found := strings.Cut(selector, ":")
+	if !found || name == "" {
+		return nil, &errInvalidSelector{fmt.Errorf("selector %q must be in the form <realm>:<name>", selector)}
```
crobert-1: Since realm is allowed to be empty, is there some way to make this error message more clear (the way it's currently written seems to imply `realm` is required).

internal/configsource/splunksecretconfigsource/source.go
```diff
@@ -0,0 +1,152 @@
...
+	// https://help.splunk.com/en/data-management/splunk-enterprise-admin-manual/10.4/configuration-file-reference/10.4.1-configuration-file-reference/passwords.conf
+	//
+	// servicesNS/{user}/{app}/ pins the lookup to a specific user/app namespace. A
+	// <realm>:<name> credential is only guaranteed unique within the user/app namespace it
+	// was created in, so s.user and s.app both default to "-" (all users/apps, per the
+	// wildcard documented at https://help.splunk.com/en/?resourceId=Splunk_RESTREF_RESTlist)
+	// but can be set via the User/App config fields to disambiguate credentials that collide
+	// across namespaces. See the "Namespace" section at
+	// https://help.splunk.com/en/?resourceId=Splunk_RESTUM_RESTusing for details.
+	reqURL := fmt.Sprintf("%s/servicesNS/%s/%s/storage/passwords/%s?output_mode=json", s.endpoint, url.PathEscape(s.user), url.PathEscape(s.app), url.PathEscape(credentialID))
```
crobert-1: Maybe pull this out into its own method for testing? It would also be helpful with readability, since this long of a comment in the middle of a method makes it hard to parse mentally (for me at least)

internal/configsource/splunksecretconfigsource/config_test.go
crobert-1, line 1: For all tests: My understanding is that [t.Context](https://pkg.go.dev/testing#T.Context) should be used whenever possible for unit tests to match the proper scope of a given test's context. Please move from `context.Background` -> `t.Context` as much as possible.

(There's actually [a linter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d975955280b0345300af84e7a029004541fb89ab/.golangci.yml#L415) that `contrib` is using that [detects](https://github.com/ldez/usetesting/tree/3e637e559ec80411560a428dcf1bc5b5735976b6#contextbackground-go--124) usages of `context.Background` in tests and fails when found, that's the main reason I'm aware of this. We should probably figure out if we want to match contrib's linting so that it's actually automated for all PRs, not manually checked.)

internal/configsource/splunksecretconfigsource/source_test.go
```diff
@@ -0,0 +1,206 @@
...
+			if app == "" {
+				app = "-"
+			}
+			user := tc.user
+			if user == "" {
+				user = "-"
+			}
+			source := newConfigSource(cfg, app, user, time.Second, zap.NewNop())
+			retrieved, err := source.Retrieve(context.Background(), tc.selector, nil, nil)
+			if tc.checkErr != nil {
```
crobert-1, line 227: Tiny nit: It might be a bit easier to follow if the rest of this method was an if else, rather than having it be an if with a return inside. It concerned me at first that something was accidentally being missed by returning here.

.github/workflows/ta-v2.yaml
pjanotti: Yeah, I want to add that to KTLO: certainly doable and the workflows are getting pretty long anyway. Adding a KTLO task for this.

internal/configsource/splunksecretconfigsource/README.md
pjanotti: This is following the same pattern as other "config sources" in the repo

internal/configsource/splunksecretconfigsource/README.md
pjanotti: This is following the same pattern as other "config sources", that said I agree with you. I'll batch the suggestions to change the config sources as a single KTLO issue so we do it once.

Claude does a good rewrite in one go, but, for consistency I want to do this for all config sources at once:


Under `config_sources:` use `splunk_secret:` or `splunk_secret/<name>:` to create a
Splunk secret config source. When running as a TA modular input, no configuration
is necessary: `endpoint` and `session_key` default to the values provided by the
[TA launcher shim](../../../pkg/modularinput/ta_launcher.go), so an empty
`splunk_secret:` entry is enough. The following parameters are available:

| Parameter | Default | Description |
| --- | --- | --- |
| `endpoint` | `${env:SPLUNK_SERVER_URI}` | The splunkd management URI, e.g. `https://127.0.0.1:8089`. Defaults to the `SPLUNK_SERVER_URI` environment variable set by the TA launcher shim when running as a TA modular input. |
| `session_key` | `${env:SPLUNK_SESSION_KEY}` | Authenticates requests to the `storage/passwords` endpoint. Defaults to the `SPLUNK_SESSION_KEY` environment variable set by the TA launcher shim when running as a TA modular input. |
| `app` | `"-"` (all apps) | Scopes the `storage/passwords` lookup to a specific Splunk app namespace. A `<realm>:<name>` credential is only guaranteed unique within the app it was created in, so if multiple apps store a credential with the same realm and name, set this to the owning app to disambiguate. |
| `user` | `"-"` (all users) | Scopes the `storage/passwords` lookup to a specific Splunk user namespace. A `<realm>:<name>` credential can also be stored with user-level sharing, so if multiple users store a credential with the same realm and name, set this to the owning user to disambiguate. |
| `insecure_skip_verify` | `false` | Controls whether the TLS certificate presented by splunkd is verified. Certificate verification is enabled unless explicitly disabled. Splunk's management port typically presents a self-signed certificate, so set this to `true` only in trusted environments or after having installed a trusted certificate on splunkd. |
| `timeout` | `10s` | The maximum amount of time to wait for a response from splunkd. |

internal/configsource/splunksecretconfigsource/factory.go
pjanotti: What is a bit unusual here is that for the typical case, the TA, the user should never see this error and typical wouldn't need to create a configuration for the `splunk_secret`. That said the text can be improved together with the env var name.

internal/configsource/splunksecretconfigsource/README.md
pjanotti: The main usage is when deploying as a TA, for instance to monitor the Splunk PostgreSQL one could add the following receiver and automatically grab the needed credentials using the config source:

```yaml
receivers:
  postgresql:
    username: postgres_admin
    password: ${splunk_secret:postgres:postgres_admin}
```

internal/configsource/splunksecretconfigsource/source.go
pjanotti: This is following the pattern for config sources: `newConfigSource` allows the creation of a default config source that can be used in `Retrieve` tests without the full validation of config and handling of special cases that is implemented in `CreateConfigSource`.

.github/workflows/ta-v2.yaml
pjanotti: OTL-4507

.github/workflows/ta-v2.yaml
pjanotti: I consider both "Powershell" in general 😃  but `pwsh` is the latest version. I didn't look if it was needed for all of the job steps where it was added, but it is definitely needed for the `-SkipCertificateCheck` parameter used in the `Invoke-WebRequest` at https://github.com/signalfx/splunk-otel-collector/pull/7852/changes#diff-6202a2354f42619fcfc2bebf14ffb4f8022508233d9fbe5bc4785d076193699bR514 - After that I simply added to all new steps.

internal/configsource/splunksecretconfigsource/README.md
pjanotti: OTL-4510

internal/configsource/splunksecretconfigsource/README.md
pjanotti: OTL-4510

internal/configsource/splunksecretconfigsource/config.go
pjanotti: Pending...

internal/configsource/splunksecretconfigsource/README.md
pjanotti: 58804e879

internal/configsource/splunksecretconfigsource/README.md
pjanotti: 4bad8a8cf

internal/configsource/splunksecretconfigsource/factory.go
pjanotti: e8b363982

internal/configsource/splunksecretconfigsource/source.go
pjanotti: I can't guarantee what the REST API will return, but being paranoid and adding fadde8781

internal/configsource/splunksecretconfigsource/source.go
pjanotti: 5b359938b

internal/configsource/splunksecretconfigsource/source.go
pjanotti: dac197c1c

internal/configsource/splunksecretconfigsource/config_test.go
pjanotti: 0fcac1d38

internal/configsource/splunksecretconfigsource/source_test.go
pjanotti: 90b853856

internal/configsource/splunksecretconfigsource/source.go
pjanotti: 87449e5da

internal/configsource/splunksecretconfigsource/config.go
pjanotti: eab2d7d01

crobert-1 (approved):

PR #7857, opened 2026-07-31
[receiver/gnmireceiver] Add skeleton for the gNMI receiver

jkoronaAtCisco:

Introduces the skeleton for a new gNMI receiver under `internal/receiver/gnmireceiver`. gNMI (gRPC Network Management Interface) is the modern, streaming replacement for SNMP polling; this receiver will ingest network-device telemetry via gNMI Subscribe (dial-in) and emit OpenTelemetry metrics.

This PR is structure only — it establishes the component, wires it into the distribution, and passes lifecycle tests. It does not yet open connections or produce any telemetry. Configuration, the subscription/streaming engine, and metric conversion land in follow-up PRs.

Unit tests for the factory and config, plus the generated lifecycle test.

Metadata.yaml + README.md


internal/receiver/gnmireceiver/README.md
```diff
@@ -0,0 +1,26 @@
+<!-- status autogenerated section -->
```
crobert-1: Is this section actually being auto generated, or is this just following upstream formatting? If it's not actually being autogenerated in our repo, I'd prefer to keep the formatting (header, description, and table), but remove the header and footer so they're not misleading here.

internal/receiver/gnmireceiver/README.md
```diff
@@ -0,0 +1,26 @@
...
+# gNMI Receiver
+
+The gNMI receiver ingests streaming network telemetry from devices that expose a
+[gNMI](https://github.com/openconfig/gnmi) interface. It subscribes to the device
+(dial-in) and converts the streamed updates into OpenTelemetry metrics.
+
+| Status        |           |
```
crobert-1: The table itself is good to have here, but we need to remove information that's irrelevant or wrong (e.g. issues links to contrib, we don't have a breakdown of codecov by component in our distro, `splunk` doesn't have a link for the distribution, etc).

If the content is being generated by mdatagen incorrectly we should probably just have the table be manually maintained.

internal/receiver/gnmireceiver/metadata.yaml
```diff
@@ -0,0 +1,18 @@
+type: gnmi
+display_name: gNMI Receiver
+description: |
+  The gNMI receiver ingests streaming network telemetry from devices that expose a
+  [gNMI](https://github.com/openconfig/gnmi) interface. It subscribes to the device
+  (dial-in) and converts the streamed updates into OpenTelemetry metrics.
```
crobert-1: Is `dial-in` terminology used by gNMI itself? Receivers commonly use the wording of "push" vs. "pull" in terms of how they get metrics, so I'd want to use that wording as much as possible unless "dial-in" is especially meaningful in the context of gNMI. If dial-in is essential, we could also add push or pull to make it clear from the collector side.

crobert-1 (commented):
The actual go code and tests have lots of comments, most of which are unnecessary. Things like "this is not implemented yet" are fine with me, but please review others to see if they're actually valuable or not. (Unnecessary comments can make the code much harder to read IMO, so we should only include them when they're really explaining something we wouldn't understand otherwise.)

internal/receiver/gnmireceiver/metadata.yaml
jkoronaAtCisco: "dial-in" isn't gNMI spec terminology, so I've reworded to lead with "push-based streaming telemetry via the gNMI Subscribe RPC"

internal/receiver/gnmireceiver/README.md
jkoronaAtCisco: It was autogenerated, but I changed it to manual due to issues you [pointed out ](https://github.com/signalfx/splunk-otel-collector/pull/7857#discussion_r3691610560)

internal/receiver/gnmireceiver/README.md
jkoronaAtCisco: Changed to manual

internal/receiver/gnmireceiver/README.md
```diff
@@ -0,0 +1,19 @@
+# gNMI Receiver
+
+The gNMI receiver ingests push-based streaming network telemetry from devices via the
+[gNMI](https://github.com/openconfig/gnmi) `Subscribe` RPC, converting the streamed
+updates into OpenTelemetry metrics.
+
+| Status        |                                                          |
+| ------------- |----------------------------------------------------------|
+| Stability     | [development]: metrics                                   |
+| Distributions | [Splunk]                                                 |
```
crobert-1: After the latest revision `development` is no longer a link, we should keep it as a link to upstream's definition. We might want to make `Splunk` a link as well, just to this GitHub repo

internal/receiver/gnmireceiver/config.go
```diff
@@ -0,0 +1,26 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gnmireceiver
+
+import "go.opentelemetry.io/collector/component"
+
+// Config defines the configuration for the gNMI receiver.
```
crobert-1, line 19: Nit: I think this comment should also be removed.

internal/receiver/gnmireceiver/doc.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build mdatagen
+
+//go:generate mdatagen metadata.yaml
+
+// Package gnmireceiver implements a receiver that ingests network telemetry
+// from devices via gNMI Subscribe.
```
crobert-1, line 20: Nit: I don't think this is comment is valuable enough to keep

crobert-1 (commented):
A few more nits. BTW, if you think comments are helpful and I'm being overly aggressive in asking to remove them, please feel free to push back. If they're helping provide context that isn't obvious to others it'd be good for me to be more aware of that 👍

jkoronaAtCisco:
> A few more nits. BTW, if you think comments are helpful and I'm being overly aggressive in asking to remove them, please feel free to push back. If they're helping provide context that isn't obvious to others it'd be good for me to be more aware of that 👍

I was looking recently into many components upstream and those comments are quite common, but I'm also ok with removing them.

internal/receiver/gnmireceiver/README.md
jkoronaAtCisco: Good idea. Added links.

crobert-1 (approved):
LGTM once CI is passing 👍

PR #7863, opened 2026-07-31
[telegraf/logparser] delete the monitor

atoulme:
**Description:**
Delete the telegraf/logparser monitor, which has no support from Telegraf and is no longer in use.


.chloggen/delete_telegraf_logparser.yaml
```diff
@@ -0,0 +1,17 @@
...
+note: Delete the monitor.
+
+# One or more tracking issues related to the change
+issues: [7863]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Delete the telegraf/logparser monitor, which has no support from Telegraf and is no longer in use.
```
crobert-1, line 17: Nit: Maybe suggest any one of the upstream log receivers?

crobert-1 (approved):

.chloggen/delete_telegraf_logparser.yaml
atoulme: telemetry usage shows no use, and this is a weird one that we want folks to probably avoid and use our metricization pipelines with Ingest Processor instead.  I think we can leave this without replacement.

PR #7884, opened 2026-08-03
[chore] Remove use of bitnami images

agoallikmaa:
**Description:**
The only places where these images were referenced were unused Kafka Dockerfile setup and Prometheus image used in discovery tests. Removed the Kafka image related files, as tests have already been moved to using the stock apache kafka image, and switched Prometheus image to quay.io one.


tests/general/discoverymode/docker_observer_discovery_test.go
```diff
@@ -46,7 +46,7 @@ func TestDockerObserver(t *testing.T) {
 	require.NoError(t, err)
 
 	_, shutdownPrometheus := tc.Containers(
-		testutils.NewContainer().WithImage("bitnamilegacy/prometheus").WithLabel("test.id", tc.ID).WillWaitForLogs("Server is ready to receive web requests."),
+		testutils.NewContainer().WithImage("quay.io/prometheus/prometheus:v3.13.1").WithLabel("test.id", tc.ID).WillWaitForLogs("Server is ready to receive web requests."),
```
crobert-1, line 49: Ideally we'd be able to update this image version automatically. Maybe using an environment variable set in the GitHub workflow would work?

Otherwise, we _might_ be able to just specify latest and let it go until it starts causing too much noise.

agoallikmaa: Could move it to workflow config, but I don't think that really would make it automatic - it would just move the pinned version to GHA workflow definition. Downside of that would be that for running the test locally, it would have to be either mandatory to provide that env var, which is inconvenient, or have a default fallback. And the default fallback could then go out of sync with workflow one unless both are always bumped at the same time, which is just an additional thing to keep track of.

As for `latest`, it checks basically the whole set (at the time this test was created, additional ones probably wouldn't break the test) of metric names, their values, and their metadata, which sounds to me like it might be something that could have changes that could break this test. I suppose we could do that and just pin it back when it does actually break.

jinja2 (approved):

PR #7886, opened 2026-08-03
Support the featuregate subcommand

kelseyma:
Add support for the upstream `featuregate` subcommand that lists info about available feature gates and their status. Have subcommand follow version flag in immediate exit as parsing of config/additional setup isn't needed for output. Example use case for command: OpAMP Supervisor calls `otelcol featuregate` to determine if `service.AllowNoPipelines` is available so it can use that to start the bootstrap nop collector used to get the agent description.

- Added unit and integration tests
- Built local binary and verified output:
```
#  ./bin/otelcol featuregate -h
Display information about available feature gates and their status

Usage:
  otelcol featuregate [feature-id] [flags]

Flags:
  -h, --help   help for featuregate

#  ./bin/otelcol featuregate  
ID                                                                       Enabled  Stage   Description
cloudfoundry.resourceAttributes.allow                                    true     Beta    When enabled, envelope tags are copied to the metrics as resource attributes instead of datapoint attributes
confmap.enableMergeAppendOption                                          false    Alpha   Combines lists when resolving configs from different sources. This feature gate will not be stabilized 'as is'; the current behavior will remain the default.
...

#  ./bin/otelcol featuregate splunk.opamp.enabled
Feature: splunk.opamp.enabled
Enabled: true
Stage: Beta
Description: When enabled (default), the opamp/splunk_o11y extension is active. When disabled, the opamp/splunk_o11y extension is removed from the service.extensions configuration at startup, if it is present.
From Version: v0.151.0
To Version: v<nil>

#  ./bin/otelcol -h 
  otelcol [flags]
  otelcol [command]

Available Commands:
  featuregate  Display feature gates information
  validate     Validates the config without running the collector

Flags:
      --config string          Locations to the config file(s), note that only a single location can be set per flag entry e.g. --config=/path/to/first --config=path/to/second. (default "[]")
      --feature-gates string   Comma-delimited list of feature gate identifiers. Prefix with '-' to disable the feature. '+' or no prefix will enable the feature. (default "[]")
      --no-convert-config      Do not translate old configurations to the new format automatically. By default, old configurations are translated to the new format for backward compatibility.
      --set string             Set arbitrary component config property. The component has to be defined in the config file and the flag has a higher precedence. Array config properties are overridden and maps are joined. Example --set=processors.batch.timeout=2s (default "[]")
  -v, --version                Version of the collector.
```


.chloggen/add-featuregate-subcommand.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: otelcol
```
kelseyma, line 5: not sure if this is the correct component, can change based on feedback!

crobert-1 (commented):
Should `featuregate` be included in valid options/help menu when someone runs `otelcol --help`? Everything else looks good, but I guess I'd expect it to show up as a valid sub-option when running with the help flag.

kelseyma:
> Should `featuregate` be included in valid options/help menu when someone runs `otelcol --help`? Everything else looks good, but I guess I'd expect it to show up as a valid sub-option when running with the help flag.

Agreed, we overwrite the help output so originally I didn't change it in case we meant to not show cmds but agree it probably should be listed. Updated the help output to follow upstream format

crobert-1 (approved):

PR #7888, opened 2026-08-04
[chore][receiver/gnmireceiver] Implement gNMI Receiver configuration

jkoronaAtCisco:

Second phase of the gNMI receiver (follows the skeleton in #7857). This PR adds the full configuration schema and validation. It is configuration only — the receiver still returns a no-op and produces no telemetry; the connection/streaming engine and metric conversion land in follow-up PRs.

[README](https://github.com/signalfx/splunk-otel-collector/pull/7888/changes#diff-e5ace109f0ca02a73ad7f7293f379450ac2afadc7483fc1554b785ebad6a407d) Configuration section: full YAML example, `target/subscription` field tables, and a "metric type and unit resolution" subsection.


internal/receiver/gnmireceiver/README.md
```diff
@@ -10,10 +10,85 @@ updates into OpenTelemetry metrics.
...
-> added in follow-up changes.
+> **Note:** This receiver is in early development. Configuration is defined, but the
+> subscription/streaming engine and metric conversion are still being added, so the
+> receiver does not yet produce telemetry.
 
 ## Configuration
 
-Configuration options will be documented here as they are implemented.
+The receiver connects to one or more gNMI `targets`. Each target opens a `Subscribe` 
+stream and can carry multiple path `subscriptions`.
```
jinja2: I initially read "Each target opens a Subscribe stream" as meaning that the device initiates a connection to the collector.

My understanding from reading about the spec [here](https://openconfig.net/docs/gnmi/gnmi-specification/#35-subscribing-to-telemetry-updates) is that the receiver is the gNMI client and the collector opens an outbound, long-lived Subscribe RPC to each configured target, and the target then sends updates over that stream. If this is correct, could we reword this to make the connection direction clearer?

Perhaps:

> The receiver opens a long-lived gNMI Subscribe stream to each configured target. Each stream can contain multiple path subscriptions.

internal/receiver/gnmireceiver/README.md
```diff
@@ -10,10 +10,85 @@ updates into OpenTelemetry metrics.
...
+| `tls`           |           | Standard collector TLS client settings.                            |
+| `subscriptions` | (required)| One or more path subscriptions (below).                            |
+
+### Subscription
+
+| Field                | Default   | Description                                                                 |
+| -------------------- | --------- | --------------------------------------------------------------------------- |
+| `path`               | (required)| gNMI path to subscribe to.                                                  |
+| `origin`             |           | YANG model origin (e.g. `openconfig`).                                      |
+| `mode`               | (required)| `sample`, `on_change`, or `target_defined`.                                 |
```
jinja2: Could we clarify that this is the per-path subscription mode inside a `STREAM` SubscriptionList?

Perhaps the description could say:

> Per-path mode within a STREAM subscription: `sample`, `on_change`, or `target_defined`. The receiver does not support ONCE or POLL SubscriptionLists.

internal/receiver/gnmireceiver/config.go
```diff
@@ -14,13 +14,195 @@
...
+	return nil
+}
+
+func (s *SubscriptionConfig) Validate() error {
+	if s.Path == "" {
+		return errors.New("path is required")
+	}
+	switch s.Mode {
+	case modeSample:
+		if s.SampleInterval <= 0 {
```
jinja2, line 182: Reading the SAMPLE spec [here](https://openconfig.net/docs/gnmi/gnmi-specification/#35152-stream-subscriptions) and have two questions about `sample_interval` config.

For `sample`, the spec gives an interval of `0` a defined meaning - use the target’s lowest supported interval. Is requiring a positive value an intentional safety restriction for this receiver? That seems reasonable, but perhaps it should be documented as a difference from the protocol.

Looks like a non-zero `sample_interval` is currently accepted with `on_change` or `target_defined`. The spec says a target should reject `target_defined` when a sample interval is supplied. Since the streaming implementation is coming later, would it be better for this validation to reject a non-zero interval unless the mode is `sample`, rather than having the implementation send or silently ignore it?

internal/receiver/gnmireceiver/config.go
```diff
@@ -14,13 +14,195 @@
...
+	}
+
+	if len(t.Subscriptions) == 0 {
+		return errors.New("at least one subscription must be specified")
+	}
+	return nil
+}
+
+func (s *SubscriptionConfig) Validate() error {
+	if s.Path == "" {
```
jinja2, line 173: Do we plan to validate the configured path’s string syntax here, or will that happen when the subscription request is constructed in the follow-up implementation?

I realize that schema validity and whether the path currently exists must be handled at runtime by the target. I was only wondering whether malformed or relative path strings should fail during Collector configuration validation.

internal/receiver/gnmireceiver/config.go
```diff
@@ -14,13 +14,195 @@
...
+	ClientConfig  configgrpc.ClientConfig `mapstructure:",squash"`
+	// Redial is the delay before reconnecting after a session failure.
+	// Set to 0 to disable automatic reconnection.
+	Redial time.Duration `mapstructure:"redial"`
+}
+
+func NewDefaultTargetConfig() TargetConfig {
+	return TargetConfig{
+		ClientConfig: configgrpc.NewDefaultClientConfig(),
+		Encoding:     encodingProto,
```
jinja2, line 78: The gNMI spec defines JSON as the minimum encoding client/target should support. `proto`is probably more efficient, but is it broadly supported by targets? If it requires capability discovery and fallback, would JSON be a safer default?

internal/receiver/gnmireceiver/README.md
```diff
@@ -10,10 +10,85 @@ updates into OpenTelemetry metrics.
...
+Each target embeds the standard collector [gRPC client settings][configgrpc]
+(`endpoint`, `tls`, `keepalive`, `headers`, `compression`, ...) plus the fields below.
+
+| Field           | Default   | Description                                                        |
+| --------------- | --------- | ------------------------------------------------------------------ |
+| `endpoint`      | (required)| `host:port` of the gNMI target.                                    |
+| `username`      |           | Sent as gNMI gRPC metadata (not an `Authorization` header).        |
+| `password`      |           | Sent as gNMI gRPC metadata. Redacted in logs.                      |
+| `encoding`      | `proto`   | gNMI encoding: `proto`, `json`, or `json_ietf`.                    |
+| `redial`        | `10s`     | Delay before reconnecting after a session failure (min `1s`).      |
```
jinja2: Can we also add "Set to 0 to disable automatic reconnection." here?

internal/receiver/gnmireceiver/README.md
jkoronaAtCisco: You're right, I corrected description.

internal/receiver/gnmireceiver/README.md
jkoronaAtCisco: Agreed, I used your suggested descrption.

internal/receiver/gnmireceiver/README.md
jkoronaAtCisco: Done — added to the redial row so it matches the Go doc comment.

internal/receiver/gnmireceiver/config.go
jkoronaAtCisco: I've kept `proto` as the default because it's what our primary target platform (Cisco IOS-XE) commonly uses and it's more efficient on the wire, but I've documented the caveat: we don't do Capabilities negotiation, so users on targets without proto support need to set encoding explicitly.

internal/receiver/gnmireceiver/config.go
jkoronaAtCisco: I added some basic validation: a check that the path is absolute (starts with `/)` so malformed/relative paths fail during config validation. Deeper parsing will be implemented in followup PR.

internal/receiver/gnmireceiver/config.go
jkoronaAtCisco: Requiring `> 0` is intentional — with `0` the device picks the rate, which can produce a surprising volume of telemetry. Documented as a deviation in the new "Notes on gNMI specification behavior" section.

On non-sample modes you're right — `Validate()` now rejects a non-zero `sample_interval` unless `mode: sample`, so it fails at config time.

.chloggen/gnmi-receiver-config.yaml
```diff
@@ -0,0 +1,16 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: gnmireceiver
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add configuration schema and validation for the gNMI receiver (targets, gRPC transport, subscriptions). No telemetry is produced yet.
```
jinja2: I don't think we need a changelog entry since the receiver is not really usable yet and our changelog is user focused only

jinja2 (approved):

PR #7893, opened 2026-08-06
[receiver/gnmireceiver] Add gNMI connection engine

jkoronaAtCisco:

Third phase of the gNMI receiver. This PR adds the connection engine: the receiver now dials each configured target, opens a long-lived gNMI Subscribe stream, and reconnects on failure.

Metric conversion is not included — received responses are handed to a stub parser that returns no metrics, so the receiver still produces no telemetry. That lands in the next PR.

Added `receiver_test.go` that creates an in-process mock gNMI server and covers:
- start/shutdown
- shutdown without start 
- redial after a session failure 
- redial: 0 not retrying 
- a subscription error ending the session
- startup failing on an unparseable path
- keyed and malformed path parsing
- mode/encoding enum mappings.

`README` note updated — the receiver now maintains Subscribe streams, with metric conversion still pending.


internal/receiver/gnmireceiver/client.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	req, err := c.buildSubscribeRequest()
+	if err != nil {
+		return err
+	}
+	if err := stream.Send(req); err != nil {
+		return err
+	}
+	c.logger.Info("gNMI subscription established",
```
jinja2: Does `Send` confirm that the target accepted the subscription? Would “subscription request sent” be more accurate here, or should “established” be logged after the first response?

internal/receiver/gnmireceiver/client.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return err
+	}
+
+	req, err := c.buildSubscribeRequest()
+	if err != nil {
+		return err
+	}
+	if err := stream.Send(req); err != nil {
+		return err
```
jinja2: Should we handle `io.EOF` from `Send` by continuing to `Recv`? The grpc [doc](https://github.com/grpc/grpc-go/blob/v1.82.1/stream_interfaces.go#L110-L120) says a server-side error may surface from Send as EOF, with the actual RPC status available through Recv. I wonder if returning here could hide errors such as `Unauthenticated` or `InvalidArgument`.

internal/receiver/gnmireceiver/client.go
```diff
@@ -0,0 +1,214 @@
...
+	return nil
+}
+
+func (c *gnmiClient) subscribe(ctx context.Context) error {
+	streamCtx := ctx
+	if c.target.Username != "" {
+		streamCtx = metadata.AppendToOutgoingContext(ctx,
+			"username", string(c.target.Username),
+			"password", string(c.target.Password),
+		)
```
jinja2: Can you confirm if gnmi distinguishes username-only authorization from username+password authentication? An omitted password seems to be sent as `password: ""`, which may cause an empty-password authentication attempt. And should a password without a username also fail validation instead of being silently ignored?

internal/receiver/gnmireceiver/client.go
jkoronaAtCisco: You're right, `Send` only queues the request. I moved the log to after the first successful `Recv`.

internal/receiver/gnmireceiver/client.go
jkoronaAtCisco: Yes, I changed that. Now EOF from `Send` falls through to the receive loop and only client-generated errors return directly.

internal/receiver/gnmireceiver/client.go
jkoronaAtCisco: According to [gNMI specs](https://www.openconfig.net/docs/gnmi/gnmi-specification/#31-session-security-authentication-and-rpc-authorization) metadata must contain a username and may include a password. So I fixed it to match specs.

jinja2 (approved):
Not blocking, but I noticed that codecov is reporting 0% for the gnmi files even though we have unit tests covering them. I wonder we missing some setting which would make codecov register the new files.

crobert-1:
> Not blocking, but I noticed that codecov is reporting 0% for the gnmi files even though we have unit tests covering them. I wonder we missing some setting which would make codecov register the new files.

~I'm not sure if this will 100% solve the problem, but the component needs a Makefile in its local directory, [like other components](https://github.com/signalfx/splunk-otel-collector/blob/main/pkg/processor/timestampprocessor/Makefile). Any `make for-all` target currently skips this directory.~ Never mind, I think I'm confused about how we handle go packages in our repo.

jkoronaAtCisco:
> Not blocking, but I noticed that codecov is reporting 0% for the gnmi files even though we have unit tests covering them. I wonder we missing some setting which would make codecov register the new files.

It should be fixed by #7914

PR #7897, opened 2026-08-07
Add OpAMP Supervisor and launcher to Linux packages

kelseyma:
Update the Linux amd64 and arm64 packages to include the new `otelcollauncher` and `opampsupervisor` binaries.
Service will now start the launcher instead of the collector so that users can enable the supervisor if desired. 
- By default, service will continue with just starting the collector so users should not experience any impact. 
- When OpAMP Supervisor is enabled, it will manage the collector process and allow for remote configuration.

Windows packaging and mass deployment updates will follow in separate PRs to keep PRs easier to review. Targeting all these changes to go in next collector release

- Updated package tests to cover both direct collector setup and opamp supervisor managed setup
- Local dev testing with greenfield install (flag enablement), brownfield upgrade (env var enablement), install without supervisor flag
- Also see #7578 for previous PRs with testing performed on launcher/supervisor

Remote config feature docs will be worked on by docs team


.chloggen/package-linux-opamp-supervisor.yaml
```diff
@@ -0,0 +1,22 @@
...
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  By default, the launcher runs the collector directly, preserving existing service behavior.
+  When enabled, the launcher starts the OpAMP Supervisor to manage the collector process
+  and apply configuration updates from Splunk Observability Cloud.
+  Enable supervisor mode by either:
+  - During install, pass `--with-supervisor` to the installer script
+  - After upgrade, set `SPLUNK_OPAMP_SUPERVISOR_ENABLED=true` in the service environment and restart the service
```
kelseyma: not sure how much we wanted to advertise this since we will likely release before the server? Technically users could enable it before the server is updated and things will still work, opamp server will just ignore the extra remote config capabilities

packaging/fpm/etc/otel/collector/splunk-otel-collector.conf.example
```diff
@@ -15,6 +15,10 @@
 # Run `/usr/bin/otelcol --help` to see all available options.
 OTELCOL_OPTIONS=""
 
+# Set to true to manage the collector with the OpAMP Supervisor.
+# Leave false to run the collector directly.
+SPLUNK_OPAMP_SUPERVISOR_ENABLED=false
```
kelseyma, line 20: not sure if it was necessary to include in the example file, can remove if not needed!

packaging/fpm/postinstall.sh
```diff
@@ -22,10 +22,15 @@ if command -v setcap >/dev/null 2>&1; then
     # See: https://man7.org/linux/man-pages/man7/capabilities.7.html
     # Note: +eip gives the capability to the program regardless of who runs it.
     setcap CAP_SYS_PTRACE,CAP_DAC_READ_SEARCH=+eip /usr/bin/otelcol
+    # The launcher and supervisor need CAP_DAC_READ_SEARCH to maintain existing
+    # collector capability to read protected collector configurations.
+    setcap CAP_DAC_READ_SEARCH=+eip /usr/bin/otelcollauncher
+    setcap CAP_DAC_READ_SEARCH=+eip /usr/bin/opampsupervisor
```
kelseyma, line 28: added these mostly to maintain parity with otelcol, since both binaries need to read the config files passed in. But if it's not likely that customers are passing in custom configs that we don't have permissions to read, then this could probably be removed? 
Not sure if this was actually just needed for host_metrics receiver?

packaging/fpm/postinstall.sh
```diff
@@ -44,6 +49,11 @@ if [ -d /etc/otel/collector ]; then
     chown -R "$service_user:$service_group" /etc/otel/collector
 fi
 
+# /var/lib/otelcol is shared by collector file storage and supervisor state.
+# Preserve its contents while ensuring the active service owner can access it.
+mkdir -p "$state_dir"
+chown -R "$service_user:$service_group" "$state_dir"
```
kelseyma, line 55: ensuring the base state dir is created and set to the correct ownership so that supervisor can be enabled after install/upgrade without requiring a separate dir creation step (since that needs root). Supervisor creates a subdir when running for persistence files

packaging/installer/install.sh
```diff
@@ -1700,6 +1707,11 @@ parse_args_and_install() {
     access_token=$(request_access_token)
   fi
 
+  if [ "$with_supervisor" = "true" ] && [ -z "$access_token" ]; then
+    echo "[ERROR] --with-supervisor requires a Splunk Observability Cloud access token." >&2
```
kelseyma, line 1711: added this check so users can't accidentally enable just the platform flags + supervisor, since currently opamp server only exists in o11y and we have the opamp extension setup to report to o11y

Makefile
```diff
@@ -317,18 +317,14 @@ binaries-darwin_arm64:
 .PHONY: binaries-linux_amd64
 binaries-linux_amd64:
 	GOOS=linux   GOARCH=amd64 $(MAKE) otelcol
-ifeq ($(WITH_OPAMP_SUPERVISOR), true)
 	GOOS=linux   GOARCH=amd64 $(MAKE) otelcollauncher
 	GOOS=linux   GOARCH=amd64 $(MAKE) opampsupervisor
```
kelseyma: confirmed that no legal review is needed for packaging supervisor since we are just using the upstream package (just building locally)

tests/package/package_test.go
```diff
@@ -138,9 +204,21 @@ func TestCollectorPackageUpgrade(t *testing.T) {
 				copyFileToContainer(t, container, pkgPath)
 				upgradePackage(t, container, packageType, "/test/"+filepath.Base(pkgPath))
 
+				assertPackagedBinaries(t, container)
 				require.Eventually(t, func() bool {
 					return serviceIsRunning(t, container)
 				}, 20*time.Second, time.Second)
+				require.False(t, processIsRunningAs(t, container, serviceOwner, supervisorProcess))
+
+				// Existing installations can opt in to supervisor mode after upgrading.
```
jinja2: I’m worried about the brownfield path here.

Can we add a latest-released (this one does against 0.35 but one against the last version which did not have the launcher would be good too) -> PR package upgrade case while preserving the existing environment and configuration files? Maybe a test to enable the supervisor, then disable it again, restart, and assert that we are back to one direct `otelcol` process with no supervisor. That is the rollback path I assume we'll give users?

packaging/installer/install.sh
```diff
@@ -1119,6 +1120,8 @@ Collector:
                                         Specify this option to skip this step and use a pre-configured repo on the
                                         target system that provides the 'splunk-otel-collector' deb/rpm package.
   --test                                Use the test package repo instead of the primary.
+  --with-supervisor                     Whether to manage the Splunk OpenTelemetry Collector with OpAMP Supervisor.
```
jinja2, line 1123: I don’t see an end-to-end test for the `--with-supervisor` installer path. I noticed the pkg tests enable it by the env file after installation, but no coverage for the arg.

Can we install the PR-built deb/rpm through `install.sh --with-supervisor` and assert that the flag is persisted and both the supervisor and collector start successfully? This is the documented new-install entrypoint, so I’d like it tested directly rather than only testing the equivalent manual environment change.

packaging/installer/install.sh
jinja2: With this check, the installer will still accept O11y Cloud + Splunk Platform + `--with-supervisor`. If this is a supported combination, can we add an end-to-end test for it? If it is not supported, I think we should reject it explicitly.

packaging/fpm/postinstall.sh
jinja2: Doesn't the launcher read the original collector configs before it execs the supervisor? If so, I would keep line 27.

I am not sure about `opampsupervisor`. Do you know of any protected file it may need to read? What about tls cert files or something?

packaging/fpm/postinstall.sh
jinja2: Minor observation - this recursively walks and changes ownership of `/var/lib/otelcol` on every install or upgrade. That directory can contain a large filelogs store, so this could make an otherwise small package upgrade unexpectedly expensive.

Since supervisor creates `/var/lib/otelcol/supervisor` itself (i assume at 0700), would making the parent directory writable by the service user be enough without recursively touching existing collector state? I am wondering if a non-recursive `chown` here and in the matching installer code would be sufficient.

If recursively re-owning existing state is intentional, I think that upgrade behavior should at least be called out in the release or upgrade notes.

.chloggen/package-linux-opamp-supervisor.yaml
jinja2: Could we keep the changelog entry succinct and link to the user-facing documentation?

The Fleet Management docs should explain what enabling the supervisor provides and why someone would use it. For the Linux installation paths, I think the official docs should also cover:

- Add `--with-supervisor`, its requirements, and the default-off behavior to the Linux installer page.
- Update the manual deb/rpm installation documentation with the new `otelcollauncher` and `opampsupervisor` binaries, the systemd entrypoint change, capabilities, and generated config/state locations.
- Note that the tar bundle now includes the launcher and supervisor binaries, while making clear that it does not provide the same package-managed systemd setup.
- Update the Linux upgrade guide to explain that existing installations remain in direct mode by default, how to opt in, and how to return to direct mode.

I’m not sure how much of the lower-level package detail belongs in the existing manual installation page, so we can trim that list or move the operational details into a dedicated supervisor page.

If the docs update is planned for later, I would suggest an upgrading guide would be better than noting the details in the changelog. I do think we should mention the entrypoint change and the rollback path somewhere.

packaging/fpm/etc/otel/collector/splunk-otel-collector.conf.example
jinja2: we can keep it

.chloggen/package-linux-opamp-supervisor.yaml
kelseyma: They are starting to work on the docs update, should hopefully go out with the server release but the current prod docs keep the OpAMP enablement separate from normal collector docs: https://help.splunk.com/en/splunk-observability-cloud/manage-data/manage-otel-agents-and-collectors/manage-opentelemetry-agents-and-collectors/set-up-opentelemetry-fleet-management 

I wasn't sure if we were going to add additional sections here or if it was preferred to update the existing install pages? But I agree it's probably cleaner to link to user-facing docs if possible instead of adding this in the changelog

packaging/fpm/postinstall.sh
kelseyma: Yes launcher always reads the original collector configs, but for `opampsupervisor`, it depends. If we didn't need to sanitize the configs (aka remove the unneeded `opamp/splunk_o11y` extension) then supervisor does end up reading the original config files and would in that case need the additional capability

packaging/installer/install.sh
kelseyma: Yes but this is more of a temporary solution we were asked to support until splunk cloud has their management platform in place, at which time users will be migrated over

tests/package/package_test.go
kelseyma: yes good point, I covered this with my local dev testing but would be good to add a formal test here

packaging/installer/install.sh
kelseyma: I looked into this and found that for our current linux installer script tests, we're using the current prod package and not the locally built package. Do you know if it's preferred to keep those with the prod package and all local package testing in the other `linux-package-test` workflow? 
Wondering as in order to support the local package in installer tests, we would need to add the compile/build steps which would increase test time. Otherwise I could add it into the package test workflow and copy over the `install.sh` like `TestCollectorPackageUpgrade` and `TestCollectorPackageUpgradeWithCustomServiceOwner` currently do

packaging/fpm/postinstall.sh
kelseyma: Oh that's a really good callout on perf, I didn't realize the size `filelogs` could be. The chown is a case of how much do we want to guard against edge cases with custom users. Since supervisor needs to write files in `/var/lib/otelcol/supervisor` (you're correct on the `0700`), I think the main concern here was if customers decide to change their custom user during upgrade. Then the dir would still have the old user permissions and supervisor would fail to write the files (which all use `0600`). 
However, I guess we would've run into this already with the `filelogs` dir so if we haven't received a lot of issues  about this, can we assume this shouldn't be a common case? Then I could just change to make sure the parent `var/lib/otelcol` dir is writable as suggested

jinja2: I might have overstated it, the filelogs dir has the offsets from file_logs receiver. Your custom-user upgrade case makes sense to me. Given that supporting that transition is intentional, the recursive ownership change seems reasonable to me.

packaging/fpm/postinstall.sh
jinja2: Lets keep it then

packaging/installer/install.sh
jinja2: Adding it to the package test workflow and copying in `install.sh` should be good. Looks like `--collector-version` treats an existing file path as a local package so the test can pass the PR-built deb/rpm path with `--skip-collector-repo`.

tests/package/package_test.go
kelseyma: test added!

packaging/installer/install.sh
kelseyma: test added!

packaging/installer/install.sh
kelseyma: test added!

packaging/fpm/postinstall.sh
kelseyma: noted recursive ownership change in upgrade guideline

.chloggen/package-linux-opamp-supervisor.yaml
kelseyma: Sorry totally misread your initial comment, on rereading made these changes:
- simplified changelog entry to just point to upgrade guide (not sure timing of docs so figured it would be best to have something here first)
- added upgrade guide (have some questions on amount of detail for this, left some comments on the new changes)
- will be working with docs to tell them all the existing Linux install/upgrade pages that need updating as well as fleet management docs to add supervisor details

README.md
```diff
@@ -160,6 +160,26 @@ manually before the backward compatibility is dropped. For every configuration u
 [the default agent config](https://github.com/signalfx/splunk-otel-collector/blob/main/cmd/otelcol/config/collector/agent_config.yaml)
 as a reference.
 
+### From 0.158.0 to 0.159.0
```
kelseyma, line 163: Not sure about level of detail here? tried to give enough for context but not make this section too long (as soon we will also have windows side too). didn't note tar here for that reason but could add a small note if needed

README.md
```diff
@@ -160,6 +160,26 @@ manually before the backward compatibility is dropped. For every configuration u
...
+
+Linux DEB and RPM packages now use `otelcollauncher` as the service entrypoint instead of
+`otelcol`. By default, the launcher starts `otelcol` directly and passes through the existing `OTELCOL_OPTIONS`,
+so upgrades preserve the previous Collector service behavior. The launcher allows the service to start either `otelcol`
+directly or the OpAMP Supervisor, enabling additional Fleet Management capabilities. See
+[OpenTelemetry Fleet Management](https://help.splunk.com/en/splunk-observability-cloud/manage-data/manage-otel-agents-and-collectors/manage-opentelemetry-agents-and-collectors)
```
kelseyma, line 169: not sure if we want to link to docs if the docs updates could potentially come after our release? but wasn't sure how else to avoid putting too many details here (for both fleet management and install paths)

tests/package/package_test.go
jinja2, line 381: I might have missed this, do we have an assert for `/var/lib/otelcol` ownership with custom user/grp? If not, can we add this too?

.chloggen/package-linux-opamp-supervisor.yaml
```diff
@@ -0,0 +1,18 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: new_component
```
jinja2, line 2: I don’t think `new_component` is enough on its own. The supervisor is new and opt-in, but every existing DEB/RPM installation also changes to the launcher systemd entrypoint on upgrade.

Could we keep this entry and add a short `enhancement` entry for that existing behavior change? Something like and just link to the readme details for further info:

`Linux DEB and RPM services now start through otelcollauncher; direct Collector mode remains the default.`

I don’t think this needs to be marked as breaking.

README.md
jinja2: Its ok either ways, we can remove the installer links. But keep fleet management? I assume there is some relevant info about just what this about in there?

jinja2 (approved):
@kelseyma please resolve the conflicts. I have left some minor feedback but otherwise this LGTM

jinja2 (commented):
@kelseyma Could you also confirm that FIPS-enabled mode still works with the new launcher entrypoint? Existing Linux installations using `GODEBUG=fips140=on` should continue to start `otelcol` through `otelcollauncher` without changing the FIPS behavior. I couldn’t find a Linux/package test covering the FIPS path. Is this covered elsewhere, or should we add a small check here?

README.md
kelseyma: Fleet management doesn't have the supervisor (write path) specific info yet, currently it only has the read path (opamp extension only approach) info there. I'm working with the docs team now on the changes needed, but those will probably come at the same time as the installer updates. We can have the fm docs point to the collector docs on how to install so does make sense to remove installer links

kelseyma:
> Could you also confirm that FIPS-enabled mode still works with the new launcher entrypoint? Existing Linux installations using `GODEBUG=fips140=on` should continue to start `otelcol` through `otelcollauncher` without changing the FIPS behavior. I couldn’t find a Linux/package test covering the FIPS path. Is this covered elsewhere, or should we add a small check here?

@jinja2 Yeah that's a good callout I don't think there are any package level tests for FIPS. In terms of test coverage, I was thinking we could update `TestCollectorPackageRollbackSupervisorMode` to use fips (covers upgrade to direct launcher mode, then supervisor, then rollback) and avoid adding another full test to the matrix (I noticed the new tests have added ~10min to the linux package test runtime so far). But let me know if we'd prefer to keep it as a new separate test

.chloggen/package-linux-opamp-supervisor.yaml
kelseyma: added!

tests/package/package_test.go
kelseyma: added!

jinja2:
> > Could you also confirm that FIPS-enabled mode still works with the new launcher entrypoint? Existing Linux installations using `GODEBUG=fips140=on` should continue to start `otelcol` through `otelcollauncher` without changing the FIPS behavior. I couldn’t find a Linux/package test covering the FIPS path. Is this covered elsewhere, or should we add a small check here?
> 
> @jinja2 Yeah that's a good callout I don't think there are any package level tests for FIPS. In terms of test coverage, I was thinking we could update `TestCollectorPackageRollbackSupervisorMode` to use fips (covers upgrade to direct launcher mode, then supervisor, then rollback) and avoid adding another full test to the matrix (I noticed the new tests have added ~10min to the linux package test runtime so far). But let me know if we'd prefer to keep it as a new separate test

Updating the existing rollback test works for me

README.md
```diff
@@ -160,6 +160,26 @@ manually before the backward compatibility is dropped. For every configuration u
...
+for details. 
+
+For a new installation, pass `--with-supervisor` to the installer script. To enable the OpAMP Supervisor after upgrading, set
+`SPLUNK_OPAMP_SUPERVISOR_ENABLED=true` in `/etc/otel/collector/splunk-otel-collector.conf` and restart the service.
+
+To stop running the Collector under OpAMP Supervisor, set `SPLUNK_OPAMP_SUPERVISOR_ENABLED=false` and restart the
```
dmitryax, line 175: Should we clarify where to set it?

dmitryax (approved):

README.md
kelseyma: will clarify in the followup windows PR

PR #7917, opened 2026-08-12
Document host_observer Windows privilege requirements in discovery README

pmateos-cisco:
`host_observer` already works on Windows via `gopsutil` (`iphlpapi.dll`), which I verified by running it against real listeners on a Windows host. This documents that in the discovery README, along with the one real gap: resolving process name/command line for endpoints owned by other accounts (most built-in Windows services) requires the Collector to run as `LocalSystem` or with `SeDebugPrivilege` enabled, otherwise those endpoints are dropped entirely. The default MSI install already runs as `LocalSystem`, so this mainly matters for non-default setups.


pjanotti (approved):

pjanotti:
@pmateos-cisco could you please "sign" the CLA per https://github.com/signalfx/splunk-otel-collector/pull/7917#issuecomment-5263726500 - after that we can merge your PR. Thanks for your contribution!

PR #7918, opened 2026-08-12
[receiver/gnmireceiver] Convert gNMI updates into metrics

jkoronaAtCisco:

This PR replaces the stub parser with the real conversion, so the receiver now produces telemetry.

Value conversion

- Decodes `uint`, `int`, `float`, `double`, `bool`, `string`, JSON/JSON-IETF and leaf-lists.
- Integer values are emitted as integer datapoints rather than doubles, so large counters keep full precision.
- Booleans become `1`/`0`.

Naming and attributes

- Metric names are the gNMI path elements joined with dots, prefixed with the model origin when the target supplies one: `/interfaces/interface[name=eth0]/state/counters/in-octets` → `openconfig.interfaces.interface.state.counters.in-octets`.
- `notification.Prefix` is joined with each update's path before naming, since targets routinely split the path across the two. Keys from both are merged.
- Path keys become datapoint attributes (`name=eth0`) rather than part of the metric name; the target endpoint is recorded as the `server.address` resource attribute, and the scope name comes from the generated metadata.

Type and unit resolution

gNMI carries a value's data type but not whether it is a counter or a gauge, nor its unit — that lives in the YANG schema. Rather than inferring it, the parser resolves both from the subscription configuration: a leaf is looked up in `overrides`, falling back to `default`. A leaf matching neither is dropped. Where subscriptions overlap, the longest matching path prefix wins so the most specific configuration applies.

Non-numeric and structured values

- Configured string leaves become info metrics: an `_info` name suffix, a constant value of `1`, and the string in a `value` attribute. Unconfigured strings are dropped like any other unconfigured leaf.
- JSON payloads are flattened to one metric per leaf; nested object keys extend the name, while array elements keep the name and record their position in an `index` attribute. Leaf-lists are flattened the same way.

- unit tests provided
- additionally verified manually against a Cisco IOS-XR device (DevNet sandbox)

README gains a "Metric names and attributes" section covering naming, integer precision, info metrics and JSON/leaf-list flattening.


internal/receiver/gnmireceiver/parser.go
```diff
@@ -15,28 +15,335 @@
...
+	dp.SetIntValue(1)
+	dp.SetTimestamp(ts)
+	dp.Attributes().PutStr(infoValueAttr, value)
+	putAttrs(dp.Attributes(), keys)
+}
+
+func (p *metricParser) newNumberDataPoint(
+	sm pmetric.ScopeMetrics, origin string, elems []string, cfg MetricConfig,
+) pmetric.NumberDataPoint {
+	m := sm.Metrics().AppendEmpty()
```
jinja2, line 252: Non blocking, just wondering if there is a reason to not reuse one `Metric` per identity and append datapoints for each interface or array element. It would avoid repeating the same metric metadata and follows the general rec to aggregate identical `Metric` objects.

internal/receiver/gnmireceiver/README.md
```diff
@@ -77,6 +75,32 @@ Each target embeds the standard collector [gRPC client settings][configgrpc]
 | `default`            |           | Metric `type`/`unit` applied to leaves not matched by `overrides`.          |
 | `overrides`          |           | Map of leaf name → metric `type`/`unit`, taking precedence over `default`.  |
 
+### Metric names and attributes
```
jinja2, line 78: Not suggesting this needs to be part of the receiver now, but could the existing per-leaf override be extended later with optional output mappings? I see a lot of overlap in existing nw conventions and the counters read with this receiver.

For e.g.:

```yaml
overrides:
  in-octets:
    type: sum
    unit: By
    name: system.network.io
    attributes:
      network.io.direction: receive
    attribute_mappings:
      name: network.interface.name
```

Existing type/unit configs could continue producing raw path-derived metrics, while users who want otel semantic conventions could opt into the additional mapping (they can transform these in processor ofc).

internal/receiver/gnmireceiver/parser.go
```diff
@@ -15,28 +15,335 @@
...
+	val *gnmipb.TypedValue,
+	ts pcommon.Timestamp,
+) error {
+	if val == nil {
+		return nil
+	}
+
+	switch v := val.GetValue().(type) {
+	case *gnmipb.TypedValue_UintVal:
+		p.emitInt(sm, origin, elems, keys, int64(v.UintVal), ts) //nolint:gosec // G115: gNMI counters are uint64;
```
jinja2: Do we need overflow handling here? What happens if unit value is greater that MaxInt64?

internal/receiver/gnmireceiver/parser_test.go
```diff
@@ -0,0 +1,528 @@
...
+// default and a couple of leaf overrides.
+func countersSubscription() SubscriptionConfig {
+	return SubscriptionConfig{
+		Path:    "/interfaces/interface/state/counters",
+		Mode:    modeSample,
+		Default: &MetricConfig{Type: metricTypeSum, Unit: "1"},
+		Overrides: map[string]MetricConfig{
+			"in-octets":      {Type: metricTypeSum, Unit: "By"},
+			"in-octets-rate": {Type: metricTypeGauge, Unit: "By/s"},
+			"oper-status":    {Type: metricTypeGauge},
```
jinja2, line 43: I may be misunderstanding this mapping, but [oper-status](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-interfaces.html#interfaces-interface-state-oper-status) seems close to an otel status metric. We usually emit every possible state, with the active state set to `1` and the others to `0` (this helps in visualizing transitions). Here it looks like we only emit the current state as `1`. After `UP` changes to `DOWN`, does the receiver "clear" the previous `UP=1` timeseries?

internal/receiver/gnmireceiver/README.md
```diff
@@ -77,6 +75,32 @@ Each target embeds the standard collector [gRPC client settings][configgrpc]
...
+
+Path keys are not part of the name; they become datapoint attributes (`name=eth0`
+above). The target endpoint is recorded as the `server.address` resource attribute.
+
+Integer values (including `counter64`) are emitted as integer datapoints so large
+counters keep full precision. Booleans are emitted as `1`/`0`.
+
+Non-numeric (string) values cannot be represented as a metric value, so a configured
+string leaf is emitted as an *info metric*: the name gains an `_info` suffix, the
+datapoint value is always `1`, and the string is carried in the `value` attribute.
```
jinja2, line 103: I was reading [this blog](https://xrdocs.io/8000/tutorials/cisco-8000-npu-traps-telemetry-gnmi#telemetry-configuration) to get some understanding of existing metrics collection options. It has a warning about JSON IETF representing numeric type as strings. Would a counter like `"in-octets": "123"` reach `emitInfo` and become an `_info` metric, even when configured as a Sum in this case?

internal/receiver/gnmireceiver/parser.go
```diff
@@ -15,28 +15,335 @@
...
+		sum := m.SetEmptySum()
+		sum.SetIsMonotonic(true)
+		sum.SetAggregationTemporality(pmetric.AggregationTemporalityCumulative)
+		return sum.DataPoints().AppendEmpty()
+	}
+	return m.SetEmptyGauge().DataPoints().AppendEmpty()
+}
+
+func (p *metricParser) resolve(elems []string) (MetricConfig, bool) {
+	sub := p.subscriptionFor(elems)
```
jinja2: If two subscriptions use the same path under different origins, would they both match the same config and whichever appears first provides the type/unit? I don't see origin being referenced when selecting the configured type

jkoronaAtCisco: That was an oversight; thanks for pointing it out.

internal/receiver/gnmireceiver/README.md
jkoronaAtCisco: That's how I originally implemented it, but it makes more sense to resolve it from type. So now JSON/gNMI string is parsed rather than downgraded.

internal/receiver/gnmireceiver/parser_test.go
jkoronaAtCisco: Right now the receiver has no knowledge of a leaf's possible enum values. But it will be introduced with YANG schemas handling. Also I could add `enum_values` option to `MetricConfig`. But I would rather do it in follow up PR. WDYT?

internal/receiver/gnmireceiver/README.md
jkoronaAtCisco: I think that's good idea that fits current config model. I will add it as a followup.

internal/receiver/gnmireceiver/parser.go
jkoronaAtCisco: Current per-update model is simpler and cheap at this scale, but I think we can add batching across whole `parse()`. Should I add it to this PR or in followup?

internal/receiver/gnmireceiver/parser.go
jkoronaAtCisco: Added overflow handling.

internal/receiver/gnmireceiver/parser.go
jinja2: can be a follow up

internal/receiver/gnmireceiver/parser_test.go
jinja2: Sounds good to me. Can we just make a note for how an enum is handled? This e.g. transition showing what the output metrics will be should be good enough.

jinja2 (approved):
I have an open request to document something, otherwise LGTM

PR #7925, opened 2026-08-13
[receiver/gnmireceiver] Add integration test and DevNet sandbox testing guide

jkoronaAtCisco:

Fixes `gNMI receiver` silently overwriting datapoint attributes when two path elements reuse the same key name. Colliding keys are now namespaced by their owning element (e.g. `subinterface.index`)

Adds a container-based integration test for the gNMI receiver, plus a guide for manually validating it against a real device via the [Cisco DevNet](https://devnetsandbox.cisco.com/DevNet) sandbox.

#### Integration test

`TestGNMIReceiverPipeline` (`tests/receivers/gnmi/gnmi_pipeline_test.go`) runs the receiver inside a real collector container and compares the OTLP-exported metrics against a golden file (`testdata/pipeline_expected.yaml`).

The target is the OpenConfig fake gNMI agent (`github.com/openconfig/gnmi/testing/fake/gnmi/cmd/fake_server`), built from `testdata/server/Dockerfile` — the upstream reference implementation for testing against a gNMI collector, not a hand-written mock. It streams a fixed set of `SubscribeResponse` messages (`testdata/server/config.pb.txt`), so the expected metrics are pinned by the fixture rather than by whatever a live device reports. 

This covers: 
- metric naming (notification prefix rejoined with the update path),
- config-driven type/unit resolution, 
- path keys as datapoint attributes, 

#### DevNet testing guide

`TESTING_GUIDE.md` documents the manual procedure for validating the receiver against a real device (Cisco DevNet IOS XRv 9000 sandbox). `testdata/devnet-xrv9k.yaml` is the example config used in that procedure.


dloucasfx (approved):

internal/receiver/gnmireceiver/testdata/devnet-xrv9k.yaml
```diff
@@ -0,0 +1,119 @@
...
+        password: C1sco12345
+        # The device is configured with "no-tls", so request proto over plaintext.
+        encoding: proto
+        redial: 30s
+        tls:
+          insecure: true
+        subscriptions:
+          # OpenConfig interface counters. IOS-XR has good OpenConfig coverage,
+          # but confirm the leaves exist before relying on them (see notes below).
+          - path: /interfaces/interface/state/counters
```
jinja2, line 43: where will the interface usually be in such a counter? would `/interfaces/interface[name=eth0]/state/counters` be more realistic?

jinja2: oh, i thought this was a test fixture but looks like this is just the collector config. But now I am thinking, can we have a counter with multiple elements in path as one of the fixtures?

jkoronaAtCisco: Sure, I added multi-element keyed path fixture.

tests/receivers/gnmi/testdata/pipeline_expected.yaml
```diff
@@ -0,0 +1,61 @@
...
+            stringValue: 127.0.0.1:34847
+    scopeMetrics:
+      - metrics:
+          - name: openconfig.interfaces.interface.state.counters.in-octets
+            sum:
+              aggregationTemporality: 2
+              dataPoints:
+                - asInt: "9007199254740993"
+                  attributes:
+                    - key: name
```
jinja2, line 15: Since the element keys are so generically named (`name`, `index`), I am wondering if these paths can have repeated key names at different path levels. How will the receiver flatten these conflicting attributes in this case?

jkoronaAtCisco: It's a real bug that I missed earlier.  Fixed it in `joinPath`: key names that occur exactly once keep their plain name. Key names that occur more than once are namespaced by their owning element so they no longer collide.

jinja2:
@jkoronaAtCisco can you please update the pr description and add changelog for the attr fix if this receiver is shipped in components?

jinja2 (approved):

PR #7927, opened 2026-08-14
[receiver/gnmireceiver] Merge same-named metrics within a batch

jkoronaAtCisco:

Follow-up to #7918, addressing review comment [#discussion_r3769566400](https://github.com/signalfx/splunk-otel-collector/pull/7918#discussion_r3769566400):
> non-blocking, but asked whether the parser could reuse one `Metric` per identity and append datapoints per interface/array element, rather than creating a new `Metric` object for each one.

The same leaf reported more than once within a single `SubscribeResponse` — e.g. `in-octets` for both `eth0` and `eth1`, or elements of a JSON array or gNMI leaf-list — previously produced one `pmetric.Metric` per update, so a batch with N interfaces reporting the same counter emitted N separate metrics that shared a name. The parser now accumulates a `parseBatch` keyed by the metric's final name, creating the `Metric` once and appending a datapoint to it for every subsequent update that resolves to the same name. This matches the general OpenTelemetry recommendation to reuse a `Metric` across identical series.

### DIFF
--- a/internal/receiver/gnmireceiver/parser.go
+++ b/internal/receiver/gnmireceiver/parser.go
@@ -48,6 +48,43 @@ func newMetricParser(endpoint string, subscriptions []SubscriptionConfig) *metri
 	return &metricParser{endpoint: endpoint, subscriptions: subscriptions}
 }
 
+// parseBatch accumulates the metrics produced while converting a single SubscribeResponse.
+type parseBatch struct {
+	sm     pmetric.ScopeMetrics
+	byName map[string]pmetric.Metric
+}
+
+func newParseBatch(sm pmetric.ScopeMetrics) *parseBatch {
+	return &parseBatch{sm: sm, byName: map[string]pmetric.Metric{}}
+}
+
+func (b *parseBatch) numberDataPoint(name, unit string, wantType pmetric.MetricType) (pmetric.NumberDataPoint, error) {
+	if m, ok := b.byName[name]; ok {
+		if m.Type() != wantType {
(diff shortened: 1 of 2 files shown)


internal/receiver/gnmireceiver/parser.go
```diff
@@ -48,6 +48,43 @@ func newMetricParser(endpoint string, subscriptions []SubscriptionConfig) *metri
...
+	byName map[string]pmetric.Metric
+}
+
+func newParseBatch(sm pmetric.ScopeMetrics) *parseBatch {
+	return &parseBatch{sm: sm, byName: map[string]pmetric.Metric{}}
+}
+
+func (b *parseBatch) numberDataPoint(name, unit string, wantType pmetric.MetricType) (pmetric.NumberDataPoint, error) {
+	if m, ok := b.byName[name]; ok {
+		if m.Type() != wantType {
```
jinja2, line 63: should we compare the unit before appending since it is also part of the metric identity? From my understanding a collision is very unlikely. One (unrealistic) e.g. i can think of a string leaf named `foo` becomes `foo_info`, which collides with some numeric leaf named `foo_info`. If both are gauges, their datapoints would be merged even if their units and meanings differ

jinja2 (approved):

internal/receiver/gnmireceiver/parser.go
jkoronaAtCisco: It's quite unlikely, but also it's cheap to check it so I added it.

PR #7948, opened 2026-08-19
Add OpAMP Supervisor and launcher to Windows packages

kelseyma:
Update the Windows packages to include the new otelcollauncher and opampsupervisor executables.
Service will now start the launcher instead of the collector so that users can enable the supervisor if desired. This is the windows counterpart to the Linux packaging changes from #7897.

- By default, service will continue just starting the collector so users should not experience any functional impact. 
- When OpAMP Supervisor is enabled, it will manage the collector process and allow for remote configuration.
 
Also some cleanup changes: 
- Fix direct collector mode logging under launcher to have the collector directly output to event log, which preserves log level
- Remove `WITH_OPAMP_SUPERVISOR` build env var since now all the existing usages now build with supervisor by default

- Updated e2e tests to cover supervisor enabled with:
  - install via script: o11y only, o11y + platform flags
  - choco test: install, upgrade
  - direct msi install
  - install pre-launcher version -> upgrade (direct) -> enable supervisor -> rollback to direct: all with fips on
- Local dev testing with greenfield install (flag enablement), brownfield upgrade (env var enablement), install without supervisor flag
- Also see #7578 for previous PRs with testing performed on launcher/supervisor

**Documentation:**
Working with docs team on the docs updates (both existing collector install/upgrade pages and new opamp supervisor pages under fleet management)


.github/workflows/scripts/win-test-services.ps1
```diff
@@ -102,6 +140,22 @@ if ((service_running -name "splunk-otel-collector")) {
 $uninstallProperties = Get-ChildItem -Path "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall" |
     ForEach-Object { Get-ItemProperty $_.PSPath } |
     Where-Object { $_.DisplayName -eq "Splunk OpenTelemetry Collector" }
+$installed_version = [Version]$uninstallProperties.DisplayVersion
```
kelseyma, line 143: existing line moved up from original location so it could be used here for the exe check. For the `expect_launcher` check, wanted to have this be flexible for both supervisor and future releases so we don't need to change it

.github/workflows/win-package-test.yml
```diff
@@ -151,6 +154,7 @@ jobs:
           & ${{ github.workspace }}\.github\workflows\scripts\win-required-ports.ps1
 
       - name: Installation test
+        shell: pwsh
```
kelseyma: this is the default shell for windows, just specifying so it'll pass the actionlint checks

cmd/otelcollauncher/child_output.go
```diff
@@ -22,16 +22,10 @@ import (
...
-	Info(msg string)
-	Error(msg string)
-}
-
 // forwardChildOutput drains both child process output streams and writes complete
-// lines to the provided sink. The returned channel closes after both streams
+// lines with the provided function. The returned channel closes after both streams
 // reach EOF so callers can wait for output to drain after the child exits.
-func forwardChildOutput(stdout, stderr io.Reader, sink childLogSink) <-chan error {
+func forwardChildOutput(stdout, stderr io.Reader, write func(string)) <-chan error {
```
kelseyma: Simplifying windows launcher log forwarding to just send info events, since by default collector outputs to stderr so sending those as error events could be confusing

tests/windows-install-script/windows_install_script_test.go
```diff
@@ -74,16 +86,85 @@ func TestUpgradeAndUninstallFromNonMachineWideVersion(t *testing.T) {
...
+	require.NoError(t, err)
+	require.NoError(t, configFile.Close())
+	configHash := hashFile(t, configPath)
+
+	fipsEnvironment := "GODEBUG=" + fipsMode
+	verifyServiceEnvironmentContains(t, fipsEnvironment)
+
+	msiInstallerPath := getFilePathFromEnvVar(t, "MSI_COLLECTOR_PATH")
+	t.Logf(" *** Upgrading collector from %q", msiInstallerPath)
+	installCollector(
```
kelseyma, line 123: is running install script again ok for windows upgrade? other existing test in this file is doing that so I followed the behavior

.github/workflows/win-package-test.yml
pjanotti: This called my attention that this step should be using `shell: powershell` instead since it runs the `install.ps1`. `pwsh` is not installed by default on Windows boxes, but is the latest more modern "Powershell", `powershell` is the legacy one that comes OOTB w/ Windows. If this exposes some errors in the other scripts requiring `pwsh` ensure that at least `install.ps1` runs on the legacy one.

.github/workflows/win-package-test.yml
```diff
@@ -278,6 +299,7 @@ jobs:
           & ${{ github.workspace }}\.github\workflows\scripts\win-required-ports.ps1
 
       - name: Chocolatey ${{ matrix.SCENARIO }} test
+        shell: pwsh
```
pjanotti: Similarly, give it a try with `shell: powershell`

cmd/otelcollauncher/main_windows.go
```diff
@@ -218,8 +204,8 @@ type childProcess struct {
 }
 
 // startChild launches the selected binary in its own process group. Interactive
-// runs inherit console stdout/stderr; service runs bridge child stdout/stderr
-// to Event Log through the provided event log handle.
+// runs inherit console stdout/stderr; service runs bridge both child streams to
+// Information events because stream identity does not indicate log severity.
```
pjanotti: The fact that we lose log level for the collector is not good, the levels are definitely used to alert and monitor the collector. Perhaps the solution is to have the Windows launcher instead of passing its own `serviceHandler` if should pass the collector one, `otelcol`, with the params updated according to the supervisor config. This should avoid the loss of the log level keep a single process and avoids implementing the service handler here too.

pjanotti (commented):
Let's try to pass directly the collector service handler on Windows, this should avoid loss of log level when running as service.

cmd/otelcol/eventlog_windows.go
kelseyma, line 1: Heavily borrowed from upstream [collector_windows.go](https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/collector_windows.go) for the buildZapLogger override and windowsEventLogCore

cmd/otelcol/main_windows.go
```diff
@@ -52,6 +60,9 @@ func run(params otelcol.CollectorSettings) error {
 		if ok && errno == windows.ERROR_FAILED_SERVICE_CONTROLLER_CONNECT {
 			// Per https://learn.microsoft.com/en-us/windows/win32/api/winsvc/nf-winsvc-startservicectrldispatchera#return-value
 			// this means that the process is not running as a service, so run interactively.
+			if isParentProcessOtelcolLauncher() {
+				return runInteractiveWithWindowsEventLog(params)
```
kelseyma, line 64: This does mean that if launcher is run as a process, then collector will log to event log instead of console. However, I think that should be ok since launcher is mainly here for the service entrypoint, and if users want to run the collector as a process then they should just continue to use the `otelcol.exe` (as noted in docs) instead of starting through the launcher. 
I did look into how to support some sort of differentiation here for launcher process, but seemed to add more complexity for a use case that potentially isn't really applicable

cmd/otelcol/main_windows.go
```diff
@@ -60,3 +71,22 @@ func run(params otelcol.CollectorSettings) error {
 
 	return nil
 }
+
+// isParentProcessOtelcolLauncher reports whether the immediate parent is the launcher.
+// Lookup failures are treated as a non-match so startup can continue with default logging behavior.
+func isParentProcessOtelcolLauncher() bool {
```
kelseyma, line 77: followed `isParentProcessSplunkd` as example for this

cmd/otelcollauncher/child_output.go
kelseyma: removed this change

tests/windows-install-script/windows_install_script_test.go
pjanotti: yes, "upgrade" on Windows at least requires uninstall first.

cmd/otelcol/main_windows.go
pjanotti: I even think that there may be cases that some one wants to use the event log even if running as a process, but that is a separate issue.

pjanotti (approved):

PR #7953, opened 2026-08-20
feat: use query parameter for windowseventlog receiver

weliasz:
Use Query parameter of windows event log receiver.
https://splunk.atlassian.net/browse/ADDON-88165


cmd/otelcol/config/collector/splunk_logs_config_windows.yaml
```diff
@@ -1,39 +1,60 @@
 receivers:
   windows_event_log:
-    channel: Application
+    query: |
+      <QueryList>
+        <Query Id="0">
+          <Select Path="Application">*</Select>
+          <Suppress Path="Application">*[System[Provider[@Name='XENAGENT']]]</Suppress>
+          <Suppress Path="Application">*[System[Provider[@Name='Tenable Nessus Agent']]]</Suppress>
```
pjanotti, line 8: I still keep questioning if this is something that we want in general, or if it is particular to our test environment. Does the UF has these same exclusions?

pjanotti (approved):

cmd/otelcol/config/collector/splunk_logs_config_windows.yaml
weliasz: The only exceptions in splunk-add-on-for-microsoft-windows are for EventCode="4662" and EventCode="566". There are no exceptions for XENAGENT and Tenable Nessus Agent' so I agree that OTEL config should match config of UF with addon.

This exception was added by AI and indeed without it exceptions are logged because of missing DLL metadata.

I'll comment out these lines.

PR #7955, opened 2026-08-20
Update OpAMP Supervisor version to pinned commit

kelseyma:
Temporarily pinning OpAMP Supervisor to a version that includes a bug fix. The bugfix was the first commit after `0.159.0` release so there is minimal change in what this pseudo-version includes. 
The next update-deps job should be able to update to `0.160.0` and bring us back in line with upstream.

- Built and tested binary locally to ensure bugfix was included


internal/tools/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/golangci/golangci-lint/v2 v2.7.2
 	github.com/google/addlicense v1.2.0
 	github.com/jstemmer/go-junit-report v1.0.0
-	github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor v0.159.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor v0.159.1-0.20260817215956-bf8d1bce8395
```
kelseyma, line 10: Currently set skip changelog for this PR similar to the normal update-deps , let me know if we actually need to include this in changelog

dmitryax (approved):

PR #7956, opened 2026-08-21
Add a storage extension designed for persistent queue

atoulme:
**Description:**
This storage extension is specifically meant for the exporter queue persistent storage.

It stores on disk using append-only file storage, compacting old files after they are read.

The extension fails to run if it is used outside of the use persistent queue use case.


internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,393 @@
...
+	doneChan chan struct{}
+	closed   bool
+}
+
+// New instantiates a diskQueue, restoring its index from the filesystem if
+// present, and starts a background goroutine that periodically flushes
+// pending writes according to syncEvery/syncTimeout.
+func New(name, dataPath string, maxBytesPerFile int64,
+	syncEvery int64, syncTimeout time.Duration, logger *zap.Logger,
+) Interface {
```
MovieStoreGuy: Should this use an options pattern or config type instead of specifying each value?

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,393 @@
...
+		liveCount:       make(map[int64]int64),
+		exitChan:        make(chan struct{}),
+		doneChan:        make(chan struct{}),
+	}
+
+	if err := d.retrieveIndex(); err != nil && !os.IsNotExist(err) {
+		d.logger.Error("failed to retrieve index", zap.String("name", d.name), zap.Error(err))
+	}
+
+	go d.syncLoop()
```
MovieStoreGuy: An unbounded go routine? Does goleak approve?

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,393 @@
...
+
+// Put appends data to the current segment file and records its location
+// under key, making it independently retrievable and deletable.
+func (d *diskQueue) Put(key uint64, data []byte) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	if d.closed {
+		return errors.New("exiting")
+	}
```
MovieStoreGuy: Is it worth using an atomic reference here instead of a bool to avoid a full lock if closed?

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,393 @@
...
+		return err
+	}
+
+	d.index[key] = loc
+	d.liveCount[loc.fileNum]++
+	d.pendingSync++
+
+	if d.pendingSync >= d.syncEvery {
+		if syncErr := d.syncLocked(); syncErr != nil {
+			d.logger.Error("failed to sync", zap.String("name", d.name), zap.Error(syncErr))
```
MovieStoreGuy: Should a sync error cause some of error to be bubbled up, or metric to be captured?

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,393 @@
...
+func (d *diskQueue) Delete(key uint64) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	loc, ok := d.index[key]
+	if !ok {
+		return nil
+	}
+
+	delete(d.index, key)
```
MovieStoreGuy: Keep an eye on map performance and underlying capacity size since it doesn't shrink maps storage on delete.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,393 @@
...
+	d.liveCount[loc.fileNum]--
+	d.pendingSync++
+
+	var err error
+	if d.liveCount[loc.fileNum] <= 0 && loc.fileNum != d.writeFileNum {
```
MovieStoreGuy: Should the count ever drop below zero?

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,393 @@
...
+	d.mu.Lock()
+	if d.closed {
+		d.mu.Unlock()
+		return nil
+	}
+	d.closed = true
+	d.mu.Unlock()
+
+	close(d.exitChan)
+	<-d.doneChan
```
MovieStoreGuy: Would this be better worked with a `once.Do` instead of locking?

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,388 @@
...
+	if err != nil {
+		return err
+	}
+
+	w := bufio.NewWriter(f)
+	if _, err := fmt.Fprintf(w, "%d,%d\n%d\n", d.writeFileNum, d.writePos, len(d.index)); err != nil {
+		_ = f.Close()
+		return err
+	}
+	for key, loc := range d.index {
```
dmitryax: Hm.. this gives us O(N^2) complexity for filling and draining the queue. This will likely significantly slow down writes for long queues. We need to benchmark this against the regular storage queue

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,388 @@
...
+	length  int64
+}
+
+// diskQueue implements a filesystem backed, key-addressable record store.
+// Records are appended to sequentially numbered segment files; an in-memory
+// index (persisted alongside the data) maps each key to its segment file and
+// byte range, so a record can be read or deleted independently of any other
+// record, in any order.
+type diskQueue struct {
+	index           map[uint64]location
```
dmitryax: Why do we have to index every item? This looks like we reimplement the generic storage without taking advantage of sequential write/read pattern in this part

dmitryax: To clarify, deletes can still be out of order because we have multiple workers finishing requests async.

In that case I think we should be able to apply some segmentation and don't track per-item index...

atoulme: yes something like that is happening.

atoulme: I'll revert the per-index portion and go back to a simple set of chans.

MovieStoreGuy:
Sorry I noticed my comments hadn't posted, clicked late and it looks like they are outdated.

atoulme:
> Sorry I noticed my comments hadn't posted, clicked late and it looks like they are outdated.

they still largely are applicable. There's some Go bad practices like the use of go without waitGroup there. I'll look.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: that's gone

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: I am using a proper sync.WaitGroup now.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: yes definitely, the lock + int flag is old school.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: I think that whole mess is gone

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: this is internal API - users are expected to suffer.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: this code is gone

atoulme:
Starting to make this faster than bbolt. Here is the output of the benchmarks:
```
BenchmarkExtensionAsPersistentQueueWithWorkers
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-100
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-100-12         	       1	1917113000 ns/op	212051008 B/op	   47573 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-200
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-200-12         	       1	3533127500 ns/op	427915352 B/op	   92468 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-300
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-300-12         	       1	5196371542 ns/op	567356416 B/op	  134483 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-400
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-400-12         	       1	6802544625 ns/op	718583728 B/op	  178473 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-500
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-500-12         	       1	8614722667 ns/op	919217056 B/op	  223060 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-1000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-1000-12        	       1	16557961417 ns/op	1684419368 B/op	  439075 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-100
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-100-12             	       1	2550667208 ns/op	30656040 B/op	   50282 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-200
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-200-12             	       1	4920075458 ns/op	55891936 B/op	   98616 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-300
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-300-12             	       1	7532003542 ns/op	56374552 B/op	  146136 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-400
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-400-12             	       1	10095236000 ns/op	100051080 B/op	  195043 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-500
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-500-12             	       1	12720987458 ns/op	107678368 B/op	  242724 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-1000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-1000-12            	       1	25606708166 ns/op	168572104 B/op	  482606 allocs/op
```

Rendered in Excel charts:
image

The extension is better than file_storage, except for its B/op. I think it's because we zip before storing to disk, maybe.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+	"os"
+	"path"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+type Message struct {
```
MovieStoreGuy: Could I ask you to document you're public values?

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+type Message struct {
+	ConsumeCallback func()
+	Payload         func() []byte
```
MovieStoreGuy: These values are not safe when using the zero value, this should be converted to an interface, have a constructor, or move these to private with accessors

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+	ConsumeCallback func()
+	Payload         func() []byte
+}
+
+type Queue interface {
+	Put([]byte) error
+	PeekChan() <-chan Message // this is expected to be an *unbuffered* channel
+	Close() error
+	Depth() int64
+}
```
MovieStoreGuy: Can I ask you to document these values and also provide details if it is safe to use after being closed?

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+
+type Queue interface {
+	Put([]byte) error
+	PeekChan() <-chan Message // this is expected to be an *unbuffered* channel
+	Close() error
+	Depth() int64
+}
+
+type Metadata struct {
+	Segments    []*segment
```
MovieStoreGuy: This type is export a package private value, Plus move Segments to match Metadata exported status.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+	exitFlag              atomic.Bool
+	needSync              bool
+	exitWG                sync.WaitGroup
+	metadataTruncateEvery int
+	metadataWrites        int
+}
+
+// New instantiates an instance of diskQueue, retrieving metadata
+// from the filesystem and starting the read ahead goroutine
+func New(name, dataPath string, maxBytesPerFile int64,
```
MovieStoreGuy: I am am honestly not a fan of how many params are being passed here, but I don't see a better option.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+func (d *diskQueue) Close() error {
+	err := d.exit()
+	if err != nil {
+		return err
+	}
+	return d.sync()
+}
+
+func (d *diskQueue) exit() error {
+	d.exitFlag.Store(true)
```
MovieStoreGuy: Should this compare the previous value incase it has already been stopped?

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+		d.logger.Debug("peekOne() opened", zap.String("name", d.name), zap.String("filename", curFileName))
+	}
+
+	var msgSize int64
+	for _, s := range d.metadata.Segments {
+		if s.FileNum == d.metadata.PeekFileNum && s.Pos == d.metadata.PeekPos {
+			msgSize = s.MessageLen
+			break
+		}
+	}
```
MovieStoreGuy: I am not 100% certain what this is trying to do, do you mind explaining it for me?

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+			d.logger.Error(" failed to sync - %s", zap.String("name", d.name), zap.Error(err))
+		}
+
+		if d.writeFile != nil {
+			_ = d.writeFile.Close()
+			d.writeFile = nil
+		}
+	}
+
+	if d.writeFile == nil {
```
MovieStoreGuy, line 203: Are we expecting the file to be nil or not most of the time? 

This method feels like it is doing a lot of work in order to writeOne?

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+			MessageLen: 0,
+			Consumed:   true,
+		}
+	}
+	return d.metadata.Segments[len(d.metadata.Segments)-1]
+}
+
+// writeOne performs a low level filesystem write for a single []byte
+// while advancing write positions and rolling files, if necessary
+func (d *diskQueue) writeOne(data []byte) error {
```
MovieStoreGuy: This method name name needs improvement, writeOne, doesn't really explain what it is trying to do. 

To me , it looks like a mix between `Write` and `Sync`?

atoulme:
Now with a draining scenario
```
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-100-12                       1        1964872125 ns/op        38474744 B/op      42022 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-200-12                       1        3613453542 ns/op        39125800 B/op      81094 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-300-12                       1        5236001667 ns/op        47986832 B/op     120217 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-400-12                       1        6901463583 ns/op        63332288 B/op     160020 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-500-12                       1        8457635667 ns/op        58096328 B/op     198735 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-1000-12                      1        17434704541 ns/op       118741352 B/op    396298 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-100-12                 1        7913292208 ns/op        23340024 B/op      57404 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-200-12                 1        7765089833 ns/op        43112864 B/op     100410 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-300-12                 1        7689955500 ns/op        50137152 B/op     142205 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-400-12                 1        13786758083 ns/op       77043952 B/op     210462 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-500-12                 1        14124517500 ns/op       99109712 B/op     254042 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-1000-12                1        19684578000 ns/op       200264792 B/op    475828 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-100-12                           1        2545255375 ns/op        14066512 B/op      49687 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-200-12                           1        5045991167 ns/op        22177208 B/op      97697 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-300-12                           1        6801030791 ns/op        68096880 B/op     146411 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-400-12                           1        8942181750 ns/op        70847360 B/op     194219 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-500-12                           1        11098227917 ns/op       71985472 B/op     241630 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-1000-12                          1        22910531833 ns/op       152392736 B/op    482369 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-100-12                     1        7706293542 ns/op        28679488 B/op      66323 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-200-12                     1        7926099291 ns/op        62309056 B/op     118282 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-300-12                     1        7659289833 ns/op        74510104 B/op     173272 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-400-12                     1        10237320542 ns/op       121807456 B/op    235657 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-500-12                     1        12975692584 ns/op       175657696 B/op    303218 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-1000-12                    1        23707244583 ns/op       314500176 B/op    631600 allocs/op
PASS
ok      github.com/signalfx/splunk-otel-collector/internal/extension/diskqueuestorageextension  243.255s
```

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+		lastSegment = s
+	}
+	// place at the tip.
+	if lastSegment != nil {
+		d.metadata.PeekFileNum = lastSegment.FileNum
+		d.metadata.PeekPos = lastSegment.Pos + lastSegment.MessageLen
+	}
+}
+
+type segment struct {
```
MovieStoreGuy: Move this type definition higher in the file to avoid it being missed and keeps a clear structure to the file.

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+	d.metadata.Segments = compactedSegments
+}
+
+func (d *diskQueue) ioLoop() {
+	var peekDataRead []byte
+	var err error
+	var count int64
+	var p chan Message
+	var messagePeekFileNum int64
+	var messagePeekPos int64
```
MovieStoreGuy: Can you move this to a group definition instead of a var definition for each?

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+		}
+	}
+	d.metadata.Segments = compactedSegments
+}
+
+func (d *diskQueue) ioLoop() {
+	var peekDataRead []byte
+	var err error
+	var count int64
+	var p chan Message
```
MovieStoreGuy: This is going to be nil by default, is that going to be an issue?

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+	recomputePeek := false
+
+	syncTicker := time.NewTicker(d.syncTimeout)
+	defer syncTicker.Stop()
+
+	type callback struct {
+		pos     int64
+		len     int64
+		fileNum int64
+	}
```
MovieStoreGuy: Is it possible to move this definition outside the method so it is easier to find?

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+ioLoop:
+	for {
+		if count == d.syncEvery {
+			d.needSync = true
+		}
+
+		if d.needSync {
+			err = d.sync()
+			if err != nil {
+				d.logger.Error("failed to sync", zap.String("name", d.name), zap.Error(err))
```
MovieStoreGuy: This feels like a critical error, should this report a metric?

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+			count = 0
+		}
+
+		if recomputePeek {
+			recomputePeek = false
+			lastSegment := d.lastSegment()
+			if d.metadata.PeekFileNum < lastSegment.FileNum || (d.metadata.PeekFileNum == lastSegment.FileNum && d.metadata.PeekPos < lastSegment.Pos+lastSegment.MessageLen) {
+				peekDataRead, err = d.peekOne()
+				if err != nil {
+					d.logger.Error(fmt.Sprintf("failed peeking at %d of %s", d.metadata.PeekPos, d.fileName(d.metadata.PeekFileNum)),
```
MovieStoreGuy: Avoid using Sprintf within the log filed, use the attributes instead for dynamic values.

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+				callbackLen := int64(len(peekDataRead))
+				callbackFileNum := messagePeekFileNum
+				return func() {
+					callbackChan <- callback{
+						pos:     callbackPos,
+						len:     callbackLen,
+						fileNum: callbackFileNum,
+					}
+				}
+			}(),
```
MovieStoreGuy: I am concerned that the inline functions of functions is going to cause confusion or potentially hold state it shouldn't. 

From memory, anon functions are discouraged from upstream, we should try to minimise this.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
```diff
@@ -0,0 +1,560 @@
...
+			}
+			d.needSync = true
+		case <-d.exitChan:
+			d.logger.Debug("closing", zap.String("name", d.name))
+			break ioLoop
+		}
+	}
+}
+
+const separator = "\n\n"
```
MovieStoreGuy: Move to the top of the file.

internal/extension/diskqueuestorageextension/diskqueue.go
MovieStoreGuy, line 1: I think DiskQueue as a whole is doing too much and should be abstracted further, at least to seperate metadata and file management.

internal/extension/diskqueuestorageextension/extension.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+// special keys used by the persistent queue to store metadata.
+const (
+	metadataKey = "qmv0"
+
+	// all legacy keys - ignored.
+	legacyReadIndexKey                = "ri"
+	legacyWriteIndexKey               = "wi"
+	legacyCurrentlyDispatchedItemsKey = "di"
```
MovieStoreGuy, line 38: If these are legacy keys, why do we need then for a new extension?

internal/extension/diskqueuestorageextension/extension.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (d *diskQueueStorageExtension) Start(_ context.Context, _ component.Host) error {
+	return nil
+}
+
+func (d *diskQueueStorageExtension) Shutdown(_ context.Context) error {
+	return nil
+}
+
+const separator = "\n\n"
```
MovieStoreGuy: This has already been declared once, why has been re added here?

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: No, this is an internal folder. But point taken - I have folded the internal folder into the extension package. No more exported types.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: this interface is now gone

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: I mean, we can also create the struct ourselves and set it to start. Might be more idiomatic. Thoughts?

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: unexported - the reason the Segments field is exported is for JSON serialization.

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: read the metadata to find the segment you're on, get the message length from it. Use that length to know how many bytes to read.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: no that's a fixture of the thing. When there's nothing to peek, the select statement works with a p that is nil, which is then skipped.

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: done, agree it's not nice, but note it's only used in that function

internal/extension/diskqueuestorageextension/extension.go
atoulme: to explicitly skip them. The persistent_queue.go code checks for them.

internal/extension/diskqueuestorageextension/extension.go
atoulme: 2 different metadata: one for the persistent queue, one for the disk queue. I merged them anyway since we're merging packages.

internal/extension/diskqueuestorageextension/extension.go
atoulme: if that's not clear, we're jerry-rigging this to work with persistent queue.

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: removed!

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: agreed

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: given this is all very internal and hidden now I think we can close

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: when we sync we close it, and set it to nil

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: renamed to "write", still a mess

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: no longer in play now

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: I put a panic("foo") in there

atoulme:
Latest benchmarks:
```
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-100-12                       6         174201410 ns/op        250626234 B/op     45385 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-200-12                       5         218387008 ns/op        469331881 B/op     88237 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-300-12                       4         273830510 ns/op        728162904 B/op    131890 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-400-12                       4         332693719 ns/op        975086014 B/op    175508 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-500-12                       3         393848625 ns/op        1211336928 B/op   218999 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-1000-12                      2         661156708 ns/op        2274157840 B/op   431793 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-5000-12                      1        3783957584 ns/op        10619193136 B/op         2155159 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-10000-12                     1        12237352291 ns/op       21078542008 B/op         4357130 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-100-12                 1        7358754333 ns/op        227146176 B/op     61447 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-200-12                 1        7376869542 ns/op        490511744 B/op    109050 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-300-12                 1        7486695333 ns/op        669080648 B/op    157447 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-400-12                 1        7511258292 ns/op        934148648 B/op    204932 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-500-12                 1        7516366417 ns/op        1058685544 B/op   249717 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-1000-12                1        7506699667 ns/op        2075456944 B/op   481870 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-5000-12                1        7296733041 ns/op        9924298544 B/op  2282930 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-10000-12               1        17352292167 ns/op       21184654528 B/op         5241078 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-100-12                           1        2527203125 ns/op        210356696 B/op     54894 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-200-12                           1        4897558916 ns/op        384612040 B/op    107055 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-300-12                           1        7279207791 ns/op        598256000 B/op    160067 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-400-12                           1        9657410792 ns/op        792897888 B/op    212430 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-500-12                           1        12693397834 ns/op       999643136 B/op    265886 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-1000-12                          1        24188124792 ns/op       1853586104 B/op   526261 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-5000-12                          1        117063732000 ns/op      7780092728 B/op  2580841 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-10000-12                         1        258312716417 ns/op      14126599088 B/op         5116422 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-100-12                     1        7732944334 ns/op        215620808 B/op     71374 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-200-12                     1        7939542750 ns/op        422994224 B/op    129260 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-300-12                     1        7877928042 ns/op        621089936 B/op    187628 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-400-12                     1        10586222333 ns/op       820166208 B/op    251824 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-500-12                     1        13708559958 ns/op       1006823296 B/op   327067 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-draining-1000-12                    1        25425094125 ns/op       2029859120 B/op   680956 allocs/op
```

atoulme:
```
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-100-12                       6         169631160 ns/op        252372366 B/op     44284 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-200-12                       5         230592600 ns/op        484333641 B/op     86124 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-300-12                       4         313730386 ns/op        635005122 B/op    125830 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-400-12                       3         381912681 ns/op        846439760 B/op    167291 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-500-12                       2         541470125 ns/op        1040168140 B/op   208412 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-1000-12                      2         695356084 ns/op        2277385828 B/op   419933 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-5000-12                      1        2511482875 ns/op        10081665272 B/op         2061878 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-10000-12                     1        5270154334 ns/op        18709966816 B/op         4089807 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-100-12                 1        7445021334 ns/op        245893616 B/op     61375 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-200-12                 1        7573911417 ns/op        428968424 B/op    107040 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-300-12                 1        7610477459 ns/op        617340824 B/op    153358 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-400-12                 1        7504115209 ns/op        822524680 B/op    199870 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-500-12                 1        7555185417 ns/op        953382352 B/op    243613 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-1000-12                1        7523389208 ns/op        1649234568 B/op   462939 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-5000-12                1        7490890083 ns/op        7111277464 B/op  2187985 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-draining-10000-12               1        7557347959 ns/op        13663545880 B/op         4273126 allocs/op
```

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: look a few lines above - we have the logic to rotate files there

internal/extension/diskqueuestorageextension/extension.go
```diff
@@ -0,0 +1,227 @@
...
+	path                  string
+	callbacks             []map[string]func()
+	metadataWrites        int
+	metadataTruncateEvery int
+	checkFirstGet         sync.Once
+}
+
+func (c *client) Get(_ context.Context, key string) ([]byte, error) {
+	// This is a kludgy way to detect that the extension is not used for persistent queueing.
+	c.checkFirstGet.Do(func() {
```
dmitryax, line 73: TODO (as a follow-up): we need an integration test to confirm this contract. It should fail whenever the expected behavior changed in the exporterhelper

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,629 @@
...
+	len     int64
+	fileNum int64
+}
+
+// diskQueue implements a filesystem backed FIFO queue
+type diskQueue struct {
+	writeFile             *os.File
+	writeChan             chan []byte
+	exitChan              chan int
+	recomputePeekChan     chan struct{}
```
dmitryax: unused

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,629 @@
...
+
+func (d *diskQueue) writeLoop() {
+	syncTicker := time.NewTicker(d.syncTimeout)
+	defer syncTicker.Stop()
+	opCount := int64(0)
+	for {
+		select {
+		case dataWrite := <-d.writeChan:
+			opCount++
+			data := d.write(dataWrite)
```
dmitryax:
```suggestion
			err := d.write(dataWrite)
```

internal/extension/diskqueuestorageextension/config.go
```diff
@@ -0,0 +1,43 @@
...
+type Config struct {
+	Path            string        `mapstructure:"path"`
+	MaxBytesPerFile int64         `mapstructure:"max_bytes_per_file"`
+	SyncEvery       int64         `mapstructure:"sync_every"`
+	SyncTimeout     time.Duration `mapstructure:"sync_timeout"`
+}
+
+func (c *Config) Validate() error {
+	if c.Path == "" {
+		return errors.New("path must be a valid folder")
```
pjanotti: nit: stat it or change the message to say that it can't be empty

internal/extension/diskqueuestorageextension/config.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"errors"
+	"time"
+)
+
+type Config struct {
+	Path            string        `mapstructure:"path"`
+	MaxBytesPerFile int64         `mapstructure:"max_bytes_per_file"`
+	SyncEvery       int64         `mapstructure:"sync_every"`
```
pjanotti, line 25: every what? I guess I will see somewhere, but having it explicit in the config is nice

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,629 @@
...
+		return errors.New("exiting")
+	}
+	buf := bufPool.Get().(*bytes.Buffer)
+	buf.Reset()
+	defer bufPool.Put(buf)
+	if _, err := gzipWriterPool.Write(buf, data); err != nil {
+		return err
+	}
+	d.writeChan <- buf.Bytes()
+	return <-d.writeResponseChan
```
pjanotti: How does it ensure that the response corresponds to the bytes sent to writeChan? Or it doesn't matter, i.e.: the response for any other write is good?

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,629 @@
...
+	if lastSegment != nil {
+		d.peekMetadata.PeekFileNum = lastSegment.FileNum
+		d.peekMetadata.PeekPos = lastSegment.Pos + lastSegment.MessageLen
+		// set when we write the next segment
+		d.peekMetadata.PeekLen = 0
+	}
+}
+
+func (d *diskQueue) moveForward(fileNum, pos, messageLen int64) {
+	d.metadataLock.Lock()
```
dmitryax: It's confusing that concurrency is controlled by a mix of channels and locks, put/peek/callbacks are synchronous round-trips already. I'd say we should pick one approach:
- one goroutine handling put/peek/callback/tick via a single select, and then drop metadataLock entirely so the goroutine is the sole owner of the state.
- drop both loops, keep one mutex for put/peek/callback and a single small goroutine only for the sync ticker.

Let me know if I'm missing something why it has to be done this way

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,629 @@
...
+		}
+		d.logger.Debug("peekData() opened", zap.String("name", d.name), zap.String("filename", curFileName))
+	}
+
+	_, err = d.peekFile.Seek(d.peekMetadata.PeekPos, 0)
+	if err != nil {
+		_ = d.peekFile.Close()
+		d.peekFile = nil
+		return nil, err
+	}
```
dmitryax: Why do we need this if we call ReadAt right after?

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: Fixed now - no more locks. I have 2 loops to control writes and reads, and they interact only in the case that peek reaches head of the queue by passing a message to relocate. I also simplified the data structures by writing length of objects in the file, removed the notion of segments.

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: removed, good catch

internal/extension/diskqueuestorageextension/diskqueue.go
atoulme: it uses a single for loop as a way to initiate the request, and the response only can come from the peek. The rub is that if the queue has no elements, then you need to wait until a write happens and then initiate the peek. That's done well now.

internal/extension/diskqueuestorageextension/config.go
atoulme: every "object", every "element" of the queue. Not sure if worth exposing.

internal/extension/diskqueuestorageextension/config.go
atoulme: actually, an empty value means current folder. It's fine. Removing that validation.

internal/extension/diskqueuestorageextension/internal/diskqueue.go
atoulme: we rotate maps now

internal/extension/diskqueuestorageextension/extension.go
atoulme: added a test for that, not integration but at least testing the code path

internal/extension/diskqueuestorageextension/internal/compression/writer_pool.go
```diff
@@ -0,0 +1,87 @@
...
+// NewGZIPWriterPool returns a WriterPool that compresses data with gzip.BestSpeed.
+func NewGZIPWriterPool() WriterPool {
+	return NewWriterPool(func() *gzip.Writer {
+		w, _ := gzip.NewWriterLevel(io.Discard, gzip.BestSpeed)
+		return w
+	})
+}
+
+// NewLZ4WriterPool returns a WriterPool that compresses data with LZ4.
+func NewLZ4WriterPool() WriterPool {
```
dmitryax: This isn't used

internal/extension/diskqueuestorageextension/internal/compression/writer_pool.go
```diff
@@ -0,0 +1,87 @@
...
+	Write(dst io.Writer, data []byte) (int64, error)
+}
+
+type writerPool[RWC ResetWriteCloser] struct {
+	_    struct{}
+	pool sync.Pool
+}
+
+// NewGZIPWriterPool returns a WriterPool that compresses data with gzip.BestSpeed.
+func NewGZIPWriterPool() WriterPool {
```
dmitryax: Why do we need to compress messages? We're optimizing for durability/throughput not for disk footprint, right? Also, if we apply it on the individual messages it won't even help with the size for small ones, only slow down throughput

atoulme: Are we not batching beforehand? I think we'd like to keep the byte footprint lower at the expense of the CPU usage. I can run with/without and compare.

atoulme:
```
GOROOT=/Users/atoulme/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.6.darwin-arm64 #gosetup
GOPATH=/Users/atoulme/go #gosetup
GOPROXY=https://proxy.golang.org/,direct #gosetup
/Users/atoulme/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.6.darwin-arm64/bin/go test -c -o /Users/atoulme/Library/Caches/JetBrains/IntelliJIdea2026.2/tmp/GoLand/___BenchmarkExtensionAsPersistentQueueWithWorkers_in_github_com_signalfx_splunk_otel_collector_internal_extension_diskqueuestorageextension.test github.com/signalfx/splunk-otel-collector/internal/extension/diskqueuestorageextension #gosetup
/Users/atoulme/Library/Caches/JetBrains/IntelliJIdea2026.2/tmp/GoLand/___BenchmarkExtensionAsPersistentQueueWithWorkers_in_github_com_signalfx_splunk_otel_collector_internal_extension_diskqueuestorageextension.test -test.v -test.paniconexit0 -test.bench ^\QBenchmarkExtensionAsPersistentQueueWithWorkers\E$ -test.run ^$ #gosetup
goos: darwin
goarch: arm64
pkg: github.com/signalfx/splunk-otel-collector/internal/extension/diskqueuestorageextension
cpu: Apple M4 Pro
BenchmarkExtensionAsPersistentQueueWithWorkers
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-100
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-100-12         	       8	 138404969 ns/op	 9593980 B/op	   36837 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-200
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-200-12         	       7	 146711185 ns/op	10724416 B/op	   71409 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-300
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-300-12         	       7	 149693607 ns/op	14675716 B/op	  106134 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-400
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-400-12         	       6	 168318000 ns/op	20076002 B/op	  141075 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-500
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-500-12         	       6	 175279465 ns/op	18875698 B/op	  175866 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-1000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-1000-12        	       5	 238510192 ns/op	45069371 B/op	  350782 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-5000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-5000-12        	       2	 662236896 ns/op	160032584 B/op	 1744311 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-10000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-ok-10000-12       	       1	1202939084 ns/op	265608504 B/op	 3492059 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-100
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-100-12   	       1	7325264750 ns/op	 7366680 B/op	   52860 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-200
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-200-12   	       1	7539003000 ns/op	15023640 B/op	   94067 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-300
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-300-12   	       1	7543996500 ns/op	29250936 B/op	  135812 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-400
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-400-12   	       1	7551298416 ns/op	49376464 B/op	  177770 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-500
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-500-12   	       1	7449806792 ns/op	69043728 B/op	  216508 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-1000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-1000-12  	       1	6972928500 ns/op	88968776 B/op	  414261 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-5000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-5000-12  	       1	7514368291 ns/op	289752744 B/op	 2026328 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-10000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-nocompress-draining-10000-12 	       1	7581063000 ns/op	470294784 B/op	 4002371 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-100
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-100-12           	       8	 138093594 ns/op	12181193 B/op	   36963 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-200
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-200-12           	       7	 147593458 ns/op	12583526 B/op	   71851 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-300
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-300-12           	       7	 159007655 ns/op	12849838 B/op	  106769 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-400
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-400-12           	       7	 169199714 ns/op	23391652 B/op	  142082 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-500
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-500-12           	       6	 178962514 ns/op	18486460 B/op	  176944 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-1000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-1000-12          	       5	 237100750 ns/op	40322822 B/op	  352519 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-5000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-5000-12          	       2	 689870521 ns/op	182878468 B/op	 1755693 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-10000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-ok-10000-12         	       1	1220949334 ns/op	310330552 B/op	 3506136 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-100
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-100-12     	       1	7531629292 ns/op	 8732960 B/op	   54223 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-200
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-200-12     	       1	7533515583 ns/op	11838920 B/op	   95118 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-300
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-300-12     	       1	7538328583 ns/op	25512104 B/op	  136519 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-400
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-400-12     	       1	7548792833 ns/op	52500600 B/op	  178732 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-500
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-500-12     	       1	7458495042 ns/op	72005416 B/op	  218423 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-1000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-1000-12    	       1	7481054583 ns/op	94419784 B/op	  423006 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-5000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-5000-12    	       1	7528020500 ns/op	288349280 B/op	 2019133 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-10000
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-compress-draining-10000-12   	       1	7495823584 ns/op	487658184 B/op	 4000304 allocs/op
PASS

Process finished with the exit code 0
```

atoulme: I offer that we go without compression by default ; the tests show a difference, but it's not overwhelming. We would need messier data to see how this affects performance.

internal/extension/diskqueuestorageextension/internal/compression/writer_pool.go
atoulme: removing

internal/extension/diskqueuestorageextension/extension.go
```diff
@@ -0,0 +1,231 @@
...
+				}
+				return nil
+			}
+		}
+
+		return errors.New("cannot delete " + key)
+	}
+}
+
+func (c *client) Batch(ctx context.Context, ops ...*storage.Operation) error {
```
dmitryax, line 144: Batch must be applied atomically. The write/read operations come from the exporterhelper along with the metadata updates. If we don't don't respect that, the metadata and actual queue can diverge and result in deadlocks.

internal/extension/diskqueuestorageextension/extension.go
```diff
@@ -0,0 +1,231 @@
...
+		for i := callbackLen - 1; i >= 0; i-- {
+			cbMap := c.callbacks[i]
+			if callback, ok := cbMap[key]; ok {
+				callback()
+				delete(cbMap, key)
+				if len(cbMap) == 0 && i != callbackLen-1 {
+					c.callbacks[i] = make(map[string]func(), callbacksSize)
+				}
+				return nil
+			}
```
dmitryax, line 137: Why can't we maintain just one map for the callbacks? It'll be capped by the number of exporter workers which isn't a high number. So it won't grow out of band. And we don't need to pre-initialize with 10k

atoulme: Maps continue to use more room over time: https://github.com/golang/go/issues/20135 see https://github.com/signalfx/splunk-otel-collector/pull/7956#discussion_r3827975059

atoulme: we can preinitialize to number of workers. I don't have a strong take on a good initial size.

internal/extension/diskqueuestorageextension/extension.go
atoulme: is the batching specifically for writes or are there times where we mix writes and reads in a batch?

dmitryax: We have to at least support metadata_write+item_read, metadata_write+item_delete, metadata_write+item_write.

But I'm afraid of adding another implicit contract here. It'll be hard to maintain.

I suggest we implement batch properly for any combination unless there are strong reasons against that

atoulme: I mean, we can try, but given we write in different files, I don't quite see how we do that, unless you just mean having a write lock around Batch?

atoulme: ok this is now implemented and ready for review. Note performance suffers since we pass bytes around, though it's still orders of magnitudes faster than the bbolt alternative.

```
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-100-12                       7         143875250 ns/op        108635744 B/op     39452 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-200-12                       7         154930631 ns/op        194786755 B/op     76217 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-300-12                       6         169866257 ns/op        289917254 B/op    113318 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-400-12                       6         179971833 ns/op        368651036 B/op    150023 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-500-12                       6         195322910 ns/op        459840429 B/op    187141 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-1000-12                      4         266292802 ns/op        877880950 B/op    371153 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-5000-12                      2         793737479 ns/op        3941308872 B/op  1833743 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-diskqueue-ok-10000-12                     1        1431292292 ns/op        7490799232 B/op  3652602 allocs/op
```
vs
```
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-100-12                           1        2784037834 ns/op        190057600 B/op     54917 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-200-12                           1        5056939708 ns/op        396214672 B/op    108035 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-300-12                           1        7918490750 ns/op        594004640 B/op    160804 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-400-12                           1        9941767500 ns/op        808627368 B/op    213526 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-500-12                           1        12567918500 ns/op       970593912 B/op    265689 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-1000-12                          1        25531277791 ns/op       1902430208 B/op   528527 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-5000-12                          1        116076831167 ns/op      7697970024 B/op  2579080 allocs/op
BenchmarkExtensionAsPersistentQueueWithWorkers/bench-bbolt-ok-10000-12                         1        221522530458 ns/op      13998794896 B/op         5113022 allocs/op
```

internal/extension/diskqueuestorageextension/extension.go
```diff
@@ -0,0 +1,179 @@
...
+
+	var setMetadata *storage.Operation
+	var changeOp *storage.Operation
+	for _, op := range ops {
+		if op.Type == storage.Set && op.Key == metadataKey {
+			setMetadata = op
+		} else {
+			changeOp = op
+		}
+	}
```
dmitryax, line 109: This is not enough to cover all calls. One important path is the crash recovery where the Batch is called with a set of Delete and Set operations. See `enqueueNotDispatchedReqs`. 

Can be done as a follow up

internal/extension/diskqueuestorageextension/extension.go
```diff
@@ -0,0 +1,179 @@
...
+	if setMetadata == nil || changeOp == nil {
+		return errors.New("invalid batch")
+	}
+
+	switch changeOp.Type {
+	case storage.Set:
+		return c.queue.put(setMetadata.Value, changeOp.Value)
+	case storage.Get:
+		message := <-c.queue.peek(setMetadata.Value)
```
dmitryax, line 118: Looks like this updates the in-memory metadata, but doesn't persist it to make the "read_data + write_metadata" call atomically durable

internal/extension/diskqueuestorageextension/diskqueue.go
```diff
@@ -0,0 +1,570 @@
...
+					d.logger.Error(" failed to Remove", zap.String("name", d.name), zap.String("filename", f), zap.Error(err))
+				}
+			}
+			d.metadata.metadata.Store(c.metadata)
+		case <-syncTicker.C:
+			if peekOps == 0 {
+				// avoid sync when there's no activity
+				continue
+			}
+			if peekOps == d.syncEvery {
```
dmitryax, line 469:
```suggestion
			if peekOps >= d.syncEvery {
```

I can see how the counter can overshoot in this switch section

dmitryax (commented):
Good to merge as a first version. It's only durable on writes though: read progress is split across peek.dat and qmv0 in meta.dat, written by two separate goroutines. I don't think reads can be durable and atomic while they commit concurrently from a separate goroutine. I think this should be changed to one state file (write pos + read pos + qmv0) written by a single writer. The file would be an append-only log of self-describing records that include all the metadata

dmitryax (approved):

PR #7957, opened 2026-08-21
feat: example of both redaction and transform processors

weliasz:
Commented example of redaction processor that could be used to mask sensitive data.
https://splunk.atlassian.net/browse/ADDON-89072

### DIFF
--- a/cmd/otelcol/config/collector/splunk_logs_config_windows.yaml
+++ b/cmd/otelcol/config/collector/splunk_logs_config_windows.yaml
@@ -167,14 +167,38 @@ processors:
     detectors: [gcp, ecs, ec2, azure, system]
     override: true
 
-  # Use below example redaction processor to mask sensitive data
+  # The processors below assume the Windows Event Log body is structured JSON (raw: false). They can target nested fields such as "event_data.ProcessId".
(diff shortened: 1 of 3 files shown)


cmd/otelcol/config/collector/splunk_logs_config_windows.yaml
```diff
@@ -167,6 +167,15 @@ processors:
     detectors: [gcp, ecs, ec2, azure, system]
     override: true
 
+  # Use below example redaction processor to mask sensitive data
+  # https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/redactionprocessor/README.md
+  #  redaction/windows_sensitive_logs:
+  #    allow_all_keys: true
+  #    summary: silent
+  #    blocked_values:
+  #      AWS IPv4 addresses
```
pjanotti: nit suggestion so people can just copy and paste and uncomment the example
```suggestion
  #      # AWS IPv4 addresses
```

pjanotti (commented):
Is it possible to create an example redacting specific parts of the event payload? I mean there is the "message" but the event layout can have "event_data" and which will have specific ids. It will be great if the example is actually against a Windows event.

pjanotti (approved):

PR #7977, opened 2026-08-25
Prevent creation of invalid metrics pipeline by discovery

kelseyma:
Currently, discovery can create an invalid metrics pipeline if there is no default metrics pipeline configured. This results in the collector crashing with error: `invalid configuration: service::pipelines::metrics: must have at least one exporter`

This was seen in tests that enable both OpAMP Supervisor + discovery because supervisor will first create a nop bootstrap collector to get AgentDescription data. The nop collector doesn't have a metrics pipeline and would fail with this error.

Behavior after fix:
- `service::pipeline::metrics` exists: activate metrics discovery
- `service::pipelines::logs/entities` exists: activate continuous entity discovery
- Both exist: activate both
- Neither exists: cleanup-only return for Supervisor bootstrap

- Added unit tests
- Added e2e test with `--discovery` flag with installer script
- Tested locally with dev build: verified supervisor and collector were able to start

### DIFF
--- a/internal/configconverter/discovery.go
+++ b/internal/configconverter/discovery.go
@@ -26,8 +26,9 @@ import (
 
 // SetupDiscovery will find `service::<extensions|receivers>/splunk.discovery` entries
 // provided by the discovery confmap.Provider and relocate them to
-// `service::extensions` and `service::pipelines::metrics::receivers`,
-// by appending them to existing sequences, if any.
+// `service::extensions` and, when the default metrics pipeline exists,
+// `service::pipelines::metrics::receivers`, by appending them to existing
+// sequences, if any.
 func SetupDiscovery(_ context.Context, in *confmap.Conf) error {
 	if in == nil {
 		return nil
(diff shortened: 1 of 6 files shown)


agoallikmaa (approved):

internal/configconverter/discovery.go
```diff
@@ -129,11 +130,14 @@ func getDiscoReceivers(service map[string]any) (bool, []any, error) {
 }
 
 func getMetricsPipelineAndReceivers(pipelines map[string]any) (map[string]any, []any, error) {
-	metricsPipeline := map[string]any{}
-	if mp, ok := pipelines["metrics"]; ok && mp != nil {
-		metricsPipeline = mp.(map[string]any)
+	mp, ok := pipelines["metrics"]
+	if !ok || mp == nil {
```
dmitryax, line 140: I think this check should gate execution of `SetupDiscovery` not just silently ignore this step

kelseyma: @dmitryax Updated to have the following behavior:
- `service::pipeline::metrics` exists: activate metrics discovery
- `service::pipelines::logs/entities` exists: activate continuous entity discovery (was existing expected behavior, wasn't sure if we wanted to change this)
- Both exist: activate both
- Neither exists: cleanup-only return for Supervisor bootstrap

dmitryax (approved):
LGTM

PR #7995, opened 2026-08-29
Use SHA-256 for RPM per-file digests

jinja2:

Use SHA-256 instead of MD5 for per-file digests in newly built Collector and auto-instrumentation RPMs. Both RPM builders verify their output and fail if SHA-256 file digests are missing. Existing published RPMs are unchanged, and this does not change the packaged files, permissions, scripts, dependencies, configuration, or runtime behavior. Newly built RPMs require RPM 4.6 or newer.

The existing GitHub package integration test verifies the installed Collector RPM’s digest algorithm. After GitLab publishes signed RPMs, new jobs install the exact candidate and newest previous versions of both packages on x86_64 and aarch64. Candidate checks also verify package ownership and individual SHA-256 digests for the Collector, launcher, OpAMP Supervisor, injector library, and packaged Java, Node.js, and .NET instrumentation assets.

The repository-install jobs run against the test stage on normal `main` and non-main dry-run pipelines, and against the beta stage for beta/RC pipelines. Stable release-stage installation is intentionally excluded until the job runtime is understood.

### DIFF
--- a/.chloggen/rpm-sha256-file-digest.yaml
+++ b/.chloggen/rpm-sha256-file-digest.yaml
@@ -1,0 +1,6 @@
+change_type: enhancement
+component: packaging
+note: Use SHA-256 instead of MD5 for per-file digests in newly built Collector and auto-instrumentation RPMs.
+issues: []
+subtext: |
+  Existing published RPMs are unchanged. Newly built packages require RPM 4.6 or later.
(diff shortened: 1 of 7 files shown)


.chloggen/rpm-sha256-file-digest.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: enhancement
+component: packaging
+note: Use SHA-256 instead of MD5 for per-file digests in newly built Collector and auto-instrumentation RPMs.
+issues: []
```
jinja2:
```suggestion
issues: [7995]
```

kelseyma (approved):

pjanotti (approved):

PR #8013, opened 2026-09-01
[chore] fix Makefile tgz target

atoulme:
Fix how we silence the script and make sure we don't have long paths in the archive.

### DIFF
--- a/packaging/otlpinput/Makefile
+++ b/packaging/otlpinput/Makefile
@@ -1,5 +1,3 @@
-include ../../Makefile.common
-
 GOCMD?=go
 
 VERSION ?= $(shell git describe --match "v[0-9]*" HEAD)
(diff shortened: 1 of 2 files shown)


pjanotti (approved):

packaging/otlpinput/Makefile
```diff
@@ -1,22 +1,20 @@
-include ../../Makefile.common
```
dmitryax, line 1: Why do we exclude this?

atoulme: it's not needed by this package. Similar to https://github.com/signalfx/splunk-otel-collector/blob/main/packaging/ta-v2/Makefile#L1

dmitryax (approved):

PR #8020, opened 2026-09-03
[chore][release] Auto-tag the baseline module on release version bumps

dmitryax:
## Problem

`baseline` (`github.com/signalfx/splunk-otel-collector/baseline`) is consumed in-repo via a `replace` directive but is meant to be published as a standalone Go module that flavours resolve by tag (`go get .../baseline@vX.Y.Z`).

## Proposed fix

Keep `replace` for the in-repo build and make the published tag drift-free by requiring every change to baseline's shipping files to bump a declared version, then tagging on merge. The approach is inspired by the Helm `chart-releaser` model.

- **`baseline/VERSION`** declares the module version (seeded at `v0.159.0`, matching the current tree).
- **`baseline-version-guard.yml`** (PR check): if a PR changes baseline shipping files (`**/*.go`, `go.mod`, `go.sum`), `baseline/VERSION` must be strictly greater than the latest `baseline/*` tag. Doc-only changes under `baseline/` don't require a bump.
- **`baseline-tag.yml`** (push to `main`): read `baseline/VERSION`; create signed `baseline/vX.Y.Z` if absent; no-op if present and the tree matches; fail if present but the tree differs.
- **`update-deps`** derives the next baseline version from `OTEL_VERSION` and writes `baseline/VERSION` when the dep sweep changed baseline's shipping files, so the dep-update PR carries the bump. Stable tracks upstream major+minor at `vX.Y.0`; nightly increments the rc. Cherry-picks need a manual bump; the guard fails with an explicit hint.


update-deps
```diff
@@ -57,3 +57,20 @@ for gomod in $( find "$REPO_DIR" -name "go.mod" | grep -v "/examples/" | sort );
...
+source "${REPO_DIR}/.github/workflows/scripts/baseline-version.sh"
+(
+    cd "$REPO_DIR"
+    latest_tag="$( latest_baseline_tag )"
+    if [ -n "$latest_tag" ] && ! baseline_shipping_changed "$latest_tag"; then
+        exit 0
+    fi
+    ref="${latest_tag#baseline/}"
+    [ -z "$ref" ] && ref="$( read_version )"
+    next="$( derive_baseline_version "$OTEL_VERSION" "$ref" )"
```
jinja2, line 73: I think `latest` is getting mixed up with main here. The scheduled job passes `OTEL_VERSION=latest`, and go get resolves that to a stable release, but this line still passes the literal word latest to derive_baseline_version. Since that function only recognizes vX.Y.Z as stable, won't this end up generating an RC/nightly version?

baseline/VERSION
```diff
@@ -0,0 +1 @@
+v0.159.0
```
jinja2, line 1:
```suggestion
v0.160.0
```
the dep update was merged

.github/workflows/baseline-tag.yml
```diff
@@ -0,0 +1,53 @@
...
+            fi
+            echo "${tag} already exists and matches HEAD; nothing to do."
+            exit 0
+          fi
+
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+          echo ">>> Creating tag ${tag} at $(git rev-parse HEAD) ..."
+          git tag -a "$tag" -m "Release ${tag}"
```
jinja2, line 52: The tag won't be signed, but i think that it is okay with `go get`?

.chloggen/baseline-auto-tag.yaml
```diff
@@ -0,0 +1,14 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. crosslink)
+component: baseline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Publish the `baseline` module on a version tag. `baseline/VERSION` declares the module version, a CI guard requires a bump when baseline's shipping files change, and merges to `main` auto-tag `baseline/vX.Y.Z`.
```
jinja2, line 8: Can we briefly explain what baseline provides and remove the ci details? Maybe:

```suggestion
note: Publish the `baseline` module, the shared set of upstream OpenTelemetry Collector components that Splunk Collector distributions build on, for reuse by other distributions.
```

dmitryax:
Closed in favor of https://github.com/signalfx/splunk-otel-collector/pull/8044

PR #8025, opened 2026-09-04
[chore] implement promql receiver

atoulme:
Implement more functionality associated with the promql receiver.


internal/receiver/promqlreceiver/scraper.go
```diff
@@ -130,69 +139,200 @@ func (s *scraper) runOneQuery(_ context.Context, endpointURL *url.URL, q string,
...
+	case parser.ValueTypeString:
+		s.settings.Logger.Error("PromQL response of type string is not supported:", zap.String("query", q.Query))
+	case parser.ValueTypeScalar:
+		s.settings.Logger.Error("PromQL response of type scalar is not supported:", zap.String("query", q.Query))
+	case parser.ValueTypeMatrix:
+		var matrix promql.Matrix
+		if err := json.Unmarshal(qData.Result, &m); err != nil { //nolint:staticcheck // staticcheck fails to check promql.Matrix
+			return err
+		}
+		if sm, ok := convertMatrix(matrix, q.MetricName); ok {
```
pjanotti: Given the implementation of `convertMatrix` the last parameter, `q.MetricName` seems more like a fallback metric name, if the `matrix` doesn't provide one. The name in the config doesn't imply anything of this behavior.

internal/receiver/promqlreceiver/scraper.go
```diff
@@ -130,69 +139,200 @@ func (s *scraper) runOneQuery(_ context.Context, endpointURL *url.URL, q string,
...
-		switch sample.Metric.Get(model.MetricTypeLabel) {
+		switch metricType {
 		case string(model.MetricTypeGauge):
 			if metric.Type() == pmetric.MetricTypeEmpty {
 				metric.SetEmptyGauge()
 			}
 			dp := metric.Gauge().DataPoints().AppendEmpty()
 			dp.SetDoubleValue(sample.F)
-			dp.SetTimestamp(pcommon.Timestamp(sample.T)) //nolint:gosec // disable G115
+			dp.SetTimestamp(pcommon.Timestamp(sample.T * 1e+6)) //nolint:gosec // disable G115 // convert from ms to ns
```
pjanotti: The individual data point code is very similar to the one for the loop to avoid the risk of inconsistency between matrix and vector it will be good to have the common code in helper functions.

pjanotti (approved):
Two suggestions but they shouldn't block merge.

internal/receiver/promqlreceiver/scraper.go
atoulme: naming things is hard. Is fallback metric name a better approach? Should we set it always if it's set? I don't think I got to a good place.

pjanotti: any understanding when the matrix may not return a name for the metric?

pjanotti: what is the downside of enforcing the name from the config? (besides it becoming a required non-empty field)

internal/receiver/promqlreceiver/scraper.go
atoulme: This is now done.

internal/receiver/promqlreceiver/scraper.go
atoulme: It can happen if we perform transformations as part of the query. I would like to have an integration test specifically targeting this - will do.

PR #8027, opened 2026-09-04
[receiver/gnmireceiver] Add enum_values metric config option to emit the full state set

jkoronaAtCisco:
**Description:**

Adds an `enum_values` option to a leaf's metric config (`default/overrides`) for string leaves with a known, closed set of possible values (e.g. `oper-status: UP/DOWN/TESTING`). When set, the receiver emits an `_state` metric with one datapoint per declared value (`1` for the active value, `0` for the rest) and a `state` attribute, instead of the plain `_info/value` metric used for unconfigured strings. This makes state transitions unambiguous in a metrics pipeline: the same notification that reports the new value also zeroes out the previous one, avoiding the stale-timeseries problem noted as a follow-up in the receiver's README. Values are matched after trimming whitespace and stripping any module: prefix (so `openconfig-interfaces:UP` and `UP` are  equivalent); a received value outside the declared set emits `0` for every state and is reported as an error rather than silently minting a new series. `enum_values` is only valid with `type: gauge` and is rejected at config-validation time for `type: sum`.

It's follow up to [#7918](https://github.com/signalfx/splunk-otel-collector/pull/7918#pullrequestreview-4920873389).

Added unit tests and extended integration test to cover `enum_values` use cases.

 Updated `README.md`.


internal/receiver/gnmireceiver/config.go
```diff
@@ -120,6 +120,13 @@ type MetricConfig struct {
 	// Unit is the metric unit, ideally in UCUM (e.g. "By", "1", "By/s").
 	// Optional.
 	Unit string `mapstructure:"unit"`
+
+	// EnumValues declares the closed set of values a string leaf can take.
+	// When set, the leaf is emitted as a "_state" metric with one datapoint
+	// per declared value (1 for the active value, 0 for the rest).
+	// Applies only to leaves that arrive as strings; it has no effect
+	// on numeric or boolean  values. Optional.
```
pjanotti, line 128: nit of neatness: extra space in the comment ... (now I need to hide)
```suggestion
	// on numeric or boolean values. Optional.
```

internal/receiver/gnmireceiver/config.go
```diff
@@ -219,5 +226,22 @@ func (m *MetricConfig) Validate() error {
...
+		seen := make(map[string]string, len(m.EnumValues))
+		for _, v := range m.EnumValues {
+			if strings.TrimSpace(v) == "" {
+				return errors.New("enum_values must not contain empty values")
+			}
+			normalized := normalizeEnumValue(v)
+			if orig, ok := seen[normalized]; ok {
+				return fmt.Errorf("enum_values contains duplicate values %q and %q", orig, v)
+			}
+			seen[normalized] = v
```
pjanotti, line 243: I see that below the code enumerates over `EnumValues` and call `normalizeEnumValue` again over it. It sounds that perhaps it is worth to either save the results from here, via a private config member, or create a similar on receiver start instead of looping over the field again. Yeah, very small practical effect but some very small compute savings.

internal/receiver/gnmireceiver/parser.go
```diff
@@ -297,6 +302,50 @@ func (p *metricParser) writeInfo(
...
+	unit := cfg.Unit
+	if unit == "" {
+		unit = "1"
+	}
+
+	normalized := normalizeEnumValue(value)
+	matched := false
+	for _, enumValue := range cfg.EnumValues {
+		var n int64
+		if normalizeEnumValue(enumValue) == normalized {
```
pjanotti, line 319: This is the call to `normalizeEnumValue` that could be avoided.

internal/receiver/gnmireceiver/parser.go
```diff
@@ -297,6 +302,50 @@ func (p *metricParser) writeInfo(
...
+			matched = true
+		}
+
+		dp, err := b.numberDataPoint(name, unit, pmetric.MetricTypeGauge)
+		if err != nil {
+			return err
+		}
+		dp.SetIntValue(n)
+		dp.SetTimestamp(ts)
+		dp.Attributes().PutStr(stateValueAttr, normalizeEnumValue(enumValue))
```
pjanotti, line 330: Another one

internal/receiver/gnmireceiver/parser.go
```diff
@@ -297,6 +302,50 @@ func (p *metricParser) writeInfo(
...
+		putAttrs(dp.Attributes(), keys)
+	}
+
+	if !matched {
+		return fmt.Errorf("value %q for %q is not in enum_values %v; emitted all states as 0",
+			value, metricName(origin, elems), cfg.EnumValues)
```
pjanotti, line 336: Why not use the `name` created at the top?


```suggestion
		return fmt.Errorf("value %q for %q is not in enum_values %v; emitted all states as 0",
			value, name, cfg.EnumValues)
```

pjanotti: Thought: should we consider a fallback state to represent this? I don't want to mix that with this PR but something that crossed my mind.

internal/receiver/gnmireceiver/parser.go
```diff
@@ -297,6 +302,50 @@ func (p *metricParser) writeInfo(
 	return nil
 }
 
+func (p *metricParser) writeEnumState(
+	b *parseBatch, origin string, elems []string,
+	keys map[string]string, cfg MetricConfig, value string, ts pcommon.Timestamp,
+) error {
+	name := metricName(origin, elems) + stateMetricSuffix
```
pjanotti, line 309: nit:
```suggestion
	metricName := buildMetricName(origin, elems) + stateMetricSuffix
```

internal/receiver/gnmireceiver/parser.go
```diff
@@ -297,6 +302,50 @@ func (p *metricParser) writeInfo(
...
+	}
+	return nil
+}
+
+func normalizeEnumValue(value string) string {
+	value = strings.TrimSpace(value)
+	if idx := strings.LastIndexByte(value, ':'); idx >= 0 {
+		return value[idx+1:]
+	}
+	return value
```
pjanotti, line 346: nit: the practical difference should be invisible, but... the trim left (if it ever happens) is wasted when there is a `:` in the string... 
```suggestion
	if idx := strings.LastIndexByte(value, ':'); idx >= 0 {
		value = value[idx+1:]
	}
	return strings.TrimSpace(value)
```

internal/receiver/gnmireceiver/README.md
```diff
@@ -45,6 +45,13 @@ receivers:
               out-octets:
                 type: sum
                 unit: By
+          - path: /interfaces/interface/state/oper-status
+            mode: on_change
+            overrides:
+              oper-status:
+                type: gauge
+                # closed set of values this leaf can take; see "Enum leaves" below
+                enum_values: [UP, DOWN, TESTING]
```
pjanotti, line 54: If a fallback state is desirable it probably should have a configurable name, but as mentioned before, it is more a future consideration.

pjanotti (approved):

PR #8035, opened 2026-09-07
[chore][ci] Remove debian-bullseye, aka debian 11, from CI

pjanotti:
Assisted-by: Codex


packaging/installer/install.sh
```diff
@@ -40,9 +40,6 @@ get_distro_codename() {
 
   if [ "$distro" = "debian" ] && [ -z "$codename" ]; then
     case "$( get_distro_version )" in
```
pjanotti, line 42: I'm not sure if we removing from the install script as soon as the OS reaches EOL or not. I will check that. Anyway, if we remove it from the script I will add a chlog file.

pjanotti: I opted to keep the debian 11 support on the install script for now

jinja2 (approved):

PR #8037, opened 2026-09-08
Move splunk_outputs exporter into pkg/exporter/splunkoutputsexporter

dmitryax:
Brings the `splunk_outputs` exporter into this repo as a new module under `pkg/exporter/splunkoutputsexporter`, instead of consuming it from [tarunner](https://github.com/splunk/tarunner). Mirrors the `splunk_inputs` receiver move (#8029); a follow-up removes the copy from tarunner (splunk/tarunner#136).

The exporter still depends on `github.com/splunk/tarunner/pkg/splunkta` (`conf`, `stanza`, `tabuilder`) for TA config parsing and output-exporter construction, pinned via a tarunner pseudo-version. `internal/components` registers `splunk_outputs` behind the existing `enableTARunner` gate, alongside `splunk_inputs`. Gated off by default; no behavior change.

The Go source is a 1:1 copy from tarunner. Changes over the copy: module path repointed to this repo, dropped tarunner's local `replace`, pinned `github.com/splunk/tarunner` to the pseudo-version, self-import in the external test repointed, and a `Makefile` added (`include ../../../Makefile.Common`) for the per-module targets. No formatting or lint fixes were needed.


pkg/exporter/splunkoutputsexporter/go.mod
```diff
@@ -0,0 +1,111 @@
+module github.com/signalfx/splunk-otel-collector/pkg/exporter/splunkoutputsexporter
+
+go 1.26.4
```
pjanotti: We are already at 1.26.6, btw, splunk_inputs should be updated too. Anyway, it can be done separately if we want to pull as-it-is first.

pjanotti (approved):

PR #8043, opened 2026-09-09
[chore] Move splunkta core into pkg/splunkta and drop tarunner dependency

dmitryax:
Move the shared splunkta core out of `github.com/splunk/tarunner` into a new `pkg/splunkta` module and repoint the `splunk_inputs` receiver and `splunk_outputs` exporter onto it. This drops `github.com/splunk/tarunner` from the collector entirely and reverses the dependency: tarunner will import this code from SOC instead.

The moved packages are `conf`, `stanza`, `tabuilder`, `receiver/*` (batch, monitor, script, tcp, udp, wineventlog, filter), `operator/prop`, `operator/transform`, `script`, `scriptedinput`, and `featuregates`.

## Changes on top of the verbatim copy

The core is otherwise a verbatim copy. Everything that differs from the tarunner source:

- **Import path rewrite.** Every `github.com/splunk/tarunner/...` import in the moved files becomes `github.com/signalfx/splunk-otel-collector/pkg/splunkta/...`. Mechanical prefix swap, no logic change.
- **Module wiring.** New `pkg/splunkta/go.mod` (`module .../pkg/splunkta`, `go 1.26.6`) and `Makefile` (`include ../../Makefile.Common`). Both plugins add a `require` on `pkg/splunkta` plus a relative `replace => ../../splunkta`; the main module adds the same `replace` and drops `github.com/splunk/tarunner` from its require block after `tidy`.
- **Scoped `pkg/splunkta/.golangci.yml`.** Relaxes the linters the pre-existing code trips (`gocritic`, `gosec`, `perfsprint`, `revive`, `testifylint`, govet `shadow`) so this lands as a pure move. Precedent: `pkg/receiver/smartagentreceiver/.golangci.yml`.
- **`//nolint:contextcheck` in `splunkoutputsexporter/subexporter.go`.** `tabuilder.CreateOutputExporter` takes no context; now that `tabuilder` is a local package, contextcheck sees through it.
- **`scriptedinput` race skip.** `Test_ScriptedInput` is skipped under the race detector via a build-tagged `raceDetectorEnabled` flag (`race_on_test.go` / `race_off_test.go`). The moved code has a data race between `_execute`'s `cmd.Wait` and the stdout reader goroutine; tarunner CI never ran `-race`, so it stayed latent.
- **`fo` -> `fakeOut` rename in `scriptedinput/input_test.go`** and **`Memeory` -> `Memory` in `operator/transform/parser_test.go`.** SOC's spellcheck (codespell and typos, not run in tarunner) flagged both.

These components are dev-stability behind the `enableTARunner` feature gate, so no changelog entry (matching the earlier receiver/exporter move PRs).

The `codecov/patch` check fails because codecov treats the whole moved package as new lines and applies the 85% patch target; the code carries its original tarunner coverage. It is not a required check.


jinja2:
@dmitryax do we not want to add it to https://github.com/signalfx/splunk-otel-collector/blob/main/versions.yaml?

dmitryax:
@jinja2 good catch. Thanks, Jina. The file updated. This reminds me that we need some enforcement for the versions.yaml completeness

jinja2 (approved):

pjanotti (approved):

PR #8044, opened 2026-09-09
[chore][release] Tag baseline and pkg/* modules on release via multimod

dmitryax:
Publish the `baseline` and `pkg/*` Go modules as version tags on each release so external consumers can import them at a stable released version, instead of only the root module being tagged.

Adopt an OpenTelemetry-style `versions.yaml` declaring a single `splunk-otel-collector` module set (root, `baseline`, and the seven `pkg/*` modules) at one version. `.gitlab/create-github-tag.sh` now runs `multimod tag` to create and push the per-module tags (`vX.Y.Z`, `baseline/vX.Y.Z`, `pkg/*/vX.Y.Z`) from the existing GitLab release job. Tags are signed the same way as before (`multimod` shells out to `git tag -a -s`).

The `manual_release_trigger` job asserts `versions.yaml` was bumped to `RELEASE_VERSION` and that version is not already tagged, failing early with a hint otherwise. So a release can't publish a stale or duplicate module version. `RELEASE_VERSION` stays the release input (empty still means a test run); bump `versions.yaml` to the next release as part of the release prep.

`baseline` and `pkg/*` keep their in-repo `replace` directives; the tags exist only for external consumers. `internal/tools`, `tests*`, `packaging/ta-v2/tests`, `internal/signalfx-agent`, and the `examples/*` modules are excluded from the set.

`versions.yaml` seeds at `v0.160.0` (the current latest release), so the next release bumps it to `v0.160.0` and tags the full set for the first time.

### Follow-ups

- Make `versions.yaml` the single source of truth and drop the manual `RELEASE_VERSION` input. Out of scope here because empty `RELEASE_VERSION` currently doubles as the test-run flag across the pipeline.


.gitlab-ci.yml
```diff
@@ -95,6 +95,15 @@ manual_release_trigger:
         fi
       fi
 
+      # versions.yaml drives the module set tags created at release time; assert
+      # it was bumped to this release so the tags match.
+      SET_VERSION=$(awk '/^  splunk-otel-collector:/{f=1} f&&/^    version:/{print $2; exit}' versions.yaml)
+      if [ "$SET_VERSION" != "$RELEASE_VERSION" ]; then
+        echo "versions.yaml has the splunk-otel-collector module set at $SET_VERSION but release is $RELEASE_VERSION." >&2
```
pjanotti, line 102: We can improve this later, but ideally this should be verified before we get to GitLab, perhpas on the "prepare release" PR.

pjanotti (approved):

.gitlab-ci.yml
dmitryax: This is initial check in the release step which verifies that the version submitted manually to the Gitlab CI job matches the versions.yaml file. Similar to the check above

.gitlab/create-github-tag.sh
```diff
@@ -31,8 +38,10 @@ cd repo-tmp
...
-git tag -s "$VERSION_TAG" "$COMMIT_SHA" -m "Release $VERSION_TAG"
-
-echo ">>> Pushing tag $VERSION_TAG to GitHub ..."
-git push origin "$VERSION_TAG"
+echo ">>> Creating signed tags for module set $MODULE_SET ($VERSION_TAG) at $COMMIT_SHA ..."
+multimod tag --module-set-name "$MODULE_SET" --commit-hash "$COMMIT_SHA" --print-tags \
+  | sort -u \
+  | while IFS= read -r tag; do
+      echo ">>> Pushing tag $tag to GitHub ..."
+      git push origin "$tag"
```
jinja2: Not blocking, but I think we can use `--atomic` option here to push all tags in same cmd so that either the full module set is published or none of it is.

.chloggen/tag-modules-on-release.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: enhancement
+component: collector
+note: Publish the baseline and pkg/* Go modules as version tags on each release, so external consumers can import them at a stable released version
+issues: [7954]
```
jinja2:
```suggestion
issues: [8044]
```

jinja2 (approved):

versions.yaml
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+
+# Module sets for the multimod releaser (go.opentelemetry.io/build-tools/multimod).
+# The release job tags every module in the splunk-otel-collector set at this
+# version on each release. Bump `version` to the next release before releasing;
+# the release job fails if it is still at an already-tagged version.
+module-sets:
+  splunk-otel-collector:
+    version: v0.160.0
+    modules:
```
jinja2, line 22: Just to confirm, we don't need to run the [prerelease cmd](https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/multimod/README.md#prepare-a-prerelease-commit) since we don't have any dependencies between any of these modules in this set?

PR #8047, opened 2026-09-10
[receiver/gnmi] Implement yang-based type & unit resolution

jkoronaAtCisco:

Adds a `yang_modules` config option to the gNMI receiver. Today, every leaf's metric type (gauge/sum), unit, and enum value set must be hand-declared in `default`/`overrides`, which is redundant (the device's YANG models already state it) and a correctness hazard — a hand-maintained `enum_values` list silently goes stale when a vendor adds a value, and numeric-ness of string leaves is guessed via `ParseInt`/`ParseFloat`.

`yang_modules` points at YANG module files/directories loaded once at startup; the receiver derives each leaf's type, unit, and enum/identityref value set from the schema.

Resolution is layered: an explicit `overrides` entry always wins outright; the schema is checked next; `default` fills in anything the schema left blank (commonly `unit`, since OpenConfig models rarely declare a `units` statement). A leaf the schema can't resolve falls back to `overrides`/`default` exactly as before. Without `yang_modules` configured, behavior is unchanged — `type` and at least one of `default`/`overrides` are still required per leaf; with it set, both become optional.

Added unit tests for verify schema-related functionality.

Updated `README.md` with a new "YANG schema resolution (`yang_modules`)" section.


internal/receiver/gnmireceiver/README.md
```diff
@@ -190,17 +190,64 @@ Notes:
...
-`default`. **A leaf matched by neither is dropped** (no metric is emitted). A subscription
-must therefore define at least one of `default` or `overrides`.
+The receiver resolves these either from the YANG schema (`yang_modules`, below) or from
+configuration, with an explicit `overrides` entry always taking precedence over both:
+
+1. `overrides[<leaf name>]`, if present, wins outright.
+2. Otherwise, the loaded YANG schema, if `yang_modules` is set and it resolves the leaf.
+   `default` still fills in any of `type`/`unit`/`enum_values` the schema left blank
+   (most commonly `unit`: OpenConfig models rarely declare a `units` statement).
+3. Otherwise, `default`, if present.
```
jinja2, line 207: Could we safely merge non-empty override fields onto the schema result?

A unit-only override seems like a genuine use case:
generic shared YANG schema:
```
in-octets:
  type: counter64
  unit: bytes
```

config:
```
overrides:
  in-octets:
    unit: By
```

Ideally, the effective configuration would be:
```
type: sum  # derived from YANG
unit: By   # supplied by the override
```

internal/receiver/gnmireceiver/config.go
```diff
@@ -153,6 +165,29 @@ func (cfg *Config) Validate() error {
...
+			prefix := fmt.Sprintf("targets[%d].subscriptions[%d]", ti, si)
+			if sub.Default == nil && len(sub.Overrides) == 0 {
+				return fmt.Errorf("%s: at least one of \"default\" or \"overrides\" must be specified when yang_modules is not set", prefix)
+			}
+			if sub.Default != nil && sub.Default.Type == "" {
+				return fmt.Errorf("%s.default: type is required when yang_modules is not set", prefix)
+			}
+			for leaf, override := range sub.Overrides {
+				if override.Type == "" {
+					return fmt.Errorf("%s.overrides[%q]: type is required when yang_modules is not set", prefix, leaf)
```
jinja2, line 186: Since override wins even when yang schema is given, I don't understand why we have this exception for type being conditionally not needed

internal/receiver/gnmireceiver/testdata/yang/test-interfaces.yang
```diff
@@ -0,0 +1,90 @@
+module test-interfaces {
+  yang-version 1.1;
+  namespace "urn:test:interfaces";
+  prefix ti;
+
+  import test-identities { prefix tid; }
+
+  typedef counter64 {
```
jinja2, line 8: I see from the [YANG types used by OpenConfig](https://github.com/openconfig/public/blob/master/third_party/ietf/ietf-yang-types.yang#L116-L175) that typedefs can be chained (for e.g. `zero-based-counter64` is derived from `counter64`)

Could we add a test that parses a similar typedef chain and verifies that a leaf using `zero-based-counter64` resolves to a sum? I’m not familiar enough with goyang to tell whether the current check preserves the intermediate counter64 type, so testing the parsed model would help confirm that this works as expected.

internal/receiver/gnmireceiver/testdata/yang/test-interfaces.yang
```diff
@@ -0,0 +1,90 @@
+module test-interfaces {
```
jinja2, line 1: Could we add a test covering two loaded modules that define the same unqualified path? For e.g., one module could define `/interfaces/state/value` as a numeric value with unit bytes, while another defines the same path as a string. I’m wondering whether the prefix stripping when indexing could cause issues for such cases (not sure how likely it is to be given such input).

internal/receiver/gnmireceiver/README.md
jkoronaAtCisco: Good point, it's real issue not just documentation gap. I fixed it in `resolve()`: overrides now merge onto the schema/default result per field. The override wins for whatever field it sets, the schema fills in what's left blank, and default fills in anything still blank after that.

internal/receiver/gnmireceiver/config.go
jkoronaAtCisco: This exception is intentional, and now that `resolve()` merges overrides/default onto the schema result per field, the rationale actually holds: a type left blank in overrides or default is expected to be filled in by the schema at runtime, not a validation gap. 

Added a comment in code that explains it.

internal/receiver/gnmireceiver/testdata/yang/test-interfaces.yang
jkoronaAtCisco: Confirmed — `t.Root` collapses to the ultimate primitive type (e.g. `uint64`), never an intermediate typedef name. So the fallback check could never fire for a multi-level typedef chain: a leaf typed as `zero-based-counter64` was silently resolving to gauge instead of sum. 

Fixed `isCounterType` to walk `t.Base` instead, which does thread through intermediate typedefs. Added `zero-based-counter64` to the test schema with a leaf using it, plus a test proving it resolves to sum.

internal/receiver/gnmireceiver/testdata/yang/test-interfaces.yang
jkoronaAtCisco: Confirmed — two loaded modules can collide today. Two independently loaded modules defining the same path silently have the last-processed module's resolution win, with no warning.

I added validation for that case: if two modules resolve the same path to different types/units/enums, startup  fails with an error naming the conflicting path instead of picking one silently.

internal/receiver/gnmireceiver/parser.go
```diff
@@ -372,19 +394,81 @@ func metricDataType(cfgType string) pmetric.MetricType {
...
 	sub := p.subscriptionFor(origin, elems)
 	if sub == nil {
-		return MetricConfig{}, false
+		return resolvedMetric{}, false
 	}
+
 	leaf := elems[len(elems)-1]
-	if cfg, ok := sub.Overrides[leaf]; ok {
-		return cfg, true
+	schemaRM, schemaOK := p.schema.lookup(elems)
```
jinja2, line 404: Could you clarify how `origin` is expected to interact with `yang_modules`? From docs, my understanding is that an origin identifies schema/data tree a path belongs to, and YANG modules may use `oc-ext:origin` to associate themselves with such a tree. Since schema lookup here only uses the path elements, would two origins that expose the same path conflict? Is origin-aware schema resolution something this receiver should support, or is yang_modules intentionally limited to a single schema tree? If it is intentionally limited, I think we should document it.

.chloggen/gnmireceiver-yang-modules.yaml
```diff
@@ -0,0 +1,19 @@
...
+note: Add `yang_modules` config option to derive each leaf's metric type (gauge/sum), unit, and enum/identityref value set directly from YANG schema files instead of requiring them in `default`/`overrides`.
+
+# One or more tracking issues related to the change
+issues: [8047]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  An explicit `overrides` entry still wins outright over the schema, and `default` still fills in
```
jinja2, line 17: Please update this since there is merge support now

jkoronaAtCisco: Thanks!

internal/receiver/gnmireceiver/parser.go
jkoronaAtCisco: This is intentional for now, not an oversight: `yang_modules` is scoped to a single schema tree. Two origins that happen to expose the same path are not disambiguated — and per the cross-module conflict check from the earlier thread, if they disagree on type, loading now fails rather than silently picking one. I've documented this explicitly in a new "yang_modules and origin" README section and on the yangSchema type, rather than changing behavior. 

Per-origin schema resolution would be a reasonable follow-up if it turns out to be needed in practice, but I'd rather not expand this PR's scope for a case we don't have a concrete need for yet.

jinja2: Ack, but I don't see the readme or changelog update in the PR

jinja2 (approved):
LGTM but it would be good to have the changelog/readme updates as mentioned in the other comment

jkoronaAtCisco:
> LGTM but it would be good to have the changelog/readme updates as mentioned in the other comme

Forgot to push 🤦

PR #8048, opened 2026-09-10
feat: run otel on dedicated Windows service user

weliasz:
Use a dedicated Windows virtual service account, NT SERVICE\splunk-otel-collector, for fresh Collector MSI installations. 
Add SPLUNK_SERVICE_ACCOUNT_TYPE=virtual|localsystem and the corresponding install.ps1 -service_account_type option; upgrades preserve the existing account by default and warn with the explicit migration command when remaining on LocalSystem.

Provision only receiver-required access for the virtual account: performance counters, Windows event logs including Security, default file-log sources, and file-storage checkpoints. Add upgrade/account-selection coverage and custom-action unit tests.

Zrzut ekranu 2026-09-10 o 18 43 06

### DIFF
--- a/cmd/otelcol/config/collector/splunk_logs_config_windows.yaml
+++ b/cmd/otelcol/config/collector/splunk_logs_config_windows.yaml
@@ -361,7 +361,7 @@ extensions:
     directory: "${ProgramData}/Splunk/OpenTelemetry Collector/FileStorage"
     compaction:
       on_start: true
-      directory: ${env:TEMP}
+      directory: "${ProgramData}/Splunk/OpenTelemetry Collector/FileStorage/Temp"
     create_directory: true
 
 service:
--- a/packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
+++ b/packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
@@ -1,0 +1,431 @@
+﻿// Copyright Splunk Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Microsoft.Win32;
+using System.ComponentModel;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Security.AccessControl;
+using System.Security.Principal;
+
+internal enum ServiceAccountType
+{
+    Preserve,
+    LocalSystem,
+    Virtual,
+}
+
+internal static class ServiceAccountConfiguration
+{
+    internal const string ServiceName = "splunk-otel-collector";
+    internal const string VirtualAccountName = @"NT SERVICE\splunk-otel-collector";
+    internal const string LocalSystemAccountName = "LocalSystem";
+
+    private const uint ScManagerConnect = 0x0001;
+    private const uint ServiceChangeConfig = 0x0002;
+    private const uint ServiceNoChange = 0xffffffff;
+    private const uint ServiceConfigServiceSidInfo = 5;
+    private const uint ServiceSidTypeUnrestricted = 1;
+    private const uint PolicyCreateAccount = 0x0010;
+    private const uint PolicyLookupNames = 0x0800;
+    private const int ErrorMemberInAlias = 1378;
+    private const int ErrorNoSuchAlias = 1376;
+    private const int NerrGroupNotFound = 2220;
+
+    internal static ServiceAccountType ParseAccountType(string? accountType)
+    {
+        switch (accountType ?? string.Empty)
+        {
+            case "":
+                return ServiceAccountType.Preserve;
+            case "virtual":
+                return ServiceAccountType.Virtual;
+            case "localsystem":
+                return ServiceAccountType.LocalSystem;
+            default:
+                throw new ArgumentException("SPLUNK_SERVICE_ACCOUNT_TYPE must be empty, 'virtual', or 'localsystem'.", nameof(accountType));
+        }
+    }
+
+    internal static string SelectAccount(string? configuredAccountType, string? existingAccount)
+    {
+        string existingAccountName = existingAccount ?? string.Empty;
+        switch (ParseAccountType(configuredAccountType))
+        {
+            case ServiceAccountType.Virtual:
+                return VirtualAccountName;
+            case ServiceAccountType.LocalSystem:
+                return LocalSystemAccountName;
+            case ServiceAccountType.Preserve:
+                if (string.IsNullOrWhiteSpace(existingAccountName))
+                {
+                    return VirtualAccountName;
+                }
+
+                if (IsLocalSystem(existingAccountName))
+                {
+                    return LocalSystemAccountName;
+                }
+
+                if (string.Equals(existingAccountName, VirtualAccountName, StringComparison.OrdinalIgnoreCase))
+                {
+                    return VirtualAccountName;
+                }
+
+                throw new InvalidOperationException(
+                    $"The existing service account '{existingAccountName}' cannot be preserved because Windows does not expose its password. " +
+                    "Set SPLUNK_SERVICE_ACCOUNT_TYPE to 'virtual' or 'localsystem' explicitly.");
+            default:
+                throw new InvalidOperationException("Unknown service account type.");
+        }
+    }
+
+    internal static bool IsLocalSystem(string accountName) =>
+        string.Equals(accountName, LocalSystemAccountName, StringComparison.OrdinalIgnoreCase) ||
+        string.Equals(accountName, @"NT AUTHORITY\SYSTEM", StringComparison.OrdinalIgnoreCase);
+
+    internal static string GetExistingServiceAccount()
+    {
+        using (RegistryKey? serviceKey = Registry.LocalMachine.OpenSubKey($@"SYSTEM\CurrentControlSet\Services\{ServiceName}"))
+        {
+            return serviceKey?.GetValue("ObjectName") as string ?? string.Empty;
+        }
+    }
+
+    internal static string LocalSystemRecommendation =>
+        "The Splunk OpenTelemetry Collector service remains configured as LocalSystem. " +
+        "To use the dedicated service account, run: msiexec.exe /i <path-to-splunk-otel-collector.msi> /qn SPLUNK_SERVICE_ACCOUNT_TYPE=virtual";
+
+    internal static void Configure(string selectedAccount, string programDataPath, Action<string> log)
+    {
+        ConfigureServiceAccount(selectedAccount);
+
+        if (!string.Equals(selectedAccount, VirtualAccountName, StringComparison.OrdinalIgnoreCase))
+        {
+            log($"Info: Splunk OpenTelemetry Collector service account is {selectedAccount}.");
+            return;
+        }
+
+        SetUnrestrictedServiceSid();
+        AddToLocalGroup("Performance Monitor Users", log);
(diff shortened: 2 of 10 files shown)


cmd/otelcol/config/collector/splunk_logs_config_windows.yaml
```diff
@@ -361,7 +361,7 @@ extensions:
     directory: "${ProgramData}/Splunk/OpenTelemetry Collector/FileStorage"
     compaction:
       on_start: true
-      directory: ${env:TEMP}
+      directory: "${ProgramData}/Splunk/OpenTelemetry Collector/FileStorage/Temp"
```
pjanotti, line 364: The change itself seems good, but I'm curious if there was anything specific triggering it: is this because the default temp location is not available to the service account?

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
```diff
@@ -0,0 +1,431 @@
...
+    private const uint PolicyLookupNames = 0x0800;
+    private const int ErrorMemberInAlias = 1378;
+    private const int ErrorNoSuchAlias = 1376;
+    private const int NerrGroupNotFound = 2220;
+
+    internal static ServiceAccountType ParseAccountType(string? accountType)
+    {
+        switch (accountType ?? string.Empty)
+        {
+            case "":
```
pjanotti, line 50: nit:
```suggestion
            case string.Empty:
```

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
```diff
@@ -0,0 +1,431 @@
...
+        switch (ParseAccountType(configuredAccountType))
+        {
+            case ServiceAccountType.Virtual:
+                return VirtualAccountName;
+            case ServiceAccountType.LocalSystem:
+                return LocalSystemAccountName;
+            case ServiceAccountType.Preserve:
+                if (string.IsNullOrWhiteSpace(existingAccountName))
+                {
+                    return VirtualAccountName;
```
pjanotti, line 73: Not sure exactly how we can hit this, but it seems worth that it is trying to preserve the current account used for the service, but failing to get the current account name, and failing back to the virtual account.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
```diff
@@ -0,0 +1,431 @@
...
+                }
+
+                if (string.Equals(existingAccountName, VirtualAccountName, StringComparison.OrdinalIgnoreCase))
+                {
+                    return VirtualAccountName;
+                }
+
+                throw new InvalidOperationException(
+                    $"The existing service account '{existingAccountName}' cannot be preserved because Windows does not expose its password. " +
+                    "Set SPLUNK_SERVICE_ACCOUNT_TYPE to 'virtual' or 'localsystem' explicitly.");
```
pjanotti, line 88: Isn't possible to just leave credentials untouched in this case? I mean there are two entries in the registry: one for the account and another for the password if those are not changed the previous settings can be preserved.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    private static void GrantDefaultFileLogAccess(SecurityIdentifier serviceSid, Action<string> log)
+    {
+        string windowsDirectory = Environment.GetFolderPath(Environment.SpecialFolder.Windows);
+        GrantExistingFileAccess(Path.Combine(windowsDirectory, "System32", "DHCP"), serviceSid, true, log);
+        GrantExistingFileAccess(Path.Combine(windowsDirectory, "WindowsUpdate.log"), serviceSid, false, log);
+        GrantExistingFileAccess(Path.Combine(windowsDirectory, "debug", "netlogon.log"), serviceSid, false, log);
+        GrantExistingFileAccess(Path.Combine(windowsDirectory, "System32", "LogFiles", "Firewall"), serviceSid, true, log);
+    }
```
pjanotti, line 310: Isn't better to give `SeBackupPrivilege` instead of these very specific settings? Are some of these locations not covered by `SeBackupPrivilege`. Look at the defaults used by UF, these should be our reference: https://help.splunk.com/en/splunk-cloud-platform/forward-and-process-data/universal-forwarder-manual/10.4/install-the-universal-forwarder/install-a-windows-universal-forwarder#ariaid-title2

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
```diff
@@ -0,0 +1,431 @@
...
+        AddToLocalGroup("Performance Monitor Users", log);
+        AddToLocalGroup("Event Log Readers", log);
+
+        SecurityIdentifier serviceSid = ResolveVirtualAccountSid();
+        GrantSecurityEventLogPrivilege(serviceSid, log);
+        GrantFileAccess(
+            Path.Combine(programDataPath, "Splunk", "OpenTelemetry Collector"),
+            serviceSid,
+            FileSystemRights.ReadAndExecute,
+            true);
```
kelseyma, line 130: Just wanted to check if this will also cover the `supervisor` subdir that gets created when supervisor mode is enabled? or if that needs to be added similar to `FileStorage` below?

pjanotti: yes, the method used to set the rights specify the flags for directories and objects to inherit the rights.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
```diff
@@ -0,0 +1,431 @@
...
+            Directory.CreateDirectory(path);
+            DirectoryInfo directory = new DirectoryInfo(path);
+            DirectorySecurity security = directory.GetAccessControl();
+            security.AddAccessRule(new FileSystemAccessRule(
+                serviceSid,
+                rights,
+                InheritanceFlags.ContainerInherit | InheritanceFlags.ObjectInherit,
+                PropagationFlags.None,
+                AccessControlType.Allow));
+            directory.SetAccessControl(security);
```
pjanotti, line 336: Q.: if you start a subdirectory that has very broad rights, don't you need to stop the current location from inherit from parent first? I'm not sure but considering a bad practice that someone has given broad rights to `ProgramFiles` and/or `ProgramData` and made inherit. At first I would like those rights to not be inherited during install.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
```diff
@@ -0,0 +1,431 @@
...
+        if ((isDirectory && !Directory.Exists(path)) || (!isDirectory && !File.Exists(path)))
+        {
+            log($"Warning: Default file log path '{path}' does not exist; grant {VirtualAccountName} read access before enabling it.");
+            return;
+        }
+
+        GrantFileAccess(path, serviceSid, FileSystemRights.ReadAndExecute, isDirectory);
+    }
+
+    private static void GrantFileAccess(string path, SecurityIdentifier serviceSid, FileSystemRights rights, bool isDirectory)
```
pjanotti, line 323: nit: the name is `GrantFileAccess` but there is a `isDirectory` parameter, and inside the there is a decision based on the parameter. It makes me wonder if instead of the parameter we should have different methods instead.

cmd/otelcol/config/collector/splunk_logs_config_windows.yaml
weliasz: ${env:TEMP} may work for a virtual service account, but Windows does not guarantee that the resolved temp path exists or is writable by that service identity.

The change therefore provides a deterministic, installer-owned directory where the MSI explicitly grants the virtual account Modify access. This avoids relying on ambient TEMP ACLs and avoids granting the Collector broad write access to a shared system temp directory.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
weliasz: User with SeBackupPrivilege can access any file/directory bypassing ACLs.
It was not used due to security reasons. With SeBackupPrivilege enabled process can read user accounts and passwords together with key used to encrypt them.

SeBackupPrivilege is optional during Splunk UF installation as I guess there may be clients that would not accept giving this permission to Splunk service user.

Without SeBackupPrivilege, access would have to be granted to all filelog entries using dedicated setting.
If new entries would be added to the config ACL would have to be modified accordingly.

We could use SeBackupPrivilege but only if user explicitly ask for it during installation.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
weliasz: install.ps1 explicitly uninstalls the old MSI before installing the new one.
After removal, there is no existing service configuration or credential to preserve. The new service is initially created as LocalSystem.

Supporting account preservation would require redesigning the upgrade path.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
weliasz: This case would happen on fresh install.

Should we consider support for preserving on upgrade account names other than: NT SERVICE\splunk-otel-collector and LocalSystem?

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
weliasz: If we stop the inheritance we would need to consider which users/services would have to be explicitly granted access to new directories e.g. LocalService, NetworkService, ordinary Users, parent groups.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
pjanotti: Yes, that would be ideal.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
pjanotti: makes sense, so no preservation possible via install.ps1, that is fine. Can we do it via the MSI directly?

pjanotti: I mean the main question is: does the MSI upgrade preserves any account that the user may have setup? If not then we could separate the support of "preserving existing account during upgrade" a desirable feature for the future.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
pjanotti: My concern is that if the installation is made to a directory with low protection, then all of the collector files will inherit the same exposing another risk. Perhaps we should look at the default permissions for ProgramFiles and always start the collector folder always with those permissions.

weliasz: There are 2 directories the Collector needs permission to 

- %ProgramFiles%\Splunk\OpenTelemetry Collector - execution of otel binary
- %ProgramData%\Splunk\OpenTelemetry Collector - configuration must be readable and FileStorage writable by service account.

So we could choose a template for each of them (maybe ProgramFiles and ProgramData respectively) plus add permissions for service account. ACLs would be copied and made immutable after installation.

packaging/msi/SplunkCustomActions/src/ServiceAccountConfiguration.cs
weliasz: When string.Empty is used following error in [Ansible / msi-custom-actions / msi-custom-actions (pull_request)] is reported:

Error: D:\a\splunk-otel-collector\splunk-otel-collector\packaging\msi\SplunkCustomActions\src\ServiceAccountConfiguration.cs(50,18): error CS9135: A constant value of type 'string' is expected [D:\a\splunk-otel-collector\splunk-otel-collector\packaging\msi\SplunkCustomActions\src\SplunkCustomActions.csproj]

PR #8057, opened 2026-09-11
add splunk_outputs observer

omrozowicz-splunk:
**Description:**   
Add filesystem observer to splunk_outputs exporter that watches `etc/system/default` and `etc/system/local` for `outputs.conf` changes and hot-reloads the underlying exporter without restarting the collector.

- Single active `exporter.Logs` under `sync.RWMutex`; reconcile atomically swaps in a fresh exporter and shuts down the old one
- `fsnotify-backed` watcher with 500ms debounce, mirroring `splunkinputsreceiver` structure
- Watches parent dirs (etc, etc/system) so late creation of etc/system/local is detected; reconcile retries `watcher.Add` on every call
- aggregateExporter handles multiple output stanzas (e.g. [tcpout:server1] + [tcpout:server2])
- All real work deferred from CreateLogs to Start, consistent with OTel component lifecycle

**Link to Splunk idea:** N/A

**Testing:** Unit tests, manual tests

**Documentation:** This is more an internal change


dhruvanshus-crest (approved):

dmitryax:
I'm wondering if dynamic reload of output.conf files are actually supported by the UF... We can have it regardless though

dmitryax (approved):

omrozowicz-splunk:
> I'm wondering if dynamic reload of output.conf files are actually supported by the UF... We can have it regardless though

Yes, this is a good question 😅 Looks like UF requires a restart when modifying `outputs.conf`. As work is done already, let me merge it.

PR #8060, opened 2026-09-11
[chore] Fix data race in scriptedinput

dmitryax:
`_execute` read stdout in a goroutine that raced `cmd.Wait`, and `si.command` was accessed from both `_execute` and `Stop` without synchronization. The stdout race also dropped output when the process exited quickly, which flaked `Test_ScriptedInput`.

Read stdout to EOF synchronously before `cmd.Wait` (as `exec.Cmd.StdoutPipe` requires) and guard `si.command` with a mutex. With the race gone, the `-race` test skip and its `race_on`/`race_off` build-tag files are removed.

### DIFF
--- a/pkg/splunkta/scriptedinput/input.go
+++ b/pkg/splunkta/scriptedinput/input.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os/exec"
 	"strconv"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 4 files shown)


atoulme (approved):

dmitryax:
Superseded by the stacked pair #8090 (context-based lifecycle rework, lands first) and #8091 (the flake fix on top, no mutex needed). Closing this one.

PR #8092, opened 2026-09-15
improvements to partial reload for splunk inputs

dhruvanshus-crest:
**Description:** 
This change introduces partial reload support for TA input configurations.
Previously, changing one input stanza caused all receivers in the affected TA to restart. Now, each TA input stanza is compared independently, and only receivers with changed effective configuration are recreated.

What changed
- Store the effective configuration for each TA input stanza.
- Compare inputs.conf, props.conf, and transforms.conf values during reload.
- Preserve unchanged receivers.
- Restart only added, removed, or modified receivers.
- Continue checking all TAs for system-level configuration changes, while restarting only affected receivers.
- Add documentation and regression tests.

**Link to ADDON Jira:** [ADDON-90732](https://splunk.atlassian.net/browse/ADDON-90732)

**Testing:** Unit tests and manual tests

**Documentation:** This is an internal change


docs/partial-reload.md
dmitryax, line 1: Even though this doc is useful, it describes the changeset. Should be in a PR description, no need to check it in the code.

pkg/receiver/splunkinputsreceiver/receiver.go
```diff
@@ -71,6 +93,13 @@ func (r *splunkInputsReceiver) Shutdown(ctx context.Context) error {
 	return nil
 }
 
+// Reload reconciles all system and TA input stanzas against the currently
+// running receiver instances. Only stanzas whose effective configuration
+// changed are rebuilt.
+func (r *splunkInputsReceiver) Reload(ctx context.Context) error {
```
dmitryax: Why do we need to expose this API in this PR? How is it going to be used?

docs/partial-reload.md
dhruvanshus-crest: ok, I'll remove this file.

pkg/receiver/splunkinputsreceiver/receiver.go
dhruvanshus-crest: I thought it might be necessary to avoid waiting on debounce period, something like immediate reload. But may be it is not required for now. I will remove it.

pkg/receiver/splunkinputsreceiver/receiver.go
```diff
@@ -210,18 +220,25 @@ func (r *splunkInputsReceiver) reconcile(ctx context.Context, pending map[string
 	if len(added) > 0 {
 		if err := r.handler.OnAdd(ctx, added); err != nil {
 			logger.Error("splunk_inputs: failed to start receivers for new TAs", zap.Error(err))
+			return err
```
omrozowicz-splunk: now when I think about it, wouldn't it stop the reconcile process when `splunk_inputs` from one of the TA don't work? What about other TAs? shouldn't we just log and continue?

dhruvanshus-crest: yeah, nice catch will update accordingly

omrozowicz-splunk (approved):

dmitryax (approved):

PR #8098, opened 2026-09-16
[chore] Use Microsoft-Go for Windows binaries

pjanotti:
Testing if there is any impact in switching tool chain to Microsoft-Go to build Windows binaries


pjanotti:
No need for this experiment anymore.

